PR #787, opened 2016-09-14
resolve Safari / WebKit sizing issue for toolbars

kevinushey:
This PR resolves a Safari / WebKit issue where `<td>` elements containing a single child element styled with `display: none` would still be 'sized' (seemingly a single pixel in size). This causes the toolbar used in the source pane to appear to be inconsistently sized across different document types, as for disabled commands we still add these to the DOM but toggle their visibility with e.g. `display: none`.

Here's a small screencast showcasing the issue -- the effect is most prominent on the 'document outline' toolbar button, but notice how the icons seem to 'jump' a bit as the document is changed:

toolbar-sizing

Probably too late in the game / inconsequential for the v1.0 release, but let me know what you think.

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/dom/DomUtils.java
+++ b/src/gwt/src/org/rstudio/core/client/dom/DomUtils.java
@@ -42,6 +42,7 @@ import org.rstudio.core.client.dom.impl.DomUtilsImpl;
 import org.rstudio.core.client.dom.impl.NodeRelativePosition;
 import org.rstudio.core.client.regex.Match;
 import org.rstudio.core.client.regex.Pattern;
+import org.rstudio.core.client.theme.res.ThemeStyles;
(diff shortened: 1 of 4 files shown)


jjallaire:
Let's definitely hold this one for post v1.0

kevinushey:
I think this would be safe to turn on in all browsers. FWIW this doesn't really need to be a 'global' fix, since the only time a user would 'see' this bug is when tabbing through the source pane -- each  tab shares icons that normally lie in the same space, and the spacing between those icons will appear to change as the document is changed. I don't think that occurs in any other Toolbar UI we present.

Also agree that holding for post v1.0 is the way to go.

kevinushey:
FWIW, this issue still exists with Safari 10.0.

jmcphers:
Merged onto 1.0-patch.

PR #789, opened 2016-09-16
avoid prompting for spark install or show versions when spark home set

javierluraschi:
Another one to address issues form https://github.com/rstudio/sparklyr/issues/218 - Safe fix with broken functionality worth merging into 1.0 branch.

The issue is that one can configure `SPARK_HOME` and still get this dialog:

screen shot 2016-09-16 at 1 30 05 pm

When the versions will be effectively ignored; therefore, this dialog makes more sense for this fix:

screen shot 2016-09-16 at 1 29 10 pm

Validated that we still have the original dialog when `SPARK_HOME` is not present.


javierluraschi:
Updated screenshot with updated dialog.

PR #790, opened 2016-09-20
Avoid auto adding spark protocol to allow yarn-client

javierluraschi:
Most Spark cluster URLS start with the prefix `spark://` so we added support in the UI to auto add this in cluster mode. However, some modes like `yarn-client` and `yarn-cluster` are valid and don't require this protocol prefix. Removing this since users can always type their own protocol prefix as needed.

Related to https://github.com/rstudio/sparklyr/issues/218

screen shot 2016-09-20 at 3 41 37 pm

screen shot 2016-09-20 at 3 41 45 pm


javierluraschi:
It's safe enough to merge and specially since it blocks a common case of urls (`yarn-client`) from the UI.

PR #792, opened 2016-09-22
use correct completion result for 'filter' when dplyr loaded

kevinushey:
This PR resolves an issue where, even if `dplyr` is loaded, the help topic for `stats::filter` was shown in preference to that for `dplyr::filter`.

screen shot 2016-09-22 at 3 33 00 pm

### DIFF
--- a/src/cpp/session/modules/SessionRCompletions.R
+++ b/src/cpp/session/modules/SessionRCompletions.R
@@ -1460,10 +1460,17 @@ assign(x = ".rs.acCompletionTypes",
    # Keywords are really from the base package
    packages[packages == "keywords"] <- "base"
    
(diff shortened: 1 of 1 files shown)


jmcphers:
What's your assessment of the risk here? i.e. should we take this now (given we've pushed out the release a week) or delay for the patch release?

kevinushey:
I think this should be okay to take for v1.0, although I'd be fine deferring for the patch release in the interest of being extra conservative. (we've been living with this bug for quite a while and we've only heard about it recently)

jjallaire:
Yeah, let's keep it in patch.

jmcphers:
Okay, merged into 1.0-patch.

PR #793, opened 2016-09-23
fix HOME, R_USER, R_LIBS_USER encoding issue on Windows

kevinushey:
This PR fixes an issue where the values of the environment variables `HOME`, `R_USER`, and `R_LIBS_USER` would be mis-encoded on Windows for users with non-ASCII characters in their names. This should ultimately fix a large number of issues for such users on Windows (e.g. inability to install packages).

We may want to take this for v1.0, although we should confirm that `utf8ToSystem` won't cause something terrible to happen on failure.

### DIFF
--- a/src/cpp/r/session/REmbeddedWin32.cpp
+++ b/src/cpp/r/session/REmbeddedWin32.cpp
@@ -31,6 +31,7 @@
 
 #include <core/FilePath.hpp>
 #include <core/Exec.hpp>
+#include <core/StringUtils.hpp>
 
 #include <r/RInterface.hpp>
 #include <r/RFunctionHook.hpp>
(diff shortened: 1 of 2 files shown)


jmcphers:
This seems safe and looks good, but I'd hesitate to take it for 1.0 because:

a) this is a day zero bug (i.e. it's always been this way, so failing to fix it won't break anyone)

b) there's a lot of downstream code here, so if this were to affect something else, it may do so in a way we'll miss.

How would you feel about merging this onto 1.0-patch?

jjallaire:
I agree we shouldn't merge anything else onto 1.0 unless it's a major
bugfix for new features or a regression fix. +1 on merging it onto patch.

jmcphers:
Merged onto 1.0-patch.

PR #797, opened 2016-09-27
resolve issue with latex tokenization within '{}'

kevinushey:
This PR resolves a tokenization issue where math markup within a latex command, e.g. `\foo{$bar$}`, would not be tokenized correctly.

screen shot 2016-09-27 at 10 12 47 am

The overarching issue -- our highlighter previously attempted to consume an entire 'line' containing a matching pair of `{` and `}`, tokenizing the inner contents as plain text. However, this is suboptimal for a couple reasons:

1. It's possible for a latex command to span multiple lines,
2. It's possible for non-textual content to be included within `{}`.

The solution here is to simply tokenize `{` and `}` separately as brackets, and to avoid including `{}` when tokenizing a region of text.

### DIFF
--- a/src/gwt/acesupport/acemode/markdown_highlight_rules.js
+++ b/src/gwt/acesupport/acemode/markdown_highlight_rules.js
@@ -214,22 +214,22 @@ var MarkdownHighlightRules = function() {
             token : "constant",
             regex : "^[ ]{0,2}(?:[ ]?\\_[ ]?){3,}\\s*$"
         }, { // MathJax native display \[ ... \]
-            token : "latex.markup.list.begin",
+            token : "latex.markup.list.string.begin",
             regex : "\\\\\\[",
             next  : "mathjaxnativedisplay"
         }, { // MathJax native inline \( ... \)
-            token : "latex.markup.list.begin",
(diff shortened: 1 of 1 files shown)


jjallaire:
LGTM but probably also worth a look from @jmcphers before we merge.

jjallaire:
@mine-cetinkaya-rundel

kevinushey:
As an aside, should we consider a different highlight color for the mathjax boundaries? It seems like it would be useful to color them distinctly just so the boundaries were more obvious (especially for e.g. `\[ \]` style boundaries)

jjallaire:
We use the pinkish colors for $, what about that?

kevinushey:
It looks like `textmate` is one of the only themes that actually supplies highlight rules for `markup.list`; the simple change here is to append `string` so that this highlight rule is used for themes which do not have rules for `markup.list`. (The `string` highlighting for most themes tends to be soft, and I think fits well)

screen shot 2016-09-27 at 10 55 23 am

jmcphers:
LGTM too. Thanks for spotting this, @mine-cetinkaya-rundel!

PR #798, opened 2016-09-27
improve resilience against malformed YAML header

kevinushey:
Currently, malformed YAML headers can cause the rest of the document to be mis-tokenized:

screen shot 2016-09-27 at 1 10 43 pm

Note that the R chunk below is not properly highlighted as R code, and it also does not have a chunk toolbar + the other 'goodies' normally afforded to R chunks. This is occurring because the second instance of `---` is being interpreted as the start of a YAML block, and hence the rest of the following code also as YAML.

We had code that intended to ensure that YAML headers would only be detected at the very start of the document, but this did not always function in practice. This PR remedies that, so that a `---` is only ever interpreted as the start of a YAML block at the very start of the document.

With this PR:

screen shot 2016-09-27 at 1 15 28 pm

### DIFF
--- a/src/gwt/acesupport/acemode/rmarkdown_highlight_rules.js
+++ b/src/gwt/acesupport/acemode/rmarkdown_highlight_rules.js
@@ -139,10 +139,7 @@ var RMarkdownHighlightRules = function() {
        ["start", "listblock", "allowBlock"]
    );
 
-   // Embed YAML highlight rules, but ensure that they can only be
-   // found at the start of a document. We do this by moving all of the
-   // start rules to a second '$start' state, and ensuring that YAML headers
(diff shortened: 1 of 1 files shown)


jjallaire:
do you like this for 1.0-patch or v1.1?

kevinushey:
I would prefer v1.0-patch just because it might not be obvious to a user why their document "doesn't work" after accidentally making a typo / change that breaks the YAML header. They would hopefully discover the cause after attempting a document render, but it may not be obvious until then.

jjallaire:
Okay, sounds good.

jmcphers:
Looks good, merged to 1.0-patch.

kevinushey:
Thanks! (was just about to try to change the base but you beat me to it :))

PR #807, opened 2016-10-03
ship pagedtable.js embedded instead than from rmarkdown

javierluraschi:
This PR ships `pagedtable.js` embedded in the IDE instead than fetching it from `rmarkdown`. This helps resolve issues from taking a dependency to rmarkdown, especially, to avoid requiring additional restarts after the `rmarkdown` package gets installed or other cross-package dependency loading issues.

### DIFF
--- a/src/cpp/session/modules/rmarkdown/NotebookData.cpp
+++ b/src/cpp/session/modules/rmarkdown/NotebookData.cpp
@@ -60,21 +60,12 @@ void handleNotebookDataResReq(const http::Request& request,
 {
    std::string resourceName = http::util::pathAfterPrefix(request, kNotebookDataResourceLocation);
 
-   std::string resource;
+   std::string resourcePath("pagedtable/");
+   resourcePath.append(http::util::pathAfterPrefix(request, kNotebookDataResourceLocation));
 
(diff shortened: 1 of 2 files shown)


kevinushey:
Do we want to perhaps do this as a fallback? Ie, attempt to get `pagedtables.js` from `rmarkdown`, and only if that fails use the RStudio-embedded version?

jjallaire:
It's not that we can't get it from rmarkdown, it's that it's the wrong
version. I think to correctly detect a wrong version would require more
(regression-prone) code which we don't want to introduce at this point.
Let's take this PR now and then eventually depend fully on rmarkdown once
the dependency stabilizes.

kevinushey:
Right, my point is just that with this PR the only way to ship updates to `pagedtable.js` is with new versions of the IDE (ie we'll have to ask users to update RStudio); just want to make sure that's intended. Do we plan to change this in the patch release or not until v1.1?

jjallaire:
Yes, it's fine for improvements in pagedtable.js within the IDE to require
an RStudio update for now.

I think we'll change this is v1.1 when we are convinced that pagedtable.js
is more or less done.

kevinushey:
Sounds good, thanks!

PR #809, opened 2016-10-10
handle 'gvis' output in notebooks

kevinushey:
Not quite ready for merge...

This PR allows for `googleVis`-generated outputs to be rendered within the notebook.

Some issues that we might want to figure out how to solve in this PR:

1. `gvis` objects are not responsive -- should we re-render / refresh them when the editor window is resized, or is it okay to live with the lack of responsiveness?

2. The height of the line widget container is not computed correctly for `gvis` widgets -- that is, the container size does not resize to match the size of the generated `gvis` plot.

Current state (note the empty space below the plot is part of the line widget's container)

screen shot 2016-10-10 at 1 18 51 pm

### DIFF
--- a/src/cpp/session/modules/NotebookHtmlWidgets.R
+++ b/src/cpp/session/modules/NotebookHtmlWidgets.R
@@ -38,6 +38,27 @@
    htmlfile <- tempfile("_rs_gvis_", tmpdir = ctx$outputFolder, fileext = ".html")
    depfile  <- tempfile("_rs_gvis_deps_", tmpdir = ctx$outputFolder, fileext = ".json")
    
+   # inject some of our own JavaScript event handlers in the gvis object
+   # mutates 'x' on success
+   try(silent = TRUE, {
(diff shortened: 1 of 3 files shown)


kevinushey:
I think the problem with the height here is a timing one -- we're attempting to read the height of the document before the HTML within has been properly loaded + resized.

kevinushey:
@jmcphers, a quick question -- where should I look to find the code that sets the icons + text for the 'multi-output' display? E.g. here:

screen shot 2016-10-11 at 10 13 20 am

We'd probably like to use a special icon for these, and perhaps have the underlying text just read `gvis`. Let me know what you think.

jmcphers:
No other HTML widget gets a special icon so I don't think this needs one either (alternately we could try to pick the top N widgets and make special icons for them, but I'm not sure gvis would make the list!). 

Here's where you'd tweak the text on the thumbnail:

https://github.com/rstudio/rstudio/blob/88e8188569d08017f635252a669b02f53c35e4fe/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/ChunkHtmlPage.java#L43-L45

PR #815, opened 2016-10-12
report R launch errors

kevinushey:
This PR modifies some of the R startup error reporting such that non-zero exit statuses are now reported to the user. (Note that processes that run to completion, but return a non-zero exit status, return a `Success()` error object -- hence they are often not reported as one might expect).

The overall goal is to ensure that initialization errors are better reported to users.

### DIFF
--- a/src/cpp/core/include/core/system/Process.hpp
+++ b/src/cpp/core/include/core/system/Process.hpp
@@ -140,6 +140,9 @@ struct ProcessResult
    int exitStatus;
 };
 
+// Return a user-friendly error message for process results.
+std::string processExitStatusMessage(const ProcessResult& result,
+                                     const std::string& command);
 
 // Run a program synchronously. Note that if executable is not an absolute
 // path then runProgram will duplicate the actions of the shell in searching
--- a/src/cpp/core/r_util/REnvironmentPosix.cpp
+++ b/src/cpp/core/r_util/REnvironmentPosix.cpp
@@ -195,6 +195,10 @@ std::string extraLibraryPaths(const FilePath& ldPathsScript,
    Error error = runCommand(command, core::system::ProcessOptions(), &result);
    if (error)
       LOG_ERROR(error);
+   
+   if (result.exitStatus)
+      LOG_ERROR_MESSAGE(system::processExitStatusMessage(result, command));
+   
    std::string libraryPaths = result.stdOut;
    boost::algorithm::trim(libraryPaths);
    return libraryPaths;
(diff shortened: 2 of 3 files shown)


jjallaire:
I don't love the idea of changing whether we return true or false based on the presence a non-success exitStatus (perhaps there are situations where we currently rely on this behavior, I'd have to look at all of the client code to know for sure). What about if we just log the error messages and let the control flow stay as is? That would give us the additional diagnostics without the possibility of unintended regressions.

Of course you might be attempting to address a bug in control-flow here, if so let's discuss.

kevinushey:
The main goal is to just improve error reporting when RStudio fails to launch the underlying R session. (It appears that this can occur if R and RStudio were built with incompatible toolchains, although I haven't ascertained a root cause; at least, this is what I saw in a couple cases locally)

The problem is that users can see error messages like:

```
Jul 23 13:00:40 <user> rserver[8587]: ERROR R did not return any output when queried for directory location information; LOGGED FROM: bool rstudio::core::r_util::<unnamed>::detectRLocationsUsingR(const std::string&, rstudio::core::FilePath*, rstudio::core::FilePath*, rstudio::core::config_utils::Variables*, std::string*) /root/rstudio-pro/src/cpp/core/r_util/REnvironmentPosix.cpp:488 

Jul 23 13:00:40 <user> rserver[8587]: ERROR system error 71 (Protocol error) [description=Unable to parse version from R, version-info=]; OCCURRED AT: rstudio::core::Error rstudio::core::r_util::rVersion(const rstudio::core::FilePath&, const rstudio::core::FilePath&, std::string*) /root/rstudio-pro/src/cpp/core/r_util/REnvironmentPosix.cpp:773; LOGGED FROM: bool rstudio::core::r_util::detectREnvironment(const rstudio::core::FilePath&, const rstudio::core::FilePath&, const std::string&, std::string*, std::string*, rstudio::core::r_util::EnvironmentVars*, std::string*) /root/rstudio-pro/src/cpp/core/r_util/REnvironmentPosix.cpp:681
```

However, I think these messages might imply that R was launched successfully + ran some commands, but those commands did not produce any output, when the real root cause is simply that R died immediately after being launched (or exited early for some other reason). Reporting those early / unexpected exits could be helpful for some users. Let me know what you think.

jjallaire:
Okay, let's discuss realtime on Monday (I'm out tomorrow). My main
sensitivity is making changes to the control flow of code that runs at the
start of every R session (we can certainly do so if warranted, I just want
to make sure we think through all lower-risk paths first).

PR #825, opened 2016-10-19
allow customization of home, scratch directories

kevinushey:
This PR allows users to customize the default home directory, alongside the RStudio 'scratch' directory, using options passed to `rsession.conf`. This should be useful for RStudio Server users with limited space in their home directories, who instead want to tell RStudio to use + write configuration / state to a separate directory.

We likely want to tweak this PR a bit if we end up taking it, e.g.

1. We might want to allow 'templates' for the path; e.g. allow users to provide something like `/usr/local/rstudio/rstudio-$USER-$PID/`, and expand those shell variables (so one global config file can serve + separate for multiple users),

2. We might also want to allow users to set this with an environment variable.

Thoughts, @jjallaire @jmcphers?

### DIFF
--- a/src/cpp/session/SessionOptions.cpp
+++ b/src/cpp/session/SessionOptions.cpp
@@ -200,7 +200,13 @@ core::ProgramStatus Options::read(int argc, char * const argv[])
        "default directory for new projects")
       ("show-help-home",
        value<bool>(&showHelpHome_)->default_value(false),
-         "show help home page at startup");
+         "show help home page at startup")
+      ("session-user-home-path",
+       value<std::string>(&userHomePath_)->default_value(""),
+       "default user home path")
+      ("session-user-scratch-path",
+       value<std::string>(&userScratchPath_)->default_value(""),
+       "default user scratch path");
 
    // allow options
    options_description allow("allow");
@@ -448,8 +454,13 @@ core::ProgramStatus Options::read(int argc, char * const argv[])
                                     r_util::SessionTypeServer;
 
    r_util::UserDirectories userDirs = r_util::userDirectories(sessionType);
-   userHomePath_ = userDirs.homePath;
-   userScratchPath_ = userDirs.scratchPath;
+   
+   // use user paths (unless specified by option)
+   if (userHomePath_.empty())
+      userHomePath_ = userDirs.homePath;
+   
+   if (userScratchPath_.empty())
+      userScratchPath_ = userDirs.scratchPath;
 
    // set HOME if we are in standalone mode (this enables us to reflect
    // R_USER back into HOME on Linux)


jjallaire:
There is actually considerably more than meets the eye to this and I don't think we should attempt to do it now. The user home and scratch directories are used throughout the server binaries for various things including session locks, collaboration, and more. Further, they are computed rather than read from a setting. Therefore, any attempt to make this customizable needs to be at the rserver.conf level and then the change must be successfully propagated to all of the other processes that use it. This is an intrusive and risky change.

There's a further issue here, which is that there is *already* a way to customize the `~` directory, which is the specification of the R_USER environment variable (see https://github.com/rstudio/rstudio/blob/54cd3abcfc58837b433464c793fe9b03a87f0bb4/src/cpp/core/r_util/RUserData.cpp#L40). This method has the benefit of working for both R at the terminal and for RStudio.

In terms of specifying a custom scratch path, so far the environments that have requested this have some administrative/policy related restriction on the size of user home directories so the scratch path is a way to get around this. Unfortunately, a custom scratch path means maintaining a parallel home directory-style filesystem (per user permissions, quotas, and backup). It also doesn't solve the root problem which is that if my R dataset sizes are large I need *more storage* (and I'll need that in the home directory as well for saved .RData). So the custom scratch path is as far as I can tell so far a band-aid over a broken internal storage policy. I would like to see a better use-case / argument for this which *isn't* solved by R_USER before we do anything here.

Final note: this PR is against master which will become v1.0-patch. If we don't want the patch release to drag on for 3 or 4 months we need to hold changes that have high potential risk of regression, which this one absolutely would if taken to it's conclusion throughout the codebase.

kevinushey:
Thanks for the extra information, JJ. The primary motivation for this PR was indeed to help users who have limited space in their home directory (ie, if we do this work then perhaps only customizing the 'scratch' space directory is warranted).

We also definitely don't have to take this for v1.0-patch (I'll make sure I base my PRs against the intended merge patch in the future).

I'll close this PR for now and we can follow up in the future when / if we decide to pursue this.

jjallaire:
Okay, let's discuss realtime if we do decide to revisit this. One other
thing most users don't realize is that if they are using projects then
there are actually multiple scratch directories (one for each project) so
moving the global scratch directory doesn't really solve their problem.

My 2 cents is that the real feature here is moving the ~ directory and that
since we already have it we just need to document it better.

nwstephens:
I would like to document this better, but I was unable to get `R_USER` working on the terminal and in RStudio. I set the following under `R_HOME` in `/usr/lib64/R/etc/Rprofile.site`:

`Sys.setenv(R_USER="/home2")`

When I start R in the terminal and in RStudio, I see that the `R_USER` environment variable has been set appropriately when I run `Sys.getenv()`. However, nothing happens; R still associates `~` with `/home` not with `/home2`. Running `setwd("~")` for example sends me to `/home` not `/home2`. When I start up a new session in RStudio my file browser goes to `/home` not `/home2`.

kevinushey:
I wonder if the `Rprofile.site` is being executed by R too late for this? Does setting it within `Renviron` / Renviron.site` work?

nwstephens:
I also tried to set `R_USER` in `/etc/rstudio/rsession-profile`. But this lead RSP to give me a flashing grey screen as it tried to reconnect.

nwstephens:
Yep, I got the same behavior in `Renviron`:

`R_USER=/home2`

I see `R_USER` in `Sys.getenv()` but `setwd("~")` sends me to the old home.

nwstephens:
I should say that I'm trying this on RSP Version 1.1.230.

kevinushey:
Hunting through the R Manuals a bit, it looks like the use of `R_USER` is documented as a primarily Windows-only thing. For Unix, you might need to just set the `HOME` environment variable explicitly.

nwstephens:
Thanks, Kevin. I will try that. I also noticed that R_USER was pegged to Windows documentation.

nwstephens:
I set `HOME=/home2` in `Renviron`. Now when I start R in the terminal I get `/home2` as my current working directory. But in RSP I still get `/home` as the current working directory and in the file pane (see attached). But if I change directories to `~` then my working directory is set to home. But the `.rstudio` folder is still dropped into `/home` and not `/home2`. (Note I used a fresh login here to test.)

screen shot 2017-05-24 at 9 22 07 pm

What I really want is to map `.rstudio` into `/home2` and also have `/home2` as my working directory.

Thing I haven't tried yet are to set the `HOME` in different startup files.

nwstephens:
I'm getting the same behavior when I put `Sys.setenv(HOME="/home2")` into `Rprofile.site`. 

It appears that `.rstudio` still maps to `/home` while the rest map to `/home2` (`R`, `pki`, `.subversion`). Everything appears to be mapping to `/home2` except the one thing I really want, which is `.rstudio`.

jjallaire:
I think you need to use system level environment variable setting rather
then Renviron, etc as we don't see those values until it's too late

nwstephens:
Thanks. `/etc/profile` and `~/.bash_profile` both give me the flashing grey screen in RStudio when I set `export R_USER=home`. I'm not sure how `R_USER` is being used by R on Linux. Oddly, when I set `R_USER` in `/etc/profile` it doesn't even show up in my R environment in the terminal.

PR #826, opened 2016-10-20
fix escaping of shell characters

kevinushey:
This PR resolves an issue where elements were incorrectly escaped when passed to shell commands on POSIX systems.

One way to reproduce this -- in a Git project, attempt to stage a file with the name `$$$.R`. This fails because the filename is 'escaped' as `$1$1$1.R`, and an attempt to commit that file fails. The issue here is that we attempted to perform regex substitution without actually specifying a capture group in the regular expression used.

I think this is a candidate for v1.0-patch but let me know what you think.

### DIFF
--- a/src/cpp/core/CMakeLists.txt
+++ b/src/cpp/core/CMakeLists.txt
@@ -104,7 +104,6 @@ set (CORE_SOURCE_FILES
    r_util/RSessionContext.cpp
    r_util/RTokenizer.cpp
    r_util/RSourceIndex.cpp
-   r_util/RTokenizerTests.cpp
    r_util/RUserData.cpp
    spelling/HunspellCustomDictionaries.cpp
    spelling/HunspellDictionaryManager.cpp
--- a/src/cpp/core/system/PosixShellUtils.cpp
+++ b/src/cpp/core/system/PosixShellUtils.cpp
@@ -23,9 +23,8 @@ namespace shell_utils {
 
 std::string escape(const std::string& arg)
 {
-   using namespace boost;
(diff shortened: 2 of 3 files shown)


src/cpp/core/CMakeLists.txt
```diff
@@ -104,7 +104,6 @@ set (CORE_SOURCE_FILES
    r_util/RSessionContext.cpp
    r_util/RTokenizer.cpp
    r_util/RSourceIndex.cpp
-   r_util/RTokenizerTests.cpp
```
kevinushey: Note that all files matching the glob `*Tests.cpp` are implicitly added to the test executables built so this entry was unnecessary.

src/cpp/core/system/PosixShellUtils.cpp
```diff
@@ -23,9 +23,8 @@ namespace shell_utils {
 
 std::string escape(const std::string& arg)
 {
-   using namespace boost;
-   regex pattern("[\\\\$`!\\n\"]", regex_constants::normal);
-   return "\"" + regex_replace(arg, pattern, "\\$1") + "\"";
+   boost::regex pattern("(\\\\|\\$|`|!|\n|\")");
```
kevinushey: We might actually need to replace newlines explicitly with the characters `\n` here as well (since I'm not sure if placing a `\` in front a literal newline character will have the intended effect)

jjallaire:
The original regex was from @jcheng5. Joe, could you weight in here to make sure we aren't missing anything we should be?

Yes, I think this is a good candidate for v1.0-patch!

jcheng5:
LGTM

PR #828, opened 2016-10-20
add support for html chunk sizing with knitr.figure=FALSE

javierluraschi:
Adds support for html chunk sizing with `knitr.figure=FALSE` by persisting the sizing policy in the notebook metadata, loading as a 25px height chunk, loading for the html iframe to render and then adjust the size to it's content while disabling `?viewer_pane=1` in the URL to avoid `viewer.fill` (http://www.htmlwidgets.org/develop_sizing.html#examples) to trigger which would cause the iframe and wrapping elements to set `height=100%`.

screen shot 2016-10-20 at 3 57 45 pm


hadley:
🎉

PR #829, opened 2016-10-20
remove key-column special-case alignment in pagedtables

javierluraschi:
On notebook data chunks, functionality was added to left align a numeric column if it looked like a key column. This was performed by checking if numeric and incrementing. However, this introduced to problems:

1) Users (Hadley for instance), suggest that this is confusing since simple tables with similar first columns, say, `[1, 2, 3]` and `[1, 1, 2]` will have different alignments.

2) In `tibble`, which we've tried to have partify with in notebook data tables, does not perform this alignment.

Sent also PR to `rmarkdown`: https://github.com/rstudio/rmarkdown/pull/843

### DIFF
--- a/src/cpp/session/modules/NotebookData.R
+++ b/src/cpp/session/modules/NotebookData.R
@@ -196,12 +196,6 @@
         paste0("<", summary, ">")
       })
 
-  if (length(columns) > 0) {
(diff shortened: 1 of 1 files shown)


jjallaire:
Merged on the rmarkdown side

PR #830, opened 2016-10-21
save + restore document selection

kevinushey:
NOTE: not quite ready for merge (see comments at end)

This PR attempts to save + restore the document selection for documents opened in the IDE.

---

The main snag to handle here -- we need to detect if the document has changed between the time the user saved the document with RStudio, and is opening it now; otherwise, we could end up attempting to restore a bogus cursor position. (separate comments from @jmcphers suggest that this likely isn't worth worrying about)

As an aside, the document position is currently only saved when the document is saved; we might instead want to do this somewhat more eagerly as currently navigation in a file without actually mutating the document won't save the cursor position. This may be fine, however.

We may also want to hide this behavior behind a UI preference, although I think this behavior will be preferred in most circumstances.


src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTarget.java
```diff
@@ -1295,6 +1295,21 @@ public void run()
...
+         final String rangeSpec = document.getProperties().getString("selection");
+         final Range range = Range.decode(rangeSpec);
+         new Timer()
+         {
+            @Override
+            public void run()
+            {
+               docDisplay_.setSelectionRange(range);
+            }
+         }.schedule(100);
```
jmcphers: It looks like this is going to stomp on the source position used for popin/popout since it runs on a timer that will fire after we've already set the cursor position there. It probably shouldn't do that since the popin/popout source position is more accurate. Can we defer to that position if it's available, or unify the mechanisms so they do not conflict? 

https://github.com/rstudio/rstudio/blob/0d7b3def3844265a4d807351b3f3f17bc489fbeb/src/gwt/src/org/rstudio/studio/client/workbench/views/source/Source.java#L843-L844

https://github.com/rstudio/rstudio/blob/0d7b3def3844265a4d807351b3f3f17bc489fbeb/src/gwt/src/org/rstudio/studio/client/workbench/views/source/Source.java#L1883

src/gwt/src/org/rstudio/studio/client/workbench/views/source/model/DocUpdateSentinel.java
```diff
@@ -327,12 +328,16 @@ private boolean doSave(String path,
          }
       }
       
-      // Update marks after document save
+      final boolean saved = didSave;
```
jmcphers: My main concern with doing this on save is that autosaves happen half a second after the edit, which is plenty of time to move your cursor. So the position we're going to save and restore is neither (a) where you made your last edit, nor (b) where you last left your cursor, but "somewhere near" your last edit, which may not feel predictable.

src/gwt/src/org/rstudio/studio/client/workbench/views/source/model/DocUpdateSentinel.java
```diff
@@ -343,9 +348,19 @@ public void execute()
...
             }
+            
+            // Update cursor selection position after document save
+            String oldRangeSpec = "";
+            if (hasProperty("selection"))
+               oldRangeSpec = getProperty("selection");
+            
+            String newRangeSpec = Range.encode(docDisplay_.getSelectionRange());
+            if (!oldRangeSpec.equals(newRangeSpec))
+               setProperty("selection", newRangeSpec);
```
jmcphers: This is going to effectively double the amount of disk I/O for most saves -- `setProperty` calls `modifyDocumentProperties` which eventually calls `source_database::put(...)` on the server, which writes out the whole doc contents to the source database. If we're going to use the save codepath, I think the fold spec would be a better model to follow than the vim marks.

jmcphers: see e.g.: https://github.com/rstudio/rstudio/blob/0d7b3def3844265a4d807351b3f3f17bc489fbeb/src/gwt/src/org/rstudio/studio/client/workbench/views/source/model/DocUpdateSentinel.java#L367-L372

kevinushey:
Alright, I've modified things here to match how folds are handled. Note that this does have one significant downside: cursor position, folds, and marks are all lost once the document is closed (they are not persistent properties) and so this might be surprising / frustrating for some users.

jmcphers:
I think that's actually very reasonable as it matches what most editors do after a document is closed -- do you know of any that save your cursor position when you close a document, and restore it when you re-open it?

kevinushey:
I guess you're right! I have Vim configured to do that, but I guess it's not a default action for most editors (beyond saving the current state of open files). (http://vim.wikia.com/wiki/Restore_cursor_to_file_position_in_previous_editing_session)

kevinushey:
Alright, given that caveat, I think we should be good to go with this PR. Note that the cursor restoration logic also detects more eager changes in the cursor position now as well, so the cursor position should be preserved as before with popped-out documents.

src/cpp/session/modules/SessionSource.cpp
```diff
@@ -289,6 +291,18 @@ Error saveDocumentCore(const std::string& contents,
...
+   bool hasMarksSpec = json::isType<std::string>(jsonMarksSpec);
+   if (hasMarksSpec)
+   {
+      pDoc->setMarks(jsonMarksSpec.get_str());
+   }
+
+   bool hasSelectionSpec = json::isType<std::string>(jsonSelectionSpec);
+   if (hasSelectionSpec)
+   {
+      pDoc->setSelection(jsonSelectionSpec.get_str());
```
jmcphers: Can we remove the code that sets and restores the `SourcePosition` when popping out/in now that it's done here?

src/gwt/src/org/rstudio/studio/client/workbench/views/source/model/DocUpdateSentinel.java
```diff
@@ -377,7 +370,9 @@ private boolean doSaveImpl(final String path,
       // Don't auto-save when there are no changes. In addition to being
       // wasteful, it causes the server to think the document is dirty.
       if (path == null && fileType == null && diff.isEmpty()
-          && foldSpec.equals(oldFoldSpec) 
+          && newFoldSpec.equals(oldFoldSpec)
+          && newSelectionSpec.equals(oldSelectionSpec)
```
jmcphers: If it turns out to be too heavyweight to trigger autosaves as the cursor moves, we could try triggering a cursor position save only when the page unloads and/or the tab loses focus -- this would introduce a couple of pathological situations (e.g. you'd lose position after a crash) but would also dramatically reduce the overhead.

jjallaire:
I'd like to inject one more thing here while we are actively reasoning about and modifying this code: currently, whenever you Knit or Run Document on an Rmd it updates the file modified time (even if the file has not been modified). This causes problems in 3 scenarios:

1. When editing with an external editor e.g. Emacs. @aronatkins Has reported this a couple of times.

2. There is a scenario for bookdown which I'm not quite familiar with (@yihui) where this touching of the file causes an extra Knit.

3. In `runtime: shiny_prerendered` we look at file modification times to determine whether we need to do a render. As things currently stand we *always* do a render for Run Document within the IDE whether we need to or not.

I took a look at this a couple of weeks ago and ran into some of the code we are dealing with now for the cursor persistence. I didn't feel confident enough to make a small change without unrolling all of the implications. Perhaps as part of this work we can.

jjallaire:
I recall many years ago attempting to persist scroll position (just during the normal course of autosave) and recall having some trouble with the restoration (moving the scroll position around during restore was interacting with document initialization and causing some exceptions). So I would say to be very paranoid and do lots of testing around restoration. @dfalty

src/cpp/session/modules/SessionSource.cpp
kevinushey: Good point; I believe we should be able to!

src/gwt/src/org/rstudio/studio/client/workbench/views/source/model/DocUpdateSentinel.java
jmcphers: It's not the overhead of the RPCs I'm worried about (they are indeed differential) but the disk I/O on the server. An autosave causes the entire document contents to be written to disk (even though only the diff travels over the wire). This change adds a lot of disk activity which is triggered by nothing other than the user moving their cursor around. It might not be a big deal (we already tolerate this while typing) but worth keeping an eye on.

jjallaire:
That's what I was thinking, but I didn't want to make that change myself
just in case I got something wrong and caused an unintended issue up or
downstream.

kevinushey:
I think the overall sense for this PR is that we should do the work to separate document properties and document contents into separate files first, to avoid over-eagerly updating potentially large files in the source database as triggered just by cursor movement. Does that sound right?

jjallaire:
Yes, I think that's exactly right.

Externalizing the files in the source database shouldn't be too, too
difficult. What I'd do is use the exact same "body" field but use a preface
+ sentinel token (e.g. a uuid) to point to an external file. That file of
course needs to have it's storage scope married to the lifetime of the
source database so needs to be right in the same directory (a specially
named subdirectory).  This would mean that if a user migrates to an older
version of the IDE they'll "lose" their file contents but I don't see any
way around this.

kevinushey:
Closing this in favor of a simpler incoming PR.

PR #838, opened 2016-10-25
add support for 'gvis' outputs in R Notebook

kevinushey:
NOTE: if we take this, we'll likely want to do the associated work in R Markdown to support this as `gvis` outputs are not properly rendered (by default) in their current form.

See https://github.com/rstudio/rstudio/pull/809 for some details on prior work.

TODO:
- [ ] Fix up vertical sizing behaviors (sometimes the widget opens too small, too large, etc.)
- [ ] Support `gvis` outputs in `rmarkdown` package (consider adding a `knit_print` method)

### DIFF
--- a/src/cpp/session/modules/NotebookHtmlWidgets.R
+++ b/src/cpp/session/modules/NotebookHtmlWidgets.R
@@ -50,6 +50,8 @@
       jsDrawChart[[injectionLoc]] <- paste(
          paste(
             "google.visualization.events.addListener(chart, 'ready', function() {",
+            "    if (typeof parent === 'undefined') return;",
+            "    if (typeof parent.$onGvisChartReady === 'undefined') return;",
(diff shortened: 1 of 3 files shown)


jjallaire:
This seems like a pretty aggressive change for the patch release (which will receive almost no external testing).

jjallaire:
Note also that this will only support static gvis outputs (as dynamic ones are embedded shiny apps). My point being that users will ultimately hit a wall attempting to use ggviz in notebooks so papering over the current problems, while useful, doesn't necessarily get us to a satisfactory endpoint.

kevinushey:
Right, the intention would be to take this in the future for v1.1 (assuming we want to do the associated work in `rmarkdown` as well). I'll rebase the PR to make that apparent.

jjallaire:
Okay, sounds good!

PR #841, opened 2016-10-25
Enable recording profiler state while ran as profvis command

javierluraschi:
Currently, when one runs:

```{r}
x <- profvis::profvis({
  profvis::pause(5)
})
```

The state of the ide does not change into "profiling" since the `Rprof` hook was added through `registerNotifyHook` which registers this hook in the environment, not the namespace (thanks @kevinushey for explaining this). So in order to track this state, we need to also get notifications for the times when `rprofvis` calls `Rprof` to start/stop profiling.

The bug that this fixes is a corner case tracked by fogbugz 5689, in which starting the profiler, then starting another profile with `profvis` and then stopping the profiler would cause us to try to read a file that was not created and a not very useful error to trigger as a pop up.

### DIFF
--- a/src/cpp/r/R/Tools.R
+++ b/src/cpp/r/R/Tools.R
@@ -397,7 +397,7 @@ assign(envir = .rs.Env, ".rs.getVar", function(name)
 
 
 # hook an internal R function
-.rs.addFunction( "registerHook", function(name, package, hookFactory)
+.rs.addFunction( "registerHook", function(name, package, hookFactory, namespace = FALSE)
 {
    # get the original function  
    packageName = paste("package:", package, sep="")
@@ -414,6 +414,13 @@ assign(envir = .rs.Env, ".rs.getVar", function(name)
       unlockBinding(name, packageEnv)
       assign(name, new, packageName)
       lockBinding(name, packageEnv)
+
+      if (namespace && package %in% loadedNamespaces()) {
+         ns <- asNamespace(package)
+         unlockBinding(name, ns)
+         assign(name, new, envir = ns)
+         lockBinding(name, ns)
+      }
    }
    else
    {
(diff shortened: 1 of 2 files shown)


src/cpp/r/R/Tools.R
```diff
@@ -414,6 +414,13 @@ assign(envir = .rs.Env, ".rs.getVar", function(name)
       unlockBinding(name, packageEnv)
       assign(name, new, packageName)
       lockBinding(name, packageEnv)
+
+      if (namespace && package %in% loadedNamespaces()) {
```
kevinushey: FWIW, calling `asNamespace()` will also explicitly load that package's namespace if it's not already loaded. I think we'd want to register our hooks unconditionally (ie, load the package if it's not loaded and register our hook), so I think we can remove the `package %in% loadedNamespaces()` check. Does that sound right?

javierluraschi: @kevinushey Thanks, from our slack chat, it does seem to be the case that `package %in% loadedNamespaces()` could be removed, in this case, since we are checking the `utils` package. However, to leave `registerHook` generic enough that others could use this functionality, we would still want to check the package is loaded since other packages might not be loaded by default, correct?

I don't fully understand the side effects from calling `asNamespace()`, in its documentation it mentioned that: `Internal namespace support functions. Not intended to be called directly, and only visible because of the special nature of the base namespace.`; however, this is used throughout the codebase.

I'm inclined to keep this as it is, but let me know if I might be missing something here.

src/cpp/r/R/Tools.R
```diff
@@ -414,6 +414,13 @@ assign(envir = .rs.Env, ".rs.getVar", function(name)
...
+         ns <- asNamespace(package)
```
javierluraschi: @jmcphers  I see. Yes, I guess a function could be defined in the package but removed from the environment, correct? PR updated.

PR #846, opened 2016-10-26
add support for attributes in Image preview

kevinushey:
NOTE: Not quite ready for merge; needs some more tweaking + testing of behaviors.

This PR implements support for HTML attributes in an image preview, e.g.

```
![](/an/image.png){width=300 height=200}
```

This helps ensure the way the image is previewed in the Notebook matches what is produced after a pandoc render.


kevinushey:
Okay, I think this PR is ready for review. We might want to hold this for v1.1 given that the PR does have some failure potential (the PR contains a somewhat ad-hoc HTML attributes parser) and given the short ramp-up time for v1.0-patch we might want to push it up in the interest of safety.

jjallaire:
+1 that we should move this to v1.1

PR #856, opened 2016-10-28
apply theme colors to html notebook chunks with retries to wait for load

javierluraschi:
Fixing two timing issues where html chunks had no color themes applied since the page loads after the call to apply themes. Fix applied to stream and gallery since it repros in both. With fix:

screen shot 2016-10-28 at 3 47 44 pm
screen shot 2016-10-28 at 3 47 56 pm

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/ChunkHtmlPage.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/ChunkHtmlPage.java
@@ -107,17 +107,43 @@ public class ChunkHtmlPage extends ChunkOutputPage
    public void onSelected()
    {
       // no action necessary for HTML widgets
+      updateHtmlPageTheme();
    }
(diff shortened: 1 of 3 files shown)


javierluraschi:
@jmcphers I believe we can't piggy back on `onFrameLoaded` without changing the default behavior since it's correctly called when the document finishes rendering; however, for `htmlwidgets`, this means that the `js` resources are loaded but does not mean that the code that generates the widget has run. In the case of `kable`, it actually changes the body tag once the `js` code runs after `onload` completes, so it overwrites the theme settings we set here and therefore, we either need to change when `onCompleted` is called, or add an extra timer to check for the widget to finish loading to apply resources. 

That said, there was a bit of code duplication by adding this new timer in the gallery and stream, refactored this timer and made it more general since other code might want to trigger code after the widget renders. Also, noticed that padding/margin was also being lost which is fixed now, and fixed an issue where the theme was not changing on the widget.

PR #858, opened 2016-10-31
Avoid nested restarts to prevent crashes during restart

javierluraschi:
Multiple subsequent restarts crash the IDE. While I spent some time trying to narrow this down, the code in https://github.com/rstudio/rstudio/blob/master/src/cpp/r/session/RSession.cpp#L1598-L1654 does not seem safe to reentrancy to me. The fact that we are saving the session state and allowing another overlapping restart seems dangerous enough to be blocked. Thoughts? Can anyone think of a case where a user would actually be interested in restarting during a restart that currently works as expected?

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/application/ApplicationQuit.java
+++ b/src/gwt/src/org/rstudio/studio/client/application/ApplicationQuit.java
@@ -67,6 +67,7 @@ import com.google.gwt.core.client.Scheduler;
 import com.google.gwt.core.client.Scheduler.RepeatingCommand;
 import com.google.gwt.resources.client.ImageResource;
 import com.google.gwt.user.client.Command;
+import com.google.gwt.user.client.Timer;
 import com.google.inject.Inject;
 import com.google.inject.Provider;
 import com.google.inject.Singleton;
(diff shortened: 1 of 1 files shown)


kevinushey:
One thing to consider (albeit unlikely) -- if a JavaScript exception were to get thrown during an attempt to suspend, it's possible the `suspend` flag here could get stuck 'on' and the user would no longer be able to restart at all.

Out of an abundance of caution, we might want to wrap the underlying implementation in a `try-finally` block?

jjallaire:
I could be missing something but I'm not sure what we would wrap in
try/finally (as I don't think any of the code here will throw unless there
is a bug in our implementation). We generally assume that code doesn't
throw (unless there is a bug) and errors are reported via the async
callback. I'd therefore prefer not to add more error handling logic as it
would just complicated the code further.

LGTM.

kevinushey:
LGTM as well (sorry; my paranoia is likely unwarranted on this code path!)

javierluraschi:
I agree, I don't think the try/catch block would help us here, I did spent some time looking at the code paths to see if we could be left in a bad state, which from the code, seems unlikely.

That said, while looking at the code I did notice that we stop the progress indicator before we get the restart ping, so I've refactored this code to match progress and restart-prevention together which would be the correct progress indication in cases where a restart takes too long.

jjallaire:
LGTM!

PR #860, opened 2016-11-01
null check for notebook html chunk themes fix

javierluraschi:
A null-check and nit cr comment for https://github.com/rstudio/rstudio/pull/856

/cc @kevinushey


kevinushey:
The exception log I saw even included an entry where the attempt to retrieve the document with `getDocument()` failed. Is it possible that this `runAfterRender()` command is being called before `onFrameLoaded()` has been called?

Maybe `onFrameLoaded()` should set a `loaded_` flag, and `runAfterRender()` should listen to that?

Note that we're pretty conservative in our checks for loadedness in the `DynamicIFrame` superclass: https://github.com/rstudio/rstudio/blob/d83043b9e026d742e53d43533a5ec5ea24719625/src/gwt/src/org/rstudio/core/client/widget/DynamicIFrame.java#L56-L57

javierluraschi:
@kevinushey since `getDocument()` wasn't even available, then yes!

PR #862, opened 2016-11-01
highlight '?' as operator in R mode

kevinushey:
This PR ensures that `?` is highlighted as an operator in R modes.

(let me know if not highlighting `?` was actually intentional or if it was just an oversight!)


jjallaire:
Good catch!

PR #865, opened 2016-11-02
remove dependency to tibble

javierluraschi:
Consider removing dependency to `tibble` since `tibble` currently requires `3.1.2`.

**Dependent**: Dependent on https://github.com/rstudio/rmarkdown/pull/853

### DIFF
--- a/src/cpp/session/modules/NotebookData.R
+++ b/src/cpp/session/modules/NotebookData.R
@@ -143,6 +143,86 @@
 
 .rs.addFunction("readDataCapture", function(path)
 {
+  type_sum <- function(x) {
+    format_sum <- switch (class(x)[[1]],
+                          ordered = "ord",
+                          factor = "fctr",
+                          POSIXt = "dttm",
+                          difftime = "time",
+                          Date = date,
+                          data.frame = class(x)[[1]],
+                          tbl_df = "tibble",
+                          NULL
+    )
+    if (!is.null(format_sum)) {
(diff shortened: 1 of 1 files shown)


src/cpp/session/modules/NotebookData.R
```diff
@@ -143,6 +175,53 @@
 
 .rs.addFunction("readDataCapture", function(path)
 {
+  "%||%" <- function(x, y) {
```
jjallaire: Is there a way we could factor the rmarkdown code so that we could call this code within the rmarkdown package? (since it's duplicated there). Note that it's fine for the IDE to call private methods of packages since we aren't subjected to CRAN checks.

javierluraschi: Makes sense, using private code form rmarkdown now.

kevinushey:
Do we need an associated version bump (to ensure that new builds of the IDE use the version of `rmarkdown` exporting these functions)?

jjallaire:
My bad -- I just realized that calling code within rmarkdown might yield the same problem we had with pagedtable.js (unable to render after an IDE update). So do we need to make a copy of this code after all?

javierluraschi:
@jjallaire hard to tell, put to be sure, I agree that duplicating this code is the safest path. Changed.

javierluraschi:
@jmcphers rmarkdown PR merged, this one can now be merged as well.

PR #866, opened 2016-11-02
don't save + restore workspace for 'install.packages'

kevinushey:
This PR seeks to resolve a loop that can occur when attempting to install a package that is currently loaded.

The overarching issue -- when RStudio restarts the session when installing a package, it also saves and restores the global environment. However, certain objects can force the load of a package namespace -- if the namespace of the package which we want to install is inadvertently loaded while restoring the global environment, then RStudio will present us with the same prompt again.

install-loop

This PR seeks to resolve the issue by passing along the `saveWorkspace` argument through the regular `save` call. Previously, this argument was ignored (from the comments before it appears this was intentional, although I don't know why exactly); so the change here just makes sure that we respect this option in the regular `save` code path.

One major change with this PR is that we now actually discard the global environment when restarting for a package installation -- we could also defer restoration of the global environment instead until after the `install.packages()` command has run to completion.

### DIFF
--- a/src/cpp/r/session/RSearchPath.cpp
+++ b/src/cpp/r/session/RSearchPath.cpp
@@ -226,17 +226,20 @@ void attachEnvironmentData(const FilePath& dataFilePath,
 } // anonymous namespace
    
 
-Error save(const FilePath& statePath)
+Error save(const FilePath& statePath, bool saveWorkspace)
 {
-   // save the global environment
-   FilePath environmentFile = statePath.complete(kEnvironmentFile);
-   Error error = saveGlobalEnvironmentToFile(environmentFile);
-   if (error)
-      return error;
+   // save the global environment if requested
+   if (saveWorkspace)
+   {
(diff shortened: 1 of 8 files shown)


kevinushey:
It's also possible that rather than doing a 'full' save we just want to do a 'minimal' save when installing a package (thereby making most of the changes in this PR unnecessary; ie, we could just use the `saveMinimal` codepath)

jjallaire:
I don't think we can do this without a bunch of hand waving around the fact that the global environment is going to get tossed out (as for many users this still represents the saved state of their work). I'd rather have the occasional looping dialog than discard their work.

Let's discuss realtime before we pursue this further.

kevinushey:
I think this PR is definitely out of scope for v1.0-patch at this point.

An alternative we might consider is simply deferring the restoration of the workspace until after the `install.packages()` call is invoked; however, that may be trickier than it looks to get right. I'm going to close this for now and we can revisit for v1.1 in the future.

jjallaire:
Sounds good!

PR #879, opened 2016-11-07
fix backslash escaping in 'Test Package'

kevinushey:
This PR fixes an issue when attempting to run `Test Package` in R packages which are not using `devtools` / `testthat` for unit tests. Note that the problem only came to light with the fix in https://github.com/rstudio/rstudio/commit/1b3228a89cfbe565a5444b16a0493f5229bc756.

Ironically, with the previous 'broken' escape functionality, we were sending code to the shell like:

```
list.files(pattern = '$1$1.R$1')
```

which was parsed by the shell fine and handled by R, although did not actually filter the set of files correctly. After this fix, we were constructing the pattern as e.g.

```
"list.files(pattern = '\\\\.R\$')"
```

However, we need to double-escape the backslashes here, as otherwise R will only see

```
> list.files(pattern = '\.R$')
```

and that will produce an error like:

```
==> Sourcing R files in 'tests' directory

Error: '\.' is an unrecognized escape in character string starting "'\."
Execution halted

Exited with status 1.
```

### DIFF
--- a/src/cpp/core/system/PosixShellUtils.cpp
+++ b/src/cpp/core/system/PosixShellUtils.cpp
@@ -23,8 +23,8 @@ namespace shell_utils {
 
 std::string escape(const std::string& arg)
 {
-   boost::regex pattern("(\\\\|\\$|`|!|\n|\")");
-   return "\"" + boost::regex_replace(arg, pattern, "\\\\$1") + "\"";
+   boost::regex pattern("[\\\\$`!\n\"]");
+   return "\"" + boost::regex_replace(arg, pattern, "\\\\$0") + "\"";
 }
 
 std::string escape(const core::FilePath &path)
--- a/src/cpp/core/system/PosixShellUtilsTests.cpp
+++ b/src/cpp/core/system/PosixShellUtilsTests.cpp
@@ -37,6 +37,14 @@ context("Shell Escaping")
       std::string expected = "\"\\$\\$\\$\"";
       expect_true(escaped == expected);
    }
+   
+   test_that("Commands with backslashes are escaped")
+   {
+      std::string backslashes = "\\\\\\";
+      std::string escaped = escape(backslashes);
+      std::string expected = "\"\\\\\\\\\\\\\"";
+      expect_true(escaped == expected);
+   }
 }
 
 } // end namespace tests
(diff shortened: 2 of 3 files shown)


src/cpp/core/system/PosixShellUtils.cpp
```diff
@@ -23,8 +23,8 @@ namespace shell_utils {
 
 std::string escape(const std::string& arg)
 {
-   boost::regex pattern("(\\\\|\\$|`|!|\n|\")");
-   return "\"" + boost::regex_replace(arg, pattern, "\\\\$1") + "\"";
+   boost::regex pattern("[\\\\$`!\n\"]");
```
kevinushey: The changes here just tweak the pattern to use the original `[]`-enclosed characters as per @jcheng5's initial comment.

kevinushey:
I should double-check that this doesn't regress on Windows.

jjallaire:
I was going to make the same comment when the original change came through.
Should we move both of these to v1.1 to be on the safe side?

kevinushey:
This PR actually fixes a bug that currently only exists on the patch-release branch (ie master); that bug now being present with https://github.com/rstudio/rstudio/commit/1b3228a89cfbe565a5444b16a0493f5229bc756.

Note that the changes in the `regex_replace` bit effectively bring back the old form of the pattern, but fix the replacement string (ie, `\\$1` to `\\\\$0`). (sorry for making this so confusing!)

jmcphers:
IIRC https://github.com/rstudio/rstudio/commit/1b3228a89cfbe565a5444b16a0493f5229bc756 allows Git staging of filenames with special characters such as `$`. If that's the full scope of the fix then I'd support rolling it back on the patch release (as it's a long-standing bug affecting few users), and moving both that commit and this PR to 1.1.

jjallaire:
Yes, I agree. Let's put both fixes in v1.1. Kevin, if you want to fix the
git staging issue I'd just use a local version of escape than add a
reminder to migrate to the global once once we are on v1.1

kevinushey:
Okay, I'll revert https://github.com/rstudio/rstudio/commit/1b3228a89cfbe565a5444b16a0493f5229bc756 and reapply on v1.1.

PR #881, opened 2016-11-08
Initial work on xterm-based terminal tab

gtritchie:
Initial chunk of functionality for the improved terminal support in RStudio v1.1.

This work is hiding behind a feature flag, execute `.rs.writeUiPref("enable_xterm", TRUE)` in Console and refresh browser. This displays a new menu command Tools->Shell Tab, which shows a new tab alongside the Console, and launches an xterm-compatible shell session. You can now enter shell commands, including ncurses-based things like vim and tmux, and paging of results with less, and man pages, etc. Color is also supported.

Due to persistence not being wired up yet (see list of known work below), if you refresh the browser or reattach after showing the terminal tab, it will reconnect it to the existing modal shell dialog and look very odd.

There are a number of work items remaining (in addition, of course, to bug fixing). Listed below, will elaborate further in an update to the dev document. Expect these to happen in more discrete check-in chunks, now that I've got the basics in place.

- As-needed loading of xterm.js and its components; currently have an issue (and a TODO in the code) where I had to load it very early (at same place Ace gets loaded), but we don't want to incur that cost up-front.
- MS-Windows desktop support (exact behavior TBD, but currently not implemented at all)
- Always show terminal tab next to console, with no close button
- Lazy-loading of terminal; doesn't start the remote process until user clicks on tab to make it visible
- Mouse selection of terminal contents to enable Copy
- Browser refresh support (reattach, re-populate terminal buffer)
- Session persistence and resume support
- Termination handling when user terminates shell via `exit`, etc.
- Minify xterm.js and its support files (build-time work)
- Review and resolve various TODO's I have put in the code, followed by (gary) so I can find them easily in Eclipse task list
- Performance evaluation of current RPC technique for sending keystrokes and obtaining echo'd output, especially in browser-based scenarios
- Prevent session from sleeping when terminal has active child process
- Possibly remove the existing Tools/Shell menu for showing the modal dialog
- Ensure there is a way to get focus to the terminal tab via keyboard (new View/Move Focus menu item for that?)
- Support multiple terminal sessions, possibly within a single Terminal tab (thinking a toolbar, with ability to switch to a session, add new sessions, close a session)
- Consider using the xterm.js fullscreen add-on to allow temporary expansion of terminal to fill entire browser.
- Theming; might have a fixed dark-background color scheme in v1.1, or could look at syncing with Ace themes.


src/cpp/core/include/core/system/Process.hpp
```diff
@@ -59,14 +59,19 @@ struct ProcessOptions
    ProcessOptions()
 #ifdef _WIN32
       : terminateChildren(false),
```
jjallaire: Do we need to initialize `cols` and `rows` in the `_WIN32` codepath as well?

src/gwt/src/org/rstudio/studio/client/workbench/views/terminal/TerminalPane.java
```diff
@@ -0,0 +1,256 @@
...
+import org.rstudio.studio.client.workbench.views.terminal.events.TerminalDataInputEvent;
+
+import com.google.gwt.event.dom.client.ClickEvent;
+import com.google.gwt.event.dom.client.ClickHandler;
+import com.google.gwt.event.shared.HandlerRegistration;
+import com.google.gwt.user.client.Command;
+import com.google.gwt.user.client.ui.ResizeLayoutPanel;
+import com.google.gwt.user.client.ui.Widget;
+
+public class TerminalPane extends WorkbenchPane
```
jjallaire: Is the terminal a side-by-side tab with the console? (that's what I was thinking we'd do)

gtritchie:
Yes, it's side by side. At the moment it isn't initially displayed and can be closed but will be changing that (see my list of planned work items in the pull request notes).

Get Outlook for iOS




On Tue, Nov 8, 2016 at 7:41 AM -0800, "JJ Allaire" <notifications@github.com> wrote:












@jjallaire commented on this pull request.





In src/gwt/src/org/rstudio/studio/client/workbench/views/terminal/TerminalPane.java:
> +import org.rstudio.studio.client.server.ServerRequestCallback;
+import org.rstudio.studio.client.server.Void;
+import org.rstudio.studio.client.server.VoidServerRequestCallback;
+import org.rstudio.studio.client.workbench.model.WorkbenchServerOperations;
+import org.rstudio.studio.client.workbench.ui.WorkbenchPane;
+import org.rstudio.studio.client.workbench.views.terminal.events.ResizeTerminalEvent;
+import org.rstudio.studio.client.workbench.views.terminal.events.TerminalDataInputEvent;
+
+import com.google.gwt.event.dom.client.ClickEvent;
+import com.google.gwt.event.dom.client.ClickHandler;
+import com.google.gwt.event.shared.HandlerRegistration;
+import com.google.gwt.user.client.Command;
+import com.google.gwt.user.client.ui.ResizeLayoutPanel;
+import com.google.gwt.user.client.ui.Widget;
+
+public class TerminalPane extends WorkbenchPane



Is the terminal a side-by-side tab with the console? (that's what I was thinking we'd do)

gtritchie:
Yes, good idea.

Get Outlook for iOS




On Tue, Nov 8, 2016 at 7:36 AM -0800, "JJ Allaire" <notifications@github.com> wrote:












@jjallaire commented on this pull request.





In src/cpp/session/include/session/SessionConsoleProcess.hpp:
> @@ -173,6 +174,11 @@ class ConsoleProcess : boost::noncopyable,
 
    // Whether the process should be stopped
    bool interrupt_;
+   
+   // Whether the tty should be notified of a resize
+   bool resize_;



Could you get rid of the resize_ flag by just using sentinel values in newCols_ and newRows_? (i.e. resize whenever they are not == -1)

gtritchie:
Yes, good catch. Haven't fleshed out the win32 code yet.

Get Outlook for iOS




On Tue, Nov 8, 2016 at 7:00 AM -0800, "JJ Allaire" <notifications@github.com> wrote:












@jjallaire commented on this pull request.





In src/cpp/core/include/core/system/Process.hpp:
> @@ -59,14 +59,19 @@ struct ProcessOptions
    ProcessOptions()
 #ifdef _WIN32
       : terminateChildren(false),



Do we need to initialize cols and rows in the _WIN32 codepath as well?

src/gwt/src/org/rstudio/studio/client/workbench/views/terminal/TerminalPane.java
```diff
@@ -0,0 +1,256 @@
...
+      {
+         @Override
+         public void onResponseReceived(ConsoleProcess consoleProcess)
+         {
+            consoleProcess_ = consoleProcess;
+            cryptoServer_ = server_;
+            
+            if (getInteractionMode() != ConsoleProcessInfo.INTERACTION_ALWAYS)
+            {
+               throw new IllegalArgumentException("Unsupport ConsoleProcess interaction mode");
```
jmcphers: Where is this likely to be caught? If we fail to launch the terminal for some reason we still need to respond to the user's gesture in some way (even if it's just to show a dialog or error bar that says "sorry, this didn't work").

src/cpp/core/include/core/system/Process.hpp
gtritchie: @jjallaire no change needed until I do the win32 work; the cols and rows fields are only in the non-Win32 build

src/gwt/src/org/rstudio/studio/client/workbench/views/terminal/XTermWidget.java
```diff
@@ -0,0 +1,286 @@
...
+import com.google.gwt.user.client.ui.Widget;
+
+/**
+ * Xterm-compatible terminal emulator
+ */
+public class XTermWidget extends Widget implements RequiresResize,
+                                                   ResizeTerminalEvent.HasHandlers,
+                                                   TerminalDataInputEvent.HasHandlers
+{
+   public enum AnsiColor
```
jmcphers: Oh the memories! ☎️

src/gwt/src/org/rstudio/studio/client/workbench/views/terminal/events/CreateTerminalEvent.java
```diff
@@ -0,0 +1,50 @@
...
+
+package org.rstudio.studio.client.workbench.views.terminal.events;
+
+import com.google.gwt.event.shared.EventHandler;
+import com.google.gwt.event.shared.GwtEvent;
+
+/**
+ * Event to trigger creation of a terminal session.
+ */
+public class CreateTerminalEvent extends GwtEvent<CreateTerminalEvent.Handler>
```
jmcphers: This should be a `CrossWindowEvent` since only the main window is going to handle it (but it could be invoked via menus or shortcuts in other windows). For now all that's needed here is to extend `CrossWindowEvent` rather than `GwtEvent`.

src/gwt/src/org/rstudio/studio/client/workbench/views/terminal/xterm/XTermNative.java
```diff
@@ -0,0 +1,123 @@
...
+
+   public final native void onData(CommandWithArg<String> command) /*-{
+      this.handler = 
+         $entry(function(data) {
+            command.@org.rstudio.core.client.CommandWithArg::execute(Ljava/lang/Object;)(data);
+         });
+   }-*/;
+   
+   public final native void onTerminalData(CommandWithArg<String> command) /*-{
+      this.handler =
```
jmcphers: What's the difference between `onData` and `onTerminalData` (their bodies look the same)? Also, is it possible to stack these? (i.e. can you ever have multiple handlers?)

src/gwt/src/org/rstudio/studio/client/workbench/views/terminal/events/CreateTerminalEvent.java
gtritchie: Done.

src/gwt/src/org/rstudio/studio/client/workbench/views/terminal/xterm/XTermNative.java
gtritchie: Yes, redundant method left over from some refactoring. Removing it.

jmcphers:
There are a lot of TODOs still in the code -- let's open a blanket bug against v1.1 to make sure we address all of those.

PR #891, opened 2016-11-10
refactor addins indexer to provide extensible package resource indexer

kevinushey:
NOTE: not quite ready for merge; needs some testing to ensure nothing is regressed on the addins side; we might also want to tweak the overall implementation as well depending on whether we're happy with the current separation of concerns.

---

This PR attempts to refactor the indexer we use for R addins into a more general base class that should be useful for handling of project templates provided by R packages.

Most of the machinery used in the R addins indexer has been moved to a more general class, which is tentatively called the 'package provided extension' indexer (or `ppe::Indexer` for short). The general idea is that users will subclass this indexer, implementing a couple of virtual methods to control how indexing is handled, and the classes will define a method that describes how to rebuild those objects from a file provided by an R package.

### DIFF
--- a/src/cpp/session/include/session/SessionPackageProvidedExtension.hpp
+++ b/src/cpp/session/include/session/SessionPackageProvidedExtension.hpp
@@ -39,6 +39,7 @@ class Indexer : public boost::noncopyable
    virtual void onIndexingStarted() = 0;
    virtual void onWork(const std::string& pkgName, const core::FilePath& resourcePath) = 0;
    virtual void onIndexingCompleted() = 0;
+   virtual void ~Indexer() {}
    
 public:
    explicit Indexer(const std::string& resourcePath)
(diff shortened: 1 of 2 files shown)


kevinushey:
I think this PR should be ready for review; @jjallaire, do you want to sanity check?

jjallaire:
Yup, will do tomorrow or first thing Monday.

src/cpp/session/include/session/SessionPackageProvidedExtension.hpp
```diff
@@ -0,0 +1,156 @@
...
+#include <session/SessionModuleContext.hpp>
+
+namespace rstudio {
+namespace session {
+namespace modules {
+namespace ppe {
+
+class Indexer : public boost::noncopyable
+{
+   virtual void onIndexingStarted() = 0;
```
jjallaire: You'd want a virtual destructor here as well

kevinushey: Good catch; thanks!

jjallaire:
I think this looks good. Merge when ready!

PR #897, opened 2016-11-14
respect use of 'output.var' in chunk header

kevinushey:
This PR resolves an issue where attempts to write a Stan chunk header of the following form could fail:

    ```{stan, output.var="output"}

Previously, we had assumed the `output.var` field would be specified as part of the `engine.opts` chunk option; it appears that `knitr` now also allows for `output.var` at the 'top level' for the chunk options. It looks like this behavior is also documented at http://rmarkdown.rstudio.com/authoring_knitr_engines.html#stan.

### DIFF
--- a/src/cpp/session/modules/rmarkdown/NotebookAlternateEngines.cpp
+++ b/src/cpp/session/modules/rmarkdown/NotebookAlternateEngines.cpp
@@ -227,7 +227,7 @@ Error executeStanEngineChunk(const std::string& docId,
    
    // write code to file
    FilePath tempFile = module_context::tempFile("stan-", ".stan");
-   error = writeStringToFile(tempFile, code);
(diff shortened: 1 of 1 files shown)


src/cpp/session/modules/rmarkdown/NotebookAlternateEngines.cpp
```diff
@@ -227,7 +227,7 @@ Error executeStanEngineChunk(const std::string& docId,
    
    // write code to file
    FilePath tempFile = module_context::tempFile("stan-", ".stan");
-   error = writeStringToFile(tempFile, code);
+   error = writeStringToFile(tempFile, code + "\n");
```
kevinushey: The extra newline here is added to avoid errors that can occur during a call to `readLines(..., warn = TRUE)` -- I didn't hunt down where exactly that call was coming from however.

PR #907, opened 2016-11-21
Install rmarkdown and rsconnect from CRAN

jjallaire:
NOTE: We shouldn't take this PR until after both packages are accepted to CRAN (rmarkdown v1.2 has been accepted but rsconnect v0.6 is still pending).

This change is only for the patch branch (it shouldn't also be merged into master b/c master is by convention using the development version of both packages).


jjallaire:
rsconnect is now on CRAN so I just merged

PR #921, opened 2016-12-07
Generalize new connections dialog to use shiny app

javierluraschi:
I believe this is ready for merge, with a few caveats worth noting:

 1. This will break `sparklyr` create new connections. The plan would be to push to CRAN `sparklyr` next week and then merge this [branch](https://github.com/rstudio/sparklyr/commits/feature/create-connection-ui) back to master. Older versions of sparklyr will remain broken with the following error. Once `sparklyr` releases to CRAN, we can recommend users to install the devel branch of `sparklyr`:

screen shot 2016-12-07 at 2 52 15 pm

(( Just changed the dialog to read "Please upgrade to sparklyr 0.5.1 or newer." since "0.5" will be the CRAN version which won't have support for this yet. ))

 2. There is more work to do in `sparklyr` to handle displaying `hadoop_versions` in sync with `spark_versions`, warnings, allow to select master, etc. However, those changes don't need to be performed with the IDE, necessarily.

In order to get the ui back today, one can install `sparklyr` from this branch:

```{r}
devtools::install_github("rstudio/sparklyr", ref = "feature/create-connection-ui")
```

Then we should get this functionality back as expected but running as a shiny app:

screen shot 2016-12-07 at 3 46 05 pm


src/cpp/session/modules/SessionConnections.R
```diff
@@ -125,53 +125,50 @@ options(connectionViewer = list(
...
+   if (package == "sparklyr" & packageVersion("sparklyr") < "0.5") {
+      return(.rs.error(
+         "sparklyr ", packageVersion("sparklyr"), " does not support this functionality. ",
+         "Please upgrade to sparklyr 0.5.1 or newer."
+      ))
+   }
+
+   consoleCommand <- quote(.rs.launchEmbeddedShinyConnectionUI(package = package))
+
+   .rs.enqueClientEvent("send_to_console", list(
```
jmcphers: Presumably this isn't the end game as we don't want to expose this command to the user. I'd consider using a separate R session for this, it's a little more overhead but frees you from having to worry about polluting (or being polluted by) the user's state. `AsyncRProcess` can handle this. Alternately, you could emit this to the console with some decoration that prevents it from being echoed or committed to history.

src/gwt/src/org/rstudio/studio/client/workbench/views/connections/ConnectionsPresenter.java
```diff
@@ -248,123 +247,61 @@ public void onActiveConnectionsChanged(ActiveConnectionsChangedEvent event)
    {
       updateActiveConnections(event.getActiveConnections());
    }
+
+   private void terminateShinyApp()
```
jmcphers: Interrupting R is asynchronous and might not succeed -- In the context used here (interrupt, then perform an action) I think `ApplicationInterrupt.interruptR` is what you need.

src/cpp/session/modules/SessionConnections.R
jmcphers: (Okay with this as-is for now, just thinking about how we want this to work in the final state)

src/gwt/src/org/rstudio/studio/client/workbench/views/connections/ConnectionsPresenter.java
javierluraschi: I have a mental todo on this one, @jmcphers and I talked about this, will merge and address this couple comments afterwards to start working in the same branch.

jjallaire:
We do indeed run everything at the console for interrupt-ability. We
probably just need to add a nicer looking wrapper function so the user sees
something pleasant and high level at the console.

src/cpp/session/modules/SessionConnections.R
```diff
@@ -125,53 +125,50 @@ options(connectionViewer = list(
...
+      "code" = .rs.scalar(deparse(consoleCommand)),
```
kevinushey: `deparse()` can return a vector of length > 1 if the command is too long; you might want something like `paste(deparse(consoleCommand), collapse = " ")`. We might also want to put this in a helper function, e.g. `.rs.deparse`.

javierluraschi: Thanks Kevin, changed.

src/cpp/session/modules/SessionConnections.R
javierluraschi: @jmcphers You have a point that the `launchEmbeddedShinyConnectionUI` going to console is confusing. Found a better solution by sending to console the actual package command, as in:

screen shot 2016-12-08 at 5 41 46 pm

In this case, I think is very informative for users to understand that we are running something outside the IDE and if the UI breaks, they would have a better context from where the error came from and where to ask for help.

Already pushed this change to `master`.

javierluraschi:
Thanks JJ, yes, changed this to a nicer looking call.

PR #927, opened 2016-12-14
Implement new connection wizard

javierluraschi:
This PR implements the core functionality for a new connections wizards. Fully functional for `sparklyr` under https://github.com/rstudio/sparklyr/tree/feature/create-connection-ui branch.

screen shot 2016-12-13 at 5 59 52 pm

screen shot 2016-12-13 at 6 00 00 pm

screen shot 2016-12-13 at 4 05 02 pm


jjallaire:
Presuming that the UX that provides the list of connection types will end up looking more like Kevin's new list of project types, correct? In general I think this listing benefits from density in both cases so we can continue tweaking the presentation until we get enough items above the fold in the initial display.

javierluraschi:
@jjallaire correct! I didn't spend much time styling this since I don't have a realistic way of adding many connections yet. But I'll style this list to look more like Kevins (thinner) in subsequent iterations.

PR #933, opened 2016-12-23
support session suspend when terminals are idle

gtritchie:
Suspend
---------
If no terminal sessions have subprocesses (i.e. no subprocesses of bash) allow:

- session to suspend
- terminal tab to be closed without warning

This is a Mac-only (dev environment) implementation. Linux implementation will be done separately.

Other Fixes
-----------

- Show an asterisk after the terminal name in the dropdown if that terminal has subprocesses
- Trim terminal names before showing in dropdown
- Scroll terminal buffer to bottom before displaying new input
- Support pasting in text longer than 117 characters by chunking it up
- Fix ability to enter passwords on Linux
- switch terminal from xterm to xterm-256color
- implement command for clearing scrollback buffer


src/cpp/core/system/PosixChildProcess.cpp
```diff
@@ -660,14 +665,81 @@ struct AsyncChildProcess::AsyncImpl
...
+   }
+
+   bool computeHasSubProcess(pid_t pid)
+   {
+#ifdef _WIN32
+      return true;
+#else
+      // pgrep -P ppid returns 0 if there are matches, non-zero
+      // otherwise
+      shell_utils::ShellCommand cmd("pgrep");
```
jmcphers: It should be possible to have the kernel deliver notifications to us when the shell process forks or execs, so we can keep this updated in real time instead of polling. See e.g.:

http://netsplit.com/the-proc-connector-and-socket-filters

However it doesn't seem like this is very straightforward (and may wind up being somewhat non-portable and complicated) so maybe just worth filing a note to revisit. If we don't wind up with an event-driven approach we  also might be able to find a way to query for child processes using system calls rather than spawning a whole new process to count.

gtritchie:
Sounds good. Yes, having to run a command feels icky. 

On Linux, I could parse contents of /proc/pid/tasks/pid/children (or something like that, just off the top of my head), and it contains a list of the direct child process ids. Then I’d leave the existing code as a Mac-only dev environment workaround.

Still not a system call, but opening and determining if this virtual file has > 0 bytes in it should be more lightweight.

src/gwt/src/org/rstudio/studio/client/workbench/views/terminal/TerminalPopupMenu.java
```diff
@@ -127,7 +127,7 @@ public void setActiveTerminal(String caption, String handle)
    {
       activeTerminalCaption_ = caption;
       activeTerminalHandle_ = handle;
-      toolbarButton_.setText(activeTerminalCaption_);
+      toolbarButton_.setText(trimCaption(activeTerminalCaption_));
```
jmcphers: Typically we do this in CSS via `text-overflow` which allows the limit to be specified in pixels rather than characters and has some slightly better typographic treatment for the ellipsis

PR #934, opened 2017-01-03
unification of package extension code

kevinushey:
This PR seeks to unify the code driving the various indexers into a single indexer object. The general idea is now that:

1. We have a single indexer object that is responsible for discovering installed packages + extensions within,

2. 'Workers' can be registered on this indexer object that, given a package directory, do some 'work' (ie -- reading + parsing a specific type of `.dcf` file).

The overall outcome -- rather than having multiple indexer objects individually crawling the R library for package extensions, we now have a single indexer crawling the R library, and this indexer delegates to the registered workers.

The code handling registries has not yet been unified; this is less of a priority since the main goal of this PR is to avoid having multiple indexers running in tandem.

### DIFF
--- a/src/cpp/session/modules/SessionPackageProvidedExtension.cpp
+++ b/src/cpp/session/modules/SessionPackageProvidedExtension.cpp
@@ -201,7 +201,7 @@ void onConsoleInput(const std::string& input)
    if (module_context::disablePackages())
       return;
    
-   const char* const commands[] = {
+   static const char* const commands[] = {
       "install.packages",
       "utils::install.packages",
       "remove.packages",
--- a/src/cpp/session/modules/SessionPackageProvidedExtension.cpp
+++ b/src/cpp/session/modules/SessionPackageProvidedExtension.cpp
@@ -119,14 +119,21 @@ void Indexer::beginIndexing()
 void Indexer::endIndexing()
 {
    running_ = false;
+   
+   json::Object payload;
    BOOST_FOREACH(boost::shared_ptr<Worker> pWorker, workers_)
    {
       try
       {
-         pWorker->onIndexingCompleted();
+         pWorker->onIndexingCompleted(&payload);
       }
       CATCH_UNEXPECTED_EXCEPTION
    }
+   
+   ClientEvent event(
+            client_events::kPackageExtensionIndexingCompleted,
+            payload);
+   module_context::enqueClientEvent(event);
 }
 
 namespace {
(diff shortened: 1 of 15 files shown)


kevinushey:
I think this PR should be okay to review; @jmcphers, let me know if you have a chance to look it over.

src/cpp/session/modules/SessionPackageProvidedExtension.cpp
```diff
@@ -0,0 +1,250 @@
...
+   }
+   
+   return Success();
+}
+
+Indexer::Indexer() : index_(0), n_(0), running_(false) {}
+
+void Indexer::addWorker(boost::shared_ptr<Worker> pWorker)
+{
+   workers_.push_back(pWorker);
```
jmcphers: Should this accept the worker if it's already started indexing?

kevinushey: Probably not, but clients should generally be registering their workers quite early in the startup process (before deferred init, and hence before the indexer ever gets a chance to run) so in practice that shouldn't be an issue.

jjallaire:
There's one more type of addin we should create a handler for: currently R Markdown templates are scanned for on every invocation of the **New R Markdown** dialog. We should replace this with an indexer based implementation. @jmcphers will have the details as he implemented the original R Markdown template scanning code.

jmcphers:
That's a great idea. The template scanning code is currently implemented as an async R process; it lives here:

https://github.com/rstudio/rstudio/blob/d833ee983876575bde6cd61719e1d4c0f69d4b9e/src/cpp/session/modules/rmarkdown/SessionRMarkdown.cpp#L674-L689

kevinushey:
I spent a couple hours here this morning -- the server-side changes are relatively straightforward; however, the client-side code seems a bit trickier since most of it's built around retrieving the set of available templates on demand (ie the discovery is initiated only within the New RMarkdown dialog), and the handlers used for receiving changes in the indexer state are only available + attached while that dialog is present.

I'll file an issue in fogbugz for this; @jmcphers you might have a clearer picture on how to handle this.

jjallaire:
Yeah, I think a ton of code will need to get ripped out of the client side
(it was previously managing async streaming of the templates). @jmcphers
may indeed be better suited to do this safely.

jmcphers:
Happy to tackle this (@kevinushey opened a bug so it's tracked against 1.1). :-)

PR #937, opened 2017-01-04
detect bookdown website in precedence to package

kevinushey:
This PR ensures that `bookdown` skeletons (as generated through the associated skeleton function in https://github.com/rstudio/bookdown/pull/280) use the 'Website'-style build UI.

Currently, we check for a `DESCRIPTION` file first, and because the `bookdown` skeleton comes with such a file (mainly for dependency installation?) RStudio assumes the generated project is an R package project.

The diff looks weird but all I'm doing here is ensuring that we check whether this is a website project before an R package project.

### DIFF
--- a/src/cpp/core/r_util/RProjectFile.cpp
+++ b/src/cpp/core/r_util/RProjectFile.cpp
@@ -213,9 +213,14 @@ std::string detectBuildType(const FilePath& projectFilePath,
                             RProjectConfig* pConfig)
 {
    FilePath projectDir = projectFilePath.parent();
-   if (r_util::isPackageDirectory(projectDir))
+   if (isWebsiteDirectory(projectDir))
    {
-      setBuildPackageDefaults("", buildDefaults ,pConfig);
+      pConfig->buildType = kBuildTypeWebsite;
+      pConfig->websitePath = "";
+   }
+   else if (r_util::isPackageDirectory(projectDir))
+   {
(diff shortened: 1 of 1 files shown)


jjallaire:
Note that our existing check should be fine assuming that bookdown isn't actually adding `Type: Package` to it's DESCRIPTION file. See https://github.com/rstudio/rstudio/blob/54cd3abcfc58837b433464c793fe9b03a87f0bb4/src/cpp/core/r_util/RPackageInfo.cpp#L113-L131

kevinushey:
In this case, the bookdown package skeleton did indeed have a `DESCRIPTION` file with `Type: Package` (IIUC to power some Travis-related scripts), despite primarily being a bookdown project.

If this isn't a pattern we expect to see in the wild much then we can probably leave this PR off; I was just concerned that this might be more common for `bookdown` projects. (Users can always just manually change the build type to Website as well)

jjallaire:
There are other places in the product (and some R packages including
rprojroot) where we check for Type: Package and I think it's important that
this have integrity (i.e. it always means it's a package). The solution
should be to not propagate this use case and/or to improve the Travis R
environment so this hack isn't necessary. I'd say let's leave this PR out
for now.

kevinushey:
Sounds good -- I'll close this and let's only revisit if we see packages in the wild that adopt this scheme.

PR #945, opened 2017-01-13
use '--no-save --no-restore' w/build_book()

kevinushey:
This PR intends to resolve an issue where calls to e.g. `build_book()` would fail for users who initialize their R library paths in a user startup file (e.g. `.Rprofile` or `.Renviron`). (In such a case, because the library paths will not contain an `rmarkdown` installation, attempts to load + call `rmarkdown` functions will fail)

Rather than executing the build commands in a plain new R session with `--vanilla --slave`, we instead use `--no-save --no-restore --slave` -- that is, we still ask R to load user startup files.

### DIFF
--- a/src/cpp/session/modules/build/SessionBuild.cpp
+++ b/src/cpp/session/modules/build/SessionBuild.cpp
@@ -828,6 +828,12 @@ private:
       args.push_back("--vanilla");
       args.push_back("-e");
       args.push_back(command);
+      
+      // forward R_LIBS so the child process has access to the same libraries
+      // we do
+      std::string libPaths = module_context::libPathsString();
+      if (!libPaths.empty())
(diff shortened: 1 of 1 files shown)


jjallaire:
Typically the way we ensure that library paths are forwarded during build operations is by propagating the R_LIBS environment variable. Here's where we do that in the AsyncRProcess class: https://github.com/rstudio/rstudio/blob/54cd3abcfc58837b433464c793fe9b03a87f0bb4/src/cpp/session/SessionAsyncRProcess.cpp#L146-L154.

I think we should start with this so that we can avoid other side effects of profile files (that's the approach we've taken in other build scenarios). That said, if you think this change is the right one for this use case I'm fine with that too.

kevinushey:
Sounds good -- I just rebased the PR to that effect.

PR #947, opened 2017-01-17
disable keyboard shortcut processing when typing in terminal

gtritchie:
Don't process keyboard shortcuts when typing in the terminal. Lets programs (vim, emacs, and so on) so this themselves. Other minor cleanup on server-side.


src/gwt/src/org/rstudio/core/client/command/ShortcutManager.java
```diff
@@ -443,7 +444,8 @@ public void execute()
       }
 
       // Bail if this is an ignored key combination.
-      if (isIgnoredKeyCombination(keyCombination))
+      if (isIgnoredKeyCombination(keyCombination) || 
+            XTermWidget.isXTerm(Element.as(event.getEventTarget())))
```
kevinushey: This may be over-aggressive -- I think this will imply that commands which have a more global scope (e.g. `Build and Reload`) won't work while an XTermWidget has focus.

That said, I'm not sure what the best way to white-list (black-list?) commands here would be -- maybe some extra metadata in `commands.xml`?

jjallaire:
I'm not 100% sure that this is a good idea. There are lots of keyboard shortcuts that are still useful even when in the terminal including shortcuts for switching panes, build and reload, etc. won't be active.

jjallaire:
If we could only do this when in a full screen editor like Vim or EMacs I think it would be more reasonable.

gtritchie:
Checked in a less blunt approach; obey Workbench and Addin command shortcuts, others are passed through to terminal when it has focus.

gtritchie:
Added PackageDevelopment command context to list.

PR #948, opened 2017-01-17
Implement scroll support to new connections wizard

javierluraschi:
Before,

screen shot 2017-01-17 at 9 59 15 am

after with scroll support:

screen shot 2017-01-17 at 9 59 06 am

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/connections/ui/NewConnectionNavigationPage.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/connections/ui/NewConnectionNavigationPage.java
@@ -138,11 +138,10 @@ public class NewConnectionNavigationPage
          
          DockLayoutPanel panel = new DockLayoutPanel(Unit.PX);
          panel.addStyleName(styles.wizardPageSelectorItem());
-         
-         panel.setSize("100%", "38px");
+         panel.addStyleName(styles.wizardPageSelectorItemSize());
          
          Image rightArrow = new Image(WizardResources.INSTANCE.wizardDisclosureArrow());
-         rightArrow.addStyleName(RES.styles().rightArrow());
+         rightArrow.addStyleName(styles.wizardPageSelectorItemRightArrow());
          panel.addEast(rightArrow, 28);
          
          if (page.getImage() == null)
@@ -152,12 +151,12 @@ public class NewConnectionNavigationPage
          else
          {
             Image icon = new Image(page.getImage());
-            icon.addStyleName(RES.styles().leftIcon());
+            icon.addStyleName(styles.wizardPageSelectorItemLeftIcon());
             panel.addWest(icon, 28);
          }
          
          Label mainLabel = new Label(page.getTitle());
-         mainLabel.addStyleName(RES.styles().label());
+         mainLabel.addStyleName(WizardResources.INSTANCE.styles().wizardPageSelectorItemLabel());
          mainLabel.getElement().setAttribute("title", page.getSubTitle());
          panel.add(mainLabel);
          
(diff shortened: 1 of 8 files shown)


jjallaire:
I'd use even less vertical space for the connection types.

javierluraschi:
Other dialogs, like projects, use this much spacing:

screen shot 2017-01-17 at 10 29 55 am

Which felt too close for the connections dialog, that said, this commit reduces the spacing as follows, which seems better to me:

screen shot 2017-01-17 at 10 28 53 am

Do we want to reduce this even more?

jjallaire:
I think that your most recent treatment looks good.

I agree that the New Project dialog is too densely packed (@kevinushey and I have already discussed this). Let's actually introduce a constant that ensures that these values are synced.

src/gwt/src/org/rstudio/studio/client/workbench/views/connections/ui/NewConnectionNavigationPage.java
```diff
@@ -53,4 +81,105 @@ public NewConnectionNavigationPage(String title,
...
+
+         for (int i = 0, n = pages.size(); i < n; i++)
+         {
+            final WizardPage<NewConnectionContext, ConnectionOptions> page = pages.get(i);
+            SelectorItem item = new SelectorItem(page, new ClickHandler()
+            {
+               @Override
+               public void onClick(ClickEvent event)
+               {
+                  onSelected_.execute(page);
```
jmcphers: does this need a null guard?

javierluraschi: Not really, `setSelectionHandler` is called during `addAndInitializePage` while constructing the page before any click event would reach this. Also, the eisting code does not check for this, something bad might be going on is this is actually NULL.

javierluraschi:
Much better, sharing spacing and styles now, new dialogs:

screen shot 2017-01-17 at 11 17 35 am

screen shot 2017-01-17 at 11 17 44 am

kevinushey:
TBH I'm still a fan of the more compact dialog. The icons and text now look too small for their container (in my eyes)

javierluraschi:
We could **bold** both of them?

javierluraschi:
Bolded? (Haven't committed a change yet)

screen shot 2017-01-17 at 11 32 21 am

screen shot 2017-01-17 at 11 32 30 am

javierluraschi:
I think it looks worse bolded...

kevinushey:
I think bold is a bit heavy -- bumping the font size up a bit might help.

It's also possible nudging the icons to the right a bit would help (I just think the vertical spacing + spacing to left of each icon should be roughly equal, or at least feel comfortable).

javierluraschi:
It does look a bit better with the icons pushed right, to me...

screen shot 2017-01-17 at 11 39 31 am

screen shot 2017-01-17 at 11 39 37 am

javierluraschi:
Reducing 4 more pixels... original:

screen shot 2017-01-17 at 11 49 38 am

Latest:

screen shot 2017-01-17 at 11 48 30 am

screen shot 2017-01-17 at 11 48 34 am

All good? It does look a tiny bit smaller that I would want, but is worth to fit all the project options without scrolling, at least for now.

jjallaire:
Yes, I like the last iteration!

kevinushey:
Looks good to me as well! (nit: I would nudge the icons back to the left ~2 pixels)

javierluraschi:
Done.

screen shot 2017-01-17 at 11 48 30 am

screen shot 2017-01-17 at 12 50 07 pm

src/gwt/src/org/rstudio/studio/client/workbench/views/connections/ui/NewConnectionNavigationPage.java
jmcphers: Great, thanks!

jmcphers:
Thanks for the UX feedback everyone!

PR #953, opened 2017-01-23
Implement basic support for enumerating and connecting to odbc

javierluraschi:
Implement basic support for enumerating and connecting to `ODBC`:

 - [x] Use `odbc` package to enumerate `odbcinst.ini` connections.
 - [x] Provide basic UI for connecting trough ODBC.
 - [x] Style snippets dialog.
 - [x] Fixed issue where debug could trigger while launching shiny app dialog.
 - [x] Add support to read R snippets from `connections-path = /etc/rstudio/connections/*.R`.
 - [x] Generate dymaic UI bsed on R snippets.
 - [x] Polish css spacing for elements
 - [x] Preview connection code as user types
 - [x] Add support for "extended options" dialog
 - [x] Implement support for connections provided by system admin
 - [x] Increase wizard size for new connection dialog by subclassing dialog and adjusting background image.
 - [x] Replaced push button with themed button "Advanced..."
 - [x] Added support in snippet for connection string fields to avoid adding empty params
 - [x] Add support to escape `:` and `=` in snippets using `$colon$` and `$equal$`
 - [x] Other style adjustments throught the wizard

My setup was to:
 1. Upgrade `sparklyr` to latest version since it includes a new `Livy` connection and,
 2. The following `ODBC` configuration an snippets:

```
odbc_config --odbcinstini
/usr/local/Cellar/unixodbc/2.3.2_1/etc/odbcinst.ini
```

```
[SQLite]
Driver          = /usr/local/lib/libsqlite3odbc.dylib

[Spark ODBC]
Driver          = /usr/local/lib/libsparkodbc_sbu.dylib

[PostgreSQL]
Driver          = /usr/local/lib/libtdbcpostgres1.0.0.dylib
```

```
# /etc/rstudio/connections/SQLite.R
library(DBI)
con <- dbConnect(odbc::odbc(), .connection_string = "Driver={SQLite};Data Source=${1:Data Source=$colon$memory$colon$};Version=${2:Version=3};New=${3:New=True}")=
```

```
# /etc/rstudio/connections/PostgreSQL.R
library(DBI)
con <- dbConnect(odbc::odbc(), .connection_string = "Driver={PostgreSQL};${1:Server Name:Servername}${1:Port=5432:Port}${2:Database:Database}${3:User:UserName}${4:Password:Password}${5:Protocol=6.4:Protocol}${6:Row Versioning=No:RowVersioning}${7:Show System Tables=No:ShowSystemTables}${8:Show OID Column=No:ShowOidColumn}${9:Fake OID Index=No:FakeOidIndex}${10:Trace=Yes:Trace}${11:Trace File=sql.log:TraceFIle}")
```

```
# /etc/rstudio/connections/Twitter.R
library(twitteR)
setup_twitter_oauth(consumer_key = "${1:Consumer Key}", consumer_secret = "${2:Consumer Secret}", access_token = "${3:Access Token}", access_secret = "${4:Access Secret}")
```

Some screenshots under this configuration:

screen shot 2017-01-30 at 5 07 56 pm
screen shot 2017-01-31 at 10 38 07 am
screen shot 2017-01-31 at 10 38 18 am
screen shot 2017-01-30 at 5 08 01 pm
screen shot 2017-01-30 at 5 08 06 pm
screen shot 2017-01-30 at 5 08 16 pm
screen shot 2017-01-30 at 5 08 20 pm
screen shot 2017-01-30 at 5 08 24 pm


javierluraschi:
Progress today getting a generic connection string working:

screen shot 2017-01-24 at 5 13 32 pm

Ideally, if the drivers/r-files are correctly configured, we should never see this dialog since we would rather create individual fields; however, in reality, we will probably see packages or drivers with missing snippets that will result in this dialog.

javierluraschi:
More progress to auto generate the UI. Works for some cases, but some other ones will require a "More Options" button to present the additional settings in a separate dialog. Also, css needs more polishing, but here are a couple examples:

screen shot 2017-01-25 at 5 07 32 pm
screen shot 2017-01-25 at 5 07 39 pm
screen shot 2017-01-25 at 5 07 44 pm
screen shot 2017-01-25 at 5 07 52 pm

javierluraschi:
Polished layout (for instance, to be able to include `Port` field on the header) and added a new configuration push button that opens a full list of connection configuration options. The goal is to keep the main configuration options in the main dialog, and all the remaining ones in a separate dialog:

screen shot 2017-01-26 at 5 30 45 pm
screen shot 2017-01-26 at 5 30 52 pm

jmcphers:
A few points of UI feedback:

- Although the syntax isn't very discoverable, supporting server and port on the same line is a nice touch! Should we support 2 inputs on any row, not just the header? (server/port might not always be the first two params)

- The full params dialog needs a max height and escape hatch for connections with lots of parameters (e.g. vertical scroll)

- The placement of the toolbox icon makes it look like it configures just the Password field in your screenshot. I think this needs more aggressive treatment. Maybe a "Set All Parameters..." button with the same icon in the lower left of the dialog?

javierluraschi:
@jmcphers Thanks!
 1. I'm hesitant to allow us to configure secondary fields in any other row than the header since it looks really bad, aesthetically.
 2. Yes, good catch. Added scroll and max height, just in case.
 3. I agree; however, I didn't mean to make that icon final. Just asked @jjallaire if we could have a different one that would prevent users from thinking they are configuring a specific field. If changing the icon is not enough, I'll think of something else, the button might work out if it's nicely styled.

src/cpp/session/modules/SessionConnections.R
```diff
@@ -160,26 +160,107 @@ options(connectionObserver = list(
...
 })
 
+.rs.addFunction("connectionReadSnippets", function() {
+   snippetsPath <- getOption("connections-path", "/etc/rstudio/connections/")
+   snippetsFiles <- list()
+
+   if (!is.null(getOption("connections-path")) && !dir.exists(snippetsPath)) {
+      warning(
+         "Path '", snippetsPath, "' does not exist. ",
+         "Configure the connections-path option appropiately.")
```
kevinushey: Typo here: 'appropiately' => 'appropriately'

src/cpp/session/modules/SessionConnections.R
```diff
@@ -160,26 +160,107 @@ options(connectionObserver = list(
    NULL
 })
 
+.rs.addFunction("connectionReadSnippets", function() {
+   snippetsPath <- getOption("connections-path", "/etc/rstudio/connections/")
+   snippetsFiles <- list()
+
+   if (!is.null(getOption("connections-path")) && !dir.exists(snippetsPath)) {
```
kevinushey: nit: test the value of `snippetsPath` as the option was retrieved above?

src/cpp/session/modules/SessionConnections.R
```diff
@@ -160,26 +160,107 @@ options(connectionObserver = list(
    NULL
 })
 
+.rs.addFunction("connectionReadSnippets", function() {
+   snippetsPath <- getOption("connections-path", "/etc/rstudio/connections/")
```
jmcphers: In my view this option should be set on the server so that there's no way resetting R options inside a session could hide the connections supplied by an admin. What do you think? If we stick with an R option, we should probably prefix it with `rstudio.` for clarity.

src/gwt/src/org/rstudio/studio/client/application/ApplicationInterrupt.java
```diff
@@ -108,6 +116,61 @@ public void onFailure()
...
+                     public void onInterruptFinished()
+                     {
+                        errorManager_.setDebugSessionHandlerType(
+                           originalDebugType,
+                           new ServerRequestCallback<Void>()
+                           {
+                              @Override
+                              public void onResponseReceived(Void v)
+                              {
+                                 interruptR(new InterruptHandler() {
```
jmcphers: Why do we need to interrupt R a second time here after restoring the handler type?

src/gwt/src/org/rstudio/studio/client/application/ApplicationInterrupt.java
```diff
@@ -108,6 +116,61 @@ public void onFailure()
...
    }
    
+   public void interruptRNoDebug(final InterruptHandler handler) {
+      final int originalDebugType = errorManager_.getErrorHandlerType();
+      
+      if (originalDebugType != ErrorHandlerType.ERRORS_BREAK) {
+         interruptR(handler);
+      }
+      else {
+         errorManager_.setDebugSessionHandlerType(
```
jmcphers: Is there a possibility that this RPC won't return until R stops being busy? (It may depend on what kind of "busy" the R session is)

I understand the goal of this code is to make it possible to interrupt the Shiny apps driving connections panes w/o tripping the debugger, but I think there might be another way to do this -- let's chat offline (Joe might also have some ideas)

src/gwt/src/org/rstudio/studio/client/application/ApplicationInterrupt.java
```diff
@@ -108,6 +116,61 @@ public void onFailure()
       }
    }
    
+   public void interruptRNoDebug(final InterruptHandler handler) {
+      final int originalDebugType = errorManager_.getErrorHandlerType();
+      
+      if (originalDebugType != ErrorHandlerType.ERRORS_BREAK) {
```
jmcphers: This catches the most common case but we might want to reset user-supplied handlers as well (e.g. `options(error = browser)` would cause trouble here)

src/gwt/src/org/rstudio/studio/client/workbench/views/connections/ui/NewConnectionSnippetDialog.css
```diff
@@ -0,0 +1,33 @@
+@CHARSET "UTF-8";
+
+@external gwt-Label;
```
jmcphers: Nit: These lines are unnecessary unless you actually reference the `gwt-Label` class in your CSS (it just tells GWT not to obfuscate the class name when it sees it later)

src/gwt/src/org/rstudio/studio/client/workbench/views/connections/ui/NewConnectionSnippetHost.java
```diff
@@ -0,0 +1,398 @@
...
+            parts.add(new NewConnectionSnippetParts(order, key, value, connStringField));
+         }
+      }
+
+      Collections.sort(parts, new Comparator<NewConnectionSnippetParts>()
+      {
+         @Override
+         public int compare(NewConnectionSnippetParts p1, NewConnectionSnippetParts p2)
+         {
+            return p1.getOrder() == p2.getOrder() ? 0 : p1.getOrder() < p2.getOrder() ? -1 : 1;
```
jmcphers: You could also just do `p1.getOrder() - p2.getOrder()` :-)

src/gwt/src/org/rstudio/studio/client/workbench/views/connections/ui/NewConnectionSnippetHost.java
```diff
@@ -0,0 +1,398 @@
...
+      }
+
+      return connGrid;
+   }
+
+   private void updateCodePanel()
+   {
+      String input = info_.getSnippet();
+      
+      String pattern = "\\$\\{([0-9]+):([^:=}]+)(=([^:}]+))?(:([^}]+))?\\}";
```
jmcphers: Nit: If this needs to stay in sync w/ above copy of regex, use a constant

jjallaire:
Maybe change the caption of the "Advanced" button to "Advanced Options..." (and also the title of the dialog which appears as a result of clicking on the button).

hadley:
What happens if you enter a password? Does it get included in the connection string?

jjallaire:
I think at this point it will have to.

We need to create the end-to-end infrastructure to store and replay these
securely (as discussed in FL).

In the meantime however we could redact any R console statement we see with
password = within RStudio. We could either simply not store these in the
history or we could substitute "XXXXXX" for the password. This is by no
means perfect but is probably better than allowing passwords to enter the
history.

hadley:
I'm worried about it both entering the history and entering the source. I think it would be better to simply not support password in the connection builder at this time. (Gabor is going to start work on the keyring package soon so it would only be crippled for a couple of weeks)

jjallaire:
That might indeed be the best course. Note however that users attempting to
connect to a database that requires a password will indeed enter the
password (in the console, in the source, or both) with or without this
dialog. Given this, the confusion created by not supporting the password
field might not be worth it.

hadley:
True. We should at least kick off discussion of how the password UI should work so it can be hooked up as soon as possible.

Also, the create connection pane should indicate that those drivers are all ODBC drivers, and we should consider adding support for other DBI backends (at least SQLite, MySQL, and PostgreSQL).

jjallaire:
Yes, we definitely plan on adding support for not only other DBI backends
but even other types of non-relational data stores (e.g. mongo, etc.). For
the big ones we'll provide PRs to the package authors then hopefully others
will follow suit.

src/cpp/session/modules/SessionConnections.R
javierluraschi: We probably need two settings then, one for users to add their custom connections with the code as it is, but another one for the system admin to add their very own settings. @jmcphers  would about we let them configure a `connections-path` as a RSP profile (http://docs.rstudio.com/ide/server-pro/r-sessions.html#user-and-group-profiles)?

jmcphers: I'd start with a global session option which defines connections available to everyone (this can be done in open source), and later add a Pro profile option which defines *additional* connections available to particular groups.

src/gwt/src/org/rstudio/studio/client/application/ApplicationInterrupt.java
javierluraschi: There should not since the RPC is not going back to R code, to my knowledge. It eventually calls: https://github.com/rstudio/rstudio/blob/54cd3abcfc58837b433464c793fe9b03a87f0bb4/src/cpp/session/modules/SessionErrors.cpp#L70-L82 which seems R-free to this point.

src/gwt/src/org/rstudio/studio/client/application/ApplicationInterrupt.java
javierluraschi: There should not since the RPC is not going back to R code, to my knowledge. It eventually calls: https://github.com/rstudio/rstudio/blob/54cd3abcfc58837b433464c793fe9b03a87f0bb4/src/cpp/session/modules/SessionErrors.cpp#L70-L82 which seems R-free to this point.

Sure, I'll ping you on Slack.

jmcphers: That RPC eventually winds up in R here:

https://github.com/rstudio/rstudio/blob/54cd3abcfc58837b433464c793fe9b03a87f0bb4/src/cpp/session/modules/SessionErrors.cpp#L61

src/gwt/src/org/rstudio/studio/client/workbench/views/connections/ui/NewConnectionSnippetDialog.css
javierluraschi: Changed.

src/cpp/session/modules/SessionConnections.R
javierluraschi: Thanks - This was a very *appropriate* comment!

src/gwt/src/org/rstudio/studio/client/application/ApplicationInterrupt.java
javierluraschi: Generalized this function, I still think it will be helpful to have wheather we end up using it or not.

src/gwt/src/org/rstudio/studio/client/workbench/views/connections/ui/NewConnectionSnippetHost.java
javierluraschi: I think it requires more brain cycles to read, but changed.

src/gwt/src/org/rstudio/studio/client/workbench/views/connections/ui/NewConnectionSnippetHost.java
javierluraschi: Ah yes, refactored.

javierluraschi:
@jjallaire dialog updated with "Advanced Options..." feels a tad heavy to me, so let me know if we rather revert the button.

@jmcphers feedback addressed, except for the Slack thread regarding canceling the dialog without interrupting the r session.

javierluraschi:
Removed the password field for now since I agree it can be misleading to think we are doing something secure when we are not, yet.

For the password discussion, another option would be to replace the connection code with `getSessionPassword(#)` where # is an identifier for this particular dialog session. The password is stored in memory such that it is only used during this session.

For the connection string, us and system admins would have to create a connection string like:

```
dbConnect(paste("param1=${1:param1};password=", getSessionPassword(${2:password}), sep = "")
```

and the dialog would insert the correct password session-id without revealing the password.

That said, out of the scope of this PR to give us some time to think more about this.

src/cpp/session/modules/SessionConnections.R
javierluraschi: @jmcphers that's what we already have in this PR, correct? Just checking I'm not missing any change form this comment.

src/gwt/src/org/rstudio/studio/client/application/ApplicationInterrupt.java
javierluraschi: @jmcphers I ended up adding the RPC within `ShinyServerOperations`, suing `shiny::stopApp()` seems to work just fine!

src/cpp/session/modules/SessionConnections.R
jmcphers: I do think we want a change at the rsession level, let's discuss in person.

PR #958, opened 2017-01-26
Fix: Bash not processing startup files in terminal tab

gtritchie:
In terminal tabs, start bash as an interactive login shell so regular startup file processing happens (versus launching with —norc and a pre-determined prompt, which still happens for non-terminal cases).


gtritchie:
Also fixed bug where starting up a previously terminated session was leaving zombie terminals showing previous buffer but not connected to a new bash process.

PR #967, opened 2017-02-01
Expose dialog apis as rstudioapis

javierluraschi:
We are missing to expose the dialog apis as `.rs.api.*` functions to be consumable from `rstudioapi` package.

See also: https://github.com/rstudio/rstudioapi/pull/36

### DIFF
--- a/src/cpp/r/R/Api.R
+++ b/src/cpp/r/R/Api.R
@@ -332,7 +332,7 @@
   )
 })
 
-.rs.addFunction("showDialog", function(title, message, url = "") {
+.rs.addApiFunction("showDialog", function(title, message, url = "") {
    .Call("rs_showDialog",
       title = title,
       message = message,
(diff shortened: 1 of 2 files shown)


src/cpp/r/R/Api.R
```diff
@@ -368,6 +376,15 @@
       url = NULL)
 })
 
+.rs.addApiFunction("writePreference", function(name, value) {
```
jjallaire: Not sure we want to expose the preference functions as it would also require our documenting the names of all the UI preferences and create an alternate channel for mutating them that we don't control. What's the motivation for exposing these?

javierluraschi: @jjallaire I think you already figured out the motivation of this change from the `rstudioapi` PR already: To allow `sparklyr`/users to persist their `dplyr` or `DBI` preference in the UI.

jjallaire:
Yup, let's just do the namespacing I suggested on the RStudio side.

src/cpp/r/R/Api.R
jjallaire: Looks like the namespacing is in. This is good to go!

PR #972, opened 2017-02-03
Remove the weak encryption used on user input sent to the server

gtritchie:
Per conversations today; this encryption is weak and adds more overhead to each keystroke sent from the terminal (only a tiny amount, but getting rid of code is always nice).


jjallaire:
LGTM

PR #994, opened 2017-02-10
Fix Windows build

gtritchie:
Noticed a build break on my Windows dev box. This does the bare minimum to get it building; I haven't actually exercised the impacted code on Windows.

### DIFF
--- a/src/cpp/core/system/Win32System.cpp
+++ b/src/cpp/core/system/Win32System.cpp
@@ -498,6 +498,13 @@ Error realPath(const FilePath& filePath, FilePath* pRealPath)
    return Success();
 }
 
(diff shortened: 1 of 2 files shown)


src/cpp/core/system/Win32System.cpp
```diff
@@ -498,6 +498,13 @@ Error realPath(const FilePath& filePath, FilePath* pRealPath)
    return Success();
 }
 
+// TODO (review) this is a stopgap to get Windows compiling again!
+Error realPath(const std::string& path, FilePath* pRealPath)
+{
+   *pRealPath = FilePath(path);
```
jjallaire: I think this should just be:

```cpp
return realPath(FilePath(path), pRealPath);
```

PR #995, opened 2017-02-10
Adds dataimport menu entries for csv using base

javierluraschi:
From user feedback, bringing back the `utils` package `read.table` import dialog:

screen shot 2017-02-09 at 9 56 04 pm

screen shot 2017-02-09 at 9 56 12 pm

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/widget/ToolbarPopupMenu.java
+++ b/src/gwt/src/org/rstudio/core/client/widget/ToolbarPopupMenu.java
@@ -21,11 +21,15 @@ import com.google.gwt.core.client.Scheduler.ScheduledCommand;
 import com.google.gwt.dom.client.Document;
 import com.google.gwt.dom.client.NativeEvent;
 import com.google.gwt.event.dom.client.KeyCodes;
+import com.google.gwt.event.dom.client.MouseOverEvent;
+import com.google.gwt.event.dom.client.MouseOverHandler;
 import com.google.gwt.event.shared.HandlerRegistration;
 import com.google.gwt.safehtml.shared.SafeHtml;
+import com.google.gwt.safehtml.shared.SafeHtmlUtils;
 import com.google.gwt.user.client.Event;
 import com.google.gwt.user.client.Event.NativePreviewEvent;
 import com.google.gwt.user.client.Event.NativePreviewHandler;
+import com.google.gwt.user.client.Window;
 import com.google.gwt.user.client.ui.MenuBar;
 import com.google.gwt.user.client.ui.MenuItem;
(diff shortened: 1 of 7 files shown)


jjallaire:
How about:

Using read.csv...
Using readr...

Do other people agree that a submenu is preferred here? I am 50/50 on that proposition.

hadley:
I'm not a huge fan of the submenu either - hardly seems worth it for two elements.  

Definitely should be lower case readr, I don't have strong feelings about base vs read.csv

OTOH the readr dialog is quite a lot more general than just CSV - it's any delimited file.

jjallaire:
Maybe:

Read Text (base)....
Read Text (readr)....

javierluraschi:
Last commit without submenu:

screen shot 2017-02-10 at 9 34 19 am

screen shot 2017-02-10 at 9 34 10 am

kevinushey:
Honestly my preference would be for something like (no submenu)

```
From Text File...
From Delimited File...
```

Or even

```
```
From Text File...
From Delimited File (readr)...
```

We don't want to use the label `Import Text` because that's now inconsistent with the other menu items, and I feel somewhat strongly about preserving the old label `From Text File...` (since that's what was used in older versions of the IDE).

javierluraschi:
Is this a reasonable compromise? I do agree that would be good to keep `From Text File...` for users that have been using this for a while; however, this is not descriptive in the long term. Happy to tweak this more if needed...

screen shot 2017-02-10 at 9 55 32 am

screen shot 2017-02-10 at 9 55 27 am

jjallaire:
I like Javier's last proposal (I think "From Delimited" is too obscure a
reference).

jmcphers:
My only concern about this approach is that it offers no guidance to someone just looking to get text into R. The (base) vs (readr) distinction will not be important to them, but they will be afraid to choose the wrong one since they look like peers.

I'd be in favor of just dropping the `readr` label so we have:

From Text (base) ...
From Text ...

This gives unsure users a clear default choice without removing the escape hatch for users who truly need read.table. 

I'll merge this as is since we're getting close. Let's tweak on master if we decide to change the wording any further.

jjallaire:
I'm not sure which is the clear default choice though? (`base` seems like a name that would be associated with the default). I don't think there's an easy way to finess this if we want to have both and not have awkward and further confusing UI. I'd say if we should just keep it as (base) and (readr)

jjallaire:
I'd also say that if we want to promote a clear default choice we should hide base and make it a preference to make it visible. Yes instructors will have to tell users to take a UI action but they are already telling the same users to take potentially dozens more UI actions so it hardly seems like a relevant tax.

PR #1006, opened 2017-02-14
Add a 'create branch' dialog

kevinushey:
Note: not ready for merge, pending UI feedback.

---

This PR adds a simple 'create branch' dialog, intended to make it possible to create and checkout a new branch without popping into the shell. A quick preview of what the UI looks like so far:

screen shot 2017-02-14 at 11 31 01 am

screen shot 2017-02-14 at 11 31 24 am

When a branch of the same name already exists, we prompt the user as to whether they'd like to check out that branch, or overwrite it with a fresh branch from their current checkout.

screen shot 2017-02-14 at 11 31 30 am

## Comments

Should we allow the user to set an upstream remote as well? E.g. with a dialog like:

```
Create Branch: < ___ >
Set Upstream: < ___ >
[ ] Push to remote
```

I'm trying to imagine a workflow where we attempt to set an upstream remote (if one exists); otherwise, we automatically push the repository to the remote and automatically set the upstream. Does something like that make sense?

We probably also want a dedicated branch icon (I'm reusing the mermaid icon as it fits somewhat well as a placeholder)

### DIFF
--- a/src/cpp/core/StringUtils.cpp
+++ b/src/cpp/core/StringUtils.cpp
@@ -15,10 +15,10 @@
 
 #include <core/StringUtils.hpp>
 
+#include <algorithm>
 #include <map>
 #include <ostream>
 
-#include <algorithm>
 #include <boost/algorithm/string.hpp>
 #include <boost/algorithm/string/case_conv.hpp>
 #include <boost/algorithm/string/classification.hpp>
@@ -311,6 +311,13 @@ std::string systemToUtf8(const std::string& str, int codepage)
 std::string systemToUtf8(const std::string& str)
 {
    return systemToUtf8(str, CP_ACP);
+}
+
+std::string toUpper(const std::string& str)
+{
+   std::string upper = str;
+   std::transform(upper.begin(), upper.end(), upper.begin(), ::toupper);
+   return upper;
 }
 
 std::string toLower(const std::string& str)
--- a/src/cpp/core/include/core/StringUtils.hpp
+++ b/src/cpp/core/include/core/StringUtils.hpp
@@ -80,6 +80,7 @@ std::string systemToUtf8(const std::string& str);
 std::string systemToUtf8(const std::string& str, int codepage);
 
 std::string toLower(const std::string& str);
+std::string toUpper(const std::string& str);
 std::string textToHtml(const std::string& str);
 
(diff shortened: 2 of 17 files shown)


jjallaire:
I think the caption should be "New Branch". Paul can design us an icon.

We absolutely should include linking to a remote! (that's the main motivation for this feature). If there is a remote already defined we should default the picker to the remote. We can then list other remotes as well as an option for "(None)".

jmcphers:
Very nice!

Your git pane looks pretty wide there -- we will probably need to collapse to just the icon at some widths if you haven't set that up already, or move the command into one of the existing menus. 

It might be nice to put it inside the existing Branch droplist as the top entry, in the same way that New Project is the top entry in the project switch menu.

kevinushey:
A second iteration on the UI for auto-push-on-create:

screen shot 2017-02-15 at 2 40 47 pm

screen shot 2017-02-15 at 2 41 00 pm

The dialog could likely be made a little bit larger.

kevinushey:
I also feel like it could be useful for us to always display the Git command(s) we executed when showing these in a process dialog -- ie, we could let the user know exactly what we're executing behind the scenes through our Git UI.

jjallaire:
Agreed the dialog should be larger. Definitely a select box for remotes but it looks like you already have that covered!

Maybe we could say "Sync branch with remote" rather than "Push branch to remote" as sync implies on ongoing relationship (set-upstream) rather than a one time action.

I think showing the git commands we are going to execute in a read only text area below might be nice but we'll have to see if it clutters up the dialog too much.

kevinushey:
Another iteration on this (this feature blew up quite a bit more than I initially expected!)

screen shot 2017-02-16 at 11 30 28 am

screen shot 2017-02-16 at 11 30 37 am

I felt adding an `Add Remote...` button here was necessary to close the loop in the case where a user has e.g. created a new Git repository locally without cloning it from e.g. GitHub first -- such users can now manually add their GitHub repository (or otherwise) as a remote and then push + set upstream.

javierluraschi:
Would be nice to align the start of the textbox with the start of the dropdown and the end of the textbox with the end of the "Add Remove..." button.

jmcphers:
Looks like this is getting close! Should we merge and apply any remaining UX polish on master?

kevinushey:
Sounds good to me! I think this PR should be ready to go.

PR #1019, opened 2017-02-17
Add missing icon to clear formats in r source

javierluraschi:
We have a missing toolbar icon in r files:

screen shot 2017-02-17 at 12 23 17 pm

Fixed:

screen shot 2017-02-17 at 11 46 38 am


jjallaire:
It's not missing -- we can and do have many menu items without icons. Typically the criteria for including an icon on a menu is whether it also has it's own toolbar button. Clearly we've gone a bit beyond that, but I don't think we should be running around trying to fill in icons where they aren't. I still use icons as a point of emphasis within menus and if we give every item an icon we lose that.

javierluraschi:
@jjallaire are we talking about menu icons or toolbar icons? This fix adds a toolbar icon, not a menu icon. Am I misunderstanding something? I haven't seen any missing toolbar icons yet in the IDE.

Maybe the suggestion here is to add the toolbar icon but remove the menu icon?

jjallaire:
Ah, got it! Yes, toolbar icons should always be there!

javierluraschi:
@jjallaire k, thank you!

jmcphers:
Why do we show these commands in plain R files in the first place? I believe they're R Markdown specific, in which case the right fix is probably just to hide them for other file types (see DynamicCommands).

jjallaire:
Yes, actually the bug is that those items shouldn't be showing up at all in
plain R source files.

BTW the second command is just for shiny_prerendered which is pretty
obscure. I'm wondering if a regression got introduced somewhere else that
triggered this showing up in plain R files.

javierluraschi:
Seems they come from these two:

 1. https://github.com/rstudio/rstudio/commit/64e22ca80573a9d0dde7b3dd1d48b21eefe367d1
 2. https://github.com/rstudio/rstudio/commit/dfc3713e3ea400b490a51aa1b27addc9101912dc

Should I scope (1) to shiny docs and (2) to r-markdown docs and not display any clearing options otherwise?

jjallaire:
What puzzles me is that those were already scoped that way (at least they
were intended to be). I think there is probably a simple bug here rather
than the need to do new work to manage the menus. I'll take a look at this
to see if I can spot the bug.

javierluraschi:
OK, thanks JJ, if it's not obvious let me know and I'll keep digging.

jjallaire:
Superseded by https://github.com/rstudio/rstudio/pull/1022

PR #1041, opened 2017-03-03
Initial support for retina icons

javierluraschi:
This PR adds support for "Use Retina Icons" appearance option and includes the first batch of retina icons.

To work in the main branch, I've added support to handle standard and non-retina icons at the same time. This is achived by:
 1. Wrapping resources under a `ImageResource2x` class to support both resolutions and wrap the commands in this class when a `@2x` resource is availalble.
 2. Loading this preference on boot time, for this I had to introduce a new RPC call `boot_init` which is called as soon as we launch RStudio and does not depend on the `RStudioGinjector` code since using any of this classes causes the Commands to be injected and the resources to be loaded. I took a couple page load time numbers and was not able to see any slowdown from adding this call. I got the following results:

| Page Load Time (Before)  | (After Change) |
| ------------- | ------------- |
| 5.86s  | 5.65s  |
| 5.65s  | 5.61s  |
| 5.69s  | 5.55s  |

Here is how the option looks like, the user preference is turned off by default.

screen shot 2017-03-02 at 4 08 09 pm

Once enabled and after restarting the session we get the following retina icons:

screen shot 2017-03-02 at 4 07 17 pm


src/gwt/src/org/rstudio/studio/client/workbench/model/BootInfo.java
```diff
@@ -0,0 +1,28 @@
...
+import com.google.gwt.core.client.JavaScriptObject;
+
+public class BootInfo extends JavaScriptObject
+{
+   protected BootInfo()
+   {
+   }
+
+   public final native boolean getUseRetinaIcons() /*-{
+      return typeof(this.use_retina_icons) !== "undefined" ? this.use_retina_icons : true;
```
jmcphers: If we fail to get a BootInfo object then `this` itself will be undefined and we'll throw here -- maybe add a guard for `this`?

src/gwt/src/org/rstudio/core/rebind/command/CommandBundleGenerator.java
```diff
@@ -420,12 +425,20 @@ private ImageResourceInfo generateImageBundle()
...
+
+         if (resourceNames.contains(key + ".png"))
          {
             writer.println("ImageResource " + commandId + "();");
             iri.addImage(commandId);
          }
+
+         if (resourceNames.contains(key + "@2x.png"))
+         {
+            writer.println("@Source(\"" + commandId + "@2x.png\")");
```
jmcphers: Some versions of CMake don't like files with `@` in their names; this problem isn't obvious when you build locally but can break on the Linux build machines in the lab. We can take the change and see how it goes but my guess is that we will want to use a different naming scheme; for e.g. pdf.js (which includes retina icons) we would up going with `_2x` instead of `@2x` to overcome this issue.

https://github.com/rstudio/rstudio/blob/f9ef3c3caf056bfcc601508920279b5dfe642ac2/src/gwt/tools/sync-pdfjs#L46-L51

jmcphers:
@gtritchie That's a good question; in my view Retina has become the Kleenex of high-DPI displays but that's just my two cents. :-) 

My view is that this preference should be temporary. It's cool to have the option to use old school icons, but the Retina icons are going to be different not only in resolution but also in style (a modern, flatter look). This puts us in a rough spot going forward as we either have to create a "retro" version of each new icon we add to the product, or accept that users that prefer the old icons are going to eventually see a mix of old and new icons which they are unlikely to find tasteful.

jjallaire:
I'm actually wondering if we shouldn't just always use Retina icons? (especially given Jonathan's point about carrying the flatter vibe all the way through the UI). Is the downside just page load time (seems that impact would be minimal).

I haven't thought it all the way through but any extra calls/logic at page load risks breaking some things that work now (as I recall we have quite a bit of logic surrounding detection of hung sessions, etc.) that we'd have to reason fully about before adding this extra RPC call. I'd say unless there is an overwhelming sentiment that we should keep the old icons we should not perform this high wire act to enable that preference.

src/gwt/src/org/rstudio/studio/client/RStudio.java
```diff
@@ -232,6 +241,34 @@ else if (view != null &&
...
+                                             null,            /* kwparams */
+                                             false,           /* redactLog */
+                                             null,            /* sourceWindow */
+                                             null,            /* clientId */
+                                             "");             /* clientVersion */
+
+      rpcRequest.send(new RpcRequestCallback() {
+         public void onError(RpcRequest request, RpcError error)
+         {
+            GWT.runAsync(runCallback);
```
javierluraschi: Agree!

src/gwt/src/org/rstudio/studio/client/workbench/model/BootInfo.java
javierluraschi: Ah, right! Yes, I'm being paranoid making sure that even is the `boot_init` call fails, we can also keep working as expected, will add this.

src/gwt/src/org/rstudio/core/rebind/command/CommandBundleGenerator.java
javierluraschi: If this has already broken the build, then I agree, `_2x` makes more sense, I'll update the PR.

javierluraschi:
@jjallaire Regarding reducing code churn, I believe we will have to adjust for the retina size anyways. To my knowledge, there is no automatic way of converting width/height correctly since the codebase currently assume that the size of the icon is the size to be displayed; so I believe we will need to make changes everywhere in the code where there are `ImageResource` references that should support retina. I'm planning to make the `ImageResource2x` class intelligent enough to be a simple replace in our codebase that handles all cases appropriately, still figuring this one out.

@jjallaire Regarding the overwhelming sentiment, we had Kevin and Sean proactively requesting the retro mode where the UI is left unchanged. Therefore, even though the `boot_init` call scares me a bit, my preference would be to add this call and remove it later if we find regressions coming out of this change that are not trivial to fix. Also, this would allow me to merge changes while working in the new theme rather than keeping everything on a branch.

@gtritchie @jmcphers I think it makes sense to remove the term retina from our codebase and use `2x` instead. For the UI, I have less of a strong opinion, Retina has become pretty common, but in retrospect, we might want to name this "Use High Resolution" to accomodate other platforms and not just icons but potentially other UI elements.

@jjallaire Regarding loading times, seems to me GWT is intelligent enough not to fetch the URLs until they are used, I see about the same number of requests and can't spot any of the low-res images.

jjallaire:
I don't mind the code churn as the changes are pretty safe and straightforward.

I don't mind the idea of a retro mode, but why can't the retro mode just use retina icons?

We have a whole bunch of logic triggered to client_init that determines whether the R session is alive at all (and offers early termination of it if it's not):

https://github.com/rstudio/rstudio/blob/master/src/gwt/src/org/rstudio/studio/client/application/ApplicationClientInit.java

You are going to have to adapt this logic for boot_init.

In short, I think the boot_init thing is asking for lots of unexpected trouble. Even if you want to keep retro icons I think we need to find a way to do it without boot_init.

javierluraschi:
Mostly since in retro mode the icons are a bit different (flatter) and not pixelated, which seems some users will miss.

Regarding the `ApplicationClientInit.java` code pointer, `boot_init` does not currently require a running r-session nor projects, so I believe we wouldn't need those prompts nor retry logic, we would need a comment in the server code to remind us of the intent of `boot_init`.

I'm OK going either way: (1) Push retina icons to everyone and later introduce retro mode if needed, or (2) Introduce retro mode from the beginning and remove it later if needed.

@jjallaire sounds like you prefer going for (1), correct? Happy to make the change.

javierluraschi:
Here are a couple side-by-side example of the icon differences:

screen shot 2017-03-03 at 11 22 14 am

screen shot 2017-03-03 at 11 21 59 am

screen shot 2017-03-03 at 11 20 25 am

screen shot 2017-03-03 at 11 18 42 am

javierluraschi:
Code review comments addressed, including removal of `boot_init` in the client and `Use High Resolution` option, we can revert this commit https://github.com/rstudio/rstudio/pull/1041/commits/b3f5c3ae328b3e76cdf4488289536f274d8f9bbe if we decide to bring this back.

jjallaire:
FYI boot_init was implemented in the rsession binary so absolutely did require a running R session. In general trying to put anything in front of client_init will likely create a rats nest of problems.

Yes I was leaning towards just using retina icons everywhere. I think flatter icons will actually go just fine with the existing chrome so am hopeful this won't be something we need to revisit.

javierluraschi:
Sounds good, I'll clean up the server side session of `boot_init` in the next change, we can either wait for this change or merge this PR and will follow up in a separate PR.

So @jjallaire, if we really had to bring back a retro mode option... what would be the way to implement this? I can obviously put an API in the `server` project; however, at the very minimum, I would have to save a `UserSetting` to load/save the `UiPref`, but do we even have the concept of a user in the `server` binary? Or is the `server` binary designed to be only a proxy to the `session` binary? Assuming is only a proxy, I guess we would have to support the `boot_init` call and fix any issues, or find a way to reload `ResourceImages` with a different url after boot, seems like the reload would be more reasonable.

BTW. Changing the icons to retina for commands was straightforward since I was able to tap into the command generation code. However, images that don't rely in this code will require either: (1) A bunch of wrappers over `ImageResource` or (2) some sort of `ClientBundle2x` class that supports wrapping retina images. I'm trying to support (2) since that would mean we only need to rename `ClientBundle` to `ClientBundle2x` or something like that; however, it's not trivial to do this, but also not that complicated, someone already tried creating an extension for this that happens to explain this process with some clarity: https://github.com/kDCYorke/RetinaImages - Will address additional retina icons in additional PRs.

jjallaire:
If we ever have to bring back the retro icons I'll take a close look at the
best way to accomplish this. As you suspected, it's actually impossible to
access user settings from the rserver binary -- partly this is because the
disk upon which user settings is stored may not even yet be mounted! (PAM
sessions often mount user home directories). In RSP we actually have a
special peer process to rsession that is used just so that we can access
the user home directory without launching an R session (but this isn't
available in the open source version).

javierluraschi:
Finished cleaning up server changes; however, I need one more fix for this PR before merging since some resources using `background-image` are not rendering properly.

javierluraschi:
Fixed a menu icons clipping issue, verified in web and desktop. We can merge this one.

javierluraschi:
Ah, yes, `BootInfo` should go away as well, will push that next week with the next batch of icons.

PR #1043, opened 2017-03-05
Handle out-of-order RPC calls for terminal input messages

gtritchie:
Have only seen this happening on Windows IDE, where it is very easy to repro and user input is often messed up (e.g. echo shows up as ehco).

Each writeStandardInput call from the client now includes a sequence count (per terminal instance) which increments with each call. The server uses this to notice out-of-order messages, and waits until the gaps are filled before it releases the input from its queue.

For other existing use cases, a flag is sent for the sequence which triggers the original behavior with pure FIFO.

Watch for overflow of int on Java side, and send a flag telling server to flush and restart the expected sequence (highly unlikely, requires 2 billion+ keystrokes, but hey, who knows). On server, also prevent locking up the queue if an expected sequence never arrives, and just emit what we have. Unlikely, but not impossible, if a message gets dropped somehow.

Unit tests for the new behavior.

### DIFF
--- a/src/cpp/session/SessionConsoleProcess.cpp
+++ b/src/cpp/session/SessionConsoleProcess.cpp
@@ -42,7 +42,7 @@ void saveConsoleProcesses();
 
 ConsoleProcess::ConsoleProcess(boost::shared_ptr<ConsoleProcessInfo> procInfo)
    : procInfo_(procInfo), interrupt_(false), newCols_(-1), newRows_(-1),
-     childProcsSent_(false), started_(false)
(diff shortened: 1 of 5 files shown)


jjallaire:
Note that it is in theory possible for a desktop session to reconnect to an existing server (a page reload would do it, and there are a couple places where we actually cause a page reload of a running session so we can sync prefs that need to be present at page load time).

kevinushey:
Looks good to me -- is this PR ready for merge?

gtritchie:
@kevinushey Yes please. Ready to go.

PR #1044, opened 2017-03-07
Integrate additional retina icons

javierluraschi:
After much investigation on making this automatic, I've gone for this more verbose approach.

The automatic approach that I investigated was around changing/customizing `ClientBundle` and `ImageResource` to automatically detect retina icons and provide the right bindings and dimensions on compile time. While this is still possible, I found out today that `ImageResourceGenerator` is a `final` class and `ClientBundle` is not designed to be extensible at this level. Therefore, the only way to extend `ImageResource` is to copy the whole bundling sources which total about 2K lines of GWT code, this approach was taking in this plugin: https://github.com/kDCYorke/RetinaImages; however, it will require us to borrow and compile [ImageResourceGenerator.java](https://github.com/gwtproject/gwt/blob/3a41a8067767965ff72f4e49eb015f587fcfee16/user/src/com/google/gwt/resources/rg/ImageResourceGenerator.java) ourselves. This is still possible, but have a feeling that it might be better to bite the bullet and wrap each `ImageResource` ourselves, might be a bit of manual work but the complexity is much simpler that maintaining and fixing issues in the `ImageResourceGenerator` class.

The changes in this PR are pretty manual and boring, mostly adding the icon and wrapping them on ` ImageResource2x` class. Therefore, **would me more interesting to discuss if we think it has merit to bring `ImageResourceGenerator.java` and compile this code ourselves**.

@jjallaire your feedback here would be much appreciated.


jjallaire:
I would agree that the manual approach is best here. It's a one time cost that is simpler and less future bug-prone.

javierluraschi:
Sounds good! I'm planning to merge this branch end of the day. More commits to follow but mostly, repetitive tasks to wrap resources over `ImageResource2x`, will merge additional commits to master without notice since there is little value in code reviewing these changes.

PR #1045, opened 2017-03-07
shell detection for win32 terminal

gtritchie:
New rpc, get_terminal_shells, returns list of detected shells on Windows; not being used yet; client will invoke to populate list of available shells in preferences, menus, etc.


kevinushey:
Awesome to see that we'll be supporting WSL Bash!

PR #1046, opened 2017-03-08
Fix compiler warnings on non-Windows build

gtritchie:
Mostly several signed/unsigned issues, an unused-code warning, and an actual bug with no return from getGitBashShell( ) when git not found.


kevinushey:
LGTM!

PR #1050, opened 2017-03-10
Update xterm.js to release 2.4.0

gtritchie:
New drop of xterm.js.

Main visible difference is that urls detected in the terminal can now be clicked on to open in a new tab. May have to suppress this on desktop clients, but we'll see how it goes.

Release notes at https://github.com/sourcelair/xterm.js/releases/tag/2.4.0


gtritchie:
I'm going to merge this one. It's a mechanical change (updating to a newer xterm.js release), so nothing to review, really, at this pre-1.1 release stage.

PR #1052, opened 2017-03-10
Fix regression with terminal buffer trimming

gtritchie:
When returning terminal buffer, it is supposed to be trimming down to the max lines supported by the terminal (currently 1000 lines). I introduced a (silly) regression with  https://github.com/rstudio/rstudio/pull/1023 in which the trimming didn't happen properly. 

Easy fix, and added a unit test that would have caught this.


gtritchie:
The xterm update earlier today includes feature to convert web links in terminal buffer to <a>, meaning they can be clicked and will open in default browser for desktop IDEs, and in a new tab for server version. Added to NEWS.md.

PR #1061, opened 2017-03-15
build boost using custom 'rstudio_boost' namespace

kevinushey:
NOTE: Not yet ready for merge, but ready for some review + discussion.

---

This PR attempts to build RStudio using a custom build of Boost which uses the `rstudio_boost` namespace, rather than the `boost` namespace. This should help resolve crashes where R packages making use of Boost inadvertently end up using / calling symbols within the `rsession` process, and a mismatch in Boost versions causes a crash.

Some overview: The `bcp` utility basically allows us to copy Boost headers to a new folder, with the `boost` namespace renamed (using `rstudio_boost` in this case), and an alias back to `boost` used. This is literally done by just replacing calls of the form (for example)

```cpp
namespace boost {
```

with e.g.

```cpp
namespace rstudio_boost{} namespace boost = rstudio_boost; namespace rstudio_boost{
```

in the Boost sources.

While the `boost` namespace is declared as an alias for `rstudio_boost`, it's still necessary for us to change our code to re-define things in the `rstudio_boost` namespace in our sources. This is done by using the macro `RSTUDIO_BOOST_NAMESPACE` and toggling that to `rstudio_boost` or `boost`, depending on whether we're building against the system Boost or our own bundled Boost.


jjallaire:
I think that if we want to take a run at this we should also update to as recent a version of boost as we can as this will maximize our chance of fixes related to bcp as well as fixes related to linking of multiple versions of boost within a target. I'd say give it a go with boost 1.63.

kevinushey:
Sounds good!

javierluraschi:
@kevinushey do we have a repro for this one that this fix resolves? 

I'm a bit skeptic that this would happen, but I've also never dug into the internals of loading dynamic libraries. My guess is that `rcpp` creates a separate dynamic library from the running `rsession` by calling a compiler into some code, all good. Then this library gets loaded dynamically by `rsession` into its own running process. However, I'm surprised that the compiled library can make use of RStudios `rsession` boost entry points since the symbols might not even be available to this new library. I think a more likely case would be that we load a version of boost, the package is compiled with a slightly different version of boost and somehow the difference in versions crash in runtime.

That said, no harm in changing RStudio's namespace.

kevinushey:
Here's an example on StackOverflow where R package code inadvertently ends up calling into the Boost filesystem symbols compiled into the `rsession` executable (and crashing):

http://stackoverflow.com/questions/41747611/rstudio-crashes-when-reading-mzml-file-r-terminal-doesnt

javierluraschi:
Thanks for the reference, interesting issue!

kevinushey:
I'm seeing some warnings during compilation while using Boost 1.63 on Windows:

> C:/Users/kevin/rstudio/dependencies/windows/boost-1.63-win-rtools33-gcc493/boost32/lib/libboost_thread-mgw49-mt-1_63.a(thread.o): duplicate section `.rdata$_ZTVN13rstudio_boost16exception_detail10clone_implINS0_19error_info_injectorINS_9gregorian8bad_yearEEEEE[__ZTVN13rstudio_boost16exception_detail10clone_implINS0_19error_info_injectorINS_9gregorian8bad_yearEEEEE]' has different size

> C:/Users/kevin/rstudio/dependencies/windows/boost-1.63-win-rtools33-gcc493/boost32/lib/libboost_thread-mgw49-mt-1_63.a(thread.o): duplicate section `.rdata$_ZTVN13rstudio_boost16exception_detail10clone_implINS0_19error_info_injectorINS_9gregorian9bad_monthEEEEE[__ZTVN13rstudio_boost16exception_detail10clone_implINS0_19error_info_injectorINS_9gregorian9bad_monthEEEEE]' has different size

My initial guess is some kind of mismatch in compiler flags used for Boost vs. RStudio / rsession; but will try to dig in a bit more.

jjallaire:
This Google query turned up a few interesting possibilities:

https://www.google.com/?q=boost+thread+duplicate+section

kevinushey:
It looks like I was able to resolve the 'duplicate section' warnings by performing static linking to the C++ runtime (ie, by using `runtime-link=static` in the `b2` invocation). This will of course require us to do some extra testing to make sure this doesn't regress anything unexpectedly!

FWIW there is some precedent for doing this -- e.g. the version of MinGW maintained by STL does the same; https://github.com/StephanTLavavej/mingw-distro/blob/0428923c6f4b3e7d59c30eda0f0b99d8a8ae4819/boost.sh#L13-L15.

kevinushey:
I can also confirm on Ubuntu 16.04 that this fixes the crashes reported by Dirk (specifically, the ones he reported regarding the `anytime` package). Phew!

jmcphers:
Nice! Is this ready for final review?

jjallaire:
That's fantastic!!!!

There are at least half a dozen similar problems that have come up over the
years which I am hopeful this will also resolve.

I tried to do this ~ 4 years ago and actually merged it into master.
Shortly thereafter I started seeing hangs on the Mac desktop when switching
projects and I immediately backed out the change. We are 10 versions of
boost/bcp later so hopefully whatever strangeness I saw is no longer in
play, but it is definitely worth some concerted testing around that
scenario to make sure that issue isn't still lurking.

kevinushey:
I haven't seen any Mac hangs in the (relatively light) testing I've done, but thus far the IDE has survived switching projects so we may be safe there. Definitely needs some stress testing though!

kevinushey:
I think this should be ready for final review; I'll add some comments to clarify some of the changes made.

CMakeCompiler.txt
```diff
@@ -31,9 +31,14 @@ if(WIN32 AND NOT DEFINED RSTUDIO_PACKAGE_BUILD)
   set(RTOOLS_PATH "${CMAKE_CURRENT_SOURCE_DIR}/../../dependencies/windows/Rtools33")
   string(REGEX REPLACE "/" "\\\\" RTOOLS_TOOLCHAIN_32BIT_PATH "${RTOOLS_PATH}/mingw_32/bin")
   set(ENV{PATH} "${RTOOLS_TOOLCHAIN_32BIT_PATH}")
+  set(WINDRES "${RTOOLS_TOOLCHAIN_32BIT_PATH}\\windres.exe")
```
kevinushey: For some reason, on my Windows machine I was seeing that CMake 3.8 couldn't find `windres.exe` when executed at e.g.

https://github.com/rstudio/rstudio/blob/f92d5a37a69ab5d51376e9cef35c6d528cf49519/src/cpp/desktop/CMakeLists.txt#L251-L268

I'm not yet sure if that was something locally weird in my environment, or if `add_custom_command()` no longer searches for commands on the PATH (or somehow that wasn't propagated to that CMake call).

dependencies/common/install-boost
```diff
@@ -24,22 +24,29 @@ INSTALL_DIR=`pwd`
...
 
 # vars
-BOOST_VERSION_NUMBER=1.50.0
-BOOST_VERSION=boost_1_50_0
+BOOST_VERSION_NUMBER=1.63.0
+BOOST_VERSION=boost_1_63_0
 BOOST_DIR=/opt/rstudio-tools/boost/$BOOST_VERSION
 BOOST_TAR=$BOOST_VERSION.tar.bz2
 BOOST_BUILD_DIR=boost-build
+BOOST_MODULES="algorithm asio array bind chrono circular_buffer context crc
```
kevinushey: Note that some of these modules aren't actually used by us (e.g. `context`) but adding them was necessary to work around a bug in `bcp`'s dependency resolution; see e.g.

https://svn.boost.org/trac/boost/ticket/11376

The issue was marked as resolved there, but I was still seeing it -- regardless this seems to function as a workaround.

dependencies/common/install-boost
```diff
@@ -65,19 +88,24 @@ then
...
 
      sudo ./bjam ${BOOST_BJAM_FLAGS} --prefix=$BOOST_DIR toolset=clang $CXX_STDLIB_FLAGS variant=release threading=multi link=static install
    else
      sudo ./bjam ${BOOST_BJAM_FLAGS} --prefix=$BOOST_DIR variant=release install
    fi
 
+   # rename libraries in the boost install dir
+   cd $BOOST_DIR/lib
+   for file in librstudio*; do
+      sudo mv $file ${file/rstudio_/}
```
kevinushey: The boost libraries are generated with names e.g. `librstudio_boost*`, so this code just removed the `rstudio_` bit.

dependencies/windows/install-boost/install-boost.R
```diff
@@ -1,63 +1,119 @@
...
+      download(boost_url, destfile = boost_archive)
+      if (!file.exists(boost_archive))
+         fatal("Failed to download '%s'", boost_archive)
+   }
    
    # extract boost (be patient, this will take a while)
    progress("Extracting archive -- please wait a moment...")
-   exec("unzip", boost_filename)
+   progress("Feel free to get up and grab a coffee.")
+   exec("7z.exe", "x -mmt4", boost_filename)
```
kevinushey: I use `7z.exe` here because both of our `zip.exe` and `tar.exe` utilities actually barf when extracting the Boost sources, as the archives contain some non-portably long paths (for documentation files).

kevinushey: (To add to the fun, even `xcopy` fails to e.g. copy the extracted directory from one place to another!)

dependencies/windows/install-boost/install-boost.R
```diff
@@ -1,63 +1,119 @@
...
+b2_build_args <- function(bitness) {
+   
+   prefix <- file.path(output_dir, sprintf("boost%s", bitness))
+   cxxflags <- "-march=core2 -mtune=generic"
+   
+   unlink(prefix, recursive = TRUE)
+   
    paste(
-      "b2",
+      "-q",
```
kevinushey: `-q` here just means 'stop on first error', so that the build dies more quickly on error.

src/cpp/CMakeLists.txt
```diff
@@ -130,14 +130,17 @@ if(APPLE OR WIN32 OR RSTUDIO_PACKAGE_BUILD)
...
+# default to boost 1.63.0
+set(BOOST_VERSION 1.63.0)
 
 # override if necessary
 if (RSTUDIO_BOOST_VERSION)
    set(BOOST_VERSION "${RSTUDIO_BOOST_VERSION}")
 endif()
 
+# disable Boost Signals deprecation warning
+add_definitions(-DBOOST_SIGNALS_NO_DEPRECATION_WARNING)
```
kevinushey: We might want to consider switching to `Boost.Signal2` as the original `Boost.Signals` library is now deprecated. (IIUC this was done because `Boost.Signals` is not thread-safe, but we never actually use Boost signals for cross-thread communication so this in theory doesn't affect us)

src/cpp/tests/cpp/tests/vendor/catch.hpp
```diff
@@ -5965,7 +5965,7 @@ namespace Catch {
         void operator=( Version const& );
     };
 
-    extern Version libraryVersion;
+    inline Version libraryVersion();
```
kevinushey: These changes were made to avoid a double-destruct crash in our `rsession` tests -- because we might load R packages that use Catch themselves, when closing the R session we can end up seeing this struct destructed twice (once by the version of Catch in the R package, and once by ourselves). I filed an associated bug on the Catch repository.

kevinushey:
TODO re: review with @jmcphers:

[ ] Put the built Windows Boost zipfile on S3
[ ] Update the `install-dependencies.cmd` script to reference new Windows Boost
[ ] Run the Boost installer script on the build machines.

kevinushey:
Alright, new version of Boost is up on S3 and I can confirm that the regular old `install-dependencies.cmd` script successfully downloads + unpacks it.

jmcphers:
Great! I'll plan to merge this tomorrow AM and update the build machines during the day. Depending on how that goes I may kick off a trial build so we're around to catch any fallout before the nightly (will let you know.) Thanks so much for tackling this project!

PR #1074, opened 2017-03-21
support R graphics engine v12

kevinushey:
This PR adds support for version 12 of the R graphics engine, as will be used with R 3.4.0.

### DIFF
--- a/src/cpp/r/session/graphics/RGraphicsDevDesc.cpp
+++ b/src/cpp/r/session/graphics/RGraphicsDevDesc.cpp
@@ -345,6 +345,103 @@ struct DevDescVersion8
    char reserved[64];
 };
 
+struct DevDescVersion9
+{
+   double left;
+   double right;
+   double bottom;
+   double top;
+   double clipLeft;
+   double clipRight;
+   double clipBottom;
+   double clipTop;
+   double xCharOffset;
+   double yCharOffset;
+   double yLineBias;
+   double ipr[2];
+   double cra[2];
+   double gamma;
+   Rboolean canClip;
+   Rboolean canChangeGamma;
+   int canHAdj;
+   double startps;
+   int startcol;
+   int startfill;
+   int startlty;
+   int startfont;
+   double startgamma;
+   void *deviceSpecific;
+   Rboolean displayListOn;
+   Rboolean canGenMouseDown;
+   Rboolean canGenMouseMove;
+   Rboolean canGenMouseUp;
+   Rboolean canGenKeybd;
+   Rboolean gettingEvent;
+
+   void (*activate)(const pDevDesc );
+   void (*circle)(double x, double y, double r, const pGEcontext gc, pDevDesc dd);
+   void (*clip)(double x0, double x1, double y0, double y1, pDevDesc dd);
+   void (*close)(pDevDesc dd);
+   void (*deactivate)(pDevDesc );
+   Rboolean (*locator)(double *x, double *y, pDevDesc dd);
+   void (*line)(double x1, double y1, double x2, double y2,
+       const pGEcontext gc, pDevDesc dd);
+   void (*metricInfo)(int c, const pGEcontext gc,
+             double* ascent, double* descent, double* width,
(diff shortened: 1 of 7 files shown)


src/cpp/r/session/graphics/RGraphicsDevDesc.cpp
```diff
@@ -487,8 +584,8 @@ pDevDesc allocate(const DevDescVersion9& devDescVersion9)
    case 9:
    default:
```
kevinushey: Note: needs update here

jmcphers:
Where is `devDescVersion12` defined?

kevinushey:
It's defined in the header file:

https://github.com/rstudio/rstudio/pull/1074/files#diff-5e6a65a803ef55ecfcf7fc71200cbb87R27

AFAICS the convention is that the most recent supported version of the graphics device is exposed there, while older backwards-compatibility versions are in the .cpp file.

jmcphers:
Ah, I missed that. Thanks, this looks good!

kevinushey:
I'll mark that commit for backporting as well.

jjallaire:
I may not be reading carefully enough here, but have we lost compatibility with a previous graphic engine version (9) with some of the renames here? The idea is that the RStudio binary should work against all known struct layouts -- again, I'm most likely just not reading / reviewing carefully enough but just wanted to make sure it was understood that in supporting a new engine version we should always continue supporting all of the old ones.

jmcphers:
That is the intent!  Unfortunately the struct layout change made in R 3.4 was made in such a way that we are going to need to use dynamic dispatch to set some of the flags (effectively the same model we currently have for the `RCNTXT` struct with which we are similarly coupled). @kevin is working on this change now (Kevin, please jump in if any of these details are incorrect!).

kevinushey:
My main worry is in this function call:

https://github.com/rstudio/rstudio/blob/a4decbee078d8ef55140d4b92fa606504a082804/src/cpp/r/session/graphics/RShadowPngGraphicsHandler.cpp#L245-L289

Here, we're trying to set fields on a `pDevDesc` pointer, but that pointer definition is based on the version of R we're compiling against, not running against -- so in theory, we might be setting the wrong fields here if the `pDevDesc` fields don't line up with the active graphics device for that session.

Let me know if I'm misunderstanding something about how this works here.

kevinushey:
Shoot -- I'm seeing a crash with this using R 3.2, so I'm going to roll this back for now:

```
* thread #1: tid = 0x26ddeb, 0x000000010eb076ee grDevices.so`RQuartz_StrWidth + 30, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x88)
  * frame #0: 0x000000010eb076ee grDevices.so`RQuartz_StrWidth + 30
    frame #1: 0x000000010a488f66 rsession`rstudio::r::session::graphics::handler::dev_desc::setSize(pDD=0x00007fd36a0008d0) + 230 at RGraphicsDevDesc.cpp:639
    frame #2: 0x000000010a48b326 rsession`rstudio::r::session::graphics::handler::shadow::setSize(pDev=0x00007fd36854d550) + 38 at RShadowPngGraphicsHandler.cpp:241
    frame #3: 0x000000010a4714e9 rsession`rstudio::r::session::graphics::device::(anonymous namespace)::createGD() + 1449 at RGraphicsDevice.cpp:517
...
```

jjallaire:
All of this should be accounted for in the existing mechanism. You can see
a set of helper methods for setting structure members here that does
runtime checks and then casts to the appropriate struct for the actual set:

https://github.com/rstudio/rstudio/blob/master/src/cpp/r/session/graphics/RGraphicsDevDesc.cpp#L500

I didn't have the time to dig into your commit enough to say, but I thought
that you had perhaps just evolved the GE version 9 struct rather than
creating a brand new one.

If the right solution isn't readily apparent I'd suggest holding off on
completing this until Monday when you and I can dig into more realtime.
Rest assured though that we've seen this exact scenario before and there is
a way to handle it within the current system.

PR #1080, opened 2017-03-22
guard against incompatible R versions on restore

kevinushey:
NOTE: not quite ready for merge; likely needs some tweaking re: how the message is reported to the user.

This PR fixes an issue where R could crash on startup when attempting to restore an workspace associated with a different version of R than the currently active version of R.

The fix is mostly straightforward -- on suspend, we now write the R version to a file; on restore, we validate against that version, and abort the restore if either the major or minor versions differ.

This should hopefully reduce some headaches that we might see with the introduction of R 3.4.0.

Right now, the error message is just plainly reported to the console:

screen shot 2017-03-22 at 12 49 02 pm

but we might want to tweak this (e.g. just change the error message, or provide a modal dialog information the user, or otherwise)

### DIFF
--- a/src/cpp/core/VersionTests.cpp
+++ b/src/cpp/core/VersionTests.cpp
@@ -1,0 +1,57 @@
+/*
+ * VersionTests.cpp
+ *
+ * Copyright (C) 2009-17 by RStudio, Inc.
+ *
+ * Unless you have received this program directly from RStudio pursuant
+ * to the terms of a commercial license agreement with RStudio, then
+ * this program is licensed to you under the terms of version 3 of the
+ * GNU Affero General Public License. This program is distributed WITHOUT
+ * ANY EXPRESS OR IMPLIED WARRANTY, INCLUDING THOSE OF NON-INFRINGEMENT,
+ * MERCHANTABILITY OR FITNESS FOR A PARTICULAR PURPOSE. Please refer to the
+ * AGPL (http://www.gnu.org/licenses/agpl-3.0.txt) for more details.
+ *
+ */
+
+#include <tests/TestThat.hpp>
+
+#include <core/Version.hpp>
+
+namespace rstudio {
+namespace core {
+namespace {
+
+void compareVersionLessThan(const std::string& lhs, const std::string& rhs)
+{
+   Version vl(lhs);
+   Version vr(rhs);
+   
+   expect_true(vl <  vr);
+   expect_true(vl <= vr);
+   expect_true(vl != vr);
+   expect_true(vr >  vl);
+   expect_true(vr >= vl);
+}
+
+void compareVersionEqual(const std::string& lhs, const std::string& rhs)
+{
+   expect_true(Version(lhs) == Version(rhs));
+}
+
+context("Version")
+{
+   test_that("Various versions are compared correctly")
+   {
+      compareVersionLessThan("3.2.0", "3.3.0");
+      compareVersionLessThan("3.2", "3.3.1");
+      compareVersionLessThan("3", "3.0.1");
+      compareVersionLessThan("3.3.0", "3.3.1");
+      
+      compareVersionEqual("3.0.0", "3.0.0.0");
+      compareVersionEqual("3", "3.0-0");
+   }
+}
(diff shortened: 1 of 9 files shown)


jmcphers:
This change looks good (and well done on avoiding the plain `major` and `minor` var names! ;-)). I'm not a fan of modals unless some action is demanded, and in this case there isn't much we can do. How about emitting a client event which triggers the global info bar to show? 

This is also going to cause users to lose data, because as soon as the new session suspends, it's going to obliterate all of the suspended session data they had previously. Mostly that's not a problem, but it's possible that they have some data in the session's .Rdata that they are not going to be happy about losing. Is there a location to which we could copy that data so that it doesn't get overwritten? We could then e.g. alert the user that we can't restore their data automatically, so we saved a copy of it in <path>.

kevinushey:
We could back up the suspended session data, but I'm slightly inclined to do nothing because it puts a couple ugly decisions on us:

1. When do we clean up the backed-up suspended session data? Or, do we put it some place like `/tmp` where it is naturally cleaned up later?

2. Right now, the only way to get out of this 'crash on start' loop is to either restore R to an older version, or manually blast the `suspended-session-data` folder. I imagine most users are fine just doing the latter.

3. In theory, users will already have their data saved in a local workspace (living outside of the suspended session data folder), and that folder will exist (and likely load) just fine -- so a restart of the R session would successfully restore that.

Let me know if I'm off-base, or if perhaps there's another avenue we could take here. Note that the ultimate cause of the crash is that we end up attempting to restore packages from the wrong library folder; that is, we end up loading R 3.2 version of a package in an R 3.3 session, so it is possible we could do something more targeted.

jmcphers:
1. I don't think we ever need to clean it up. We should just tell the user where we've put it, and let them decide what to do with it (even if that something is "keep it around indefinitely until I decide what to do with it). It's kind of annoying to litter files around, but it's better than losing data, and it shouldn't happen often. As .Rdata can contain sensitive information it shouldn't go in `/tmp`; we should probably put it somewhere in the project folder. 

2. Indeed, it'll be nice when no workaround is necessary!

3. This is indeed the case when a session is quit, but I don't think it's the case when a session is suspended. The data that I'm talking about here is the contents of the global environment at the time the session was suspended, which represents a sort of work in progress, and to the best of my knowledge is not written anywhere else (especially if your configuration doesn't write .Rdata on exit!). It's written here: https://github.com/rstudio/rstudio/blob/54cd3abcfc58837b433464c793fe9b03a87f0bb4/src/cpp/r/session/RSessionState.cpp#L378-L383

How compatible is .Rdata across R versions? If it's just the packages that are causing a problem, we might consider loading those things which are not known to cause issues. We could probably do 90% of the session resume actions, which would leave the user in a state close enough to the one they left that only a subtle informational message would be necessary.

kevinushey:
In general, `.RData` files are backwards and forwards compatible (even from e.g. R 2.11 to R 3.3).

kevinushey:
Alright, we now make sure to always restore the global environment (as that shouldn't run afoul of version change crashes), which I believe obviates the need for explicitly backing that up for the user.

I've also tweaked how we warn about restoring session state:

screen shot 2017-03-28 at 11 50 49 am

Let me know what you think. (Should we be more specific? I'm a bit worried that this might just provoke FUD)

jjallaire:
I think it might be okay to just say something like "R version change
[3.2.0 -> 3.4.0] detected when restoring session"

If we are just not restoring things in the global environment we could say:
"R version change [3.2.0 -> 3.4.0] detected when restoring session,
environment state not restored"

kevinushey:
How about "R version change [3.2.0 -> 3.4.0] detected when restoring session; search path not restored"? The primary thing that we're leaving out of the restore process is just the search path / reloading any packages that were loaded in the previous session.

jjallaire:
I think that's fine

jmcphers:
@kevinushey this looks ready to go -- anything else you want to tweak prior to merge?

kevinushey:
Nothing new incoming from me; I think this should be good to go.

PR #1090, opened 2017-03-27
support R graphics engine v12

kevinushey:
NOTE: not quite ready for merge (pending a bit of extra local testing + review).

This PR is a second stab at support for the new layout of the R graphics engine. Unfortunately, because the graphics device layout has changed relatively early in its definition, we now require indirection for effectively all of the graphics device functions -- hence why this PR has blown up with a large number of methods.

cc: @jjallaire

### DIFF
--- a/src/cpp/r/include/r/session/RSession.hpp
+++ b/src/cpp/r/include/r/session/RSession.hpp
@@ -57,7 +57,7 @@ struct ROptions
 {
    ROptions() :
          useInternet2(true),
-         rCompatibleGraphicsEngineVersion(9),
+         rCompatibleGraphicsEngineVersion(12),
          serverMode(false),
          autoReloadSource(false),
          restoreWorkspace(true),
--- a/src/cpp/r/session/graphics/RGraphicsDevDesc.hpp
+++ b/src/cpp/r/session/graphics/RGraphicsDevDesc.hpp
@@ -24,7 +24,7 @@
 
 extern "C" {
 
-struct DevDescVersion9
+struct DevDescVersion12
 {
    double left;
    double right;
@@ -55,6 +55,7 @@ struct DevDescVersion9
    Rboolean canGenMouseMove;
    Rboolean canGenMouseUp;
    Rboolean canGenKeybd;
+   Rboolean canGenIdle; // version 12
    Rboolean gettingEvent;
 
    void (*activate)(const pDevDesc );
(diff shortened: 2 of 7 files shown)


src/cpp/r/include/r/session/RSession.hpp
```diff
@@ -57,7 +57,7 @@ struct ROptions
 {
    ROptions() :
          useInternet2(true),
-         rCompatibleGraphicsEngineVersion(9),
+         rCompatibleGraphicsEngineVersion(12),
```
kevinushey: It looks like we set the compatible graphics engine version in a bunch of places, so we should sanity check that these changes make sense. (I'm a little confused because the versions declared in the various places are different)

src/cpp/r/session/graphics/RGraphicsUtils.cpp
```diff
@@ -40,7 +40,7 @@ namespace graphics {
 
 namespace {
 
-int s_compatibleEngineVersion = 8;
+int s_compatibleEngineVersion = 12;
```
kevinushey: It looks like we set the compatible graphics engine version in a bunch of places, so we should sanity check that these changes make sense. (I'm a little confused because the versions declared in the various places are different)

src/cpp/session/SessionOptions.cpp
```diff
@@ -274,7 +274,7 @@ core::ProgramStatus Options::read(int argc, char * const argv[])
          value<bool>(&autoReloadSource_)->default_value(false),
          "Reload R source if it changes during the session")
       ("r-compatible-graphics-engine-version",
-         value<int>(&rCompatibleGraphicsEngineVersion_)->default_value(11),
+         value<int>(&rCompatibleGraphicsEngineVersion_)->default_value(12),
```
kevinushey: It looks like we set the compatible graphics engine version in a bunch of places, so we should sanity check that these changes make sense. (I'm a little confused because the versions declared in the various places are different)

jjallaire: That is probably just forwarding from one place (the session config file) with some defaults that we never updated. I'd set it to 12 everywhere.

jjallaire: This all looks like the right approach to me. I'd just do some developer testing under the different versions and then have some QA testing to make sure nothing has gone amok with the device behavior (seems highly unlikely but worth some poking at).

src/cpp/r/session/graphics/RGraphicsDevDesc.hpp
```diff
@@ -55,6 +55,7 @@ struct DevDescVersion9
    Rboolean canGenMouseMove;
    Rboolean canGenMouseUp;
    Rboolean canGenKeybd;
+   Rboolean canGenIdle; // version 12
```
jjallaire: I don't see you copying or initializing `canGenIdle` anywhere (e.g. similar to https://github.com/rstudio/rstudio/blob/fef100e920af3d51344ff8e4ae10556580f3e297/src/cpp/r/session/graphics/RShadowPngGraphicsHandler.cpp#L282).

kevinushey: I'm setting it here now: https://github.com/rstudio/rstudio/pull/1090/files#diff-dcfd8bfc2c362a35a03b7913472834d8R734

Note that I moved `setDeviceAttributes()` to be a templated function with dispatch into `RGraphicsDevDesc.cpp`, just to make sure we don't run afoul of any layout differences.

jjallaire:
Okay, great.

Saw the templated setDeviceAttributes, very nice solution!

kevinushey:
Alright, I think this PR should be good for merge -- @jmcphers, can you pull the trigger? I'll mark the commit for backporting and hopefully there will be no other changes required.

PR #1092, opened 2017-03-29
catch exceptions emitted by boost::regex_match

kevinushey:
Some users have reported crashes on startup with the following error message:

> Unexpected exception: The complexity of matching the regular expression exceeded predefined bounds.  Try refactoring the regular expression to make each choice made by the state machine unambiguous.  This exception is thrown to prevent "eternal" matches that take an indefinite period time to locate.

Unfortunately, we haven't been able to discover a root cause for this issue yet, but the most likely culprit is unintentional exponential backtracking in one of our regular expressions, followed by overly-long user input of some form.

This PR attempts to catch and swallow those errors, returning `false` regarding these matches in such a case.

Let me know if:

1. This is an approach worth pursuing,
2. Whether we should consider also printing a backtrace,
3. Whether we should re-throw the exception or just return `false` for such cases.

### DIFF
--- a/src/cpp/core/Backtrace.cpp
+++ b/src/cpp/core/Backtrace.cpp
@@ -14,6 +14,7 @@
  */
 
 #include <core/Backtrace.hpp>
+#include <core/RegexUtils.hpp>
 
 #ifndef _WIN32
 # include <core/Algorithm.hpp>
@@ -82,7 +83,7 @@ void printBacktrace(std::ostream& os)
    for (std::size_t i = 0; i < stackDepth; ++i)
    {
       std::string readyForPrinting = stackStrings[i];
-      if (boost::regex_match(stackStrings[i], match, reBacktrace))
+      if (regex_utils::match(stackStrings[i], match, reBacktrace))
       {
          std::size_t n = match.length();
          if (n >= 8)
--- a/src/cpp/core/BrowserUtils.cpp
+++ b/src/cpp/core/BrowserUtils.cpp
@@ -13,10 +13,10 @@
  *
  */
 
-#include <core/BrowserUtils.hpp>
-
 #include <boost/algorithm/string/predicate.hpp>
 
+#include <core/BrowserUtils.hpp>
+#include <core/RegexUtils.hpp>
 #include <core/SafeConvert.hpp>
 
 using namespace boost::algorithm;
@@ -33,7 +33,7 @@ bool hasRequiredBrowserVersion(const std::string& userAgent,
 {
    double detectedVersion = requiredVersion;
    boost::smatch match;
-   if (boost::regex_search(userAgent, match, versionRegEx))
+   if (regex_utils::search(userAgent, match, versionRegEx))
    {
       std::string versionString = match[1];
       detectedVersion = safe_convert::stringTo<double>(versionString,
--- a/src/cpp/core/ConfigUtils.cpp
+++ b/src/cpp/core/ConfigUtils.cpp
@@ -52,10 +52,14 @@ void extractToMap(const std::string& keyAndValue,
 void extractVariables(const std::string& vars, Variables* pVariables)
 {
    // scan for variables via regex iterator
-   boost::regex var("^([A-Za-z0-9_]+=[^\n]+)$");
-   boost::sregex_token_iterator it(vars.begin(), vars.end(), var, 0);
-   boost::sregex_token_iterator end;
-   std::for_each(it, end, boost::bind(extractToMap, _1, pVariables));
+   try
+   {
+      boost::regex var("^([A-Za-z0-9_]+=[^\n]+)$");
+      boost::sregex_token_iterator it(vars.begin(), vars.end(), var, 0);
+      boost::sregex_token_iterator end;
+      std::for_each(it, end, boost::bind(extractToMap, _1, pVariables));
+   }
+   CATCH_UNEXPECTED_EXCEPTION;
 }
 
 Error extractVariables(const FilePath& file, Variables* pVariables)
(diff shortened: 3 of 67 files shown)


jjallaire:
Yes, this is excellent!!!

I think we should print what we can (what you have now may be enough, not sure if we need the backtrace) and should indeed return `false` rather than re-throw). We may want to have one other wrapper that we use when we know we are on the main thread within an R session that will print a warning to the console (so that we actually get bug reports about these issues). That could be as simple as adding a `report_stderr` flag that controls whether we print the message to stderr in addition to logging.

kevinushey:
I've added some logic around whether the code is executing on the main thread, and in such a case we write a similar message to the console -- let me know if this looks like the right approach.

https://github.com/rstudio/rstudio/pull/1092/commits/ce1f01585788367983f9ff46ee0a0fe889a198ed

(I wanted to avoid mutating the interface of `regex_utils::match()` as there are a number of 'helper' interfaces as well, and duplicating each one of those would've been a bit unwieldy)

kevinushey:
Perhaps writing directly to `std::cerr` should be sufficient then? IIUC we re-direct that output back to the R console so users will see it immediately (that is, we can avoid directly calling R APIs)

jjallaire:
Yes exactly, a plain write to stderr will do it

jmcphers:
As other writes to stderr are not guarded, is the intent of the mutex just to prevent the interleaving of two simultaneous regex errors? This seems unlikely enough that it might not be worth the additional complexity (unless I'm missing something about how these work)

kevinushey:
You're exactly right -- this is just me being extra paranoid. It's likely fine to avoid explicit synchronization here.

jjallaire:
In practice it is already thread safe, see http://stackoverflow.com/questions/14637595/iostream-thread-safety-must-cout-and-cerr-be-locked-separately

So I think we can get rid of the mutex.

jjallaire:
See also http://stackoverflow.com/questions/467938/stdout-thread-safe-in-c-on-linux?noredirect=1&lq=1

kevinushey:
I also spent some time diving into the various regular expressions we use throughout the codebase looking for any potential culprits -- nothing stands out as a definitive problem, but I thought it might be worth sharing what I learned anyhow.

The main culprits I've discovered that can lead to excessive backtracking:

1. The use of sequential greedy atoms with some intersection between what they match,
2. The use of greedy identifiers before the end of the string (e.g. `\s*$`).

## Examples

### Greedy Intersections

https://github.com/rstudio/rstudio/blob/a8dc5b3d2bdef9917b11306ad28f9f793b016280/src/cpp/session/SessionMain.cpp#L1441]

In this case, the portion of the regular expression `(\w+_\w+)` behaves poorly, as `_` is actually part of the `\w` character class, and so matching something like:

    abc123@

leads to excessive backtracking within `abc123` after failing to match at `@`.

### Greedy Ending

https://github.com/rstudio/rstudio/blob/a8dc5b3d2bdef9917b11306ad28f9f793b016280/src/cpp/core/markdown/Markdown.cpp#L216

Although it's unintuitive, a regular expression like `---\s*$` can lead to excessive backtracking, e.g. in matching

    ---                    b

Once the `\s*` section hits `b`, the regular expression will attempt to backtrack and, in effect, keep looking for the end of the line (even though it's obvious that we can't find the end of the line by backtracking!). In situations like this, we likely want to use a possessive / non-backtracking class, e.g. `---\s*+$`.

## Epilogue

Most of the examples here deal with small / likely bounded strings, so it's unlikely these are the causes of the crash -- what we likely need to look at is regular expressions we run over unbounded user-generated text (e.g. R Markdown, HTML documents).

I think any tweaks to the regular expressions we use are out of scope for this PR but worth keeping in mind if we dive in more later.

kevinushey:
Note to self: this still needs a bit more work for wrappers around `boost::regex_search()`, which calls into the same APIs that can trigger a complexity exception.

jmcphers:
@kevinushey Thanks for taking the time to investigate this thoroughly! I have adapted your remarks and added them to our coding standards doc so we have a record of these as an antipattern to watch for in reviews.

kevinushey:
I think this PR should be good to go now. A quick summary of the changes (mostly done with a blanket find + replace)

1. Replace usages of `boost::regex_search` with `regex_utils::search`,
2. Replace usages of `boost::regex_match` with `regex_utils::match`,
3. Double-check that our usages of `regex_filters` occur within try-catch blocks (they fortunately do),
4. Wrap our few usages of `boost::sregex_token_iterator`s in `try-catch` blocks (as generating a wrapper iterator definition would likely be painful)

The `regex_utils::match()` and `regex_utils::search()` functions delegate directly to their Boost counterparts, but call them within a try-catch block that reports the exception to the console (so users can inform us what's going wrong).

jmcphers:
LGTM. The bug which triggered these changes is targeted for the patch release; do you feel it still meets the risk profile given that the scope of changes has increased somewhat? (and if so, would you be willing to squash to a single commit for cherry-picking?)

kevinushey:
IMHO this doesn't quite meet the bar for the patch release, since it's no longer a targeted fix. (In theory everything here is safe, but given how many lines of code have been touched I'd rather be safe)

If we're instead able to identify which particular pattern(s) are responsible for the exception hitting users, then we can likely tweak the associated regular expression and a fix of that scope would likely be worthy of patch release.

jmcphers:
That sounds reasonable. I'll merge this change to master and leave the patch bug assigned to you. My feeling is that since this does not affect lots of users and is something we've lived with for a long time, we should probably not expend resources on a fix for the patch release (especially given that this more general fix will be forthcoming in 1.1).

PR #1096, opened 2017-03-29
Cast pDev for DevDescVersion12

javierluraschi:
@kevinushey I'm hitting this build error, looks to me like `pDev` needs a cast.


kevinushey:
Yes -- thanks for catching that!

PR #1108, opened 2017-04-04
Fix bug that assumes all libraries have at least two packages

andrie:
This is a fix for the bug originally reported at https://support.rstudio.com/hc/communities/public/questions/203052576-setting-libpath-in-Rstudio

To replicate the error, use:

```r
loc <- normalizePath(tempdir())
assign(".lib.loc", loc, envir = environment(.libPaths))
```

The resulting error message is:

```
Error in pkgs.name[[i]] : subscript out of bounds
```

This happens because the code simplifies the result of `installed.packages()`, and if the library contains zero or one row, the simplification results in a vector, not a data frame.

The fix is to set `drop=FALSE` when returning columns of the data frame.

This bug manifests in the [`checkpoint` package on CRAN](https://cran.r-project.org/package=checkpoint). In the package is a workaround to fix this bug, and this has been working without problem for the best part of two years.  See the [checkpoint repo](https://github.com/RevolutionAnalytics/checkpoint/blob/master/R/fixRstudioBug.R).

### DIFF
--- a/src/cpp/session/modules/SessionPackages.R
+++ b/src/cpp/session/modules/SessionPackages.R
@@ -312,7 +312,7 @@ if (identical(as.character(Sys.info()["sysname"]), "Darwin") &&
                          desc=pkgs.desc,
                          url=pkgs.url,
                          loaded=pkgs.loaded,
-                         check.rows = FALSE,
+                         check.rows = TRUE,
                          stringsAsFactors = FALSE)
 
    # sort and return
--- a/src/cpp/session/modules/SessionPackages.R
+++ b/src/cpp/session/modules/SessionPackages.R
@@ -298,12 +298,9 @@ if (identical(as.character(Sys.info()["sysname"]), "Darwin") &&
 
    # build up vector of package versions
    instPkgs <- as.data.frame(installed.packages(), stringsAsFactors=F)
-   pkgs.version <- character(length=length(pkgs.name))
-   for (i in 1:length(pkgs.name)) {
-      pkgs.version[[i]] <- .rs.packageVersion(pkgs.name[[i]],
-                                              pkgs.library[[i]],
-                                              instPkgs)
-   }
+   pkgs.version <- sapply(seq_along(pkgs.name), function(i){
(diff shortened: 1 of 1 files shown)


jjallaire:
@kevinushey This LGTM. Could you review and merge?

src/cpp/session/modules/SessionPackages.R
```diff
@@ -315,7 +315,7 @@ if (identical(as.character(Sys.info()["sysname"]), "Darwin") &&
                          desc=pkgs.desc,
                          url=pkgs.url,
                          loaded=pkgs.loaded,
-                         check.rows = TRUE,
+                         check.rows = FALSE,
```
kevinushey: What's the motivation for this particular change?

andrie: I'll be honest that I don't understand why this is necessary. This argument checks that row names are consistent when creating a `data.frame`, and something seems to go wrong when checking row names on an empty data frame.

But I do know that I could not get this function to work in any other way.

kevinushey:
FWIW I still see the bug with your reproducible example (thanks for providing one!)

```
Error in pkgs.name[[i]] : subscript out of bounds
```

and I think it's because this code here:

https://github.com/rstudio/rstudio/blob/538e0e4eddf3c92849f8c27df699d1410fc68a1a/src/cpp/session/modules/SessionPackages.R#L302

assumes that there is at least one package in the specified library paths, whereas the provided test case simulates a library path with no packages. I think that implies that this PR needs a few more changes before we can merge. (we should probably use `seq_along()` in the for loop there)

andrie:
Agreed. I have a fix. Coming your way shortly.

andrie:
So the fix is to replace that for loop (ugh!) with `sapply()`:

```r
pkgs.version <- sapply(seq_along(pkgs.name), function(i){
        .rs.packageVersion(pkgs.name[[i]], pkgs.library[[i]], instPkgs)
      })
```

And now the `check.rows=FALSE` is no longer necessary, so I reverted that.

kevinushey:
Awesome, thanks for that! LGTM.

PR #1113, opened 2017-04-06
Initial support for new themes

javierluraschi:
There is still more work for this feature, but it would be worth to work directly on main for those eager to provide feedback. Most changes tweak the appearance in CSS files, might start pushing directly to main since most changes are scoped to css tweaks.

Here is the preference entry point (disable by default). Which we can also consider removing this at some point once this work is complete:

screen shot 2017-04-05 at 5 15 24 pm

and a preview of this work:

screen shot 2017-04-05 at 5 13 47 pm


jjallaire:
Yeah, I'd be in favor of ditching the check box as soon as you've got the new theme looking reasonable.

PR #1118, opened 2017-04-07
Implements DSNs for new connections wizard

javierluraschi:
Implement DSNs for new connections wizard:

screen shot 2017-04-06 at 5 31 40 pm

screen shot 2017-04-06 at 5 31 49 pm

This assuming the `ODBC.ini` file is configured as:

```
[Customers]
Description     = Test to Postgres
Driver          = PostgreSQL

[Supliers]
Description     = Test to Postgres
Driver          = PostgreSQL

[Invoices]
Description     = Test to Postgres
Driver          = PostgreSQL
```

### DIFF
--- a/src/cpp/session/modules/SessionConnections.R
+++ b/src/cpp/session/modules/SessionConnections.R
@@ -259,7 +259,31 @@ options(connectionObserver = list(
 
    snippets <- .rs.connectionReadSnippets()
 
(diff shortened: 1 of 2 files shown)


jjallaire:
I think Livy should be part of the Spark workflow rather than its own entry (end users aren't likely to know what Livy is but will know "Spark")

PR #1123, opened 2017-04-12
Support ANSI color codes in Console

gtritchie:
Added console support for most commonly used ANSI codes for colors (plus bold, underline, hidden, and inverse).

Also, the console will strip out any other ANSI codes that it doesn't support.

Added environment variable RSTUDIO_CONSOLE_COLOR to indicate if running inside RStudio with color console support.


jmcphers:
When packages need to know about RStudio capabilities, we generally use the `rstudioapi` package: https://github.com/rstudio/rstudioapi

jmcphers:
Also, should we set `TERM` to a value which indicates that these sequences are supported? I'm thinking of the case wherein someone uses `system()` to run a program which emits colored output when supported. It's not worth spending a lot of time, but there's a way we can advertise this capability to spawned processes (without also making them them think we're a full-service terminal emulator), we should do so.

gtritchie:
Should I add something to `rstudioapi` or just let packages use existing version detection and assume 1.1 builds have color? Only issue being earlier pre-release builds, of course.

jmcphers:
You don't need to add anything to rstudioapi; the crayon package can ask if a specific version of RStudio is running via `isAvailable` (it can include the patch number, too, to avoid marking earlier pre-release builds) https://github.com/rstudio/rstudioapi/blob/01b4a360253cd0aef96853ae7002b8a58710ba33/R/code.R#L11

As an alternative, if you wanted to make this a little more flexible, you could add an RStudio API function which can be called to get info about the R console capabilities, and then use `haveFun("...")` instead of `isAvailable`.

gtritchie:
Setting TERM is generally not sufficient since programs will also look to see if they are outputting to a terminal (isatty). However, some commands like "ls" obey the semi-standard CLICOLOR_FORCE environment variable (TERM must also be set to something capable of color). I set TERM=vt100 and CLICOLOR_FORCE=1 and "ls -lGa" indeed shows nice color output when invoked with system().

gtritchie:
With TERM and CLICOLOR_FORCE set, ls shows color in console:

jmcphers:
That's great! We may have to reverse course on `TERM` since IIRC we don't actually support everything in the `terminfo` for `vt100-256color`, which could be confusing to some programs. Thanks for adding an option for the `TERM` setting.

gaborcsardi:
Yeah, imo it is better not to temper `TERM`.

PR #1126, opened 2017-04-12
implement tree-based object explorer

kevinushey:
This PR introduces a drill-down object explorer, primarily intended for inspection of highly-nested objects (lists, environments, and so on).

Some examples of what this looks like currently:

screen shot 2017-04-12 at 4 38 43 pm

screen shot 2017-04-12 at 4 39 53 pm

---

## TODO

- [x] Ability to generate code that extracts particular node in tree.
- [x] Use appropriate icons given object type.
- [x] Button in environment pane for inspecting lists, environments, S4 objects, ...
- [x] Consider more filters + controls.
- [ ] Consider how to open this up for extension by users (e.g. allow custom classes to define how this is populated)
- [x] Properly handle persistence across sessions (so we can close + re-open the IDE and re-inspect an old object)
- [ ] Ensure that we handle 'exotic' items (active bindings, recursive environments, promises, ...). For active bindings, we might want some UI that allows users to re-trigger the binding.
- [ ] Re-update the view if the underlying object changes.
- [ ] Consider how to handle very large tables (pagination)


jjallaire:
Looking great!  A couple comments:

1. The toolbar stuff (filter, etc.) should go on a top toolbar rather than on the bottom.

2. Let's definitely write this up for Paul so he can give us his take on how we can improve the UX.

kevinushey:
Current status:

screen shot 2017-04-24 at 10 07 52 am

I think this is ready for review now, although there is still a bit of extra feature work + testing required. Hadley indicated some interest in demo-ing this tomorrow, so it'd be great if we could get this in as part of the next daily build. User feedback will also help in motivating next steps as well.

jjallaire:
As painful as giving up the pixels might be I think we need a bit more indentation to make the hierarchy clear at a glance. Maybe 2 more pixels?

kevinushey:
That sounds good to me. With a bit more indentation:

screen shot 2017-04-24 at 10 19 01 am

kevinushey:
We might also want to consider drawing some small, grey indent guides (e.g. as one might see when executing `tree` from the command line)

jjallaire:
Looks good! I think we're okay without the indentation guides.

src/cpp/session/modules/SessionObjectExplorer.cpp
```diff
@@ -0,0 +1,226 @@
...
+{
+   Error error;
+   
+   // if we don't have a cache, nothing to do
+   if (!explorerCacheDir().exists())
+      return;
+   
+   // list source documents
+   std::vector<boost::shared_ptr<source_database::SourceDocument> > docs;
+   error = source_database::list(&docs);
```
jmcphers: This is somewhat expensive as it has to crawl the entire source database. In this case all we need is a list of the IDs in the source database (a happy result of using the ID as the cache key!) so could simply get the list of IDs via something like `source_database::path().children(&files);` (maybe wrapped behind an accessor)

src/gwt/src/org/rstudio/core/client/JsVector.java
```diff
@@ -75,7 +75,10 @@ public final native T get(int index, T defaultValue)
    
    public final native int indexOf(T value)
    /*-{
-      return this.indexOf(value);
+      for (var i = 0, n = this.length; i < n; i++)
+         if (this[i] == value)
```
jmcphers: Nit: should this use `===`?

src/gwt/src/org/rstudio/studio/client/workbench/views/environment/view/RObjectEntry.java
```diff
@@ -96,9 +98,29 @@ public String getDisplayValue()
...
+            return true;
+      return false;
+   }
 
    public static final String NO_VALUE = "NO_VALUE";
-
+   
+   private static final Set<String> DATA_CLASSES = new HashSet<String>();
+   static {
+      DATA_CLASSES.addAll(ListUtil.create(
```
jmcphers: Should this include and/or share the elements of this list? https://github.com/rstudio/rstudio/blob/54cd3abcfc58837b433464c793fe9b03a87f0bb4/src/gwt/src/org/rstudio/studio/client/workbench/views/environment/view/RObjectEntry.java#L69-L74

src/gwt/src/org/rstudio/studio/client/workbench/views/source/Source.java
```diff
@@ -970,7 +975,46 @@ public void onResponseReceived(SourceDocument response)
...
    }
+   
+   @Override
+   public void onOpenObjectExplorerEvent(OpenObjectExplorerEvent event)
+   {
+      // ignore if we're a satellite
+      if (!SourceWindowManager.isMainSourceWindow())
+         return;
+    
+      Debug.logToRConsole("Opening object explorer");
```
jmcphers: Nit: Remove debug logging

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/explorer/ObjectExplorerEditingTarget.java
```diff
@@ -0,0 +1,156 @@
...
+import org.rstudio.studio.client.workbench.views.source.editors.explorer.model.ObjectExplorerHandle;
+import org.rstudio.studio.client.workbench.views.source.editors.explorer.view.ObjectExplorerEditingTargetWidget;
+import org.rstudio.studio.client.workbench.views.source.editors.urlcontent.UrlContentEditingTarget;
+import org.rstudio.studio.client.workbench.views.source.events.PopoutDocEvent;
+import org.rstudio.studio.client.workbench.views.source.model.SourceServerOperations;
+
+public class ObjectExplorerEditingTarget
+      extends UrlContentEditingTarget
+{
+   enum QueuedRefreshType
```
jmcphers: Nit: Remove unused enum. If you want to inherit the refresh behavior of the data viewer, it might be worth considering a base class here. However this exists primarily to make sure that we don't do expensive operations when the viewer is not visible so may be less relevant in this case.

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/explorer/ObjectExplorerPresenter.java
```diff
@@ -0,0 +1,86 @@
...
+      
+      events_.addHandler(ObjectExplorerEvent.TYPE, this);
+   }
+   
+   // Handlers ----
+   
+   @Override
+   public void onObjectExplorerEvent(ObjectExplorerEvent event)
+   {
+      Debug.logToRConsole("Responding to object explorer event");
```
jmcphers: Nit: Debug logging (and elsewhere)

src/cpp/session/modules/SessionObjectExplorer.cpp
```diff
@@ -0,0 +1,226 @@
...
+      }
+   }
+}
+
+void onShutdown(bool terminatedNormally)
+{
+   if (!terminatedNormally)
+      return;
+   
+   removeOrphanedCacheItems();
```
jmcphers: Nit: It'd be slightly more efficient to remove cache entries at the time the tab is closed (instead of waiting and then sweeping them all up at shutdown). Up to you.

src/cpp/session/modules/SessionObjectExplorer.cpp
```diff
@@ -0,0 +1,226 @@
...
+   return module_context::sessionScratchPath()
+         .childPath(kExplorerCacheDir);
+}
+
+std::string explorerCacheDirSystem()
+{
+   return string_utils::utf8ToSystem(explorerCacheDir().absolutePath());
+}
+
+void removeOrphanedCacheItems()
```
kevinushey: Will do! Initially, I intended to attach cache data items to the `.rs.CachedDataEnv`, but try as I might I couldn't convince RStudio not to delete my R data objects on suspend / resume / shutdown.

src/cpp/session/modules/SessionObjectExplorer.cpp
kevinushey: Good point! I'll make that change.

kevinushey: Ah, I realize why I did this now -- currently, the cache uses different cache keys than the actual document id -- so the `"id"` document property is not the same thing as the actual document id; that is, the `"id"` document property is really the cache key. I'll try to make this change.

src/cpp/session/modules/SessionObjectExplorer.cpp
kevinushey: Agreed, although I think it makes sense to keep this here just in case (ie, to clean up just in case somehow things get out of sync)

src/gwt/src/org/rstudio/core/client/JsVector.java
kevinushey: Yes, it should!

src/gwt/src/org/rstudio/core/client/SafeHtmlUtil.java
```diff
@@ -21,6 +21,67 @@
 
 public class SafeHtmlUtil
 {
+   public static class TagBuilder
```
kevinushey: Note to self: GWT has its own HTML builders (e.g. see the `DivBuilder` class) and we can probably just use that instead.

src/gwt/src/org/rstudio/studio/client/workbench/views/environment/view/RObjectEntry.java
kevinushey: Good catch; I think you're right. (I initially thought that `class()` would return the base class for e.g. S4 `DataFrame` objects, thereby ensuring that the plain old `data.frame` check would succeed, but that is not the case -- only `DataFrame` is returned.)

kevinushey:
Alright, I think this is good to go for now (with more pending work items to be done on master). @jmcphers, can you review and merge if all looks well?

PR #1133, opened 2017-04-17
Flat dialogs for flat theme

javierluraschi:
My original plan was to remove shadow, gradients and use css borders. However, not having a shadow made the UI look off, instead, I'm reducing the shadow effect significantly but enough to add volume to the dialog.

### Browser

screen shot 2017-04-17 at 4 08 39 pm
screen shot 2017-04-17 at 4 08 52 pm
screen shot 2017-04-17 at 4 09 08 pm
screen shot 2017-04-17 at 4 09 22 pm
screen shot 2017-04-17 at 4 09 30 pm

### Desktop

screen shot 2017-04-17 at 4 01 34 pm
screen shot 2017-04-17 at 4 01 43 pm
screen shot 2017-04-17 at 4 01 51 pm
screen shot 2017-04-17 at 4 02 03 pm
screen shot 2017-04-17 at 4 02 17 pm

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/theme/res/themeStyles.css
+++ b/src/gwt/src/org/rstudio/core/client/theme/res/themeStyles.css
@@ -1672,7 +1672,7 @@ body.ubuntu_mono .searchBox {
    border-radius: 5px 5px 5px 5px;
    background: #F3F4F4;
    overflow: hidden !important;
-   box-shadow: 0 2px 20px rgba(0, 0, 0, 0.25);
+   box-shadow: 0 2px 10px rgba(0, 0, 0, 0.15);
 }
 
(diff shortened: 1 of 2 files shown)


jmcphers:
Seems pretty reasonable, especially since even operating systems that use the flat look still provide some shadows to hint at window depth.

Are we still planning to remove the 9-grid borders as part of the theming work?

javierluraschi:
@jmcphers they are "removed", as in, we render them with `display: none` which also fixes the 1px bug we intermittently hit. Removing the actual elements is challenging since GWT provide those elements so we would have to subclass the dialog. We might have to end up doing that if we want to support resizing, but for now, they are not being rendered.

jjallaire:
I agree with the approach of display: none to keep the code churn minimal.

PR #1137, opened 2017-04-18
Add support to test snippets connection

javierluraschi:
screen shot 2017-04-18 at 1 12 54 pm

screen shot 2017-04-18 at 1 31 52 pm

screen shot 2017-04-18 at 1 30 16 pm

screen shot 2017-04-18 at 1 37 17 pm

### DIFF
--- a/src/cpp/session/modules/SessionConnections.R
+++ b/src/cpp/session/modules/SessionConnections.R
@@ -517,4 +517,34 @@ options(connectionObserver = list(
    .rs.success()
 })
 
+.rs.addJsonRpcHandler("connection_test", function(code) {
+   error <- ""
 
+   oldConnections <- lapply(ls(.rs.activeConnections), function(name) {
+      get(name, envir = .rs.activeConnections)
+   })
+
+   .envir <- parent.frame(2)
+   tryCatch({
+      eval(parse(text = code), envir = .envir)
+   }, error = function(e) {
+      error <<- e$message
+   })
+
+   if (nchar(error) == 0) {
+      newConnections <- lapply(ls(.rs.activeConnections), function(name) {
+         get(name, envir = .rs.activeConnections)
+      })
+
+      oldConnectionIds <- lapply(oldConnections, function(x) paste(x$type, x$host, sep = "|"))
+      needDisconnect <- Filter(function(e) {
+         !paste(e$type, e$host, sep = "|") %in% oldConnectionIds
+      }, newConnections)
+
+      lapply(needDisconnect, function(e) {
+         .rs.connectionDisconnect(e$type, e$host)
+      })
+   }
+
+   .rs.scalar(error)
+})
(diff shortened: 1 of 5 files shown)


src/cpp/session/modules/SessionConnections.R
```diff
@@ -517,4 +517,34 @@ options(connectionObserver = list(
...
+   }, error = function(e) {
+      error <<- e$message
+   })
+
+   if (nchar(error) == 0) {
+      newConnections <- lapply(ls(.rs.activeConnections), function(name) {
+         get(name, envir = .rs.activeConnections)
+      })
+
+      oldConnectionIds <- lapply(oldConnections, function(x) paste(x$type, x$host, sep = "|"))
```
jmcphers: It will be hard to undo all the side effects of having the IDE pick up and store the connection; for instance, it will still wind up in the connection history, which we don't want if it's not a valid connection. 

Instead I think we should have the IDE completely ignore connections which are created just to test connectivity. This could happen here:

https://github.com/rstudio/rstudio/blob/0aaa25abe98e8a48f6135f984c2471dcf889dfe0/src/cpp/session/modules/SessionConnections.R#L75

We could use the environment or assigned variable name as a flag to indicate a test connection, or just check to see if we're inside this function.

javierluraschi: Good point, yes, we don't want does persisting while testing connections.

src/cpp/session/modules/SessionConnections.R
```diff
@@ -517,4 +517,36 @@ options(connectionObserver = list(
...
+         disconnectCalls <<- c(disconnectCalls, disconnect)
+      },
+      connectionClosed = function(type, host, ...) {
+
+      },
+      connectionUpdated = function(type, host, hint, ...) {
+      }
+   ))
+
+   .envir <- parent.frame(2)
```
kevinushey: If the intention is to evaluate this within the 'active' frame, you might instead want to use `.rs.getActiveFrame()`, which should make sure this works in the (rare) case where e.g. the R debugger is active or similar.

kevinushey:
I'd suggest tweaking the dialog titles to e.g.

```
Successfully Established Connection
Failed to Establish Connection
```

src/cpp/session/modules/SessionConnections.R
javierluraschi: @kevinushey based this code of https://github.com/rstudio/rstudio/blob/8b1778b0a62e989c461ca5e79eaf585f5982744d/src/cpp/session/modules/SessionCodeTools.R#L611-L630 but it looks like getActiveFrame would handle cases where one is debugging and tries to test a connection. Changed.

javierluraschi:
@kevinushey How about "Test Results"? Seems a bit more descriptive and consistent with "New Connection":

screen shot 2017-04-19 at 2 43 15 pm

javierluraschi:
I think we can merge this one... I can follow up after merge if there is additional feedback.

PR #1138, opened 2017-04-18
Separate terminal pref pane, reduce pane selector height

gtritchie:
- Move terminal preferences to a new pane in Global Settings, to avoid vertical overflow of "General" pane on Pro and to make space for future terminal options.
- Create separate terminal preferences JSON object instead of using general object
- placeholder icon for Terminal pane (one of my finer works of art and that's not saying much)
- Always show the "Default terminal shell" dropdown, even on non-Windows; it currently only shows Bash but may be extended in the future to allow users to configure different shells
- If terminal feature is disabled via the "allow-shell" feature limit, hide the preferences controls and display a message in the pane ("This feature has been disabled by your administrator.")


jmcphers:
In my view it's misleading to show an option to change the shell when it's not possible to do so (this will lead to questions like "why don't I see my shell in the list?" "how do I add my shell to the list?"). Let's leave that out unless you are planning to add support for other shells soon (we can always add it back).

I realize the pane looks pretty sparse with only a single option. We might fill it out a little by adding a short 1-sentence description of the terminal feature above the preferences, and a help link to a document where we describe how to use it (below the preferences). 

Finally, what would you think about just hiding the terminal tab entirely if the terminal is disabled? This is generally the approach we take with other disabled features (i.e. just don't show them at all.)

jmcphers:
By the way, I'm a big fan of the treatment for the preferences panel -- that saves a lot of vertical space without compromising organization. Nice work!

kevinushey:
Very minor, but can you try nudging the icons in the pane down a pixel (or two)? Right now they appear to be a bit 'too high' relative to their accompanying text.

gtritchie:
Visual updates (recommendations on text welcomed):
Without dropdown (everything but Win32):

With dropdown (Win32):

jmcphers:
Nice! I think we should leave the help link off however until we actually have something to link to (we can write a support article over the next few weeks once we're confident in the feature set).

PR #1151, opened 2017-04-25
handle ANSI color codes split across multiple VirtualConsole submits

gtritchie:
- if ANSI code is partially sent at end of one VirtualConsole submit, then the rest of the codes come in the next submit, need to store the partial code and match it when rest arrives
- also handle cases where ANSI code is complete but at end of a submit, and text to be styled comes at start of next submit call
- added support to VirtualConsole class to turn off color support, either by having it parse and discard ANSI codes, or just letting them pass through (no way to set this option to anything but "color on" yet, but will be adding a user-option)
- unit testing for new cases, and for the VirtualConsole color support setting
- show readable version of control-characters for local-echo matching diagnostics (sent to Javascript console)


src/gwt/src/org/rstudio/core/client/VirtualConsole.java
```diff
@@ -39,17 +40,38 @@
...
    
    public VirtualConsole(Element parent)
    {
+      this(parent, ANSI_COLOR_ON);
+   }
+
+   /**
+    * VirtualConsole constructor
+    * @param parent parent element
+    * @param showAnsiColors if true, apply visual styles for AnsiColor escape
```
kevinushey: nit: documentation seems to imply this is a boolean, but is really an `enum`-flavored `int`

gtritchie:
Merging. Changes are covered by unit tests and limited to the console ansi-color scenario.

PR #1160, opened 2017-04-28
implement virtualization for data grid; use with object explorer

kevinushey:
NOTE: This PR is not quite yet ready to go, but is ready for some initial review.

This PR implements DOM virtualization of a GWT DataGrid, ensuring only that the currently visible rows are drawn when rendering the data grid, and that the DOM is refreshed in response to scroll events. The goal is to allow the user to scroll through very large R objects without too much IDE lag when many rows are available.

We implement virtualization in the following way:

1. After the user has scrolled, we figure out what rows are currently in view, and how many rows are currently not visible above / below the current viewport. We tell GWT to only draw the visible rows (+ a bit of extra padding rows, for scrolling).

2. We add a top + bottom rows for padding, and set their heights such that the scroll position represented to the user matches what they would've seen if we had just drawn the row. (E.g. instead of drawing 100 rows of height 24px, we just draw 1 row of height 2400px).

One other major performance optimization -- I now use GWT sprites in the object explorer view, rather than plain old `` -- previously, each attempt to append an image in this way would actually trigger a browser `GET` request, requiring a reparse of the base64 image data included in the `src` tag! Moving to GWT Sprites (using CSS `background-image`) seems to circumvent this.


jmcphers:
LGTM so far.

kevinushey:
Okay, I think this is good to go.

I really wanted to get on-the-fly redrawing of rows during scrolling to work seamlessly, but unfortunately all the schemes for updating I tried introduce a stutter that's just annoying enough to offset the usefulness -- hence the somewhat large delay in updating on scroll. I think this is ultimately a consequence of the fact that we have to re-build the table on each re-draw, which implies re-generating the whole HTML table (as a string) and then waiting for the browser to re-parse and re-display that HTML. I'm not yet sure if there's a DataGrid-y way around this.

We might want to tweak this further in the future.

jjallaire:
I was under the impression that the GWT DataGrid automatically did this
sort of virtualization but I guess they don't :-\

kevinushey:
DataGrid does pagination rather than true virtualization, and pagination doesn't really lead to a great experience for a view like this.

jjallaire:
Ah, got it!

kevinushey:
I'm going to merge this given @jmcphers initial review + since the newest changes are not particularly risky / contentious.

PR #1164, opened 2017-05-02
Support new readxl features and help links

javierluraschi:
This PR adds support for `n_max` and `range` fields available now in `readxl 1.0`, disabled unsupported fields for older version of `readxl `. Refactored help links from `Wizard` to `ModalBase` and added links as follows:
- Reading Excel files using readxl -> http://readxl.tidyverse.org/
- Reading rectangular data using readr -> http://readr.tidyverse.org/
- Reading statistical data using haven -> http://haven.tidyverse.org/

@jennybc Feel free to post feedback on this PR or after trying this out in tomorrows build, thanks!

screen shot 2017-05-02 at 1 42 38 pm

`n_max` and `skip` get disabled when `range` is specified.

screen shot 2017-05-02 at 1 27 45 pm

`n_max` and `range` are disabled when `readxl < 1.0` is installed.

screen shot 2017-05-02 at 1 47 02 pm

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/environment/dataimport/DataImportOptionsUiSav.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/environment/dataimport/DataImportOptionsUiSav.java
@@ -124,7 +124,7 @@ public class DataImportOptionsUiSav extends DataImportOptionsUi
    public HelpLink getHelpLink()
    {
       return new HelpLink(
-         "Reading statistical data using haven",
+         "Reading data using haven",
          "import_haven",
          false,
          true);
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/environment/dataimport/DataImportOptionsUiSav.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/environment/dataimport/DataImportOptionsUiSav.java
@@ -125,9 +125,9 @@ public class DataImportOptionsUiSav extends DataImportOptionsUi
    {
       return new HelpLink(
          "Reading statistical data using haven",
-         "http://haven.tidyverse.org/",
+         "import_haven",
          false,
-         false);
+         true);
    }
    
    @UiFactory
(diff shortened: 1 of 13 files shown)


src/gwt/src/org/rstudio/studio/client/workbench/views/environment/dataimport/DataImportOptionsUiSav.java
```diff
@@ -119,6 +120,16 @@ public void setImportLocation(String importLocation)
       }
    }
    
+   @Override
+   public HelpLink getHelpLink()
+   {
+      return new HelpLink(
+         "Reading statistical data using haven",
```
kevinushey: I'm sort of iffy about calling this 'statistical data' (since it somewhat implies the data you can't read with haven is not statistical).

What about something like:

```
Reading data using haven
Reading foreign datasets using haven
```

or otherwise?

jennybc:
Very minor point: I think `readr` is used a couple times in PR description when `readxl` is the intent.

Just based on screenshots:

It's possible all the "Import Options" should bear the exact name of the `read_excel()` argument they correspond to. The one that stands out is `Max`, which is bound to `n_max`, and should not be confused with the unexposed argument `guess_max`. And if you do it for `n_max`, maybe it's better to be consistent across the board?

I look forward to giving it a whirl!

src/gwt/src/org/rstudio/studio/client/workbench/views/environment/dataimport/DataImportOptionsUiSav.java
javierluraschi: Yes, makes sense! Changed.

javierluraschi:
@jennybc thanks you, updated PR, changed `Max` to `N Max`. Other import dialogs use the verbose names, so we would have to also change those to be consistent. We can iterate on this label if we think of a better name, ping me in Slack and I'll push the change.

PR #1176, opened 2017-05-05
Several terminal-related rstudioapi's

gtritchie:
First batch of terminal rstudioapi's (I have parallel work happening in the rstudioapi project). Have a few more APIs to add, which will be relatively easy now that I've done these.

With full API set, will be easy to create, display, and close terminal instances, send text to terminals, and read text from terminals, in R. Neato. Also handy for some scripted testing via driving the UI.

In these APIs, terminal instances are identified by their caption (the string that shows up in the terminal dropdown, e.g. Terminal 1).

- `clearTerminal(id = NULL)` -- clear specified terminal buffer; NULL clears current active terminal
- `getActiveTerminalId()` -- returns id of active terminal or null if no terminals are open
- `sendToTerminal(text, id = NULL)` -- send text to specified terminal or default terminal if id = NULL
- `createTerminal(id = NULL)` -- create a terminal with specified ID (caption), passing NULL creates a default caption (Terminal 3, etc.); returns terminal ID created or NULL if unable to create terminal because caption already exists


jmcphers:
Very cool! I think these will enable some interesting add-ins. 

Can you elaborate on the `waitForMethod` approach? I'm a little concerned that there are failure modes that will leave our RPC thread dead in the water if for some reason the client doesn't do its end of the terminal creation. It seems like this is generally done in the service of being able to synchronously return the terminal ID that was created. Maybe we could structure the API in such a way that we don't need to do that, or arrange the knowledge of terminal IDs on the client and server such that it's possible to obtain a terminal ID without a client/server conversation that involves blocking a thread? 

(Not totally opposed if it turns out waitForMethod is best here!)

gtritchie:
Yes, was also a bit worried about that. It's a pattern used elsewhere, but perhaps not in such a complex orchestration back and forth between client/server multiples times in order to ultimately give a result.

Yes, the purpose is facilitate creating a terminal without specifying the terminal ID and directly getting back the one created for you.

Perhaps a better model would be to have the server create the terminal directly, then issue a client-event telling the client to connect to it. That would have the rather neat benefit of allowing terminals to be created successfully via API even if the client isn't currently attached.

Let me ponder that some more. I've got more APIs coming in so no rush to get the PR in. Mainly wanted some feedback on what I've got so far. Thanks!

jjallaire:
Another way to go would be to generate a unique terminal ID before it's
actually created, and then queue any subsequent requests that use that ID
until the terminal is ready.

gtritchie:
Have removed the use of waitForMethod when creating a new terminal. Also added remainder of planned APIs; current list from rstudioapi perspective is below.

One issue I still need to address is that certain operations are overly dependent on the IDE being connected to a terminal, and if a session has suspended and resumed, terminals don't reconnect until they are displayed by clicking on the terminal tab, then selecting that terminal in the dropdown.

## APIs
- `createTerminal(id = NULL)`: create a new terminal using supplied id, or generate one if no id is supplied; returns the id in either case
- `sendToTerminal(id, text)`: send text to a terminal
- `getAllTerminals()`: returns all terminal identifiers
- `clearTerminal(id)`: clear a terminal
- `isTerminalBusy(id)`: get busy state of one or more terminals
- `getTerminalContext(id)`: get a bunch of metadata about one or more terminals as a list of lists
-  `showTerminal(id)`: bring one terminal to the front in the UI
- `getTerminalBuffer(id, stripAnsi = FALSE)`:  return terminal buffer, optionally stripping out Ansi escape sequences
- `killTerminal(id)`: kill one or more terminals (without any prompting if they are busy)

PR #1187, opened 2017-05-10
use 'etc/ldpaths' for rJava on macOS as well

kevinushey:
This PR fixes an issue where `rJava` can fail to load when running with R 3.4.0 under RStudio on macOS.

Although this issue was originally reported by a Homebrew R user, one can actually reproduce this with R 3.4.0 installed from CRAN, using the CRAN `rJava` binary as well. Compare:

```
kevin@cdrv:~/Library/R/3.4/library/rJava/libs
$ otool -L rJava.so
rJava.so:
        rJava.so (compatibility version 0.0.0, current version 0.0.0)
        @rpath/libjvm.dylib (compatibility version 1.0.0, current version 1.0.0)
        /Library/Frameworks/R.framework/Versions/3.4/Resources/lib/libR.dylib (compatibility version 3.4.0, current version 3.4.0)
        /System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 1259.0.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1226.10.1)
```

```
kevin@cdrv:~/Library/R/3.3/library/rJava/libs
$ otool -L rJava.so
rJava.so:
        rJava.so (compatibility version 0.0.0, current version 0.0.0)
        /System/Library/Frameworks/JavaVM.framework/Versions/A/JavaVM (compatibility version 1.0.0, current version 1.0.0)
        /Library/Frameworks/R.framework/Versions/3.3/Resources/lib/libR.dylib (compatibility version 3.3.0, current version 3.3.0)
        /System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 855.17.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1197.1.1)
```

If I understand correctly, `rJava` with R 3.3.x (and below) uses the system Java installation, while newer versions attempt to use `libjvm` from the `@rpath`. This is eventually discovered on the `DYLD_FALLBACK_LIBRARY_PATH` which is set when R is run from the console, but since RStudio (prior to this PR) was not harvesting this environment variable, it would now fail in RStudio.

Altogether, I think this implies that this PR could actually be a candidate for the patch release, since the underlying issue is that, with R 3.4.0 from CRAN, `rJava` does not work by default under RStudio.

### DIFF
--- a/src/cpp/core/r_util/REnvironmentPosix.cpp
+++ b/src/cpp/core/r_util/REnvironmentPosix.cpp
@@ -70,18 +70,39 @@ FilePath scanForRScript(const std::vector<std::string>& rScriptPaths,
    return FilePath();
 }
 
+// extra paths from R (for rJava)
+std::string extraLibraryPaths(const FilePath& ldPathsScript,
+                              const std::string& rHome)
+{
+   // no-op if no script is passed
+   if (ldPathsScript.empty())
+      return std::string();
+
+   // verify that script exists
+   if (!ldPathsScript.exists())
+   {
+      LOG_WARNING_MESSAGE("r-ldpath script not found at " +
+                          ldPathsScript.absolutePath());
+      return std::string();
+   }
+
+   // run script to capture paths
+   std::string command = ldPathsScript.absolutePath() + " " + rHome;
+   system::ProcessResult result;
+   Error error = runCommand(command, core::system::ProcessOptions(), &result);
+   if (error)
+      LOG_ERROR(error);
+   std::string libraryPaths = result.stdOut;
+   boost::algorithm::trim(libraryPaths);
+   
+   return libraryPaths;
+}
+
 // MacOS X Specific
 #ifdef __APPLE__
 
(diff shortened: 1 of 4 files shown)


kevinushey:
I think this is good to go now -- there was a bit of munging around computing the `r-ldpath` script path for Debug vs. Release configurations, but I think that should be sorted out now.

jjallaire:
Note that we also set this variable here: https://github.com/rstudio/rstudio/blame/af72b0a9121d59ae445eb0914c7dd3a1366442aa/src/cpp/core/r_util/RVersionsPosix.cpp#L189

We should make sure that this new code doesn't conflict.

kevinushey:
There shouldn't be any conflict there -- the only change is that this code will also now pick up the modifications to `DYLD_FALLBACK_LIBRARY_PATH` introduced by R's `etc/ldpaths` script, which I believe is desired.

PR #1189, opened 2017-05-10
Use textbox for NAs while importing data from Excel

javierluraschi:
@jennybc the `NA`s as textbox change you suggested, double checking before the merge.

screen shot 2017-05-10 at 1 47 18 pm

screen shot 2017-05-10 at 1 47 26 pm


jennybc:
Looks good!

PR #1190, opened 2017-05-11
activateTerminal API work, error handling

gtritchie:
- renamed `showTerminal` api to `activateTerminal`; it can be used both to ensure a terminal's process is started, and optionally for showing the terminal in the UI; the former is still a work-in-progress, working through glitches
- per JJ's recommendation made terminal APIs visible without needing `rstudioapi`, via `.Options$terminal.manager`
- added some extra error logging around some cases I've seen before and during this current work where server gets out of sync with client on console-processes and throws errors
- make it so closing a terminal in client works even if the latter issue has happened (server doesn't know about the process anymore), now it will go away in the client without showing cryptic error message
- centralized code that constructs ProcessOption structure for terminals, so it can be used both in starting a new terminal via client UI, and when using `activateTerminal` api
- remove the `websockets` flag from startTerminal RPC call, as the server already has access to the information needed to determine that setting

### DIFF
--- a/src/cpp/r/R/Api.R
+++ b/src/cpp/r/R/Api.R
@@ -434,6 +434,13 @@
    .Call("rs_isTerminalBusy", id)
 })
 
+.rs.addApiFunction("isTerminalRunning", function(id) {
+   if (is.null(id) || !is.character(id))
+      stop("'id' must be a character vector")
+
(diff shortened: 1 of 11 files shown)


gtritchie:
More fixes coming, will reopen PR after another round.

PR #1194, opened 2017-05-11
add ability to select next, previous marker / search result

kevinushey:
NOTE: Not yet ready for merge.

This PR attempts to add commands for selecting the next / previous output as part of the currently active tab in the console pane. The intention is to provide a command / keyboard shortcut that can be used to e.g. select the next marker output, or select the next find result, and so on.

Unfortunately, I don't understand the wiring between tabs and panes enough to understand how I can best write code that targets whatever tab is currently active. (As far as I can see, panes that live in the console have no notion of which tab owns them, nor whether that tab is currently active, and so all the solutions I can think of thus far will imply loads of code duplication, which I'd like to avoid).

@jmcphers @jjallaire if you can suggest a way forward, that would be hugely appreciated. The overall goal is to find a way of dispatching a command handler to whichever pane is associated with the active tab in the console pane.


jjallaire:
I do believe there is a way for console panes to interact with their tabs (as they already do this when they bring themselves to the front via ensureVisible). I would have these commands handled by the code which manages all the tabs and then just have it dispatch the select next/previous to the currently active tab.

kevinushey:
Closing this as I don't think this fits the bar for v1.1.

PR #1196, opened 2017-05-12
remove custom pointer cursor for linux desktop

jjallaire:
On some older versions of Ubuntu the default hand cursor was quite ugly, so we substituted a custom hand cursor. A couple negatives of this were that we can to constantly remember to use the custom cursor and there were always parts of the product (e.g. the embedded help viewer) that used the system cursor.

Ubuntu 16.04 appears to use a better looking hand cursor, so this change may no longer be required. In any case, I think we may have over-stepped here in the first place since the user is living with the "ugly" cursor in every other app including the web browser so it's probably either not bugging them or just white noise.

**IMPORTANT NOTE**:  While I built this change on Ubuntu 16.04 I was actually unable to actually run the desktop (it came up with a blank window!). This could be because something about my change borked startup (seems unlikely) or because there is something about my setup which makes it impossible for me to run debug desktop builds).

@javierluraschi Could you check this out locally and build/test as a sanity check.


gtritchie:
I built `standard-linux-cursor` branch in Docker, and installed on Ubuntu 16.04. Here's a screenshot of what I believe is the hand cursor you're reverting. Let me know if you want the .deb package and I'll share it.

jjallaire:
I just realized that my change overstepped. I think we just need to remove
the special handling for Linux desktop in CSS rather than get rid of all
the applications of .handCursor

I'll make that revision

jjallaire:
@javierluraschi I think we should abandon this change and just address the issue that provoked it for me -- with the 2x graphics the hand cursor shows up as a "giant" hand on Linux. If we can either fix that or make it a 1x graphic that would do the job.

jmcphers:
Now tracked in case 6855.

jjallaire:
BTW this one is pretty high priority for anyone running a Linux desktop as
you pretty frequently see the "jumbo" hand cursor when using the IDE.

javierluraschi:
@jjallaire pushed fix for build 1.1.234 (tomorrows) using `-webkit-image-set`, will try this out tomorrow and iterate if needed.

jjallaire:
Great, thanks!

PR #1197, opened 2017-05-12
fix cursor blink issue by updating to xterm.js 2.6.0

gtritchie:
Terminal cursor was blinking erratically after closing full-screen programs. Fixed via new xterm.js release.

xterm.js `terminal.open` also added a parameter to specify if terminal should have focus by default. Previous behavior was `true` so I plumbed through the option to `TerminalWidget` constructor, still passing `true`. Could be handy if we ever use xterm.js in contexts other than terminal.


gtritchie:
Mostly mechanical change. Going ahead and committing.

PR #1202, opened 2017-05-15
ui preference for chunk background highlighting

kevinushey:
This PR allows users to toggle background highlighting for code chunks on / off as desired.

The primary motivation for this feature is diagnostics -- a number of users are reporting editor sluggishness when editing in large documents; this PR is mainly added to help evaluate / diagnose that sluggishness.

If preferred, we can keep this as a 'hidden' preference and not expose to the user; let me know what you think.


jjallaire:
It seems like it's better as a hidden preference as we don't want users to just turn off background highlighting and forget about it (as we really need to engage with them to figure out what's going on)

kevinushey:
Good point -- okay, sounds good to me. I'll make the preference hidden and merge.

PR #1204, opened 2017-05-15
Classic theme regression fixes

javierluraschi:
Classic theme regression fixes:
- Fix environment pane spacing.
- Fix colors under call-stack and history.
- Fix severe performance degradation in rmarkdown under windows. Found this is somewhat related with applying `border-radius` with `overflow: hidden`. Setting `overflow` to `visible` seems to fix this, notice that scrollbars are not present since other elements will clip this anyways.


kevinushey:
Nice! Thanks for figuring out what was going on with editor performance so quickly.

PR #1209, opened 2017-05-17
Bugfix/copyright update2017

gtritchie:
Not sure we want to do this in bulk, but will stop me from peeking at the start of each file I edit. Skipped some .js and .css files here and there.


jmcphers:
My two cents: I actually kind of like that we don't auto-update these. I always update them manually when I make a change to a file. That way I usually know when I open a file about how long it's been since it was touched. 

Of course, you could argue that same information (with much greater fidelity) is one git command away! 

Anyone else have a point of view/preference?

jjallaire:
I don't have a preference.

gtritchie:
I'm quite neutral on it, despite having done the search & replace. So unless someone has a strong opinion in the "let's do it" direction I'm happy to just close this. I'll leave it open rest of today and see.

javierluraschi:
I prefer to leave this as is since I use the last-modified stamp on the file to get a sense of how careful I should be changing the file and who to ask for pointers, etc. Obviously, I could read the commit and this is trivial for one change; however, if all of us start making global changes: copyright, indentation, naming, curly braces, etc. this is lost in the history. I think, ideally, we have these tasks automated by a github account specifically designed for this that runs them periodically.

gtritchie:
@jmcphers and @javierluraschi 's arguments make sense to me. Going ahead and closing it. Thanks for feedback!

PR #1210, opened 2017-05-17
Restore panel animations under themes

javierluraschi:
- Panel animations were lost under themes, this was caused by applying a background under themes that is not compatible with the way animations work (they change the parent `z-index` to `-10` which would leave them under the background).
- A few profiler color tweaks.


jjallaire:
A somewhat related issue I wanted to mention is that when the IDE is loading with themes we see the gray background of empty panels. I think this used to be white which I think looks better -- does this change get us back to white or would that be a separate change?

PR #1217, opened 2017-05-18
Set modern theme as default theme

javierluraschi:
This PR makes the "modern" theme the default theme for all users, opt-out by switching back to "Classic" theme from appearance preferences dialog.

We can use this PR to discuss the right time to make this the default and merge afterwards.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/application/Application.java
+++ b/src/gwt/src/org/rstudio/studio/client/application/Application.java
@@ -620,7 +620,7 @@ public class Application implements ApplicationEventHandlers
    @Override
    public void onThemeChanged(ThemeChangedEvent event)
    {
-      RStudioThemes.initializeThemes(uiPrefs_.get().getFlatTheme().getGlobalValue(),
+      RStudioThemes.initializeThemes(uiPrefs_.get().getFlatTheme().getValue(),
                                      Document.get(),
                                      rootPanel_.getElement());
    }
@@ -717,7 +717,7 @@ public class Application implements ApplicationEventHandlers
    
    private void initializeWorkbench()
    {
-      RStudioThemes.initializeThemes(uiPrefs_.get().getFlatTheme().getGlobalValue(),
+      RStudioThemes.initializeThemes(uiPrefs_.get().getFlatTheme().getValue(),
                                      Document.get(),
                                      rootPanel_.getElement());
 
--- a/src/gwt/src/org/rstudio/studio/client/application/ui/RStudioThemes.java
+++ b/src/gwt/src/org/rstudio/studio/client/application/ui/RStudioThemes.java
@@ -24,6 +24,8 @@ import com.google.gwt.dom.client.Style.Overflow;
 import com.google.gwt.dom.client.Style.Position;
 import com.google.gwt.dom.client.Style.Unit;
 import com.google.gwt.dom.client.Style.Visibility;
+import com.google.gwt.regexp.shared.MatchResult;
+import com.google.gwt.regexp.shared.RegExp;
 
 public class RStudioThemes
 {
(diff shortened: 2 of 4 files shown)


jjallaire:
I think we are in a good place to do this now. What does everyone else think?

javierluraschi:
@kevinushey has been the heavy hitter with the dark theme. I would just wait for Kevins feedback on Fridays build and merge if it looks good (splitters-dragging-area and missing-icon issues are fixed now, so build is current with your feedback).

gtritchie:
I think we should make the change now.

kevinushey:
My only reservation is that one user has reported editor sluggishness when running with the modern theme with RStudio Desktop (Ubuntu in his case):

> Had a chance to try out the 1.1.235 build this morning. Seems better, but to keep it performant I still have to choose *either* "show output preview == none & deselect Show output inline for all Rmd documets" or "Modern" theme, I can't have both. 
> 
> Note that even with both Modern theme and inlining it isn't as bad as before, but still notice a degradation in overall typing responsiveness.

Although we're probably safe to flip the bit here we might want to spend a bit more time tracking down any other performance regressions if possible. (Note that this user is a somewhat unique case -- vertically oriented high DPI monitor with maximized editor surface, so really straining our older JS engines)

rmflight:
Replying to @kevinushey comment: I'm actually the user, and I'll describe my actual setup, as there appears to be a misunderstanding.

I have a 4K (3840x2160) monitor, that lets me have enough space to work in RStudio while only accessing half the screen. I do use it in landscape, and have half my screen with web-browser, and half with RStudio. 

So, even with that, and the editor window at half the RStudio instance, that means that it is still on over 1000 pixels of my screen. The more pixels I have in the editor, the slower working within Rmd documents seems to be, especially with the Modern theme. If not many people are using 4K screens (on Linux), they may not be seeing this behavior.

BTW, keep up the great work guys! And thanks again to @kevinushey for helping me figure out the issues last week!

javierluraschi:
@rmflight thanks for this info! I'll start taking a look at this today. If you pop-out the source windows does the problem get fixed? Any information that could help us scope this issue would be really appreciated, will start investigating this under Ubuntu and large editor areas.

rmflight:
@javierluraschi I will try to take a look at whether popping it out makes any difference. Definitely willing to help figure this out.

javierluraschi:
Tried a couple things, but not able to reproduce the slowdown yet.

That said, I did notice a bit of a lag in both themes (while using an Ubuntu VM with 3GB of RAM under 2880x1800 resolution), so unless we can scope this to the modern theme, I'll be instead spending some time investigating performance improvements while typing under both themes.

Ran this (not very scientific) test: Copied `12345678901234567890` and kept `CTRL+V` pressed against both themes timing how long it takes to complete the operation, as in:

screen shot 2017-05-19 at 11 06 09 am

Results were similar in both cases:

Classic Theme: `20.17s`
Modern Theme: `20.55s`

kevinushey:
@javierluraschi: does it make a difference if you maximize the source editor? What about if a large number of chunks are being rendered?

javierluraschi:
With several plots and plot galleries while maximized:

Classic theme: `51.21s`
Modern theme: `52.44`

screen shot 2017-05-19 at 12 03 37 pm

Worth mentioning that `gedit` performs worse and shows longer lags by the end of the document.

GEdit: `1:02s`:

screen shot 2017-05-19 at 12 11 29 pm

jmcphers:
@javierluraschi thanks for digging in and collecting some numbers. My suspicion is that there's something document and/or session specific in play here; in the past we've had a lot of trouble caused by the chunk background highlighter, which used to e.g. become incrementally slower every save.

My view is that these changes are ready to become the default in the dailies. If this performance issue afflicts more than a few users we will know about it pretty quickly and will have more data with which to attack the problem.

rmflight:
Here is a video of me with the latest daily. On my 4K screen (again, using 1/2 of it for my RStudio instance), using a stripped down Rmd that only has a single document, with me changing amongst the various themes and options. https://youtu.be/xyalujduIXU

Biggest thing that seemed to make a difference, how much vertical space the editor had. Maximizing it (i.e. undocking it) does not seem to make it any worse.

I have 16GB of Ram by the way, so that shouldn't be an issue.

I also did this on a project I instantiated **just** for this testing, there is nothing else session wise that should be at play.

I'm curious about using copy pasting to test times @javierluraschi , can you describe how you got timings for that operation?

PR #1221, opened 2017-05-19
Stop buffer reload for Windows cmd and powershell

gtritchie:
Reloading buffer into a new Command Prompt or PowerShell session in Windows terminal causes WinPty and the underlying console to get confused. Per discussion a while back, I'm just disabling buffer reload into these terminal types.


gtritchie:
Have built and tested on Windows (and Mac). Going ahead and merging.

PR #1222, opened 2017-05-19
Make panel splitters fully draggable

javierluraschi:
Currently (and also in Classic themes), the splitters are only `3px` wide; under the Modern theme, this is a bit worse since there is more space between the panels but only `3px` are draggable, this change increases this size to fully contain the spacing. However, since `SplitLayoutPanel` only allows setting the thickness during construction of this widget, this change requires closing/reopening the IDE for the change to take effect.

screen shot 2017-05-18 at 5 05 11 pm

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/layout/DualWindowLayoutPanel.java
+++ b/src/gwt/src/org/rstudio/core/client/layout/DualWindowLayoutPanel.java
@@ -366,7 +366,8 @@ public class DualWindowLayoutPanel extends SimplePanel
                                 Session session,
                                 String clientStateKeyName,
                                 final WindowState topWindowDefaultState,
-                                final int defaultSplitterPos)
+                                final int defaultSplitterPos,
+                                final int splitterSize)
    {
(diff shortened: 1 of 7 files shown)


javierluraschi:
Since this change requires users to restart RStudio to increase/decrease `4px` the thickness of the splitters, added the following warning triggered by dropdown changes while moving to/from classic/flat themes:

screen shot 2017-05-18 at 8 35 37 pm

kevinushey:
I wonder if, instead of showing a warning bar, we just reload the IDE surface after the user has clicked 'OK' to accept the preferences?

jmcphers:
Yes, there's actually a mechanism already for triggering this:

https://github.com/rstudio/rstudio/blob/4299af2e10a319aa33152de33b0414501bdb884c/src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/PreferencesDialog.java#L112-L113

jjallaire:
There's some way for prefs panels to indicate that their changes required a
reload (forget exactly what it is though).

PR #1223, opened 2017-05-19
Use custom checkboxes under themes to enforce same rendering

javierluraschi:
Under dark themes, checkboxes are missing the background. Under linux with highDPI, checkboxes render at half the size. Looks to be the case that we would be better of drawing checkboxes everywhere; can we think of cases where we wouldn't want this?

OS X (retina):
screen shot 2017-05-19 at 10 13 39 am

Linux (HighDPI):

screen shot 2017-05-19 at 10 16 11 am


jjallaire:
LGTM!

PR #1229, opened 2017-05-22
Add .registration=TRUE to useDynLib when Rcpp >= 0.12.11

jjallaire:
Related to this PR (https://github.com/RcppCore/Rcpp/pull/694) and new CRAN requirements for registration of native routines.

This should also be backported.


kevinushey:
LGTM!

PR #1233, opened 2017-05-23
Feature: Close child processes on exit

kfeinauer:
This pull request adds a new feature than can be toggled on or off to control whether or not child processes of the session are closed when the session is exited.  The default setting is false. It can be set as a session-wide setting in rsession.conf, and can be overridden in each project via a project setting.

A few tweaks were made to PosixSystem to get additional process info I needed, such as the ppid and pgrp.

### DIFF
--- a/src/cpp/core/system/PosixSystem.cpp
+++ b/src/cpp/core/system/PosixSystem.cpp
@@ -1306,7 +1306,7 @@ Error processInfo(const std::string& process, std::vector<ProcessInfo>* pInfo, P
    // we use a colon as the separator as it is not a valid path character in OSX
    std::string cmd = process.empty() ? "ps acxj | awk '{OFS=\":\"; print $1,$2,$3,$4,$10}'"
                                      : "ps acxj | awk '{OFS=\":\"; if ($10==\"" +
-                                        process + "\") print $1,$2,$3,$4,$10}'"
+                                        process + "\") print $1,$2,$3,$4,$10}'";
 
    core::system::ProcessResult result;
    Error error = core::system::runCommand(cmd,
@@ -1323,6 +1323,8 @@ Error processInfo(const std::string& process, std::vector<ProcessInfo>* pInfo, P
 
    BOOST_FOREACH(const std::string& line, lines)
    {
+      if (line.empty()) continue;
+       
       std::vector<std::string> lineInfo;
       boost::algorithm::split(lineInfo,
                               line,
@@ -1333,7 +1335,6 @@ Error processInfo(const std::string& process, std::vector<ProcessInfo>* pInfo, P
       procInfo.pid = safe_convert::stringTo<pid_t>(lineInfo[1], 0);
       procInfo.ppid = safe_convert::stringTo<pid_t>(lineInfo[2], 0);
       procInfo.pgrp = safe_convert::stringTo<pid_t>(lineInfo[3], 0);
-      procInfo.procname = lineInfo[4];
 
       // check to see if this process info passes the filter criteria
       if (!filter || filter(procInfo))
--- a/src/cpp/core/system/PosixSystem.cpp
+++ b/src/cpp/core/system/PosixSystem.cpp
@@ -1224,7 +1224,7 @@ Error processInfo(const std::string& process, std::vector<ProcessInfo>* pInfo, P
                continue;
             }
 
-            // read the stat fields for other relevant process infostd::string str;
+            // read the stat fields for other relevant process info
             std::string statStr;
             error = core::readStringFromFile(statFile, &statStr);
             if (error)
@@ -1237,6 +1237,12 @@ Error processInfo(const std::string& process, std::vector<ProcessInfo>* pInfo, P
             boost::algorithm::split(statFields, statStr,
                                     boost::is_any_of(" "),
                                     boost::algorithm::token_compress_on);
+
+            if (statFields.size() < 5)
+            {
+               LOG_WARNING_MESSAGE("Expected at least 5 stat fields but read: " + safe_convert::numberToString<size_t>(statFields.size()));
+               continue;
+            }
 
             // get process parent id
             std::string ppidStr = statFields[3];
@@ -1329,6 +1335,12 @@ Error processInfo(const std::string& process, std::vector<ProcessInfo>* pInfo, P
       boost::algorithm::split(lineInfo,
                               line,
                               boost::algorithm::is_any_of(":"));
+
+      if (lineInfo.size() < 5)
+      {
+         LOG_WARNING_MESSAGE("Exepcted 5 items from ps output but received: " + safe_convert::numberToString<size_t>(lineInfo.size()));
+         continue;
+      }
 
       ProcessInfo procInfo;
       procInfo.username = lineInfo[0];
@@ -1745,7 +1757,7 @@ Error launchChildProcess(std::string path,
 }
 
 bool filterNonChildProcesses(pid_t pid, pid_t ppid, pid_t pgrp,
-                             const rstudio::core::system::ProcessInfo& process)
+                             const core::system::ProcessInfo& process)
 {
    // remove all but child processes
    // remove this process's parent and remove
--- a/src/cpp/core/system/PosixSystem.cpp
+++ b/src/cpp/core/system/PosixSystem.cpp
@@ -1771,7 +1771,7 @@ bool filterNonChildProcesses(pid_t pid, pid_t ppid, pid_t pgrp,
 Error getChildProcesses(std::vector<rstudio::core::system::ProcessInfo> *pOutProcesses)
 {
    if (!pOutProcesses)
-      return;
+      return systemError(EINVAL, ERROR_LOCATION);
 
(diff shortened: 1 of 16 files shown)


gtritchie:
What's the potential interaction (if any) with the terminal feature? It starts bash, etc., as children of the rsession, which are killed when the session dies.

kfeinauer:
Right when the session process is about to go down, and sends SIGTERM to
all running children. So if your terminals are still running, they get
killed too.

gtritchie:
Right; does this new work make it possible this won't happen and terminals could survive session shutdown? Sorry, haven't actually looked at the code yet, just understanding the context of the changes.

kfeinauer:
No this feature is a request for child processes of the sessions to be
killed when the session exits. If your terminals process are included in
that list, they are killed as well.

Do they need to stick around after the session exits?

src/cpp/core/system/PosixSystem.cpp
```diff
@@ -1218,11 +1224,39 @@ Error processInfo(const std::string& process, std::vector<ProcessInfo>* pInfo)
                continue;
             }
 
+            // read the stat fields for other relevant process infostd::string str;
```
jmcphers: nit: typo?

src/cpp/core/system/PosixSystem.cpp
```diff
@@ -1218,11 +1224,39 @@ Error processInfo(const std::string& process, std::vector<ProcessInfo>* pInfo)
...
+               continue;
+            }
+
+            std::vector<std::string> statFields;
+            boost::algorithm::split(statFields, statStr,
+                                    boost::is_any_of(" "),
+                                    boost::algorithm::token_compress_on);
+
+            // get process parent id
+            std::string ppidStr = statFields[3];
```
jmcphers: This will throw if we somehow read fewer fields than expected so should be guarded.

gtritchie:
Not in the current design.

src/cpp/core/system/PosixSystem.cpp
```diff
@@ -1264,19 +1298,49 @@ core::Error pidof(const std::string& process, std::vector<PidType>* pPids)
...
+      std::vector<std::string> lineInfo;
+      boost::algorithm::split(lineInfo,
+                              line,
+                              boost::algorithm::is_any_of(":"));
+
+      ProcessInfo procInfo;
+      procInfo.username = lineInfo[0];
+      procInfo.pid = safe_convert::stringTo<pid_t>(lineInfo[1], 0);
+      procInfo.ppid = safe_convert::stringTo<pid_t>(lineInfo[2], 0);
+      procInfo.pgrp = safe_convert::stringTo<pid_t>(lineInfo[3], 0);
```
jmcphers: as above (needs guard on length)

src/cpp/core/system/PosixSystem.cpp
```diff
@@ -1264,19 +1298,49 @@ core::Error pidof(const std::string& process, std::vector<PidType>* pPids)
...
-Error processInfo(const std::string& process, std::vector<ProcessInfo>* pInfo)
+Error processInfo(const std::string& process, std::vector<ProcessInfo>* pInfo, ProcessFilter filter)
 {
-   std::vector<PidType> pids;
-   Error error = pidof(process, &pids);
+   // use ps to capture process info
+   // output format
+   // USER:PID:PPID:PGID:PROCNAME
+   // we use a colon as the separator as it is not a valid path character in OSX
+   std::string cmd = process.empty() ? "ps acxj | awk '{OFS=\":\"; print $1,$2,$3,$4,$10}'"
```
jmcphers: I think this would be less fragile if implemented as a pure ps format string rather than piping to awk (maybe there's a good reason this is not the case, let me know what you think)

src/cpp/core/system/PosixSystem.cpp
```diff
@@ -1680,7 +1744,76 @@ Error launchChildProcess(std::string path,
    return Success() ;
 }
 
+bool filterNonChildProcesses(pid_t pid, pid_t ppid, pid_t pgrp,
+                             const rstudio::core::system::ProcessInfo& process)
```
jmcphers: Nit: we usually just use `core::system` rather than `rstudio::core::system` for brevity (the rstudio namespace is already in scope here)

src/cpp/core/system/Win32System.cpp
```diff
@@ -958,6 +958,41 @@ CloseHandleOnExitScope::~CloseHandleOnExitScope()
...
+      while (moreProcesses)
+      {
+          // kill child processes
+          if (processEntry.th32ParentProcessID == processId)
+          {
+              HANDLE hChildProc = ::OpenProcess(PROCESS_ALL_ACCESS, FALSE, processEntry.th32ProcessID);
+
+              if (hChildProc)
+              {
+                  TerminateProcess(hChildProc, 1);
```
jmcphers: We should probably log if this fails.

src/gwt/src/org/rstudio/studio/client/projects/model/RProjectConfig.java
```diff
@@ -73,6 +73,14 @@ public native final void setAlwaysSaveHistory(int alwaysSaveHistory) /*-{
       this.always_save_history = alwaysSaveHistory;
    }-*/;   
    
+   public native final int getQuitChildProcessesOnExit() /*-{
+      return this.quit_child_processes_on_exit;
+   }-*/;  
+   
+   public native final int setQuitChildProcessesOnExit(int quitChildProcessesOnExit) /*-{
```
jmcphers: Nit: this should return `void`

src/gwt/src/org/rstudio/studio/client/projects/ui/prefs/ProjectGeneralPreferencesPane.java
```diff
@@ -77,6 +83,25 @@ protected void initialize(RProjectOptions options)
...
+      
+      // check or uncheck the checkbox for child processes based on the configuration value
+      // if default is specified, we need to use the current session setting
+      // otherwise, yes or no indicate the check state exactly
+      int quitChildProcessesOnExit = config.getQuitChildProcessesOnExit();
+      boolean quitChildProcessesChecked = sessionInfo_.quitChildProcessesOnExit();
+      
+      switch (quitChildProcessesOnExit)
+      {
+      case 1:
```
jmcphers: We usually use named constants for these kinds of values (for readability), but it's pretty clear here so up to you.

jjallaire:
Let's slow down and discuss this one realtime tomorrow. We already terminate all children that we've spawned using the process supervisor here: https://github.com/rstudio/rstudio/blob/master/src/cpp/session/SessionMain.cpp#L995. If there are other processes we are launching that the process supervisor doesn't know about then perhaps it should. 

Let's discuss what you are attempting to accomplish and see what the best way forward is.

jjallaire:
Or if perhaps you've observed that certain processes aren't dying when they should we should figure that out as well.

kfeinauer:
Yes the process supervisor doesn't seem to know about them. I figured it
only knew about children it explicitly was told to launch, so anything
created in the r session is not known to it.

We can discuss more tomorrow. Will see up a time to discuss.

jjallaire:
You are exactly correct, processes created by R aren't known to the
supervisor.

However, in traditional R process launching the only kind of processes it
creates block the console entirely (so effectively have to be ended when
the process ends). So this hasn't generally been as issue. That said,
processx is a new package that enables async process launching/management
and there could be other packages that have ways of doing this.

As another aside, there is actually a very high level way to do this on
Linux that we take advantage of elsewhere (PR_SET_DEATHSIG, see
https://stackoverflow.com/questions/284325/how-to-make-child-process-die-after-parent-exits).
On Windows you can also use job objects to do this (we have a job object we
create in rdesktop to ensure that when it is killed the R session process
and it's children are killed).

src/cpp/core/system/PosixSystem.cpp
kfeinauer: I'm doing it this way is it is done almost exactly the same way in the pidOf function. It also makes it a consistent command regardless of if we need to search by process name or not (since in both cases we use awk).

kfeinauer:
@jmcphers Implemented feedback. Please review.

jmcphers:
@jjallaire For some context this work was primarily motivated by customer requests. The two main categories of child processes in play are:

a) forks of the R session created by parallelization packages such as mcparallel; and

b) asynchronous execution of external utilities (e.g. plink) 

Neither are tracked using the existing child process monitoring code and of course we don't have code running in either so to the best of my knowledge can't use tools like PR_SET_DEATHSIG. 

@karl It's possible we will need to tweak this to be sure it doesn't conflict with existing process shutdown behavior in processes that we already manage (such as terminals and those that we do track using our process monitoring); you should chat with Darby about this when you send him a note for testing.

kfeinauer:
@jmcphers Noted. I will let Darby know.

gtritchie:
The terminal processes are launched by the process supervisor, so hopefully nothing special needed there.

PR #1236, opened 2017-05-24
Update to pandoc 1.19.2.1 on Windows, OS X, and Ubuntu

jjallaire:
This PR updates our embedded version of pandoc to v1.19.2.1 on platforms for which a pandoc binary is available from the pandoc releases page (https://github.com/jgm/pandoc/releases).

Note that this appears to be an excellent release to embed since it was shipped several months ago and is still the "latest release". Typically there are lots and lots of pandoc releases and point releases and it's difficult to pin down which is the most stable / bug free (many releases are to fix regressions but then new regressions tend to come up). There have been 1,314 commits to master since this release so it will probably be some time before another fully stable release is available.

This introduces one change in the way we handle pandoc binaries. All of the 1.19.2.1 binaries are built by the pandoc project whereas formerly we used to build our own Linux binaries (which were capable of running on a very wide variety of older Linux distros). We did this by building pandoc on Ubuntu 10.04 with older versions of GHC/Haskell that were compatible with Ubuntu 10.04. Unfortunately pandoc now mandates more recent versions of GHC/Haskell which in turn won't install on Ubuntu 10.04.

So the new world looks like this:

- Windows, Mac, Ubuntu 14.04+ always get a recent version of pandoc (in this case 1.19.2.1)

- Other systems get pandoc 1.17.2 (the "older" pandoc version we already have in hand). 

Note that traditionally the way we propagate "older" versions is not by having lots of platform/version conditional code but rather by creating an e.g. "pandoc-1.19.2.1" bundle that just substitutes older versions wherever necessary within the bundle. We continue that here, which is why on the surface it appears as if all platforms are getting 1.19.2.1 when in fact some have substituted older versions.

Presumably this is an acceptable state of the world since the users who will end up running pandoc 1.17.2 are also running operating systems that emphasize/force use of older versions of other components.

If it turns out that there is significant demand for newer versions of pandoc on e.g. RHEL 7 we can revisit the process by which we build our own pandoc binaries (last time we did this we needed to produce binaries for RHEL 5 so needed to use a pretty obscure method of building pandoc, my guess is that building for RHEL 7 is considerably easier). Anyway, we can revisit this later as demand dictates.


jjallaire:
Note that back porting this to v1.0 would fix issues we've seen with our homespun pandoc binary on Ubuntu 16.10 and higher. I therefore think that we should back port it.

jjallaire:
@jcheng5

jmcphers:
Appreciate the background! I've added a link to this commit to the bug tracking the homespun pandoc segfault issue (already marked as a patch bug).

I've also done the Windows dependency dance with the new pandoc binaries.

jjallaire:
Awesome, thanks!!!

PR #1238, opened 2017-05-24
Play console actions back in regular order

gtritchie:
Previously when refreshing UI, we played the ConsoleActions back in reverse order, until our maximum number of console lines was displayed. Server keeps up to 1000 console actions, but each console action can generate multiple lines of output, this is why it was done this way.

This reverse playback made client Ansi parsing more difficult, often breaking reload of console with Ansi color output, and the resulting DOM had many more nodes than the DOM created during regular usage of the console, impacting performance. Another consequence was any special parsing done on console DOM had to account for both possible structures (regular and reloaded). There are some bugs in that area, and wanted to do this change before getting to those.

Now we do an initial pass through the ConsoleActions in reverse order and count newlines to determine line count, but don't output anything to the DOM. Then do a forward pass starting where it will stop at maximum desired console length (plus potential slight overrage due to size of final action's output). That is written out in normal order, thus using the same output logic as regular console output. Finally trim any excess when the reload is done.

Anecdotally, the reduced simplicity of the resulting DOM should easily overcomes the cost of doing the initial counting pass, and in general I expect this to help with reloading large console buffers.

This does fix the Ansi code parsing issue on reload.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/common/shell/ShellWidget.java
+++ b/src/gwt/src/org/rstudio/studio/client/common/shell/ShellWidget.java
@@ -389,47 +389,36 @@ public class ShellWidget extends Composite implements ShellDisplay,
 
    private boolean output(String text,
                           String className,
-                          boolean addToTop)
+                          boolean ignoreLineCount)
    {
       if (text.indexOf('\f') >= 0)
          clearOutput();
 
       Element outEl = output_.getElement();
       
-      if (addToTop)
+      // create trailing output console if it doesn't already exist 
+      if (virtualConsole_ == null)
       {
-         SpanElement leading = Document.get().createSpanElement();
-         outEl.insertFirst(leading);
-         VirtualConsole console = new VirtualConsole(leading);
-         console.submit(text, className);
-         lines_ += DomUtils.countLines(leading, true);
-      }
-      else
-      {
-         // create trailing output console if it doesn't already exist 
-         if (virtualConsole_ == null)
-         {
-            SpanElement trailing = Document.get().createSpanElement();
-            outEl.appendChild(trailing);
-            virtualConsole_ = new VirtualConsole(trailing);
-         }
-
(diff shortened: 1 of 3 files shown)


src/gwt/src/org/rstudio/studio/client/common/shell/ShellWidget.java
```diff
@@ -457,24 +454,60 @@ private boolean trimExcess()
...
+      for (; revIndex >= 0; revIndex--)
+      {
+         ConsoleAction action = actions.get(revIndex);
+         
+         if (action.getType() == ConsoleAction.INPUT)
+            lines++;
+         
+         lines = lines + StringUtil.newlineCount(action.getData());
+         
+         excessLines = excessLines(lines);
```
jmcphers: nit: could we just omit the excessLines variable and function altogether and just say `if (lines > maxLines_)` for simplicity?

src/gwt/src/org/rstudio/studio/client/common/shell/ShellWidget.java
```diff
@@ -499,12 +532,21 @@ public boolean execute()
                      break;
                }
                if (!canContinue)
-                  return false;
+               {
+                  break;
+               }                  
             }
-            if (!DomUtils.selectionExists())
-               scrollPanel_.scrollToBottom();
```
jmcphers: When I profiled the console replay performance a while back, this turned out to be one of the biggest contributors to the replay time, so I am delighted to see it gone! Now that we play everything back in order, does the scroll panel keep up automatically (or is some other mechanism taking care of scrolling to the bottom once replay is done?)

gtritchie:
The "excessLines" variable was left over from a slightly different approach, so yes, I'll get rid of it with my next PR.

Yes, the resizeCommand_.nudge() at the end of ShellWidget::output ensures the pane is scrolled to the bottom after reloading.

jmcphers:
Great, thanks!

PR #1255, opened 2017-05-31
add 'ignore whitespace' checkbox for git diff

kevinushey:
This PR adds an 'Ignore Whitespace' checkbox to the Git diff view, allowing users to toggle off / on whitespace-only diffs. Compare e.g.

screen shot 2017-05-31 at 2 10 50 pm

screen shot 2017-05-31 at 2 10 55 pm

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/vcs/git/dialog/GitReviewPresenter.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/vcs/git/dialog/GitReviewPresenter.java
@@ -18,15 +18,12 @@ import com.google.gwt.core.client.Scheduler;
 import com.google.gwt.core.client.Scheduler.ScheduledCommand;
 import com.google.gwt.dom.client.NativeEvent;
 import com.google.gwt.event.dom.client.*;
-import com.google.gwt.event.logical.shared.CloseEvent;
-import com.google.gwt.event.logical.shared.CloseHandler;
 import com.google.gwt.event.logical.shared.HasAttachHandlers;
 import com.google.gwt.event.logical.shared.ValueChangeEvent;
 import com.google.gwt.event.logical.shared.ValueChangeHandler;
 import com.google.gwt.event.shared.HandlerRegistration;
 import com.google.gwt.json.client.JSONNumber;
 import com.google.gwt.user.client.Command;
-import com.google.gwt.user.client.Window;
 import com.google.gwt.user.client.ui.HasText;
 import com.google.gwt.user.client.ui.HasValue;
 import com.google.gwt.user.client.ui.IsWidget;
@@ -289,15 +286,6 @@ public class GitReviewPresenter implements ReviewPresenter
             });
          }
       };
-      
-      Window.addCloseHandler(new CloseHandler<Window>()
-      {
-         @Override
-         public void onClose(CloseEvent<Window> event)
-         {
-            session.persistClientState();
-         }
-      });
       
       view_.getChangelistTable().addSelectionChangeHandler(new SelectionChangeEvent.Handler()
       {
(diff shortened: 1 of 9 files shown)


kevinushey:
It looks like persistence of values isn't quite working intra-session; that is, the 'Ignore Whitespace' checkbox setting isn't respected when closing + re-opening the 'Git Review' window multiple times in the same session. Is there a better way for to signal the change in client state from a satellite?

jmcphers:
Do we have a `ClientStateUpdater` in satellites? (We don't AFAIK but it's been a while since I walked through this.) I think you might be better served using a UI pref here--maybe a project UI pref so it sticks across sessions but doesn't necessarily imply a default across every IDE instance.

jjallaire:
Agree that I would make this a UIPref (those persist globally and propagate automatically to all sessions)

kevinushey:
Okay, I've tried setting this as a UI preference but that does not seem to persist automatically either (ie we don't seem to push UI pref state when closing satellites).

Do we need to add a handler to satellites that pushes UI preference state when the window is closed as well?

jmcphers:
You do need to manually push UI prefs, but I think it might be overreaching to do it every time a satellite closes. Typically we set the UI prefs when the user has committed to the new value. In this case that'd be after initiating the commit operation. You can push UI prefs using `writeUiPrefs`: https://github.com/rstudio/rstudio/blob/62b217a7115f98e933b4436a04cd5ac675a4f8c6/src/gwt/src/org/rstudio/studio/client/workbench/prefs/model/UIPrefs.java#L58

kevinushey:
`writeUiPrefs()` seems to only write global UI preferences, not project preferences. Is there a separate API I should be using here? Or am I misunderstanding something about how global + project UI preferences play together?

jmcphers:
Project UI preferences can override global ones (that's why UI prefs have both a `getValue()` and `getGlobalValue()`.  `writeProjectOptions` sets project UI prefs, but those require some additional boilerplate (see `ProjectContext::uiPrefs`). If you'd prefer to keep things simple this might be better off as a regular, global UI pref (I think you could make an argument either way!).

kevinushey:
I'll stick with global preferences for now -- IMHO it seems less likely that a user would would project-specific behavior for ignoring of diff whitespace (rather, users probably just want it to remember whatever the last value was in their session)

src/gwt/src/org/rstudio/studio/client/workbench/views/vcs/git/dialog/GitReviewPresenter.java
```diff
@@ -282,7 +289,16 @@ public void onFileChange(FileChangeEvent event)
...
          }
       };
-
+      
+      Window.addCloseHandler(new CloseHandler<Window>()
+      {
+         @Override
+         public void onClose(CloseEvent<Window> event)
+         {
+            session.persistClientState();
```
jmcphers: Can you remove this now that we're using a UI pref?

src/gwt/src/org/rstudio/studio/client/workbench/views/vcs/git/dialog/GitReviewPresenter.java
```diff
@@ -486,6 +491,23 @@ public void onValueChange(ValueChangeEvent<Boolean> event)
...
+      
+      view_.getIgnoreWhitespaceCheckBox().addValueChangeHandler(
+            new ValueChangeHandler<Boolean>()
+            {
+               @Override
+               public void onValueChange(ValueChangeEvent<Boolean> event)
+               {
+                  boolean value = event.getValue();
+                  uiPrefs_.gitDiffIgnoreWhitespace().setGlobalValue(value);
+                  uiPrefs_.writeUIPrefs();
```
jmcphers: It's a little heavyweight to push all the UI prefs to the server every time the box is checked. Would it be difficult to do this on commit instead?

jjallaire:
That said, I expect that this checkbox will get checked like once a month
so writing immediately doesn't seem so bad.

jmcphers:
Agree it's not a big deal. I'll merge as-is and @kevinushey you can decide whether to move the preference write on master afterwards.

PR #1264, opened 2017-06-02
implement filtering of toolbar menus

kevinushey:
This PR implements filtering of toolbar menus, and uses it with the Git branch toolbar menu to allow for quick search + selection of Git branches. I'm hoping this will become handy for other various toolbar menus we show where there are a large number of items to be displayed.

Feedback is welcome, especially what we can do to improve discoverability of this feature. Note that currently the search bar is activated simply by typing after the menu has been opened.

branch-filter

I'll also need to make sure we appropriately theme + style the search widget here.

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/command/BaseMenuBar.java
+++ b/src/gwt/src/org/rstudio/core/client/command/BaseMenuBar.java
@@ -64,28 +64,34 @@ public class BaseMenuBar extends MenuBar
       eventBus_ = RStudioGinjector.INSTANCE.getEventBus();
    }
    
+   private MenuItem getTargetedMenuItem(Event event)
+   {
+      Element targetEl = DOM.eventGetTarget(event);
+      if (targetEl == null)
+         return null;
+      
+      for (MenuItem item : getItems())
+         if (item.getElement().isOrHasChild(targetEl))
+            return item;
+      
+      return null;
+   }
+   
    @Override
    public void onBrowserEvent(Event event)
    {
-      if (event.getTypeInt() == Event.ONCLICK)
+      if (event.getTypeInt() == Event.ONCLICK ||
+          event.getTypeInt() == Event.ONKEYDOWN ||
+          event.getTypeInt() == Event.ONKEYPRESS ||
+          event.getTypeInt() == Event.ONKEYUP)
       {
          // By default, GWT handles click events by sending focus
          // to the menu bar instance, even when the target of the
          // click was not a menu item (e.g. it was a separator).
          // We want to avoid this behavior and so suppress click
          // handling when the click target is not a menu item.
-         MenuItem targetItem = null;
-         Element targetEl = DOM.eventGetTarget(event);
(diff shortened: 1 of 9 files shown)


jmcphers:
Nice, this will be very helpful for repos with a large number of branches! 

Re: layout, It looks like this could live to the right of "Local Branches". Search is often in the upper right so I don't think that'll be too unexpected.

    LOCAL BRANCHES         [ searchbox (x) ]

There's no way users are going to discover type-to-activate (it's very uncommon for a drop menu to accept keyboard focus when activated with the mouse!). I think we should show the searchbox all the time, and just let typing focus it automatically.

jjallaire:
I agree on letting typing focus it automatically (it would be weird if a
mouse gesture grabbed keyboard focus)

Adding it to the LOCAL BRANCHES area saves space. It's bit off key as it
searches more than just local branches  but I think still okay.

kevinushey:
Implemented now:

screen shot 2017-06-02 at 2 38 27 pm

Unfortunately, this required a lot of gymnastics as one cannot insert arbitrary widgets as separators in a GWT menu (they must be plain UIObjects); this implies that the default event handling machinery won't work here and so we have to do some extra work to route everything around.

I was also seeing some very focus weirdness where attempts to click in the search box would end up losing focus, sending it elsewhere -- I wasn't able to figure out why this was occurring and so we have this somewhat ugly native preview handler that prevents the click events causing focus loss.

jjallaire:
Could it be that our menus have some focus-responsive code that's crossing signals with this mechanism?

kevinushey:
The menubar does indeed become the active element after attempting to click the search widget (as checked with `document.activeElement`), so the behavior likely lies in our own menu bar class or within GWT's own implementation. I'll spend a bit more time to see if I can dig out why this is occurring.

kevinushey:
I think this is likely the culprit:

https://github.com/gwtproject/gwt/blob/563e9f0a20a878de6dc3156f9e601bdbeaaf0394/user/src/com/google/gwt/user/client/ui/MenuBar.java#L705-L712

If I understand correctly, by default, GWT handles click events anywhere on the menubar's surface and interprets them as a request to focus the menu bar itself (even if a sub-element within that menu could accept focus). Perhaps the best solution is for us to override `onBrowserEvent()` in our `BaseMenuBar` class and suppress click events targeted at elements which can handle them?

jjallaire:
That sounds like the right approach!

kevinushey:
Alright, this should be ready for review now.

Me implementing this PR:

t0xhtgj

jmcphers:
Thanks for putting in the time to wrestle with the browser event capture! I have the feeling this will come in handy later.

kevinushey:
I hope so! For the record, I still primarily interact with Git from the command line rather than RStudio -- I've been trying to figure out what the pain points are that keep driving me to the command line, and the inability to easily select branches from RStudio when there are a large number of branches was one of them.

PR #1268, opened 2017-06-03
Interrupt and close buttons for terminal toolbar

gtritchie:
Added a close button and a stop (a.k.a. interrupt) button to the terminal toolbar. The new interrupt/stop button will tell the pty to send an interrupt to processes running in it.

There are four possible states:

(1) no terminals open in the terminal tab (close, stop, and the clear button are hidden)

(2) current terminal is at prompt (clear and close buttons are shown)

(3) current terminal is busy running a non-full-screen program (clear and stop buttons are shown)

(4) current terminal is busy running a full-screen program (close button is shown, will warn before actually closing the terminal)

Note, on Windows desktop IDE there is currently no stop button, as the plumbing to pass through the interrupt to the pseudoterminal is NYI.


jjallaire:
Looks great!

PR #1270, opened 2017-06-05
made child process termination on windows recursive like on posix

kfeinauer:
before we were only killing direct children, now we will kill all levels of child processes

### DIFF
--- a/src/cpp/core/system/Win32System.cpp
+++ b/src/cpp/core/system/Win32System.cpp
@@ -958,10 +958,69 @@ CloseHandleOnExitScope::~CloseHandleOnExitScope()
    }
 }
 
-Error terminateChildProcesses()
+struct ProcessInfo
 {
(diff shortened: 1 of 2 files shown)


src/cpp/core/system/Win32System.cpp
```diff
@@ -958,10 +958,69 @@ CloseHandleOnExitScope::~CloseHandleOnExitScope()
    }
 }
 
-Error terminateChildProcesses()
+struct ProcessInfo
+{
```
jjallaire: Do we need a default constructor that initializes these to "null" values (i.e. 0 or -1). Fine if we don't, just asking the question.

kfeinauer: We shouldn't. The only usage for now is guaranteed to set valid values.

jjallaire:
Okay sounds good.

PR #1271, opened 2017-06-06
Custom shell setting for terminal

gtritchie:
Provide option for user to customize the shell program launched by the terminal. There were already several options on Windows (git-bash, cmd, powershell, etc.), but previously only the default Bash on Mac/Linux/Server. Now that preferences dropdown is visible on non-Windows and includes a new "Custom" option which lets user specify a command and command-line arguments for that command.

For example, normal shell setting on non-Windows:

And a custom setting:


dwhdai:
@gtritchie Would it be possible to add the custom setting option for Windows also?

gtritchie:
I need to go back and remind myself why I didn't include it for Windows. :-)

Can you provide an example of the shell you'd like to use on Windows via such a mechanism?

dwhdai:
Thanks! I personally use [cmder](http://cmder.net/) but I think [ConEmu](https://conemu.github.io/) is also a popular one.

PR #1274, opened 2017-06-07
use textbox instead of file picker for custom shell selection

gtritchie:
Use a textbox for entry of custom terminal shell. This is an advanced feature, and file pickers can make it hard to navigate to Unix-y paths (especially on Mac file picker), and may not deal well with soft-links.

Also turned off custom shell option for Windows. Not tested, and likely not very useful. Can reconsider if we get feedback on the subject.

### DIFF
--- a/src/cpp/session/modules/SessionTerminalShell.cpp
+++ b/src/cpp/session/modules/SessionTerminalShell.cpp
@@ -139,6 +139,7 @@ void scanAvailableShells(std::vector<TerminalShell>* pShells)
       }
    }
 
+#ifndef _WIN32
    // Add user-selectable shell command option
    TerminalShell customShell;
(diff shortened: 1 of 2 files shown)


kevinushey:
Does this work fine if users just type e.g. `bash`? (Ie, do we locate non-absolute-path entries on the system `PATH`?) Or is some other validation of the selected path done? (E.g. show the name in red text if it doesn't resolve to an existing shell)

gtritchie:
Currently no validation, has to be a full path. Good point, this should include some validation and/or clarification in the UI.

gtritchie:
Changed label to clarify that must enter a fully qualified path. Doing additional validation would require a server call to see if path exists and is executable. Thinking that's overkill here, as someone specifying a custom shell is likely to be an advanced user who will have a very specific shell in mind and know where it lives.

PR #1275, opened 2017-06-08
replace addin MRU w/searchable list

kevinushey:
This PR replaces the addins MRU list with a full, searachable list of available addins.

This change is primarily motivated by the fact that users have found the behavior confusing:

- Most users haven't realized the list is actually an most recently used list; instead, it just seemed as though some addins simply did not show up;

- The lack of a stable position in the menu was frustrating for users,

- Many users had requested grouping of addins by package.

Unfortunately, much of the logic from the Git Branch dialog has been duplicated here as I can't find a natural way to allow these two classes to share code -- I would appreciate thoughts on how this could be accomplished.

screen shot 2017-06-08 at 4 26 04 pm


jjallaire:
Is there still an overflow condition? (i.e. is there still a maximum number of addins you can display)

For this and the Git branch menu I'd push the search box all the way over to the right.

I think the code duplication is okay for now as if you are having a hard time seeing how to refactor then the refactor is likely high risk. We can do this at the outset of v1.3 development.

src/gwt/src/org/rstudio/studio/client/workbench/AddinsMRUList.java
```diff
@@ -1,303 +0,0 @@
...
-      command.addHandler(new AddinCommandHandler(index));
-   }
-
-   @Override
-   public void onSessionInit(SessionInitEvent sie)
-   {
-      RAddins addins = session_.getSessionInfo().getAddins();
-      update(addins);
-      
-      mruList_ = pListManager_.get().getAddinsMruList();
```
jmcphers: Should we also remove the code backing this list? https://github.com/rstudio/rstudio/blob/54cd3abcfc58837b433464c793fe9b03a87f0bb4/src/gwt/src/org/rstudio/studio/client/workbench/WorkbenchListManager.java#L101

jmcphers:
Re: code sharing, when you do get around to tackling it, I think you could do it pretty straightforwardly by creating an abstract `ToolbarPopupMenu` which includes the search widget and then dispatches to some virtual functions for population/etc.

src/gwt/src/org/rstudio/studio/client/application/ui/addins/AddinsToolbarButton.java
```diff
@@ -0,0 +1,278 @@
...
+      
+      Map<String, List<RAddin>> addinsByPkg = organizeAddins();
+      
+      if (addinsByPkg.isEmpty())
+      {
+         populateEmptyMenu();
+         return;
+      }
+      
+      MapUtil.forEach(addinsByPkg, new MapUtil.ForEachCommand<String, List<RAddin>>()
```
kevinushey: The addins are rendered in a scrollable menu clamped at 300px in height (https://github.com/rstudio/rstudio/blob/master/src/gwt/src/org/rstudio/core/client/widget/ScrollableToolbarPopupMenu.java#L76-L79), which should accommodate this.

src/gwt/src/org/rstudio/studio/client/workbench/AddinsMRUList.java
kevinushey: We should; nice catch!

kevinushey:
> Is there still an overflow condition? (i.e. is there still a maximum number of addins you can display)

There is no limit, but because the addins are rendered within a scrollable panel the experience is still fairly nice (you can scroll through the entire set of addins). In theory the dialog could blow up if someone had 1000 addins installed but I think most users will have ~100 maximum addins.

> For this and the Git branch menu I'd push the search box all the way over to the right.

I actually did this because otherwise the scrollbar can be drawn overtop the search widget when scrolling, e.g.

screen shot 2017-06-09 at 10 01 28 am

Also note that since the search widget is drawn as part of the separator, it scrolls with the dialog as well (the search widget is not absolutely positioned at the top-right of the dialog)

> I think the code duplication is okay for now as if you are having a hard time seeing how to refactor then the refactor is likely high risk. We can do this at the outset of v1.3 development.

Sounds good!

jjallaire:
All sounds good. I also think we could increase the max vertical height to
400px or 450px (bounded by actual available height of course)

jmcphers:
Could you also add a NEWS entry?

kevinushey:
I added the NEWS entry on `master` -- thanks for the review!

@yihui -- we'll be providing a better view of addins in the toolbar now; if you have time with the next daily build I'd appreciate any feedback you have.

yihui:
That looks awesome!! Thanks @kevinushey!

yihui:
Just tested it. Beautiful! 👍

PR #1276, opened 2017-06-09
Feature dark theme for sky theme

javierluraschi:
Themes under sky and dark editor do not look good; instead, use dark-grey theme for ace themes regardless of the rstudio theme selection.


JanMarvin:
I assume this breaks classic theme with dark layouts. With those you are now forced to the dark modern template, any chance this could be fixed? As a side note why don't you allow users to select their appearance? I like a dark template for my syntax still enjoy the brighter Rstudio Ui.

PR #1277, opened 2017-06-09
Add option to keep terminal pane open after process exits

gtritchie:
Added plumbing and user option to keep terminal panes open after process exits. This will make it easier to add capability to run a command in a terminal pane, without a shell, and keep pane around to show the results (this will be via a new rstudioapi, NYI).

FYI, this change added new ConsoleProcessInfo metadata, so I bumped up the storage revision number for saving open processes (console05); first time running this build you will lose any previously created terminals.

As a side effect, found and fixed a bug that had been seen occasionally before, where trying to close a terminal tab gave an error. Cause was websocket connection closing but terminal remaining open, which is much easier to get into with this work. Fix was to not dereference a null websocket object in the client.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/common/console/ConsoleProcessInfo.java
+++ b/src/gwt/src/org/rstudio/studio/client/common/console/ConsoleProcessInfo.java
@@ -117,6 +117,10 @@ public class ConsoleProcessInfo extends JavaScriptObject
    public final native int getAutoCloseMode() /*-{
       return this.autoclose;
    }-*/;
+   
+   public final native boolean getZombie() /*-{
+      return this.zombie;
+   }-*/;
 
    public static final int SEQUENCE_NO_TERMINAL = 0;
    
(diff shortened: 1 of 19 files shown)


src/gwt/src/org/rstudio/studio/client/common/console/ConsoleProcessInfo.java
```diff
@@ -30,6 +30,10 @@
    public static final int CHANNEL_WEBSOCKET = 1;
    public static final int CHANNEL_PIPE = 2;
    
+   public static final int AUTOCLOSE_DEFAULT = 0;
+   public static final int AUTOCLOSE_ALWAYS = 1;
+   public static final int AUTOCLOSE_NEVER = 0;
```
kevinushey: Should the value here be 2 or is that intentional?

gtritchie: Eek, good catch! Not intentional.

src/gwt/src/org/rstudio/studio/client/workbench/views/terminal/TerminalSession.java
```diff
@@ -740,6 +755,11 @@ public void onError(ServerError error)
          }
       });
    }
+   
+   public void showZombieMessage()
+   {
+      writeln("[Process completed]");
```
kevinushey: Would it be useful to show the exit code here? E.g. `[Process completed successfully]` vs. `[Process exited with error code n]`?

gtritchie: Good idea. I can also capture exit code and make it available via rstudioapi. I'll do this via a separate PR.

PR #1279, opened 2017-06-10
add desktop devmode

kevinushey:
There might be a reason why we didn't add this in the first place, but as far as I can see, this works (at least on macOS) for a devmode experience with the desktop product. With this PR, we can use

    ant desktop

to build the desktop version of the sources in the regular devmode configuration. One just needs to run a debug build of the desktop product, and things appear to work as expected.

### DIFF
--- a/src/gwt/build.xml
+++ b/src/gwt/build.xml
@@ -138,6 +138,23 @@
          <param name="gwt.extra.args" value="${gwt.extra.args} -draftCompile" />
       </antcall>
    </target>
+
+   <target name="desktop" depends="acesupport,javac" description="Run development mode">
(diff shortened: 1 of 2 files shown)


jjallaire:
Yes, I think this should indeed work on Safari (but not on WebKit).

kevinushey:
Playing around on my Ubuntu VM, it looks like this works fine with the desktop client as well. (That is, if you run devmode with e.g. `ant desktop` and then separately run `./rstudio-dev` to launch the desktop client, you get a development experience akin to regular devmode).

jjallaire:
That's fantastic! I didn't know that was possible, but I guess with Qt 5.4
it is :-)

PR #1298, opened 2017-06-16
use Gabor's universal pandoc binary on Linux

kevinushey:
This PR uses Gabor's universal pandoc binary for 64bit Linux builds.

I also ported some of the old Bash helper functions from the old PR (just so we have it if we need it in the future) but note that most of that is unused in this PR.

### DIFF
--- a/dependencies/common/install-pandoc
+++ b/dependencies/common/install-pandoc
@@ -20,111 +20,54 @@ set -e
 source rstudio-tools
 
 # variables that control download + installation process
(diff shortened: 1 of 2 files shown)


jjallaire:
Note that when we backport the newer pandoc to v1.0 all we need to do is replace the current 'debian' binary with this one (the debian binary is also a universal binary that works everywhere save for RHEL5, however it crashes on Ubuntu 16.10 and higher which is why we need to move to this version).

PR #1299, opened 2017-06-16
use '-z' for machine-parsable git status output

kevinushey:
This PR fixes a bug where paths containing backslashes would not be handled correctly in the RStudio Git interface.

`git` documents `core.quotepath` as (https://www.kernel.org/pub/software/scm/git/docs/git-config.html), emphasis mine:

> Commands that output paths (e.g. ls-files, diff), will quote "unusual" characters in the pathname by enclosing the pathname in double-quotes and escaping those characters with backslashes in the same way C escapes control characters (e.g. \t for TAB, \n for LF, \\ for backslash) or bytes with values larger than 0x80 (e.g. octal \302\265 for "micro" in UTF-8). If this variable is set to false, bytes higher than 0x80 are not considered "unusual" any more. **Double-quotes, backslash and control characters are always escaped regardless of the setting of this variable.** A simple space character is not considered "unusual". **Many commands can output pathnames completely verbatim using the -z option.** The default value is true.

Given that `-z` is effectively an alias for `--porcelain`, with the added bit of completely disabling path quoting, I think that's the right fix here.

### DIFF
--- a/src/cpp/session/modules/SessionGit.cpp
+++ b/src/cpp/session/modules/SessionGit.cpp
@@ -485,13 +485,7 @@ public:
       // build shell arguments
       ShellArgs arguments;
       
-      // on some platforms, git will return paths which contain characters
-      // not in the ASCII set with a so-called 'quoted octal encoding'.
-      // this is controlled by the 'core.quotepath' configuration option;
-      // by setting this to off we ensure that git will return us a
-      // plain UTF-8 encoded path which requires no further processing
-      arguments << "-c" << "core.quotepath=off"
-                << "status" << "--porcelain" << "--" << dir;
+      arguments << "status" << "-z" << "--porcelain" << "--" << dir;
       
       std::string output;
       Error error = runGit(arguments, &output);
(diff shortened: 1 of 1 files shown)


src/cpp/session/modules/SessionGit.cpp
```diff
@@ -485,21 +485,16 @@ class Git : public boost::noncopyable
...
       ShellArgs arguments;
       
-      // on some platforms, git will return paths which contain characters
-      // not in the ASCII set with a so-called 'quoted octal encoding'.
-      // this is controlled by the 'core.quotepath' configuration option;
-      // by setting this to off we ensure that git will return us a
-      // plain UTF-8 encoded path which requires no further processing
-      arguments << "-c" << "core.quotepath=off"
-                << "status" << "--porcelain" << "--" << dir;
+      arguments << "status" << "-z" << "--porcelain" << "--" << dir;
```
kevinushey: Note that we need to include both `--porcelain` and `-z` for compatibility with older versions of Git (tested with Git 1.7.1 on CentOS6)

PR #1311, opened 2017-06-20
provide support for reticulate objects in explorer

kevinushey:
This PR adds support for `reticulate` objects within the object explorer, e.g. inspecting the Python `os` module:

screen shot 2017-06-20 at 2 45 25 pm

@jjallaire, if you have time, can you take a look and provide some comments? Is this worth taking in RStudio for v1.1, and if so any thoughts on how we might better display objects (especially Python functions)? (It should be relatively inexpensive to support `reticulate` here but seems like it could be useful when diving into the internals of some Python objects)

### DIFF
--- a/src/cpp/session/modules/SessionObjectExplorer.R
+++ b/src/cpp/session/modules/SessionObjectExplorer.R
@@ -316,7 +316,10 @@
    expandable <-
       
       # is this an R list / environment with children?
-      (is.recursive(object) && !is.primitive(object) && n > 0) ||
+      (is.recursive(object) &&
+       !is.function(object) &&
+       !is.primitive(object) &&
+       n > 0) ||
       
       # is this an S4 object with one or more slots?
       (s4 && length(.rs.slotNames(object)) > 0) ||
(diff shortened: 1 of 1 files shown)


src/cpp/session/modules/SessionObjectExplorer.R
```diff
@@ -444,6 +453,25 @@
       .rs.explorer.inspectDefault(object, context)
 })
 
+.rs.addFunction("explorer.inspectReticulateObject", function(object,
+                                                             context = .rs.explorer.createContext())
+{
+   children <- NULL
```
jjallaire: I think the cost of inspecting all these types could end up being very high if you force marshalling to R (if you ask for the value via `[[` you will). So e.g. a large NumPy matrix will get copied to R here. 

I think the lightest weight way to do this is to call .DollarNames on the Python object. The return value of this will include an attribute named "types" which will be the same length as the .DollarNames result, with types encoded as follows:

https://github.com/rstudio/reticulate/blob/master/src/python.cpp#L1537-L1576

In general I'd keep the display of types very simple (just "function" should be fine).

kevinushey: Is there a way of accessing these objects without automatically marshaling them back to R?

jjallaire:
We can definitely get some info without marshalling! If it's a Python
object you can use the $ operator to access any attribute or call any
method you like on it. So any thing that you would do in Python to inspect
the object you can do within R with no marshalling (I agree 100% that we
should not force conversions here as it could be very expensive)

kevinushey:
I'm going to continue this work in a new PR.

PR #1313, opened 2017-06-21
provide API for interacting with file dialogs

kevinushey:
This PR exposes three file dialogs, eventually to be used as part of the `rstudioapi` package. The three APIs provided:

1. `openFile()` -- select an existing file on the filesystem
1. `openFolder()` -- select an existing folder on the filesystem
1. `saveFile()` -- select a file to be used for an upcoming save operation

The names are sort of misnomers, since calling those functions does not actually open or save a file; rather, they just return the path to a file. Any thoughts on how the names here could be improved? Maybe just `show*Dialog()` or similar?

### DIFF
--- a/src/cpp/desktop-mac/GwtCallbacks.mm
+++ b/src/cpp/desktop-mac/GwtCallbacks.mm
@@ -1149,88 +1149,9 @@ enum RS_NSActivityOptions : uint64_t
 
 + (NSString *) webScriptNameForSelector: (SEL) sel
 {
-   if (sel == @selector(browseUrl:))
-      return @"browseUrl";
-   else if (sel == @selector(getOpenFileName:label:dir:filter:canChooseDirectories:))
-      return @"getOpenFileName";
-   else if (sel == @selector(getSaveFileName:label:dir:defaultExtension:forceDefaultExtension:))
-      return @"getSaveFileName";
-   else if (sel == @selector(getExistingDirectory:label:dir:))
-      return @"getExistingDirectory";
-   else if (sel == @selector(getUriForPath:))
-      return @"getUriForPath";
-   else if (sel == @selector(onWorkbenchInitialized:))
-      return @"onWorkbenchInitialized";
-   else if (sel == @selector(showFolder:))
-      return @"showFolder";
-   else if (sel == @selector(showFile:))
-      return @"showFile";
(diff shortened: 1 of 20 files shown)


jmcphers:
How about just a `selectPath` API which takes flags to indicate file/folder and new/existing?

kevinushey:
I'm leaning towards having separate APIs here just because I think that makes this easier to use from a user perspective. IMO, it's somewhat easier to articulate "I want to select a file" rather than "I want a file dialog that allows me to select an existing path", if that makes sense.

In addition, some arguments (e.g. filtering on file extensions) are not really relevant when attempting to select a directory, so it make sense to hide those from the user in the exposed APIs.

Of course, this work could happen in the `rstudioapi` rather than RStudio itself, but let me know what you think.

jmcphers:
You're right, the different interface requirements are a good argument for separate API calls. How about `selectFile`/`selectPath`?

jjallaire:
Or selectFile / selectDirectory (since "path" can also refer to a file)

kevinushey:
Okay, this PR is ready for re-review.

Note that this requiring re-wiring of the dialogs used on the Qt side. Rather than relying on the static methods Qt provides for opening specific kinds of dialogs, we now manually create, configure, and execute the dialogs. This was necessary to allow for different 'Accept' button labels -- for example, the 'save' dialog is typically used to select an existing file; however, the label `Save` is not necessarily accurate in the context in which a user is asked to select an existing file.

Most of the dialog code I've added here is modeled after the code here: https://github.com/qt/qtbase/blob/5.4/src/widgets/dialogs/qfiledialog.cpp.

src/cpp/desktop-mac/GwtCallbacks.mm
```diff
@@ -1142,88 +1149,12 @@ - (NSString*) promptForText: (NSString*) title
...
-      return @"closeNamedWindow";
-   else if (sel == @selector(undo:))
-      return @"undo";
-   else if (sel == @selector(redo:))
-      return @"redo";
-   else if (sel == @selector(setPendingProject:))
-      return @"setPendingProject";
-      
-   return nil;
+   NSString* selectorName = NSStringFromSelector(sel);
```
jmcphers: Clever; didn't know this was possible!

PR #1316, opened 2017-06-22
Save and restore terminal environment variables

gtritchie:
Not ready for merge, I want to do more testing, but code is done.

Terminal now persists the environment variables, so they can be restored when a terminal has to be restarted (e.g. after a session suspend, restart, or closing and reopening a project).

This is only on Mac and Linux IDE, and Linux server. This feature is not enabled on Windows IDE; it requires plumbing that I don't have on Windows. Not planning to do it for Windows.

Short of debugging APIs, there is no clean way to get the current environment of a process (procfs, for example, can give you the initial environment, but not the current one). 

So I implemented something JJ had suggested. When terminal is idle, I put the terminal in a private command mode, trigger "/usr/bin/env", capture the output, and save it. The user doesn't see this happening. If user starts typing in the middle of this brief operation, it is queued and played back when private command mode ends.

I don't run the private command more often than the user hits the enter key, idea being if they haven't entered a command (e.g. export FOO=BAR), then the current environment hasn't changed. If they are running a program, either long-running regular program, or a full-screen program, the private command doesn't fire.

I've wired it up in a way that I could add a user preference to disable this feature, but haven't done the actual preference yet.

### DIFF
--- a/src/cpp/core/CMakeLists.txt
+++ b/src/cpp/core/CMakeLists.txt
@@ -115,6 +115,7 @@ set (CORE_SOURCE_FILES
    system/ShellUtils.cpp
    system/System.cpp
    system/file_monitor/FileMonitor.cpp
+   terminal/PrivateCommand.cpp
    tex/TexLogParser.cpp
    tex/TexMagicComment.cpp
    tex/TexSynctex.cpp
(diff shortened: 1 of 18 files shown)


gtritchie:
Pulling this back, need to reconcile some other things first.

PR #1327, opened 2017-06-28
don't index revdep subdirectories

kevinushey:
This PR ensures we don't index sub-directories within the `revdep` folder of a package project (FR from @hadley).


hadley:
Thanks!!

PR #1338, opened 2017-06-30
New command sends entire terminal buffer to new editor buffer

gtritchie:
Terminal dropdown menu now has command "Copy Terminal to New Editor Tab" which sends entire content of current terminal's buffer to a new editor buffer.

This is mainly a workaround for xterm.js lack of scrolling mouse select (i.e. selecting more than will fit on visible display) and lack of Select All / Copy.

Ansi codes are stripped out of the resulting editor buffer. At the moment backspaces are not, so you'll see some red dots in the editor. I'll look at cleaning this up separately.

Also added button to Terminal Diagnostics dialog to append the buffer to the diagnostics text, with Ansi codes preserved (and some pretty-print done, such as <ESC> for raw esc characters).

### DIFF
--- a/src/cpp/session/SessionConsoleProcessApi.cpp
+++ b/src/cpp/session/SessionConsoleProcessApi.cpp
@@ -49,6 +49,16 @@ ConsoleProcessPtr findProcByCaptionReportUnknown(const std::string& caption)
    return proc;
 }
 
+// Return buffer for a terminal, optionally stripping out Ansi codes.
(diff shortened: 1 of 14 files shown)


jjallaire:
How about we name the command "Copy Terminal to Editor" (and it just happens to create a new editor tab). Just concerned about not having that one fairly obscure command dominate the menu because of it's length.

gtritchie:
I will shorten the command name as suggested.

jjallaire:
Great, thanks!

PR #1343, opened 2017-07-03
Warn before killing active terminals

gtritchie:
There was no warning before killing busy terminals via "Restart R", "Build and Reload", or "Clean and  Rebuild".

Received enough feedback to suggest users are quite surprised that these operations kill the terminal processes, and I'm worried they could lose work in the form of long-running jobs, so added warnings.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/buildtools/BuildPresenter.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/buildtools/BuildPresenter.java
@@ -1,7 +1,7 @@
 /*
  * BuildPresenter.java
  *
- * Copyright (C) 2009-12 by RStudio, Inc.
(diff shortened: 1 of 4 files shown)


src/gwt/src/org/rstudio/studio/client/workbench/views/buildtools/BuildPresenter.java
```diff
@@ -343,29 +346,36 @@ private void executeBuild(final String type, final String subType)
    {
       // attempt to start a build (this will be a silent no-op if there
       // is already a build running)
+      final String caption = "Build";
       workbenchContext_.setBuildInProgress(true);
```
jmcphers: Seems like we'll get stuck in the "build in progress" state if the user decides not to proceed with the build after seeing that they have busy terminals?

gtritchie:
Oops, yes, good catch! Fixed.

PR #1344, opened 2017-07-03
re-focus source window if zoomed after render

kevinushey:
This PR re-zooms the Source window after a successful document knit.

The mechanism by which I provide access to the `PaneManager` feels 'wrong', but I don't see a simpler way forward without effectively duplicating a whole bunch of code in a separate class. (Let me know if I'm missing something obvious!)


kevinushey:
Would that provide the same `PaneManager` instance? (I thought that would effectively generate a provider of a separate instance of the `PaneManager` class, as `PaneManager` itself is not a singleton).

It seems like `PaneManager` could (should?) just be a singleton, though.

gtritchie:
That I'm not sure of.

jmcphers:
Let's make it an ordinary singleton instead of creating a wrapper singleton (just not an eager singleton as the existing `Provider` suggests that it needs to be lazy loaded to avoid dependency loops).

kevinushey:
Should be ready for re-review now.

PR #1350, opened 2017-07-06
auto-detect indentation when opening documents

kevinushey:
This PR adds auto-detection of the indent width when opening existing documents.

The intention for this is to support use cases where you might be editing a document living in one project that might use four-space indentation, while also editing a document living in the current project using three-space indentation. This PR ensures that the indentation detected in that document is used.


jmcphers:
Very cool, reminds me a bit of [vim-sleuth](https://github.com/tpope/vim-sleuth)!

It looks like this trumps both user and project tab settings almost all the time, for almost all files. Since it's heuristically based and cannot be overridden I'm worried that this will lead to situations in which we guess the wrong indentation settings and leave the user with no recourse for correcting them. For instance, an R markdown doc has a chunk of YAML at the beginning which could dominate the indent count (if there's little R code in the doc at present) and leave it stuck at 2-space indents. 

If we're going to heuristically set the tab width this way, we need some way for users to opt out of it; that could take the form of any of the following:

- only applying heuristics to files outside the project (those inside the project would continue to be governed by the project's tab settings)
- providing a document-level pref that indicates the desired tab width for the document, and some UX for setting it manually
- respecting existing conventions for manually indicating the tab width, such as modelines or .editorconfig

kevinushey:
Good point! Thinking a bit more, we should also make sure to skip block comments and multi-line strings when attempting to detect the indentation.

It might also be a good idea, when opening a file that belongs to a separate RStudio project, to inherit the project indentation settings from that file's project (rather than just applying whatever the current active project's options are).

jjallaire:
+1 to inheriting from other projects!

I agree that we need to be extra careful here. Maybe only do this for files
not in another project and then only if a certain threshold of confidence
on the implicit indentation level.

I believe that Vim and Emacs also support some magic comments for setting
per-file white space policy, maybe we could support those as well?

PR #1359, opened 2017-07-10
Avoid terminating installation script

mckinsel:
When errexit is set, a bash script will halt as soon as any command has
a non-zero exit status. So the checks here were just halting the script.


kfeinauer:
This looks good. Thank you for your submission!

PR #1360, opened 2017-07-11
Cleanup argument passing for terminal metadata

gtritchie:
Been wanting to do this for a while and it's getting late, but think it's worth it for maintainability.

## Client-side ##
The ConsoleProcessInfo (CPI) object contains metadata about terminals. For whatever reason, early on I started passing and storing subsets of this around in various locations in the client-code. Then it was passed to the server as a bunch of arguments to terminalStart RPC call. This resulted int a bunch of places where I needed to ensure I was building/un-building CPI objects consistently, and keeping things in sync, and lots of little consistency errors crept in which I'm still finding.

I had cleaned some of this up a while ago, but the part I didn't get to is the `TerminalSession` which is constructed with a CPI, yet immediately tears it apart and stores the fields in separate members. Even though the same info was kept as a CPI already in the TerminalList, which is the ultimate owner of this on the client.

Now TerminalSession just keeps the CPI itself and references it as necessary. This is the same CPI object stored by the `TerminalList` so no longer a problem keeping in sync.

## Server-side ##
The `startTerminal` RPC now just takes a ConsoleProcessInfo object as its parameter, instead of all the individual fields. This ripples down through several APIs, cleaning up the signatures.

### DIFF
--- a/src/cpp/session/SessionConsoleProcess.cpp
+++ b/src/cpp/session/SessionConsoleProcess.cpp
@@ -47,24 +47,20 @@ const std::string kEnvCommand = "/usr/bin/env";
 
 // create process options for a terminal
 core::system::ProcessOptions ConsoleProcess::createTerminalProcOptions(
-      TerminalShell::TerminalShellType desiredShellType,
-      int cols, int rows, int termSequence,
-      FilePath workingDir,
-      bool trackEnv,
-      const std::string& handle,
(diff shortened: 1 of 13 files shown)


gtritchie:
Did a final bit of signature cleanup, added some comments. Sanity testing has concluded that this isn't insane.

PR #1376, opened 2017-07-21
update YAML highlight rules

kevinushey:
This PR pulls in the latest version of Ace's YAML highlight rules, and adds some small patches to ensure they play nicely within embedded modes. This PR accomplishes two main things:

1. Multi-line strings are now properly highlighted,
2. Keys of the form `foo::bar` receive a special highlight rule.

screen shot 2017-07-20 at 4 59 52 pm


kevinushey:
This PR also fixes up some of the hackiness around how we attempted to embed YAML highlight rules to work only at the start of a document -- now, YAML blocks are supported anywhere in the document, as per the `pandoc` ruleset here:

> A YAML metadata block is a valid YAML object, delimited by a line of three hyphens (---) at the top and a line of three hyphens (---) or three dots (...) at the bottom. A YAML metadata block may occur anywhere in the document, but if it is not at the beginning, it must be preceded by a blank line.

screen shot 2017-07-21 at 10 35 10 am

This is coming in a bit hot for v1.1; let me know if you'd prefer we save this for v1.2 (since historically PRs that touch highlight rules have been regression prone).

PR #1384, opened 2017-07-25
Fix terminal metadata storage on multi-session

gtritchie:
On multi-session server, the terminal metadata should be strictly per-session. Previously it was per project meaning it was stomping over itself if opening multiple sessions on one project.

Now it uses the session scratch path for multi-session, meaning each new session will start clean (no existing terminals). So, unlike single-session server, if you close a project and reopen, you will not get back the previous set of terminals.


kfeinauer:
@gtritchie LGTM

PR #1390, opened 2017-07-25
fixed bug where multi session test was broken from one of my earlier commits

kfeinauer:
Updated PR for https://github.com/rstudio/rstudio/pull/1389

@jmcphers @kevinushey


kfeinauer:
@kevinushey Any thoughts?

PR #1391, opened 2017-07-26
Allow Run Selected Lines shortcut to work when terminal has focus

gtritchie:
If focus is in terminal, pressing shortcut for `Run Selected Lines` command (Cmd+Enter on Mac) does a regular enter within the terminal.

This change causes the command to run, instead, which will bring the console to the front and execute selected lines from the editor, if any, making it consistent with focus being on other locations in the IDE.


gtritchie:
This was reported by Hadley a while back.

jjallaire:
The next step here would be to to actually add the Run button (for running selected lines) to the Shell script filetype:

screen shot 2017-07-27 at 12 49 29 pm

If we do this then we don't actually need to do any special handling for when the Terminal has focus (just like we don't do special handling for running R code when the Console has focus). We'd just handle the Run command and focus the Terminal automatically whenever it occurs.

PR #1397, opened 2017-07-28
Avoid dark scrollbars on non-osx systems under popups

javierluraschi:
Two issues, 

1) On windows:

screen shot 2017-07-28 at 11 15 25 am

The problem here is that `useScrollbars()` was setting the scrollbars CSS when it was only intended for dark themes. However, since...

2) While changing between dark and light themes scrollbars don't change and require an rstudio restart.

The fix for (1) and (2) is to rather make scrollbars dynamic based on `rstudio-dark-theme` CSS property.

This works for the most part, except for a Safari bug (https://stackoverflow.com/questions/3485365/how-can-i-force-webkit-to-redraw-repaint-to-propagate-style-changes) which prevents scrollbars from changing colors based on CSS changes. The workaround is to set a parent div to `display: none` and then `display: auto`. However, I'm hesitant to implement this workaround since it could cause unintended consequences. Instead, whenswitching themes, you might have to tab to a different pane and then tab back for the effect to take change.

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/theme/res/themeStyles.css
+++ b/src/gwt/src/org/rstudio/core/client/theme/res/themeStyles.css
@@ -2441,8 +2441,9 @@ body.rstudio-themes-flat .rstudio-themes-default {
 }
 
 @if user.agent safari {
-.rstudio-themes-flat .rstudio-themes-scrollbars ::-webkit-scrollbar-thumb,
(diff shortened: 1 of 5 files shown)


javierluraschi:
Also @kevinushey pointed out:

screen shot 2017-07-28 at 4 13 06 pm

which this PR fixes as:

screen shot 2017-07-28 at 4 13 38 pm

javierluraschi:
@jmcphers merging and updating upstream repo.

PR #1408, opened 2017-08-01
(linux) Terminal processes not being killed when closing terminals

gtritchie:
When closing a terminal with running processes (e.g. sleep, vim, so on), on Linux, the processes weren't going away (but the UI was, so little hint that there's a problem).

Fixed by making some existing code not be OSX-specific. This isn't code I wrote or modified for terminal, just used as I found it. This will impact killing all PosixChildProcesses, not just the Terminal scenario. If this is a concern, I could scope this to only the terminal scenario.


gtritchie:
Scoped change to terminal scenario on Linux.

PR #1410, opened 2017-08-01
fixups for tooltip positioning + sizing

kevinushey:
This PR fixes an issue where the tooltip signature was not properly truncated when shown through a mouse idle, and also improves the general positioning behavior (better avoiding right-overflow when the browser width is small).


jmcphers:
Is this a regression in 1.1? (If not should we target for 1.2?)

kevinushey:
Not a regression; just a somewhat long-standing minor-yet-annoying bug.

jmcphers:
Let's target for 1.2 then unless you feel strongly that we should address this before shipping.

kevinushey:
FWIW it might be worth pulling this for v1.1 as this behavior is pretty ugly for some functions, e.g.

screen shot 2017-08-02 at 10 48 49 am

jmcphers:
Okay, if you're confident in the fix and it's relatively scoped (which it appears to be) we'll take it for 1.1.

PR #1423, opened 2017-08-05
Fix focus and auto-advance for send-to-terminal from .sh

gtritchie:
For .sh files executing in terminal, focus now stays in editor if the "Focus console after executing from source" option is false.

Also the cursor advances to next non-blank line (unless text was selected, same behavior as console).

One known issue, if sending text via Cmd+Enter requires the terminal to be created or re-attached, the terminal will take focus that one time. After that it is fine. I'll open a separate issue on that.


jjallaire:
Found an interesting bug, execute the following

```sh
set -e
cd fakedir
```

This blows away the terminal entirely

gtritchie:
Doesn't set -e tell the shell to exit on error? So cd fakedir had an error and the shell exits?

gtritchie:
As a mitigation (kinda) there is a terminal option to keep it open after the shell exits, and shows the exit code, too. You will need to open a new terminal after this happens, it doesn't do it automatically.

jjallaire:
Okay, that actually makes perfect sense.

PR #1435, opened 2017-08-10
additional enhancements to page viewer

jjallaire:
Several additional enhancements:

1) Ability to specify a window title for the page viewer

2) Self-contained HTML is now created directly with pandoc rather than calling the rmarkdown package (this means that all static HTML files are savable and publishable).

3) Some refactoring to give the page viewer it's own server-side codepath (previously it was masquerading as an HTML preview which worked but was fragile and hard to reason about)

4) Support for localhost URLs. This combined with processx makes it possible for packages to create add-ons that extend the API with arbitrary web apps running in other processes (shiny or otherwise)

Not as straightforward a change as the first one, however I still think it's quite safe given that it only touches the codepath dedicated to R Markdown v1 and previewing of plain HTML files, neither of which is done with any regularity by users.

### DIFF
--- a/src/cpp/session/modules/SessionHTMLPreview.cpp
+++ b/src/cpp/session/modules/SessionHTMLPreview.cpp
@@ -987,7 +987,7 @@ void handlePreviewRequest(const http::Request& request,
    }
 }
 
-SEXP rs_showPageViewer(SEXP urlSEXP, SEXP selfContainedSEXP)
+SEXP rs_showPageViewer(SEXP urlSEXP)
 {
    try
    {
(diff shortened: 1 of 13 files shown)


src/cpp/session/modules/SessionHTMLPreview.cpp
```diff
@@ -969,34 +980,52 @@ void handlePreviewRequest(const http::Request& request,
...
+      FilePath filePath, viewerFilePath;
 
-      if (selfContained)
+      // check for url
+      if (boost::algorithm::starts_with(url, "http"))
       {
-         // error if rmarkdown isn't available
-         if (!module_context::isPackageInstalled("rmarkdown"))
+         // verify local url only
+         if (!boost::algorithm::starts_with(url, "http://localhost") &&
```
jmcphers: Do we need to check for https here? (It doesn't seem likely, but `mapUrlPorts` considers it possible.)

jjallaire:
I think we do (just pushed)

src/cpp/session/modules/SessionHTMLPreview.cpp
```diff
@@ -1009,8 +1009,9 @@ SEXP rs_showPageViewer(SEXP urlSEXP, SEXP titleSEXP)
       if (boost::algorithm::starts_with(url, "http"))
       {
          // verify local url only
-         if (!boost::algorithm::starts_with(url, "http://localhost") &&
-             !boost::algorithm::starts_with(url, "http://127.0.0.1"))
+         boost::regex re("^http[s]?://(?:localhost|127\\.0\\.0\\.1):([0-9]+).*$");
```
jmcphers: This regex requires that the URL have a port component -- is that intentional? (seems like a localhost webserver on the default :80 is possible)

jjallaire:
Good point. Just changed!

jmcphers:
Thanks, this looks great!

PR #1440, opened 2017-08-10
Set $HOME to $USERPROFILE in Win32 terminal

gtritchie:
User reported that git was finding `.gitconfig` in different location (`user/documents/.gitbash`) in RStudio terminal (using git-bash as the shell) versus standalone git-bash (found in `user/.gitbash`).

Problem was we weren't setting `$HOME` to `$USERPROFILE`, so it was defaulting to looking in Documents folder.

Fixed by applying same Win32-only logic used in several other locations, both for regular terminal launch, and terminal created via `rstudioapi::terminalExecute`. Also changed the other locations to use a new shared function.


gtritchie:
Thanks @kevinushey for the tip on this one!

andybeet:
My global .gitconfig resides in

get.Sysenv("HOME")

and not 

get.Sysenv("USERPROFILE") - the default Rstudio directory

So Rstudio never sees my global settings and therefore when i commit to github in rstudion my windows "user directory name" is used as as my username, and none of my commits get assigned to my github account. Any plans to have an option in the global Rstudio settings to point to a directory other than get.Sysenv("USERPROFILE") to find gitconfig?

PR #1451, opened 2017-08-16
Fix 1.2 Mac build break

gtritchie:
I don't think the `prelaunchScript` needs to be used for anything on Mac desktop or open-source dev server but @kfeinauer should confirm.


kfeinauer:
@gtritchie That is correct, that missing was just something I missed as I never built on Mac (this change is Linux specific). This LGTM.

PR #1472, opened 2017-08-24
Add type if the custom results contain the attribute

jimhester:
Without this change the type is assumed to be `.rs.acCompletionTypes$UNKNOWN`, which the IDE will use backticks to quote if it contains non-syntastic characters. This allows custom completions to pass a type along to the IDE if desired and along with https://github.com/jimhester/completeme/commit/60a46e9786ab4e39ef3399cd96927366b3079183 makes custom completions work well in the IDE for (https://github.com/hadley/emo/pull/32, https://github.com/hadley/devtools/pull/1595).

It would be nice to have this change in the 1.1 release, but understandable if it is not possible.

### DIFF
--- a/src/cpp/session/modules/SessionRCompletions.R
+++ b/src/cpp/session/modules/SessionRCompletions.R
@@ -3323,6 +3323,7 @@ assign(x = ".rs.acCompletionTypes",
    .rs.makeCompletions(
       token = token,
       results = results.sorted,
-      packages = packages.sorted
+      packages = packages.sorted,
+      type = attr(results, "type")
    )
 })
--- a/src/cpp/session/modules/SessionRCompletions.R
+++ b/src/cpp/session/modules/SessionRCompletions.R
@@ -3309,21 +3309,26 @@ assign(x = ".rs.acCompletionTypes",
    utils:::.completeToken()
    results <- utils:::.retrieveCompletions()
    
+   packages <- NULL
+   type <- .rs.formCompletionVector(attr(results, "type"), .rs.acCompletionTypes$UNKNOWN, length(results))
+   if (type[[1]] %in% c(.rs.acCompletionTypes$UNKNOWN, .rs.acCompletionTypes$FUNCTION)) {
    packages <- sub('^package:', '', .rs.which(results))
    
    # ensure spaces around =
    results <- sub("=$", " = ", results)
    
-   choose = packages == '.GlobalEnv'
-   results.sorted = c(results[choose], results[!choose])
-   packages.sorted = c(packages[choose], packages[!choose])
+     choose <- packages == '.GlobalEnv'
+     results <- c(results[choose], results[!choose])
+     packages <- c(packages[choose], packages[!choose])
+     type <- c(type[choose], type[!choose])
    
-   packages.sorted = sub('^\\.GlobalEnv$', '', packages.sorted)
+     packages <- sub('^\\.GlobalEnv$', '', packages)
+   }
    
    .rs.makeCompletions(
       token = token,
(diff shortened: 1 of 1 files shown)


src/cpp/session/modules/SessionRCompletions.R
```diff
@@ -3323,6 +3323,7 @@ assign(x = ".rs.acCompletionTypes",
    .rs.makeCompletions(
       token = token,
       results = results.sorted,
-      packages = packages.sorted
+      packages = packages.sorted,
+      type = attr(results, "type")
```
kevinushey: Note that completions are constructed from `results.sorted`, not `results` -- I think the `type` attribute needs to be extracted beforehand and sorted to ensure that the order matches the returned completion vector.

jimhester: Actually that brings up the question on if the [scoping / sorting code](https://github.com/jimhester/rstudio/blob/6210df4ffee959fee85e71e3c220c9f765df052b/src/cpp/session/modules/SessionRCompletions.R#L3312-L3321) should be in this function at all. It assumes that the completions will be of function or argument names, which is not the case in any of the custom completions I have implemented thus far.

Perhaps we should get more of the arguments from attributes of the result with defaults if the attributes do not exist.

kevinushey: That's a good point; I think you're right.

Also note that one other underlying issue here is that the completion system only gets access to the current line, whereas a full completion system likely needs to know the whole document context (e.g. if you're requesting argument completions on a new line).

FWIW, RStudio's current completion system does a lot of the completion context generation on the front-end, and then later generates completions server-side using that context. I think it would likely be better (somewhere down the road) to completely eschew the client-side context creation and just do everything server-side.

In other words, we might want to avoid coupling `completeme` to the current state of the RStudio completion system, and instead consider how we might properly re-design this (and later just re-write RStudio to appropriate hook in to `completeme`).

What are your thoughts?

jimhester: Yes I agree, completeme is really somewhat restricted to the information command line R provides, which is only the current line.

I agree it would be nice to provide more information when running in the IDE that could be used for custom completers in the future, but for now I think even this limited functionality is ok.

jimhester:
I just updated this PR to only do the sorting for UNKNOWN and FUNCTION types and also to sort the types if it is one of those types. The code assumes all types are the same, but that seems a valid assumption here because sorting doesn't make sense if the completions are not all function types anyway.

kevinushey:
LGTM!

Given that this only affects custom completers, and is not activated by default, I think this is okay to take for v1.1. (Ie, this change would only affect users who are explicitly loading the `completeme` package).

We might also want to make sure this doesn't regress completions provided when the V8 engine is running. @jimhester can you try testing:

```R
library(V8)
context <- V8::v8()
context$console()
var abc123 = 1;
var abc456 = 2;
ab<TAB>
```

and confirm that the completion candidate is still shown / used?

What do you think, @jmcphers?

jimhester:
V8 looks good!

screenshot 2017-08-25 14 33 19

jjallaire:
@kevinushey Should we verify no interactions when reticulate is providing completions as well?

jimhester:
I tested out some simple reticulate completions and they seem to be working fine with this as well.

screenshot 2017-08-28 12 07 48

kevinushey:
`reticulate` completions don't go through the custom completions code path, so I wouldn't expect them to be affected here. (`reticulate` provides a `.DollarNames` method, which the normal RStudio autocompletion path uses to provide completions here)

PR #1484, opened 2017-08-28
Terminal API fixes (unique id, default args)

gtritchie:
Code changes are tightly scoped to Terminal API code paths, which is the last thing I'm (hopefully) making tweaks to for 1.1.

The terminal API was using "caption" as a unique identifier, but code would break if the user renamed a terminal session via the UI. This was silly, since there is an actual unique ID (8-char guid, generally named `handle`) already in place that can't be changed once created. Fixed API to accept/return this ID, instead.

`rstudioapi::terminalCreate` wasn't treating `caption` as optional as documented/intended; now it is and system will generate a caption automatically.

`rstudioapi::terminalActivate` wasn't treating `id` as optional as documented/intended; now it is, and system will activate the terminal tab without selecting a specific terminal when not specified.


src/cpp/session/SessionConsoleProcessApi.cpp
```diff
@@ -337,7 +339,7 @@ SEXP rs_terminalClear(SEXP idSEXP)
    // send the event to the client; if not connected, it will get the cleared
    // server-side buffer next time it connects.
    json::Object eventData;
-   eventData["id"] = terminalId;
+   eventData["id"] = proc->getCaption();
```
kevinushey: Is it worth making the change here to read `eventData["caption"] = proc->getCaption()`? Or is the notion of `id` on the client side still linked to the caption value? (I see we enforce that captions are unique in `terminalCreate` which I think is the right choice to avoid confusion)

gtritchie: The client-side still takes a caption. To minimize the size of the change for 1.1 I left it that way, but will clean it up in 1.2. The user can't rename the caption while this API is in flight, so won't suffer the problem I'm fixing overall.

kevinushey: Got it!

kevinushey:
LGTM! Please merge when ready.

PR #1491, opened 2017-08-31
Fix stuck console scrolling on high-dpi Windows

gtritchie:
With Windows display zooming at a variety of settings (> 100% - 225%) I observed a frequent off-by-one in the values returned by `ScrollPanel` used to determine if it is scrolled to the bottom.

So for Windows, I allow a fudge-factor of 1 pixel to prevent auto-scrolling from getting stuck.

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/widget/BottomScrollPanel.java
+++ b/src/gwt/src/org/rstudio/core/client/widget/BottomScrollPanel.java
@@ -1,7 +1,7 @@
 /*
  * BottomScrollPanel.java
  *
- * Copyright (C) 2009-12 by RStudio, Inc.
+ * Copyright (C) 2009-17 by RStudio, Inc.
  *
  * Unless you have received this program directly from RStudio pursuant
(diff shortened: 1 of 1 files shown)


jmcphers:
Does this only happen on the Windows desktop? I'd be surprised if it didn't also repro on high DPI Linux (or at least Linux at > 100% zoom) since they're both on Qt.

FWIW I'd be a fan of increasing the fudge factor *and* allowing this to apply to any OS/platform, presuming the effect is that you can be a few px off the bottom and still get auto-scroll (possibly what users want anyway if they can still see most of the last line!).

gtritchie:
I was going for minimally scoped fix, but agree there seems to be little
danger in making it platform agnostic and bumping up the fudge factor.
Stand by...

kevinushey:
I'm also a fan of just checking if `getMaximumVerticalScrollPosition() - getVerticalScrollPosition() <= 1` or something similar.

gtritchie:
@kevinushey I didn't want to assume which value was coming back slightly off, hence the `abs` (if that's what you meant).

PR #1504, opened 2017-09-08
allow non-framework R installation

randy3k:
I am on macOS 10.12 and R was installed via homebrew. Since it was not installed as a framework, `LIBR_LIBRARIES` returns `/usr/local/lib/libR.dylib` instead of `/<path to framework>/R.framework`. 

This PR extends `FindLibR.cmake` to find such R installation.

### DIFF
--- a/cmake/modules/FindLibR.cmake
+++ b/cmake/modules/FindLibR.cmake
@@ -21,11 +21,24 @@
 if(APPLE)
 
    find_library(LIBR_LIBRARIES R)
-   if(LIBR_LIBRARIES)
+
+   if(LIBR_LIBRARIES MATCHES ".*\\.framework")
       set(LIBR_HOME "${LIBR_LIBRARIES}/Resources" CACHE PATH "R home directory")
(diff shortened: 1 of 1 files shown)


jjallaire:
I don't recall all of the reasons why, but we specifically didn't support non-Framework installations in the early days b/c something else was failing downstream. I do believe that homebrew has a Framework-based version of their R package though, is that not the case?

randy3k:
It had, but the option was removed recently https://github.com/Homebrew/homebrew-core/pull/14416

I am using a non framework based R, and I cannot see anything obviously wrong now though.

jjallaire:
Okay, if things work as expected we could take this PR, but I'd encourage
you to keep running in this mode to ensure there isn't a corner we've
missed.

randy3k:
Thank you for considering it.

jmcphers:
I can't think of a reason this wouldn't work either! Appreciate the contribution and please do let us know if you run into trouble when compiling with Homebrew's R.

randy3k:
It is working perfectly. Will report if any errors occur.

PR #1553, opened 2017-10-05
Add openssl support to Windows build environment

gtritchie:
**Requires rerunning `install-dependencies` and `cmake`.**

This seeds openssl for Windows for other ongoing work not included in this PR, I have confirmed it is working as expected in that other branch.

The openssl package was built via cross-compilation using MinGW on Linux. See README inside the downloaded package for full details.

### DIFF
--- a/src/cpp/core/system/Crypto.cpp
+++ b/src/cpp/core/system/Crypto.cpp
@@ -70,18 +70,18 @@ Error lastCryptoError(const ErrorLocation& location)
                          "lastCrytpoError called with no pending error",
                          location);
    }
-
+   
    // get the error message (docs say max len is 120)
-   const int ERR_BUFF_SIZE = 250;
+   const int ERR_BUFF_SIZE = 250; 
    char errorBuffer[ERR_BUFF_SIZE];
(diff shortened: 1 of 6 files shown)


jmcphers:
Do we need to add this to `NOTICE`?

src/cpp/core/system/Crypto.cpp
```diff
@@ -70,26 +70,26 @@ Error lastCryptoError(const ErrorLocation& location)
                          "lastCrytpoError called with no pending error",
                          location);
    }
-
```
kevinushey: Can we try to avoid making whitespace-only changes in the diffs?

gtritchie:
Sure, guess I’ve got editor set to automatically clean up stray whitespace. You can add ?w=1 to github url to hide it this time around.

gtritchie:
I’ll take a look but shouldn’t it already be there since we use it on server already?

jmcphers:
We do have openssl in NOTICE already, just double-check that what you're adding has exactly the same license/terms as what exists today.

gtritchie:
Cleaned up whitespace changes and updated NOTICE. Should be good to go.

PR #1591, opened 2017-10-11
fork python highlight rules

kevinushey:
This PR forks Ace's Python highlight rules, in preparation for future improved editor support. We also implement a bit of extra highlighting, to better match how we highlight in R modes. (e.g. we now highlight brackets and `.`)

screen shot 2017-10-11 at 3 44 43 pm


kevinushey:
These are forked from our private Ace copy; I don't think any of these changes are upstream. (I haven't proposed them yet as I'm not sure if they're desired there)

PR #1720, opened 2017-11-06
upgrade to Qt 5.9.2 + Qt WebEngine

kevinushey:
NOTE: Not ready for merge but putting this up for initial review + organization.

---

This PR seeks to update Qt to 5.9.2, and alongside that port our existing Qt code base from the older, deprecated QtWebKit to the newer QtWebEngine. In this PR, we seek only to upgrade the Linux + Windows desktop components; migrating macOS to QtWebEngine will be explored in a later PR.


kevinushey:
FWIW, I haven't done any work here to ensure compatibility with the current macOS sources -- my expectation is that we'll take this PR once it's ready as-is, and then later work on getting the macOS builds up and running in a separate PR.

kevinushey:
I merged the MSVC branch back in here; please forgive me if this leads to some breakage re: work that happened in the last couple weeks!

gtritchie:
Cool, I will do a sanity check on Mac and Linux.

gtritchie:
Yes, that was just starting to occur to me. Is this something we should
consider getting into 1.1 (no idea how likely it would be to lead to
crashes in the released product, but seems like a non-zero possibility).
But also low-level so maybe best to wait for 1.2.

PR #1761, opened 2017-11-16
build Windows sources with MSVC

kevinushey:
This PR moves us over from building with MinGW to MSVC, as required by Qt 5.9.2 and QtWebEngine (Chromium).


src/cpp/desktop/DesktopMain.cpp
```diff
@@ -308,7 +308,6 @@ int main(int argc, char* argv[])
       bool devMode = false;
 
       // check for debug configuration
-#ifndef NDEBUG
```
kevinushey: This should instead be something like `#ifndef RSTUDIO_PACKAGE_BUILD` (not sure if we populate that define or something similar already)

kevinushey:
@jmcphers @gtritchie: how should we avoid further divergence with https://github.com/rstudio/rstudio/pull/1720?

gtritchie:
Could we get this merged back into a single branch and have the three of us working there? I would hope the Windows/MSVC changes are relatively separate from more general ongoing Linux/Mac work (but could be very wrong about that).

PR #1869, opened 2017-12-18
git: improve handling of multibyte paths

kevinushey:
This PR fixes an issue where attempts to stage a path whose filename is not representable in the active system encoding could fail. We accomplish this by:

- Adding a switch, allowing the `ShellArgs` class to avoid conversion of file paths to the system encoding;
- Using the wide API for creating a system process, and ensuring all the bits passed in are 'wide' (UTF-16) strings.

This PR deserves extra scrutiny since it exercises a heavily used codepath (Windows process creation). Fortunately, the changes are relatively straightforward -- we use `utf8ToWide()` to construct wide strings, and pass those on to the wide Windows APIs as needed.

### DIFF
--- a/src/cpp/core/include/core/system/ShellUtils.hpp
+++ b/src/cpp/core/include/core/system/ShellUtils.hpp
@@ -52,6 +52,12 @@ enum EscapeMode
    EscapeFilesOnly
 };
 
+enum EncodingMode
+{
+   SystemEncoding,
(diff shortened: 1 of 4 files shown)


jimhester:
@kevinushey Do you have a good test case for this issue? Similar problems happen in other projects so it would be nice to have a easy way to test for correct handling.

PR #1898, opened 2018-01-02
Fixed Windows unit tests building

kfeinauer:
The first issue prevented one of the unit tests from building due to include order due to inclusion of Boost Deadline Timer, but this was fixed by simply changing the wait method to a simple thread sleep. Second issue occurred because building a package build for Win64 was not linking the correct x64 OpenSSL lib.

Let me know if there was a reason the OPENSSL_ARCH was not being set by default for x64 - perhaps I am missing something.

### DIFF
--- a/src/cpp/CMakeLists.txt
+++ b/src/cpp/CMakeLists.txt
@@ -218,7 +218,7 @@ if(WIN32)
      include_directories(SYSTEM ${WINPTY_INCLUDEDIR})
 
      # openssl for Windows
-     if(RSTUDIO_SESSLIBRARY)
+     if(RSTUDIO_SESSION_WIN64 OR RSTUDIO_SESSLIBRARY)
         set(OPENSSL_ARCH "/x64/")
      else()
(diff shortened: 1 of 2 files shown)


kevinushey:
I suspect we're setting `RSTUDIO_SESSLIBRARY` somewhere on the build machine or in our build scripts (I can't see where we set it in the current RStudio sources).

FWIW on the Qt WebEngine branch we're doing something very similar to your proposed change:

https://github.com/rstudio/rstudio/blob/dc45253a72ae4956ce78b4b148bc5e76f5660c10/src/cpp/CMakeLists.txt#L250-L254

But there might be some other motivation that I'm not aware of. (From what I understand, when we build the 64-bit components of RStudio, we put them into a `x64` sub-directory -- perhaps that code is there to help enforce that)

kfeinauer:
@kevinushey Good to know. So it looks like this merge should be fine and whatever you have going on in the Qt WebEngine branch will supersede what I have done in `CMakeLists.txt`.

PR #2002, opened 2018-01-17
Add StringUtil.equals method for null-safe string comparison

gtritchie:
Per a recent conversation, we decided to steer in the direction of using Java-compatible string comparision conventions, even though GWT output makes str1 == str2 an ok way to compare contents of strings (unlike Java).

For cases where null is a factor, added `StringUtil.equals(str1, str2)`. This defers to the newer Java `Objects.compare(str1, str2)`` method, but has the added benefit of compile-type type-safety (and gives a place to change our approach, should we need to).

For comparisions where we know the strings are not-null, then continued use of `str1.compare(str2)`` is the way to go, avoiding the slight performance penalty of the null checks inside Objects.compare.

Also discovered that the core/client unit tests (such as for StringUtil) were not being run. Modified build.xml to include them.

Changed the Objects.equals I had previously added when IntelliJ complained about it. Did not do a sweep looking for `==` usages.

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/StringUtil.java
+++ b/src/gwt/src/org/rstudio/core/client/StringUtil.java
@@ -1,7 +1,7 @@
 /*
  * StringUtil.java
  *
- * Copyright (C) 2009-17 by RStudio, Inc.
+ * Copyright (C) 2009-18 by RStudio, Inc.
  *
  * Unless you have received this program directly from RStudio pursuant
  * to the terms of a commercial license agreement with RStudio, then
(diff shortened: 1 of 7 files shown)


src/gwt/src/org/rstudio/core/client/StringUtil.java
```diff
@@ -1165,10 +1166,20 @@ public static native int newlineCount(String str) /*-{
...
    
+   /**
+    * Compare two strings, works if one or both strings are null.
+    * @param str1
+    * @param str2
+    * @return true if non-null strings are equal, or both are null
+    */
+   public static final boolean equals(String str1, String str2)
+   {
+      return Objects.equals(str1, str2);
```
kevinushey: What do you think about using native JSNI here and directly calling `str1 == str2`? Just to avoid any potential overhead that `Object.equals()` might force upon us.

gtritchie:
@kevinushey Good idea, done.

PR #2011, opened 2018-01-18
Reimplement Mac Dock menu for opening new RStudio window

### LINKED ISSUE #2004: "New RStudio Window" command missing from dock menu
On RStudio 1.1.x for Mac, right click on the dock menu and there's a command to launch a new window:

This is missing with the 1.2.243+ builds.

gtritchie:
This was lost in the move back to Qt. Resolves #2004.

### DIFF
--- a/src/cpp/desktop/DesktopUtils.cpp
+++ b/src/cpp/desktop/DesktopUtils.cpp
@@ -35,6 +35,7 @@
 #include <core/system/Environment.hpp>
 
 #include "DesktopOptions.hpp"
+#include "DesktopMainWindow.hpp"
 
 #ifdef Q_OS_WIN
 #include <windows.h>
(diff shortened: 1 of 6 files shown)


src/cpp/desktop/DesktopUtils.cpp
```diff
@@ -45,6 +45,8 @@ using namespace rstudio::core;
 namespace rstudio {
 namespace desktop {
 
+class DesktopMainWindow;
```
kevinushey: Should this be `MainWindow`? (looks like the classname below is just `MainWindow`)

gtritchie: Oops, yes, nice catch. Already have the forward decl in the header, adding the actual include instead. Not strictly needed since non-Mac doesn't do anything with it yet, but just to be consistent.

PR #2017, opened 2018-01-19
use Q_SIGNALS / Q_SLOTS (avoid collisions with boost)

kevinushey:
This is mainly to allow us to use `boost::signals` more easily within the desktop sources. (We don't use `boost::signals` there currently but they may still be useful for 'long-range' communication between objects)


gtritchie:
FYI, this will require a bit of Pro-specific work (there are Qt-signal-using-pro-only source files for desktop). We're not building that currently, so I can clean it up after the fact.

PR #2022, opened 2018-01-19
use standalone script for installing Qt components

kevinushey:
This PR introduces our own standalone Bash script for bundling Qt components as necessary with a Linux installation.

Don't merge quite yet; need to test on our Linux machines + Docker images.


kevinushey:
Okay, I've confirmed that this:

- Produces a working `.deb` package build on `rstudio-mega`, and
- Successfully produces a `.rpm` with `docker-compile.sh` on CentOS 7.

@jmcphers would you be willing to review?

PR #2025, opened 2018-01-20
Fix Bring All to Front on Mac QtWebEngine

### LINKED ISSUE #1955: Bring All to Front not working as expected on OS X
The Bring All to Front command in the QtWebEngine version doesn't work correctly:

1) It never seems to bring secondary windows (e.g. the Git Window) to the front

2) With 2 instances of the IDE open, in one of the instances Bring All to Front is a no-op and in the second instance it brings the first instance to the front.

gtritchie:
Resolves #1955, including **Window -> Bring All to Front** menu and clicking on a single window in the Window menu.

### DIFF
--- a/src/cpp/desktop/DesktopMenuCallback.cpp
+++ b/src/cpp/desktop/DesktopMenuCallback.cpp
@@ -1,7 +1,7 @@
 /*
  * DesktopMenuCallback.cpp
  *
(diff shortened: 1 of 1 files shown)


kevinushey:
Nice! I was worries this was going to be hard to hunt down; glad that there are Qt APIs available for what we wanted here.

PR #416, opened 2015-05-08
Fix a few issues related to multibyte chars in R Markdown

yihui:
I'm not sure if I used `string_utils::utf8ToSystem()` correctly. At least the two commits in R worked for me on Windows.

For the third commit, I experimented with `Encoding(input) = 'UTF-8'` inside `rmarkdown::default_output_format` and it also worked, which means the path should have been either marked with UTF-8 or converted to system native encoding. This will make this error go away:

```
Error in file.exists(epath <- path.expand(x)) :
  file name conversion problem -- name too long?
```

PR #2028, opened 2018-01-22
handle warnings involved by Chinese Characters

XiangyunHuang:
handle warnings involved by Chinese Characters under Windows OS

### DIFF
--- a/src/cpp/session/modules/SessionCodeTools.R
+++ b/src/cpp/session/modules/SessionCodeTools.R
@@ -176,6 +176,8 @@
 
    code <- deparse(func, width.cutoff = 59, control = control)
    
+   Encoding(code) <- "UTF-8"
+   
    # if we were asked not to use source refs, or we were but there wasn't a
    # source ref to use, then format the code according to user pref
    if (!useSource || is.null(attr(func, "srcref", exact = TRUE)))
--- a/src/cpp/session/modules/SessionCodeTools.R
+++ b/src/cpp/session/modules/SessionCodeTools.R
@@ -176,8 +176,6 @@
 
    code <- deparse(func, width.cutoff = 59, control = control)
    
-   Encoding(code) <- "UTF-8"
-   
    # if we were asked not to use source refs, or we were but there wasn't a
    # source ref to use, then format the code according to user pref
    if (!useSource || is.null(attr(func, "srcref", exact = TRUE)))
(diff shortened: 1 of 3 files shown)


XiangyunHuang:
the problem is similar to #416

jmcphers:
Thank you for putting this together! Can you tell us more about the warnings you're seeing and/or give an example of the problem this PR addresses?

Are you using UTF-8 encoding for all of the files in question?

I haven't reviewed this in detail yet (so perhaps these usages are okay), but in general it isn't safe to assume that all string data on the R side is UTF-8 encoded; in some cases the data comes from the file itself, which isn't necessarily UTF-8.

XiangyunHuang:
a minimal example, as follow

example

The code used in the picture,

```r
address <- "1.待导入/11·29result.xlsx"
plot(iris[,-5], col = iris[,5], main = '鸢尾花数据')
```

The file is encoded by "UTF-8"

code

PS,
R version 3.4.3
RStudio 1.1.414
Windows 8.1

src/cpp/session/modules/SessionCodeTools.R
```diff
@@ -176,6 +176,8 @@
 
    code <- deparse(func, width.cutoff = 59, control = control)
    
+   Encoding(code) <- "UTF-8"
```
kevinushey: Are we sure that the function body is deparsed using UTF-8 encoding? It seems like this may not be true on Windows (or perhaps it depends on the value of the `getOption("encoding")` R option)

yihui: I agree. This may not be a safe change. @XiangyunHuang Could you check if the original issue https://d.cosx.org/d/419665 could be fixed without this change? In general, we tend to prefer minimal changes that fix a problem.

yihui:
I fixed similar issues in 2015: https://github.com/rstudio/rstudio/pull/416/files

To address @jmcphers's concern, I think a safer strategy is to test if UTF-8 actually works:

```r
utf8_maybe = function(x) {
  x2 = iconv(x, 'UTF-8', 'UTF-8')
  if (any(is.na(x2)) x else x2
}
```

Another strategy is to make sure the caller passes the string with a marked encoding, which will be the fundamental fix.

Yet another strategy is to factor out

```r
strsplit(code, "\n", fixed = TRUE)
```

into a separate function, and deal with the encoding problem in a single function.

The original issue was reported at https://d.cosx.org/d/419665 and I asked @XiangyunHuang to submit a PR, since two Chinese users had verified that fix worked for them.

PR #2045, opened 2018-01-23
Stop Ctrl+L from sending focus to browser address when in Terminal

### LINKED ISSUE #1636: Ctl-L in Terminal focuses browser address bar on RStudio server
In the Terminal tab, clearing the output with Ctl-L gets caught by the web browser (Chrome) and focuses the address bar. This means that Ctl-L followed by return (a common pattern) ends up reloading the page...

The Console tab has the correct behavior of keeping the cursor in the Console tab.

gtritchie:
On Chrome (Windows and Linux, not Mac) hitting Ctrl+L in the RStudio terminal clears the terminal (as it should), but then puts focus in the browser address bar. This is quite painful, especially if you are a fast typist, hit Ctrl+L in terminal, then type a command and hit enter; now the browser is going somewhere you don't want it to.

Resolves #1636.

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/command/ShortcutManager.java
+++ b/src/gwt/src/org/rstudio/core/client/command/ShortcutManager.java
@@ -1,7 +1,7 @@
 /*
  * ShortcutManager.java
  *
- * Copyright (C) 2009-17 by RStudio, Inc.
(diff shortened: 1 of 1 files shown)


kevinushey:
LGTM!

PR #2116, opened 2018-02-02
register '.ini' file type

kevinushey:
This PR adds the `.ini` file type to RStudio (which Ace fortuitously provides a highlight mode that we already bundle).


kevinushey:
Also augmented this PR with the system-reported MIME type of filetypes we haven't registered, so that we can successfully open text files through `file.edit()`.

This is necessary because we assume unknown files (when opened within RStudio Desktop) are octet streams. https://github.com/rstudio/rstudio/blob/69a22d1289046f0a11461b16e5c4065446a7797e/src/gwt/src/org/rstudio/studio/client/common/filetypes/FileTypeRegistry.java#L572-L578

PR #2123, opened 2018-02-03
properly parse document sources in .Rd doc

kevinushey:
This PR ensures we properly display the set of files used when an `.Rd` document is generated from multiple files.

screen shot 2018-02-03 at 1 03 34 pm

TODO is still to handle clicking links to navigate to the associated file.


kevinushey:
Now with clickable links:

screen shot 2018-02-03 at 3 09 23 pm

Should be ready for review. cc: @hadley

hadley:
This looks great!

@batpigandme: this should hopefully reduce the number of people editing the wrong file (at least from within RStudio)

batpigandme:
@hadley even if it means no one will ever have to see the "I-said-don't-come-in-here" gif, it's totally worth it. 🎉

PR #2172, opened 2018-02-10
fix error handling on Win32 debug builds

gtritchie:
On Win32 debug, the following pattern doesn't work:
```
Win32 call that sets an error (SetLastError)
Error error = systemError(::GetLastError(), ERROR_LOCATION);
```
The order of argument evaluation is not defined in C++ and on Win32 debug, at least, by the time ::GetLastError is called, the Win32 error code has been reset to zero, so the resulting `error` is set to an error code of zero (no-error). This leads to a variety of problems at runtime, including crashing failure of the rsession unit tests.

I opted to just manually change the calls to first store ::GetLastError in a variable which is then passed to systemError.

Doing it inside a new utility function would lose useful diagnostic information (ERROR_LOCATION would resolve to that function instead of the calling site), and a preprocessor macro that can evaluate to a return value of core:Error while capturing the original Win32 error and call-site location is tough without using non-portable behavior such as gcc "statement expressions". There may be a way, but it was faster to just do this manual search and replace.

### DIFF
--- a/src/cpp/core/FilePath.cpp
+++ b/src/cpp/core/FilePath.cpp
@@ -1076,8 +1076,7 @@ Error FilePath::open_r(boost::shared_ptr<std::istream>* pStream) const
                                    NULL);
       if (hFile == INVALID_HANDLE_VALUE)
       {
-         auto lastErr = ::GetLastError();
-         Error error = systemError(lastErr, ERROR_LOCATION);
+         Error error = LAST_SYSTEM_ERROR();
          error.addProperty("path", absolutePath());
          return error;
       }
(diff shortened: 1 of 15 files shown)


kevinushey:
I wonder if we could instead force order of evaluation by creating the error with two separate function calls, e.g.

```cpp
Error error = systemError(::GetLastError()).at(ERROR_LOCATION);
```

And then this could be wrapped up into a macro (e.g. `LAST_SYSTEM_ERROR()`). Does that sound reasonable? (Not sure if that's really an improvement; it feels like perhaps it may be better to be explicit about when and where `::GetLastError()` is called just to remind our future selves)

gtritchie:
Works via placing a self-invoking lambda in a macro. What a curious mix of modern C++ and crusty old C. Working on updating the PR.

PR #2175, opened 2018-02-11
disable accessibility by default

kevinushey:
As per https://github.com/rstudio/rstudio/issues/1539, it looks like the issue with accessibility-aware rendering in Chromium causing lagginess in the IDE affects a number of users. Because of this, we should consider disabling this by default, but allowing users who require this functionality to re-enable it (by setting the `RSTUDIO_ACCESSIBILITY` environment variable)

### DIFF
--- a/src/cpp/desktop/DesktopMain.cpp
+++ b/src/cpp/desktop/DesktopMain.cpp
@@ -261,10 +261,10 @@ int main(int argc, char* argv[])
       arguments.push_back(disableWebSecurity);
 #endif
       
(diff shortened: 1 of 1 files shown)


kevinushey:
Agree 100%!

One thing to keep in mind is that, unfortunately, the current version of Ace we use apparently does not work well with screen readers (some users have reported that they're not able to read the contents of an editor instance) but I believe this is something resolved in the newest version of Ace.

I definitely agree that we should provide something more accessible than an environment variable. I'll file that as a separate issue for v1.2 so we can discuss where that preference should live.

PR #2182, opened 2018-02-12
Fix a handful of build warnings on Windows

gtritchie:
MSVC build emits several hundred warnings. Fixing a few of them, just to start chipping away at the mountain. Mainly some type issues, and a few unused arguments in catches.

### DIFF
--- a/src/cpp/core/FilePath.cpp
+++ b/src/cpp/core/FilePath.cpp
@@ -134,7 +134,7 @@ bool copySingleItem(const FilePath& from, const FilePath& to,
    return true;
 }
 
-bool addItemSize(const FilePath& item, boost::shared_ptr<int> pTotal)
+bool addItemSize(const FilePath& item, boost::shared_ptr<uintmax_t> pTotal)
 {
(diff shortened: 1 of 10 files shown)


kevinushey:
LGTM, but I wonder if we should consider relaxing some of the warning emitted by MSVC as well? (Or just live with them and continue chipping away as time permits)

I think some of the issues are indeed worth fixing (especially related to type conversions, and those conversion that might potentially lose information) but others (unused parameters) are less important.

gtritchie:
Yes, we should probably turn off C4800 (the "performance warning" about casting non-bools to bool). That warning was removed altogether from MSVC in newer versions because it is so noisy.

The ones I changed here were some semi-legitimate type issues (narrowing conversions, and some things that were int but actually meant to be bool).

gtritchie:
Those exception parameters were the only "unused parameters" it noticed. I tend to eliminate those as I see them (CLion also flags them, and it's something I'm used to doing from when I was building with MSVC back in the day).

PR #2192, opened 2018-02-14
Add support for keyring dialog prompts

javierluraschi:
Adds support for `keyring` based dialogs to persist secrets with ease, useful with driver connections. Prompt can be triggered with `rstudioapi::askForSecret()`:

screen shot 2018-02-15 at 10 19 25 am

Installation can be performed from dialog and secret is persisted for subsequent calls if users choose the "remember" checkbox:

screen shot 2018-02-15 at 10 18 52 am

Added help redirect and support article already: https://support.rstudio.com/hc/en-us/articles/360000969634


jjallaire:
I would change "Remember using keyring?" (which sort of reads like "Do you remember when you used to use keyring?") to "Save with keyring"

src/gwt/src/org/rstudio/studio/client/common/dependencies/DependencyManager.java
```diff
@@ -675,6 +675,27 @@ public void execute(Boolean succeeded)
...
      );
    }
+
+   public void withKeyring(final Command command)
+   {
+     withDependencies(
+        "Preparing Keyring",
+        "Using keyring", 
+        new Dependency[] {
+           Dependency.cranPackage("keyring", "1.0.0", true)
```
jmcphers: Can you update the admin guide to include this package?

https://github.com/rstudio/rstudio-pro/blob/6b14277b5aabd8de1b7225291f1cf3099c56f582/docs/server/index.Rmd#L184

src/gwt/src/org/rstudio/studio/client/common/rstudioapi/AskSecretManager.java
```diff
@@ -0,0 +1,167 @@
...
+                                 result.getRemember(),
+                                 result.getHasChanged(),
+                                 new VoidServerRequestCallback(indicator));
+                              
+                           }
+
+                           @Override
+                           public void onFailure(ServerError error)
+                           {
+                              Debug.logError(error);
```
jmcphers: Since this is a waitformethod, I think we need to invoke `askSecretCompleted` even if there's an error so that the method can complete on the server side.

jjallaire:
We could also use “Remember with keyring”

src/gwt/src/org/rstudio/studio/client/common/rstudioapi/AskSecretManager.java
javierluraschi: For sure, the legacy password fialog also needs update, fixed.

PR #2298, opened 2018-03-02
Fix a couple thousand linker warnings(debug win32 desktop)

gtritchie:
The openssl debug libs for Windows did not contain debug info, and this caused thousands of unsuppressable linker warnings when building RStudio in debug config on Windows.

Tweak the build script; will push the rebuilt package to buildtools once this is approved.


gtritchie:
Hmm, I need to verify that this fixed the problem. I might have build release when I thought I was testing a debug build.

gtritchie:
Ok, confirmed this is working. I was juggling too many identically named zip files.

PR #2333, opened 2018-03-06
Restore R version selection option for Win desktop

### LINKED ISSUE #2297: Can't select R version from Global Options on Windows daily build
### System details

    RStudio Edition : Desktop 
    RStudio Version : 1.2.374
    OS Version      :  Windows 7
    R Version       :  3.4.3

Tools > Global Options > R version selector is missing

In 1.1, you could select which version of R to run from the same setting:

image

Be able to select the setting from the same place in RStudio Desktop 1.2

gtritchie:
A lingering TODO from the QtWebEngine switchover.

Fixes #2297

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/application/DesktopFrame.java
+++ b/src/gwt/src/org/rstudio/studio/client/application/DesktopFrame.java
@@ -128,6 +128,7 @@ public interface DesktopFrame extends JavaScriptPassthrough
    void bringMainFrameToFront();
    void bringMainFrameBehindActive();
 
+   // R version selection currently Win32 only
    void getRVersion(CommandWithArg<String> callback);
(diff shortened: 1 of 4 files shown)


src/gwt/src/org/rstudio/studio/client/application/DesktopFrame.java
```diff
@@ -127,7 +127,10 @@ void promptForText(String title,
 
    void bringMainFrameToFront();
    void bringMainFrameBehindActive();
-   
+
+   void getRVersion(CommandWithArg<String> callback);
```
kevinushey: Maybe add a comment warning that these are currently only implemented for Windows?

gtritchie: Those have always been Windows-only, but sure, will add a comment.

Issue #1860, opened 2017-12-15
Terminal websocket closed by proxy due to inactivity


gtritchie:
Running RStudio Server behind a proxy configured with a read-timout (for example, nginx with  proxy_read_timeout option). Start a terminal session using websockets. Now don't do anything in RStudio and wait for the timeout to expire.

Ideally, terminal should send occasional keepalive messages, but currently it does not. So, this may appear in the console, and websocket is closed.

```
[fatal] handle_write_frame error: websocketpp.transport:10 (A transport action was requested after shutdown)
```
Now, due to #1844, terminal cannot be used.

We should consider a websocket keepalive message to prevent websocket death, but don't want it to also keep the rsession alive.

We also have reports of this from containerized hosting of RStudio Server.

gtritchie:
Bumping up priority. This shows up frequently in rstudio.cloud.

dfalty:
Verified 1.2.451-1.

PR #2369, opened 2018-03-10
WebSocket keep-alive pings for terminal

gtritchie:
Proxies can be configured to close inactivate websocket connections. This can lead to error messages showing in the console (in the case of .cloud), and possibly other issues.

Now Terminal websockets will send a 1 character "ping" followed by the server sending back a 1 character "pong", to reassure proxies that the websocket is still in use. Both directions, as proxies can be configured to pay attention to one or both directions.

Default interval is 10 seconds (per each terminal connection). Can be changed or disabled with `websocket-ping-seconds=n`. Only done for Terminal websockets, but concept (and the setting) could be extended to other uses of WebSockets in the product.

Using a super-sophisticated packet format, to keep overhead minimal. If the first character of a terminal websocket message is "a", then what follows is text. If it's "b", then it's a ping/pong message. Patent-pending.

Fixes: #1860 

FYI @ssinnott @kippandrew

### DIFF
--- a/src/cpp/session/CMakeLists.txt
+++ b/src/cpp/session/CMakeLists.txt
@@ -71,6 +71,7 @@ set (SESSION_SOURCE_FILES
    SessionConsoleProcessInfo.cpp
    SessionConsoleProcessPersist.cpp
    SessionConsoleProcessSocket.cpp
+   SessionConsoleProcessSocketPacket.cpp
    SessionConsoleProcessTable.cpp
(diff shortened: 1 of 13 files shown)


jspiewak:
@dfalty could you let us know when there is a tested build with this fix please?

dfalty:
@jspiewak I verified the fix in 1.2.451-1 and that build is good to deploy.

PR #2425, opened 2018-03-17
show matplotlib plots in Plots pane

kevinushey:
This PR overrides `matplotlib.show()` such that plots are captured and shown in the Plots pane instead.

screen shot 2018-03-16 at 6 06 21 pm

There are some TODOs that we might want to think about; potentially for this PR or future PRs:

- [ ] We might want to `pickle` the figure object, such that we can regenerate the plot when the Plots pane is resized (similar to how R displaylists are handled)
- [ ] Handle high DPI displays.
- [ ] Consider whether we could write plots to `.svg` (or similar) rather than `.png`, respecting the device that the `matplotlib` plot is configured to use.
- [ ] Remind self to add `png` package as requirement for Python execution.

### DIFF
--- a/src/cpp/session/modules/SessionReticulate.R
+++ b/src/cpp/session/modules/SessionReticulate.R
@@ -79,6 +79,45 @@
    return(TRUE)
 })
 
+.rs.addFunction("reticulate.matplotlib.showHook", function(...)
+{
+   # read device size
+   size <- dev.size(units = "in")
+   width <- size[1]; height <- size[2]
+   
+   # TODO: handle high-DPI displays
+   dpi <- 72L
+   
+   # TODO: get device requested from matplotlib?
+   # TODO: handle HTML content?
+   path <- tempfile("matplotlib-plot-", fileext = ".png")
+   
+   plt <- reticulate::import("matplotlib.pyplot", convert = TRUE)
+   
+   # resize the figure
+   figure <- plt$gcf()
+   figure$set_dpi(dpi)
+   figure$set_size_inches(width, height)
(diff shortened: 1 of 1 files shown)


src/cpp/session/modules/SessionReticulate.R
```diff
@@ -79,6 +79,45 @@
...
+   plt <- reticulate::import("matplotlib.pyplot", convert = TRUE)
+   
+   # resize the figure
+   figure <- plt$gcf()
+   figure$set_dpi(dpi)
+   figure$set_size_inches(width, height)
+   plt$savefig(path, dpi = figure$dpi)
+   
+   # now, read that in and create an R plot using e.g. image
+   data <- png::readPNG(path, native = TRUE, info = TRUE)
```
jjallaire: Do we need to prompt for installation of the png package here? (or should we just make that a dependency of reticulate?)

jjallaire:
I think we can wait on pickling the objects and handling high DPI.

src/cpp/session/modules/SessionReticulate.R
jjallaire: We can add reticulate as one of the packages in the list of dependencies for reticulate within the DependencyManager (so it all comes in one shot).

kevinushey:
I'm going to add a `requireNamespace("png")` check for now (on master) and then add `png` as a required dependency in DependencyManager (planning to hook all of that up once `reticulate` hits CRAN)

jjallaire:
Okay, sounds good. I just submitted reticulate v1.6 to CRAN.

PR #2438, opened 2018-03-19
don't reindent python code on paste

jjallaire:
In at least one instance this actually breaks the code by overindenting if statements (possibly a bug in the Ace Python autoindenter?): https://raw.githubusercontent.com/datalogue/keras-attention/master/models/tdd.py

Note that there are other code oriented commands that should probably also be disabled for Python files:

screen shot 2018-03-19 at 8 41 06 am

@kevinushey


kevinushey:
We should disable most of them. I think the only one that might work out of the box is 'Comment/Uncomment Lines'. I should also make sure 'Reindent Lines' works as expected.

PR #2448, opened 2018-03-19
Fix issue with "less" and "more" handing the terminal

### LINKED ISSUE #1875: Terminal becomes terminally "busy" for multi-page man pages
If I browse a multi-page man page in an RStudio Terminal, I can neither advance through the pages, nor exit the man page. `man git rebase` will induce the behaviour. That terminal becomes terminally "busy" 🙂 and has to be closed.

gtritchie:
There was a bug in xterm.js, whereby if the terminal pane had fewer than 25 rows, and the first full-screen program run tried to do scrolling via linefeeds, it would cause JavaScript exceptions and break rendering. Only way to recover was to "q" less/more then refresh browser to get the main terminal buffer displayed again.

Made a fix to xterm.js 2.9.2 and applied to our own fork of xterm.js. You can see that change at: https://github.com/rstudio/xterm.js/commit/597a32d2fb3b5076fb21b2ef6173e3c1bb1d76bb

If and when we update to xterm.js 3.0, they have already fixed this bug, but aren't making more changes to 2.9.x.

Fixes #1875


jmcphers:
If we regularly have to take upstream drops while applying our own patches we could use a model more like `sync-ace-commits`, but I think that this is great for now. Hopefully we won't have to fix anything that isn't already taken care of upstream!

gtritchie:
They fixed this upstream in xterm 3 by rewriting the code. This issue had
been seen by others in 2.x, but never isolated as far as I can tell.
Fortunately it was relatively easy to figure out what was wrong.

Issue #2483, opened 2018-03-22
Test issues are all reported as failures


javierluraschi:
Any `testthat` warning, skip, etc, is reported as a failure which is not ideal...

<img width="471" alt="screen shot 2018-03-22 at 2 08 24 pm" src="https://user-images.githubusercontent.com/3478847/37800751-22ea507c-2de1-11e8-9225-aca29db30273.png">

dfalty:
Verified fixed 1.2.1320. 
![image](https://user-images.githubusercontent.com/15237845/53897086-fd90a600-3ffa-11e9-9d4a-7e917883ff41.png)

PR #2484, opened 2018-03-22
Use appropriate source markers for different test results

javierluraschi:
Fix for https://github.com/rstudio/rstudio/issues/2483 to support warnings and infos as well:

screen shot 2018-03-22 at 3 55 10 pm

### DIFF
--- a/src/cpp/session/modules/build/SessionBuildErrors.cpp
+++ b/src/cpp/session/modules/build/SessionBuildErrors.cpp
@@ -269,25 +269,42 @@ std::vector<module_context::SourceMarker> parseTestThatErrors(
    {
       FilePath basePathResolved = module_context::resolveAliasedPath(basePath.absolutePath());
 
-      boost::regex re("\\[1m([^:\\n]+):([0-9]+):[^:\\n]+: ([^\\n]*)\\[22m");
+      boost::regex re("\\[[0-9]+m([^:\\n]+):([0-9]+): ?([^:\\n]+): ([^\\n]*)\\[[0-9]+m");
 
       boost::sregex_iterator iter(output.begin(), output.end(), re);
       boost::sregex_iterator end;
       for (; iter != end; iter++)
       {
          boost::smatch match = *iter;
-         BOOST_ASSERT(match.size() == 4);
+         BOOST_ASSERT(match.size() == 5);
 
-         std::string file, line, type, message;
+         std::string file, line, type, message, marker, prefix;
          
          file = match[1];
          line = match[2];
-         type = "failure";
-         message = match[3];
+         type = match[3];
+
+         if (type.find("error") != std::string::npos) {
+            marker = "error";
+            prefix = "error: ";
+         } else if (type.find("failure") != std::string::npos) {
+            marker = "error";
+            prefix = "failure: ";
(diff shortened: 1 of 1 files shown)


jjallaire:
I'd say that we should only open the Issues by default if there is a warning or an error (as the navigation to issues is a cue to the user that something needs addressing)

javierluraschi:
@jjallaire looks like this is already implemented by default since we borrowed the same code for other type of warnings and errors.

javierluraschi:
Actually, the behavior today is to only open the issues if there is an error, warnings and infos remain unopened, but my guess is that this is the behavior we would want as well.

jjallaire:
Yes, I think that sounds like exactly the right behavior.

jjallaire:
A couple of things to follow up on here. In one dimension I think the new behavior may make things worse: when a test fails you often need to see the related output to diagnose it (in fact this is more valuable than navigating to the source code of the test). testthat actually does some truncation of output to ~ 12 lines (or at least it used to) and this can even obscure things (whereby you need to dig into the logs on disk to find the output). 

With this UI we are hiding the test output / error messages entirely. I don't think you can actually fix a test without seeing this output. We could try adding the output within the error item (and expanding the error item to show all of the output when you click it).

Also, highlighting info items doesn't seem that useful as I saw an info item and clicked on it -- this resulted in navigation to a source location but no actual display of the "info".

@wch and @hadley Interested to get your takes here.

hadley:
Yeah, it definitely needs to show the failing error text, and ideally it would have a call stack/tree with clickable line numbers. I'm happy to make a custom reporter for RStudio if that makes things easier.

(testthat only shows 12 lines of code when run inside R CMD check because that's what CRAN wants)

wch:
I have comments about the Output/Issues views in general. When I've been working on C/C++ code and have errors, it shows them in the Issues panel. I like that the Issues view makes it easy to see all the errors and to navigate to the corresponding code. However, sometimes the information in the Issues view isn't detailed enough, and there have been a number of times where I've puzzled over the text for a while, before finally remembering that I could click on Output to see more context. For a long time, it wasn't clear to me that (A) those were two views (full vs summarized) of the same information, and (B) that I even could click on Output at all. 

It might help a bit if Ouptut and Issues were tabs instead of buttons, but I'm not completely sure.

Another thought: what if there was an icon on the right side of each issue, where, if you clicked on it, it would switch to the Output pane and highlight the corresponding text?

jjallaire:
I think Winston's idea of the icon on the right side is a great one as this is indeed a problem with summarized views of diagnostics. We actually already have that UI for LaTeX compilation errors:

screen shot 2018-03-27 at 7 38 38 pm

hadley:
In an ideal world you'd get an expando so you could see the detail in-line, but I understand that would be more UI work (and possibly hard to generalise).

jjallaire:
I agree that the expando would be better (the selected item could show the detail). I actually don't think this would necessarily be too difficult to pull off. I'll talk to @javierluraschi about it and see what we can do.

javierluraschi:
Expando work tracked with https://github.com/rstudio/rstudio/pull/2484#issuecomment-375674412, hoping I can take a look at that for 1.2.

Issue #2502, opened 2018-03-26
No focus on opening


etienne-s:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.1.423
    OS Version      : Windows 7 x64
    R Version       : 3.4.4

### Steps to reproduce the problem

Launch RStudio.

### Describe the problem in detail

RStudio is not in focus:
The cursor is on the console but if you start typing nothing appears. Keyboard shorcuts (such as opening a new file with <kbd>Ctrl</kbd>+<kbd>O</kbd>) don't work. So you have to use the mouse to set focus on RStudio before being able to do anything.

### Describe the behavior you expected

RStudio should be in focus.

Note: this issue was already reported [here](https://support.rstudio.com/hc/en-us/community/posts/200655066-Window-Focus-on-open-R-0-97-168) 4 years ago. Also mentioned [here](https://stackoverflow.com/questions/30992433/rstudio-lost-focus-when-start-up).

rstudio-build:
This is a day 1 bug based on Qt issues (and we've never found a workaround)

Perhaps Qt in v1.2 works better?

kevinushey:
As far as I can tell, the issue here is that when RStudio is launched the window itself never actually gains focus.

dfalty:
Verified fixed on version 1.2.1090.

PR #2504, opened 2018-03-26
ensure RStudio window activated on launch

kevinushey:
Fixes https://github.com/rstudio/rstudio/issues/2502.

I have to admit I don't quite understand the mechanics around application / window launching on Windows -- hopefully someone else would be able to explain the whole ApplicationLaunch mechanism used here?

### DIFF
--- a/src/cpp/desktop/DesktopWin32ApplicationLaunch.cpp
+++ b/src/cpp/desktop/DesktopWin32ApplicationLaunch.cpp
@@ -91,7 +91,10 @@ void ApplicationLaunch::setActivationWindow(QWidget* pWindow)
 
 void ApplicationLaunch::activateWindow()
 {
-   activate((HWND)winId());
+   if (pMainWindow_)
+      activate((HWND) pMainWindow_->winId());
+   else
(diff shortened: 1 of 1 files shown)


src/cpp/desktop/DesktopWin32ApplicationLaunch.cpp
```diff
@@ -151,12 +154,7 @@ bool ApplicationLaunch::sendMessage(QString filename)
                                                        0));
       if (::IsWindow(hwnd))
       {
-         HWND hwndPopup = ::GetLastActivePopup(hwnd);
```
jjallaire: This code has to deal with activating an existing instance (e.g. when a file association is used to open a file in an existing instance of the IDE). So you should check to see whether this still works after the change (it's also possible that other aspects of the QtWebEngine migration could have broken this)

kevinushey: This does still work after the change. Note that the code here was replaced with a single call to `activate(hwnd)` as the inline code here was identical to `activate()`.

Issue #2516, opened 2018-03-27
'Save file' message box always shows on main window


kevinushey:
### System details

    RStudio Edition : Desktop
    RStudio Version : v1.2.473
    OS Version      : macOS 10.13.3
    R Version       : 3.4.4

### Steps to reproduce the problem

Within a popped-out source window, attempt to quit while an open file contains unsaved changes.

### Describe the problem in detail

The dialog informing the user of unsaved files is displayed on the main window, rather than the satellite window.

### Describe the behavior you expected

The dialog should show on the window associated with the unsaved document.

![screen shot 2018-03-27 at 10 51 00 am](https://user-images.githubusercontent.com/1976582/37985096-c4d9b354-31ac-11e8-802a-e6e6cc046e02.png)

kevinushey:
Fixed via https://github.com/rstudio/rstudio/pull/2518.

vnijs:
Seems this may be related to https://github.com/rstudio/rstudioapi/issues/91. Still seeing this issue with 1.2.502 in Windows 10

ronblum:
Verified RStudio Desktop 1.2.914 on MacOS 10.14 Mojave and Windows 10, RStudio Server 1.2.915 on Ubuntu 16.0.4.5 with Chrome 68 from MacOS.

PR #2518, opened 2018-03-27
ensure commands are fired to associated window

kevinushey:
This PR fixes a few issues:

1. Rather than having a single web channel on the main window that all windows share, each window now gets its own web channel that it can use to communicate with the Qt backend. This ensures that attempts to show dialogs (e.g. message boxes) can show with the window from which they were requested, thereby fixing https://github.com/rstudio/rstudio/issues/2516.

2. On macOS, when a command is invoked from the main menu, it is always invoked from the main window. However, we typically want to target the behavior to the last active (or currently active) window; e.g. for source editor commands that should operate on the current window. We hence also allow the main menu to request that satellites execute a command.

3. We also need to more aggressively manage commands on macOS, since an attempt to invoke a command from the main menu may actually be targeted at an active satellite window, and so the commands need to synchronize based on what is open in the popped-out source window.

### DIFF
--- a/src/cpp/desktop/DesktopMainWindow.cpp
+++ b/src/cpp/desktop/DesktopMainWindow.cpp
@@ -192,9 +192,14 @@ void MainWindow::invokeCommand(QString commandId)
 {
 #ifdef Q_OS_MAC
    QString fmt = QStringLiteral(R"EOF(
-      var wnd = window.$RStudio.last_focused_window || window;
-      wnd.desktopHooks.invokeCommand('%1')
-   )EOF");
+var wnd;
+try {
+   wnd = window.$RStudio.last_focused_window;
+} catch (e) {
+   wnd = window;
+}
+wnd.desktopHooks.invokeCommand('%1');
+)EOF");
 #else
    QString fmt = QStringLiteral("window.desktopHooks.invokeCommand('%1')");
 #endif
--- a/src/cpp/desktop/DesktopMainWindow.cpp
+++ b/src/cpp/desktop/DesktopMainWindow.cpp
@@ -17,8 +17,6 @@
 
 #include <QToolBar>
 #include <QWebChannel>
-#include <QWebEngineScript>
-#include <QWebEngineScriptCollection>
 
 #include <boost/bind.hpp>
 #include <boost/format.hpp>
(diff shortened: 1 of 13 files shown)


src/gwt/src/org/rstudio/core/client/command/AppCommand.java
```diff
@@ -186,6 +184,9 @@ public void setEnabled(boolean enabled)
          enabled_ = enabled;
          handlers_.fireEvent(new EnabledChangedEvent(this));
       }
+    
+      if (Desktop.isDesktop())
```
kevinushey: We might want to make this macOS-only (?)

src/gwt/src/org/rstudio/core/client/command/AppCommand.java
```diff
@@ -200,6 +201,9 @@ public void setVisible(boolean visible)
          visible_ = visible;
          handlers_.fireEvent(new VisibleChangedEvent(this));
       }
+      
+      if (Desktop.isDesktop())
```
kevinushey: Similar here?

src/gwt/src/org/rstudio/core/client/command/AppCommand.java
gtritchie: Why Mac-only?

kevinushey: Mainly just because this is the behaviour we've had on other platforms, and it's worked as expected.

(And yet, I'm somewhat surprised that it works as expected without these calls...)

src/cpp/desktop/DesktopMainWindow.cpp
```diff
@@ -230,11 +190,16 @@ void MainWindow::quit()
...
 void MainWindow::invokeCommand(QString commandId)
 {
-   QString command =
-         QString::fromUtf8("window.desktopHooks.invokeCommand('") +
-         commandId +
-         QString::fromUtf8("');");
-
+#ifdef Q_OS_MAC
+   QString fmt = QStringLiteral(R"EOF(
+      var wnd = window.$RStudio.last_focused_window || window;
```
jmcphers: Is it possible for `window.$RStudio` to be null?

src/gwt/src/org/rstudio/core/client/command/AppCommand.java
jmcphers: The reason it worked as expected earlier on macOS is that we had a hook that ran prior to the menu showing which did a just-in-time sync of the command state. 

https://github.com/rstudio/rstudio/blob/bf986733d06237283adb8c67223123a8b3b058cf/src/cpp/desktop-mac/MainFrameMenu.mm#L218-L244

src/gwt/src/org/rstudio/core/client/command/AppCommand.java
jmcphers: The `AppCommand` is supposed to notify the desktop when its state changes. This happens here:

https://github.com/rstudio/rstudio/blob/master/src/gwt/src/org/rstudio/core/client/command/AppCommand.java#L110-L116

src/gwt/src/org/rstudio/core/client/command/AppCommand.java
jmcphers: In 1.2 we changed this; because callbacks into JS are asynchronous but the menu is shown synchronously, we can't delay showing it until we collect the command state from the client (pull model). Instead we send the command state from the client to the desktop frame as it changes (push model).

src/cpp/desktop/DesktopMainWindow.cpp
kevinushey: It shouldn't (this is always set on the main window) but I added a fallback just in case it is missing for some reason.

src/gwt/src/org/rstudio/core/client/command/AppCommand.java
kevinushey: Gotcha. I think the behavior I was seeing was related to a satellite window thinking it wasn't a desktop window, which should be alleviated through separate changes in this PR (but I'll test to verify).

src/gwt/src/org/rstudio/core/client/command/AppCommand.java
```diff
@@ -107,12 +107,10 @@ public void onVisibleChanged(AppCommand command)
    
    public AppCommand()
    {
-      if (Desktop.isDesktopReady())
+      if (Desktop.isDesktop())
```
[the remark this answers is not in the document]
jmcphers: Originally guarded these with `isDesktopReady` because DesktopMenuCallback is not available on boot and I was getting script errors when we tried to invoke it. As long as you aren't seeing those anymore this is fine! (And if you are I think it suggests we should just move the ready check inside the callback)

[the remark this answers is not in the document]
kevinushey: I'm not seeing these anymore, so I think we should be good to go!

PR #2520, opened 2018-03-28
ensure minimal windows closed when parent window closes

### LINKED ISSUE #2137: Attempts to quit while secondary window is open (e.g. zoomed plot) fail
To reproduce:

1. Produce a plot with e.g. `plot(1)`,
2. Zoom the plot generated in the plots pane,
3. Attempt to quit the IDE with `Cmd + Q`.

The main window is closed, but the plot window persists and the application does not shut down.

kevinushey:
Closes #2137.

With Qt Webkit, we were able to rely on the `onCloseRequested()` slot for web views (or so it seems? I can't actually find that documented anywhere, now...).

Fortunately, we can still signal the request for the window to close by just hooking up the parent window's `destroyed()` signal to a close call.


gtritchie:
LGTM!

PR #2528, opened 2018-03-28
Fix Help pane popout button on desktop

### LINKED ISSUE #2117: attempts to pop out Help pane fail
### System details

    RStudio Edition : Desktop
    RStudio Version : v1.2.315
    OS Version      : macOS 10.13.2
    R Version       : 3.4.3

Attempt to pop out the help pane, using the 'Show in new window' button.

Nothing happens.

The window should pop out.

gtritchie:
This makes the help pane's "Show in new window" button work like it did in 1.1 and open in a minimal Qt browser window. Each time you click it a new instance is created, hence the name+counter; otherwise it would just reactivate the same window on subsequent clicks.

Fixes #2117

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/help/HelpPane.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/help/HelpPane.java
@@ -1,7 +1,7 @@
 /*
  * HelpPane.java
  *
- * Copyright (C) 2009-12 by RStudio, Inc.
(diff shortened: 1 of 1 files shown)


kevinushey:
LGTM, but does this imply that `globalDisplay_.openWindow()` still doesn't do what we expect?

gtritchie:
Possibly but I couldn’t find a desktop case where it mattered, but I will poke around some more.

PR #2573, opened 2018-04-03
Support for htmlwidgets in notebooks with custom knitr engines

javierluraschi:
For custom `knitr` engines that render `htmlwidgets`, notebooks currently don't render them inline nor in the notebook preview. Additionally, this makes the IDE aware of a `D3` chunk type to be able to render `D3` inline while using packages that add `D3` chunks as custom engines:

screen shot 2018-03-30 at 5 58 34 pm

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTargetChunks.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTargetChunks.java
@@ -271,7 +271,7 @@ public class TextEditingTargetChunks
    
    // runnable engines within the R Notebook mode
    private static final String RE_RUNNABLE_ENGINES =
-         "r|rscript|rcpp|python|ruby|perl|bash|sh|stan|sql|";
+         "r|rscript|rcpp|python|ruby|perl|bash|sh|stan|sql|d3|";
    
    // renderPass_ need only be unique from one pass through the scope tree to
    // the next; we wrap it at 255 to avoid the possibility of overflow
--- a/src/cpp/session/modules/NotebookHtmlWidgets.R
+++ b/src/cpp/session/modules/NotebookHtmlWidgets.R
@@ -281,6 +281,8 @@
                                             libraryFolder,
                                             chunkOptions)
 {
+   message(">> initHtmlCapture")
+   
    # cache context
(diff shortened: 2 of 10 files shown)


src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTargetChunks.java
```diff
@@ -271,7 +271,7 @@ private boolean isExecutableKnitrEngine(String engine)
    
    // runnable engines within the R Notebook mode
    private static final String RE_RUNNABLE_ENGINES =
-         "r|rscript|rcpp|python|ruby|perl|bash|sh|stan|sql|";
+         "r|rscript|rcpp|python|ruby|perl|bash|sh|stan|sql|d3|";
```
jjallaire: I wonder what additional measures we should take to ensure the r2d3 package is available here. A few options:

1) Add r2d3 to the list of packages we automagically install alongside rmarkdown (we can do this using an embedded package until we have r2d3 on CRAN)

2) Prompt to install r2d3 whenever a user attempts to run a d3 chunk within a notebook

Or I guess we could leave things as-is and just have knit fail for the user if they don't have r2d3 installed (I'm presuming that in this case interactive execution from a notebook would still work).

kevinushey: FWIW, we take approach (2) with `reticulate` -- any time you take some IDE action that requires `reticulate`, we prompt for its installation.

As an aside, we now (in theory) support _all_ engines registered to `knitr`, so we might consider synchronizing that list of engines between the server + client sides. This would allow users to define custom engines that automagically work in the IDE, beyond just `d3`.

Perhaps not for this PR but something worth thinking about?

kevinushey:
It looks like the height of the chunk output in your screenshot isn't clamped to the size of the `d3` widget. Is that something that could be easily handled?

jjallaire:
The height of the widget should be known so clamping should in fact be possible.

javierluraschi:
Going to merge this one, but will address the feedback in two follow ups:
- For package installation checks, I can merge https://github.com/rstudio/rstudio/commits/bugfix/d3-cran-install later which adds the CRAN dependency, whenever this gets to CRAN.
- Resizing might need a follow up but work is missing in the package.

PR #2575, opened 2018-04-03
remove unused desktop files

kevinushey:
These files are no longer used on Desktop, and can be safely removed. I believe these classes were all added for handling file downloads and the like, but this functionality is all handled natively by QWebEngine now, as per https://doc.qt.io/qt-5/qwebenginepage.html#download.


gtritchie:
I love deleting source code! LGTM.

kevinushey:
Me too :-)

PR #2576, opened 2018-04-03
implement download helper for webengine downloads

kevinushey:
This PR brings back a download helper for QtWebEngine downloads. Currently it's mostly just a stub; the intention is that this could be used for reporting download progress, download failures, etc.

Not sure we want to implement our own 'download manager' in RStudio but at least this leaves the door open if we do?


gtritchie:
Does this change anything from and end-user perspective, or just plumbing for potential future use?

kevinushey:
Only plumbing -- the intention is that this would be the place we provide UI for download status (e.g. download succeeded vs. download failed, and so on)

gtritchie:
Ok, might as well do the plumbing while QtWebEngine is fresh in your mind. LTGM.

PR #2578, opened 2018-04-03
match background color to theme (closes #2572)

### LINKED ISSUE #2572: set default background color for window
As a stop-gap fix / alleviation to https://github.com/rstudio/rstudio/issues/2150.

kevinushey:
This PR alleviates the issue described in https://github.com/rstudio/rstudio/issues/2150, so that the default background color is set during resize.

resize


gtritchie:
LGTM!

PR #2609, opened 2018-04-06
Bugfix/websocket fallback rpc

### LINKED ISSUE #2207: Terminal failing to automatically fall back to RPC when websockets fail
### System details

    RStudio Edition : Server
    RStudio Version : 1.1.x
    OS Version      : Linux
    R Version       : Don't know

With a proxy server that doesn't allow websockets, try to open the RStudio terminal.

Terminal is blank, and stays busy per the dropdown menu.

When websockets fail to connect should be automatically falling back on RPC. The customer in this case (a Pro customer, actually, but issue is very likely not Pro-specific) ended up having to manually disable terminal websockets.

Don't have a clear repro, other than customer said their proxy did not allow websockets, and they saw the terminal issue described above. Turning off websockets fixed it, but should be happening automatically (ideally).

Not a huge issue, perhaps, but definitely adds a support burden and we'd like to understand what's happening here to see if it can be fixed.

gtritchie:
Fixes #2207 

## Before
In some environments the attempt to connect terminal via websockets can take a long time to fail (have seen it take over two minutes). During that time there is nothing shown in the terminal UI, so user has no idea what's going on.

The code is then supposed to fall back to HTTP-based RPC for the terminal if WebSockets fails, but that also had a problem that usually resulted in no terminal (the one being loaded via RPC was instantly killed).

## After
Now there is a 5 second timeout; if websocket connection doesn't happen in 5 seconds, the terminal will switch to HTTP (fixed the bug that was preventing that), then show a message telling the user that WebSockets failed, and it switched to HTTP, and terminal performance will be reduced. Mentions the Tools / Global Options / Terminal option for turning off WebSockets to avoid the delay and the message.

Also show a message now if the fallback to RPC also fails, instead of just having a blank terminal.


jjallaire:
I'm not sure we should show a message to the user when we fall back as it's extremely unlikely that the user can actually do anything to remedy the situation (nor even grasp the difference between websockets and HTTP). It's not so much an error condition as just a fixed constraint of the environment. I think this message will probably mostly lead to lots of extra support tickets.

jjallaire:
That's a good question. On RStudio Server it seems like the failure is more likely to be transient (i.e. only for a given remote access scenario) whereas on the desktop it seems more likely to be permanent. So maybe distinct behavior depending on which version they are running?

Also, do you think you could narrow it to 3 seconds or would that risk false negatives?

PR #2629, opened 2018-04-11
Add support to insert d3 chunks

javierluraschi:
Adds support to insert d3 chunks, see also https://github.com/rstudio/r2d3/issues/6


src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTarget.java
```diff
@@ -4514,6 +4514,12 @@ public void onError(ServerError error)
       });
    }
 
+   @Handler
+   void onInsertChunkD3()
+   {
+      onInsertChunk("```{d3}\n\n```\n", 1, 0);
```
jjallaire: Do we want to include `data=` (since it's required) similar to how we include `connection=` for SQL? It would also be nice if the cursor was placed at the end of `data=` (as we do for `connection=`)

PR #2650, opened 2018-04-16
Add colors to getThemeInfo api to support better visualizations

javierluraschi:
While working with `D3`, would be nice to support rendering visualizations based on editor themes by using a header similar to:

```
// !preview r2d3 data=read.csv(system.file("samples/bubbles/flare.csv", package = "r2d3")), d3_version=4, options=rstudioapi::getThemeInfo(), sizing=htmlwidgets::sizingPolicy(padding=0)
```

screen shot 2018-04-16 at 11 38 54 am

### DIFF
--- a/src/cpp/r/R/Api.R
+++ b/src/cpp/r/R/Api.R
@@ -699,10 +699,47 @@ options(terminal.manager = list(terminalActivate = .rs.api.terminalActivate,
      editor,
      ignore.case = TRUE)
 
-   list(
-      editor = editor,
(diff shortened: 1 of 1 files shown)


jjallaire:
Note that once this is available I'd suggest we just automatically inject a `theme` variable into the D3 rendering context (rather than requiring an explicit forwarding of it). This will allow us to correctly provide theme colors in other contexts (e.g. a themed Shiny app).

PR #2658, opened 2018-04-18
Support signaling viewer pane for openfile message

javierluraschi:
Follow up to https://github.com/rstudio/rstudio/pull/2653 to:
1) Support signaling viewers that the `openfile` message is available.
2) As part of signaling, give the origin domain to allow cross-site message from a different domain/port hosting the widget.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/viewer/ViewerPane.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/viewer/ViewerPane.java
@@ -12,6 +12,9 @@
  */
 package org.rstudio.studio.client.workbench.views.viewer;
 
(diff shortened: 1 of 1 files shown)


jjallaire:
Note that we already have one other channel for signaling capabilities to JavaScript within the viewer pane: we pass a "viewer_pane" query parameter here: https://github.com/rstudio/rstudio/blob/77dac3f97feb8ec172b5c141bf81aa495c72c1e7/src/gwt/src/org/rstudio/studio/client/workbench/views/viewer/ViewerPane.java#L274-L277

So an alternative that wouldn't introduce a new protocol / code path would be to just add a "capabilities" query parameter there which has a bitmask of capabilities.

javierluraschi:
Query strings works for me, changed.

PR #2725, opened 2018-05-01
Support multiple statements in SQL preview

javierluraschi:
Dependent on https://github.com/rstudio/rstudio/pull/2720 to consider this independently.

DBI does not support executing multiple statements, see https://github.com/r-dbi/RPostgres/issues/138, but we can consider splitting them ourselves to enable more useful SQL script and modify also the default script that is created:

screen shot 2018-05-01 at 10 57 28 am

### DIFF
--- a/src/cpp/session/modules/SessionDataPreview.R
+++ b/src/cpp/session/modules/SessionDataPreview.R
@@ -45,12 +45,25 @@
    )
 
    .rs.enqueClientEvent("data_output_completed", preview)
+
+   invisible(NULL)
 })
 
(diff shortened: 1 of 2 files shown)


javierluraschi:
However, if SQL preview is mostly going to be used to create query statements since we can't do much with multi-statement SQL files in R, to my knowledge. One would have to manually split the file an execute it manually, etc. Unless we add support for SQL files in DBI.

javierluraschi:
This PR would also have to handle multiple parameters per statement, etc. Closing.

PR #2754, opened 2018-05-08
Add support for package managers

javierluraschi:
Add support to support package managers...

First, this PR splits the packages options tab into "package management" and "package development":

screen shot 2018-05-07 at 9 54 53 pm
screen shot 2018-05-08 at 5 38 31 pm

Changing the primary repo was modified to take an arbitrary URL:

screen shot 2018-05-07 at 9 56 16 pm
screen shot 2018-05-08 at 5 40 00 pm

A new secondary repos widget was created to add secondary URLs specified in `r-cran-repos-url` and expecting `text/ini` file that describes name/url repo pairs which can be used as follows:

screen shot 2018-05-07 at 11 27 01 pm
screen shot 2018-05-08 at 5 41 25 pm

A new `repos.conf` file is now read from `/etc/rstudio/` which can specify an `ini` file with a `CRAN` entry and varios secondary repos that administrators can configure instead of having to rely on users.

Finally, this PR adds a browse-to link to the packages source, most likely, this will open up the CRAN page for the package, but might also be overriden by other repos:

screen shot 2018-05-08 at 1 38 48 pm

screen shot 2018-05-07 at 9 55 27 pm


jjallaire:
Assuming that secondary repos can also use arbitrary URLs?

If there no registered secondary repos can we hide that list in the Add dialog?

I think that the Primary repo should be called "Primary CRAN Repo" since it's role is to provide a view into CRAN (and it ends up in the name list entry of "CRAN" within the `repos` option.

javierluraschi:
Yes, secondary repos can be arbitrary URLs, but they MUST be named.

Makes sense, we can hide the secondary lists unless an option is set or the primary CRAN detected as a package manager capable of listing secondary repos.

K, will rename to "Primary CRAN Repo"

jjallaire:
I'm not saying hide the UI, I'm saying hide the list if it's empty. We should always be able to specify arbitrary URLs for secondary repos!

javierluraschi:
Gotcha! So how about this for the empty use case...

screen shot 2018-05-08 at 9 02 51 am

src/cpp/r/R/Tools.R
```diff
@@ -499,13 +499,37 @@ assign(envir = .rs.Env, ".rs.hasVar", function(name)
    .rs.registerReplaceHook(name, package, unsupported)
 })
 
-.rs.addFunction( "setCRANRepos", function(reposUrl)
+.rs.addFunction( "isCRANReposList", function(repos) {
+  startsWith(repos, "CRAN|")
```
jmcphers: `startsWith` is not available until R 3.3 (if memory serves we have a polyfill)

src/cpp/session/modules/SessionPackages.R
```diff
@@ -1159,3 +1171,95 @@ if (identical(as.character(Sys.info()["sysname"]), "Darwin") &&
    # delegate to 'setInternet2'
    utils::setInternet2(value)
 })
+
+.rs.addFunction("parseSecondaryReposIni", function(conf) {
+   lines <- readLines(conf)
+   repos <- list()
+
+   for (line in lines) {
+     part <- strsplit(line, "=")[[1]]
```
jmcphers: I think we should support comments here; by convention we often include a sample config file that contains comments with a brief description of the file's purpose and some usage example.

src/cpp/session/modules/SessionPackages.R
```diff
@@ -1159,3 +1171,95 @@ if (identical(as.character(Sys.info()["sysname"]), "Darwin") &&
...
+   
+   rCranReposUrl <- .Call("rs_getCranReposUrl")
+   if (nchar(rCranReposUrl) > 0) {
+      conf <- tempfile(fileext = ".conf")
+      
+      tryCatch({
+         download.file(rCranReposUrl, conf, extra = "-H 'Accept: text/javascript'")
+         parserType <- "json"
+      }, catch = function(e) {
+         ""
```
jmcphers: Do you think we should log here if the error isn't caused by a mime type mismatch? (wondering if swallowing exceptions here might make it hart to diagnose network issues, cert issues, etc. that are preventing the URL from working)

src/cpp/session/modules/SessionPackages.R
javierluraschi: Yeah, we can use the boost parser I introduced for parsing `repos.conf` in the C++ codebase, that one does support comments.

src/cpp/session/modules/SessionPackages.R
javierluraschi: This code needs to be removed since RSPM just added support for `text/ini` files, needs to do some testing today and return the error.

src/cpp/r/R/Tools.R
javierluraschi: Thanks!

jjallaire:
It should still show a list of secondary repos as there could be more than one.

What I was saying is that if the administrator hasn't primed a list of pre-fab repos then in the add secondary repo dialog we shouldn't show that list (but rather just name and url).

javierluraschi:
I see, hopefully I got the suggestion right now. When no secondary repos are specified, dialog renders as:

screen shot 2018-05-08 at 6 31 29 pm

PR #2827, opened 2018-05-22
Use r2d3 from CRAN

javierluraschi:
Remove r2d3 as embedded package, use CRAN package instead. Need to wait for r2d3 to be available in CRAN:

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/common/dependencies/DependencyManager.java
+++ b/src/gwt/src/org/rstudio/studio/client/common/dependencies/DependencyManager.java
@@ -135,7 +135,7 @@ public class DependencyManager implements InstallShinyEvent.Handler,
          new Dependency[] {
             Dependency.cranPackage("htmltools", "0.3.6"),
             Dependency.cranPackage("htmlwidgets", "1.2", true),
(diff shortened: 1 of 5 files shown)


src/gwt/src/org/rstudio/studio/client/common/dependencies/DependencyManager.java
```diff
@@ -135,7 +135,7 @@ public void withR2D3(String userAction, final Command command)
          new Dependency[] {
             Dependency.cranPackage("htmltools", "0.3.6"),
             Dependency.cranPackage("htmlwidgets", "1.2", true),
-            Dependency.embeddedPackage("r2d3")
+            Dependency.cranPackage("r2d3", "0.2.0")
```
jjallaire: You should add the `true` parameter (as with htmlwidgets) so that we install the source package (compatible with more versions of R and eliminates the problem with lagging binaries on CRAN)

javierluraschi: Already added by JJ. I was not aware that `true` was needed to install from source, good to know!

PR #2829, opened 2018-05-22
Fix some issues with running terminal on 64-bit rsession (Windows)

### LINKED ISSUE #2817: terminal pane fails to launch any terminals on Windows
### System details

    RStudio Edition : Desktop
    RStudio Version : v1.2.650
    R Version       : R 3.5.0

Attempt to open a terminal.

No terminal is opened.

A terminal should be opened.

Unfortunately, the Terminal Diagnostics window doesn't contain any useful information; just:

```
Global Terminal Information
---------------------------
Loaded TerminalSessions: 0

Terminal List Count: 0
```

gtritchie:
Fixes #2817

This has been somewhat non-deterministic. I'm seeing cases where Win32 `executablePath()` returns an empty string, leading to mistakes in locating the winpty.dll (and presumably other problems). Also problems running terminal with 64-bit rsession build in dev environment; hadn't tried that before, always used 32-bit.

Found some mention online that using `_pgmptr` for getting at full path of current executable can have issues if used early in startup, but it is safe to use `GetModuleFileName` early. So switched to that approach.

Added an error log if there is a failure to find winpty.dll.

Made the decision on using 32-bit or 64-bit winpty.dll a compile-time thing instead of runtime, since it's already known by virtue of which rsession.exe is being built.

### DIFF
--- a/src/cpp/core/system/Win32Pty.cpp
+++ b/src/cpp/core/system/Win32Pty.cpp
@@ -111,7 +111,10 @@ Error tryLoad(const core::FilePath& libraryPath)
             libraryPath.absolutePath(),
             (void**)&hMod);
    if (error)
+   {
+      LOG_ERROR(error);
       return error;
+   }
 
    LOAD_WINPTY_SYMBOL(error_code);
    LOAD_WINPTY_SYMBOL(error_msg);
--- a/src/cpp/core/system/Win32System.cpp
+++ b/src/cpp/core/system/Win32System.cpp
@@ -672,10 +672,17 @@ PidType currentProcessId()
    return ::GetCurrentProcessId();
 }
 
-Error executablePath(const char * argv0,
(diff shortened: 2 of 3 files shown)


gtritchie:
What version of RStudio are you using?

RTMaloney:
Hi, 
I've posted more info at the RStudio community forum [here](https://community.rstudio.com/t/blank-unresponsive-terminal-pane-no-cursor/49601). 

At first I tried installing the latest **Version 1.2.5033** with the same result.

Elsewhere (#[2849](https://github.com/rstudio/rstudio/issues/2849)) a fix is suggested to install the Daily Build version.
That is **1.3.745**. 
I did so, but with a slightly different but more or less same outcome.

gtritchie:
I responded in the community forum; let's continue the troubleshooting over there.

Issue #2864, opened 2018-05-25
RSPM: Custom repo configurations do not take effect.


fereshtehRS:
### System details

    RStudio Edition : Server
    RStudio Version : 1.2.665
    OS Version      : CentOS 7
    R Version       : 3.4.2

### Steps to reproduce the problem
- Access Tools -> Global Options -> Packages and make sure `Primary CRAN repo` is set to default `Global (CDN) - RStudio`
- Edit `/etc/rstudio/rsession.conf` configuration file and add the following:

`r-cran-repos=https://<your-package-manager-server-address>/<repo-name>/latest`  (pointing to a valid RSPM host and repo)
- Restart RStudio Server
- Optional: Restart R session in the browser
- Access Tools -> Global Options -> Packages. You still see the default setting.

### Describe the problem in detail
The settings in the configuration file `rsession` or the optional file `repos.conf` does not take effect.

### Describe the behavior you expected

The custom configuration setting should have taken effect and reflected in the UI dialog.

javierluraschi:
@fereshtehRS mentioned this wasn't working for CentOS and Suse, just tried Suse 12 over EC2 following: [install-rstudio-server-suse-rmd](https://gist.github.com/javierluraschi/a52a89553e821683f95416907e415a5f#file-install-rstudio-server-suse-rmd).

and setting `/etc/rstudio/rsession.conf`:

```
r-cran-repos=http://<name>.<apackagemanager>.com/bleedingEdgeRepo/latest
```

got this working properly after changing `rsession.conf` and restarting the session:

<img width="1440" alt="screen shot 2018-05-29 at 2 06 27 pm" src="https://user-images.githubusercontent.com/3478847/40685589-88417aca-6349-11e8-8706-95531d6c9577.png">

javierluraschi:
I was able to reproduce this as well in RStudio 1.1, I know what is going on, but a fix is still needed.

The behavior of `r-cran-repos` in RStudio 1.1 is to override this setting with user preferences, so, in 1.1 we can do:

```
# rsession.conf
r-cran-repos=foo
```

Then in RStudio Server:
- Restart r session
- Run `getOption("repos")`, notice the correct value.
- Tools -> Global Options - >OK
- Run `getOption("repos")`, notice the incorrect value.

Another repro:

- Close the Terminal, make sure no tab is visible
- Restart r session
- Run `getOption("repos")`, notice the correct value.
- Create a new Terminal
[...]

PR #2891, opened 2018-05-30
Make r-cran-repos and user setting consistent

javierluraschi:
This is older issue that reproes in RStudio Server 1.1 see #2864 and https://github.com/rstudio/rstudio-pro/issues/434

When the `r-cran-repos` option is set, `SessionUserSettings.cpp` still defaults to RStudio's CRAN repo, this is harmless in many cases. However,  as shown in #2864 and https://github.com/rstudio/rstudio-pro/issues/434, whenever the user preferences get refreshed, say by OKing the Global Preferences dialog, opening a new terminal or changing R versions, etc, we change the CRAN repo to the user settings value. This yields inconsistent behavior since some r sessions might start with some CRAN repo and some with other depending on when the terminal gets created or which operation is triggered in the r session.

The fix in `SessionUserSettings.cpp` is to prevent setting the user settings default if `r-cran-repos` is set. This should fix most problems in clean systems.

However, for existing systems, most users will have their user setting set to RStudio's CRAN, and therefore, this problem of inconsistent CRAN repos will persist. The fix in `SessionMain.cpp` gives precedence to the user setting to, at least, avoid random changes in the CRAN repo. It appears to be the case that we intended to give precedence to global settings with [d348bda](https://github.com/rstudio/rstudio/commit/d348bdaa8b2d1c35fe5e24aa2c1924ba708422fd) but it's less clear why.

Also worth mentioning that from the admin guide [CRAN Repositories](http://docs.rstudio.com/ide/server-pro/r-sessions.html#cran-repositories):

> RStudio Server uses the RStudio CRAN mirror (https://cran.rstudio.com) by default. This mirror is globally distributed using Amazon S3 storage so should provide good performance for all locales. You may however wish to override the default CRAN mirror. This can be done with the r-cran-repos settings

Which makes me thing than a full override of `r-cran-repos` was not intended, but rather, the change proposed in this change (to only change RStudio's default provided CRAN mirror) seems correct since it still allows users able to manually override this default.

This PR could use some code review from @jmcphers and @jjallaire.

### DIFF
--- a/src/cpp/session/SessionMain.cpp
+++ b/src/cpp/session/SessionMain.cpp
@@ -1820,13 +1820,13 @@ int main (int argc, char * const argv[])
       rOptions.rSourcePath = options.coreRSourcePath();
       if (!desktopMode) // ignore r-libs-user in desktop mode
          rOptions.rLibsUser = options.rLibsUser();
-      // CRAN repos: repos file trumps global server option which trumps user setting
+      // CRAN repos: user setting then repos file then global server option
+      if (userSettings().cranMirror().url)
+         rOptions.rCRANRepos = userSettings().cranMirror().url;
       if (!options.rCRANMultipleRepos().empty())
          rOptions.rCRANRepos = options.rCRANMultipleRepos();
-      else if (!options.rCRANRepos().empty())
+      else (!options.rCRANRepos().empty())
          rOptions.rCRANRepos = options.rCRANRepos();
-      else
-         rOptions.rCRANRepos = userSettings().cranMirror().url;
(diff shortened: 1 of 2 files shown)


javierluraschi:
Made one more change to help mitigate previous installations where users already have a user setting overriding CRAN unintentionally: Setting `allow-r-cran-repos-edit=0` will now also override the user setting, which is consistent with the behavior in the [admin guide/cran-repositories](http://docs.rstudio.com/ide/server-pro/r-sessions.html#cran-repositories):

> Whatever the default CRAN mirror is, individual users are still able to set their own default. To discourage this, you can set the allow-r-cran-repos-edit settings.

PR #2892, opened 2018-05-31
update libclang

kevinushey:
Note: not quite ready for merge.

---

This PR updates our `libclang` integration in a couple ways:

1. We no longer bundle `libclang` on macOS and Linux; instead, we rely on the user to install a copy of `libclang` and automatically attempt to find it.

2. On Windows, we update our embedded copy of `libclang` to 5.0.2, and also allow for the usage of an official installed version of `libclang`.


kevinushey:
@jjallaire Sorry for the silly question -- For Fedora + Ubuntu, do you mean we should bundle `libclang` explicitly, or have our RStudio package depend on the default `libclang` version available in the package repository (so it gets requested / installed alongside RStudio)?

jjallaire:
I mean have RStudio's deb/rpm package depend on the libclang package available in the public repo.

kevinushey:
It looks like CentOS 6 doesn't include `clang` (or `libclang`) in the default repositories, but they are available from EPEL. However, I think R requires EPEL as well so it's probably okay to depend on `clang` from there?

CentOS 7, however, does include `clang` as part of the default repositories (although just as a single bundle; one cannot download `libclang` standalone)

On Debian, I think we can just depend on `libclang-dev` and get the version of `clang` vetted for the particular OS.

jjallaire:
We actually make a point of not depending on EPEL as many customers have an objection to configuring their servers to talk to EPEL (those customers often either build R from source or use another distribution like MSFT or Oracle R). So I'd say let's skip it entirely on RStudio Server for RPM (no worse than we are today) and only add the extra dependency for debian-based systems (`libclang-dev` as you said).

kevinushey:
Sounds good! I think this PR is ready for review, then.

For reference, on Windows, we're bundling a copy of `libclang.dll` from the official LLVM builds from https://releases.llvm.org/download.html (alongside `libc++` headers and some required Windows runtime libraries). Looks like things work just fine!

src/cpp/session/CMakeLists.txt
```diff
@@ -41,21 +41,19 @@ if(NOT EXISTS "${RSTUDIO_DEPENDENCIES_DIR}/common/plumber")
...
-if(NOT EXISTS "${LIBCLANG_HEADERS_DIR}")
-  message(FATAL_ERROR "libclang builtin-headers not found  (re-run install-dependencies script to install)")
-endif()
+   set(LIBCLANG_VERSION "5.0.2")
+   set(LIBCLANG_DIR "${RSTUDIO_DEPENDENCIES_DIR}/common/libclang/${LIBCLANG_VERSION}")
+   if(NOT EXISTS "${LIBCLANG_DIR}")
+      message(FATAL_ERROR "libclang ${LIBCLANG_VERSION} not found  (re-run install-dependencies script to install)")
+   endif()
 
+   set(LIBCLANG_HEADERS_DIR "${RSTUDIO_DEPENDENCIES_DIR}/common/libclang/builtin-headers")
```
jjallaire: It looked to me like we are no longer downloading a `builtin-headers` directory, or are we?

jjallaire:
Other than my 2 comments, LGTM. Glad that this worked out, it's a a very material upgrade in capability/compatibility!

src/cpp/CMakeLists.txt
```diff
@@ -102,6 +102,7 @@ if(UNIX)
       add_definitions(-fstack-protector --param ssp-buffer-size=4)
       add_definitions(-fPIC)
       if(CMAKE_BUILD_TYPE STREQUAL Debug)
+         add_definitions(-O1)
```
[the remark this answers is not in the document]
kevinushey: This is intentional -- `gcc` complains if you define `_FORTIFY_SOURCE` without specifying an optimization level. I think compiling with `-O1` should ensure the generated libraries / executables are still easily debuggable for debug builds.

src/cpp/session/CMakeLists.txt
kevinushey: We aren't doing this anymore; instead, I just packaged everything together into a single zipfile. This includes 32bit and 64bit `libclang.dll` (including support libraries), and also the `libc++` sources.

jjallaire: So this cmake code is still valid but points to the same core zip file, correct?

kevinushey: Right -- effectively, unpacking `libclang-windows-5.0.2.zip` will ensure both the `libclang/5.0.2` and `libclang/builtin-headers/5.0.2` folders are both populated as they're now part of the same archive.

PR #2893, opened 2018-05-31
WIP on upgrading to Qt 5.11

gtritchie:
NOT READY TO MERGE

Updating to Qt 5.11 requires we switch to MSVC 2017, and build only 64-bit on Windows. It also requires we use a newer CMake on all platforms (minimum of 3.1.0), which means building it on some Linux platforms whose repository builds of cmake are too old.

Still quite a bit of work to do here. Known issues as of this checkpoint:

- can't link Linux desktop, get a bunch of undefined dbus references (partial list here):
```
Linking CXX executable rstudio                                                                                                                                                                                                                                                             
/opt/RStudio-QtSDK/Qt5.11.0/5.11.0/gcc_64/lib/libQt5DBus.so.5.11.0: undefined reference to `dbus_message_copy@LIBDBUS_1_3'                                                                                                                                                                 
/opt/RStudio-QtSDK/Qt5.11.0/5.11.0/gcc_64/lib/libQt5DBus.so.5.11.0: undefined reference to `dbus_bus_register@LIBDBUS_1_3'                                                                                                                                                                 
/opt/RStudio-QtSDK/Qt5.11.0/5.11.0/gcc_64/lib/libQt5WebEngineCore.so.5.11.0: undefined reference to `dbus_watch_get_data@LIBDBUS_1_3'                                                                                                                                                      
/opt/RStudio-QtSDK/Qt5.11.0/5.11.0/gcc_64/lib/libQt5DBus.so.5.11.0: undefined reference to `dbus_server_set_timeout_functions@LIBDBUS_1_3'                                                                                                                                                 
/opt/RStudio-QtSDK/Qt5.11.0/5.11.0/gcc_64/lib/libQt5DBus.so.5.11.0: undefined reference to `dbus_connection_close@LIBDBUS_1_3' 
```
- get warnings about unknown compiler version from boost when building on Windows (seems to be harmless, but very noisy)
- the pure 64-bit Windows desktop build still needs some work (could remove the 32-bit command prompt and PowerShell options from terminal, remove the 32-bit R option from the selection dialog); plus, at the moment, though 64-bit desktop builds and runs, it fails to start the session (probably some logic I need to tweak)
- I'm working on a PowerShell script for setting up a Windows dev box which I need to finalize and include in this PR


gtritchie:
Closing, will submit a newer PR (the 64-bit Windows work was already merged, so the Qt 5.11 update is less involved now).

PR #2930, opened 2018-06-06
Specify shell type via terminalCreate API

### LINKED ISSUE #2903: Terminal: Feature request for rstudioapi::terminalCreate with specific type
In `cloudml`, some commands don't work nicely in git-bash nor powershell (not due to RStudio's terminal, but due to Google's tooling):

- https://github.com/rstudio/cloudml/issues/152
- https://github.com/rstudio/cloudml/issues/155

Would be really nice if I could run `rstudioapi::terminalCreate("caption", show = FALSE, type = "cmd")` maybe with `type` supporting a list of `c("cmd", "git", "powershell")`.

gtritchie:
Optional `shellType` argument to specify shell type via `terminalCreate`. The `rstudioapi` package will be updated to support this.

The shellType argument is a string and can be:

- **NULL** (default, uses shell type selected in Global Options, i.e. current behavior)
- **"default"**, same as NULL
- **"custom"**, Mac, Linux, Server: use "custom" shell specified in Global Options
- **"win-cmd"**, 64-bit Windows Command Prompt
- **"win-ps"**, 64-bit Windows PowerShell
- **"win-git-bash"** (Windows-only)
- **"win-wsl-bash"** (Windows-only)

Fixes #2903

If you request a shell type that isn't available, the default is used.

Example via rstudioapi and same thing using built-in RStudio syntax:

- `rstudioapi::terminalCreate(caption = "My Terminal", show = TRUE, shellType = "win-cmd")`
- `.Options$terminal.manager$terminalCreate("My Terminal", TRUE, "win-cmd")`

### DIFF
--- a/src/cpp/r/R/Api.R
+++ b/src/cpp/r/R/Api.R
@@ -551,18 +551,8 @@
 
    validShellType = TRUE
    if (!is.null(shellType)) {
-      validShellType <- grepl(
-         paste(
-            "default",
-            "win-cmd",
-            "win-ps",
-            "win-git-bash",
-            "win-wsl-bash",
-            "custom",
-          sep = "|"
-        ),
-        shellType,
-        ignore.case = TRUE)
+      validShellType <- tolower(shellType) %in% c("default", "win-cmd", 
+            "win-ps", "win-git-bash", "win-wsl-bash", "custom")
    }      
    if (!validShellType)
       stop("'shellType' must be NULL, or one of 'default', 'win-cmd', 'win-ps', 'win-git-bash', 'win-wsl-bash', or 'custom'.") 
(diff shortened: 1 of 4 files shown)


src/cpp/r/R/Api.R
```diff
@@ -539,14 +539,35 @@
...
 
    if (is.null(show) || !is.logical(show))
       stop("'show' must be a logical vector")
 
-   .Call("rs_terminalCreate", caption, show)
+   if (!is.null(shellType) && (!is.character(shellType) || (length(shellType) != 1)))
+      stop("'shellType' must be NULL or a character vector of length one")
+
+   validShellType = TRUE
+   if (!is.null(shellType)) {
```
kevinushey: `shellType %in% c("default", "win-cmd", <...>)` would be more idiomatic here.

PR #2935, opened 2018-06-07
Support installing odbc drivers with existing drivers with same name

javierluraschi:
Fix for https://github.com/rstudio/rstudio-pro/issues/445, in some platforms, some drivers are already installed with the same name as potential drivers installed by RStudio. Only in these cases, a `with RStudio` is appended to the driver:

screen shot 2018-06-07 at 3 40 52 pm


javierluraschi:
Renamed to `with RStudio Driver` since `with RStudio` read incomplete.

PR #3008, opened 2018-06-13
fix handling of Shift+Tab keypress

### LINKED ISSUE #2991: Shift + Tab fails to focus previous element
### System details

    RStudio Edition : Desktop
    RStudio Version : v1.2.731
    OS Version      : macOS 10.13.5
    R Version       : 3.5.0

Open the Help tab, click somewhere within the Help pane, and press <kbd>Tab</kbd> to focus one of the buttons. Then, try pressing <kbd>Shift + Tab</kbd>.

<kbd>Shift + Tab</kbd> fails to focus the previous element.

The previous element should gain focus.

---

On macOS, we're likely a victim of https://bugreports.qt.io/browse/QTBUG-65452. However, this also fails to work on other platforms, so there may be something else going on as well.

kevinushey:
This PR fixes an issue where attempts to use <kbd>Shift + Tab</kbd> to shift focus to the previous element could fail. This was because the `insertSnippet` command was being handled within `AceEditorCommandDispatcher`, and since the `ShortcutManager` saw that the keypress had been dispatched to a command, it stopped propagation of the event and prevented the browser default behavior.

This PR also works around a Qt bug whereby its attempts to dispatch `Qt::Key_Backtab` events to Chromium would fail; we accommodate this by manually translating these back to regular shift-tabs.

Closes #2991.

### DIFF
--- a/src/cpp/desktop/DesktopPosixApplicationLaunch.cpp
+++ b/src/cpp/desktop/DesktopPosixApplicationLaunch.cpp
@@ -52,6 +52,17 @@ protected:
          auto* pKeyEvent = static_cast<QKeyEvent*>(pEvent);
          auto modifiers = pKeyEvent->modifiers();
          
+         // translate backtab to regular tab
+         if (pKeyEvent->key() == Qt::Key_Backtab)
+         {
+            auto* pTabEvent = new QKeyEvent(
+                     pKeyEvent->type(),
+                     Qt::Key_Tab,
+                     pKeyEvent->modifiers() | Qt::ShiftModifier);
+            QCoreApplication::postEvent(pObject, pTabEvent);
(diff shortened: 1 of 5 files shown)


kevinushey:
For what it's worth, the `insertSnippet` command should really only be active when an Ace editor instance has focus, so this might instead be an issue related to how the commands defined here are managed:

https://github.com/rstudio/rstudio/blob/edaa34a81b026ea028f885590d2651bebd2ae66c/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/ace/AceEditorCommandDispatcher.java#L131-L137

Would this be better solved by managing these commands based on whether an Ace instance currently has focus?

Issue #2797, opened 2018-05-17
RStudio does not quit when the PDF viewer is open


aronatkins:
### System details

    RStudio Edition : Desktop
    RStudio Version : Version 1.2.639
    OS Version      : macOS 10.13.4
    R Version       : 3.4.3

### Steps to reproduce the problem

* Open an Rmd that renders to PDF. 
* Knit that PDF so it opens into a separate "RStudio: View PDF" window.
* With _either_ the primary RStudio window OR the PDF viewer in focus, hit Command-Q.
    * Using "Quit Session" from the RStudio menu has the same result.

### Describe the problem in detail

The primary RStudio IDE window closes, but the "RStudio: View PDF" window remains open.
The macOS menubar only has a single "RStudio" entry (File, Edit, etc are gone).
The "RStudio" menu has most options disabled.
![screen shot 2018-05-17 at 9 40 54 am](https://user-images.githubusercontent.com/362187/40181173-75a9483a-59b6-11e8-8ffc-178b91faee10.png)

Closing the PDF viewer does not let shutdown complete.

To quit RStudio, I Control-clicked on the RStudio icon in the dock and selected "Quit".

### Describe the behavior you expected

Command-Q quits RStudio completely.

kevinushey:
This should be fixed up in the next daily builds, with https://github.com/rstudio/rstudio/commit/7c38f3f95b707fe75ca883502563714c68dbbbf1. Thanks for reporting!

dfalty:
Verified fixed on version 1.2.1090.

PR #3025, opened 2018-06-19
ensure RStudio exits after last window closed

kevinushey:
This PR ensures that a close of a secondary window, if it is the final RStudio window opened, also quits the application.


kevinushey:
Seems like it was only an issue on Mac, since I can't reproduce the associated bug (sorry; should have linked in original post: https://github.com/rstudio/rstudio/issues/2797) on Linux with the latest v1.1 release of RStudio.

That said, I _can_ reproduce the issue on Linux with the RStudio v1.2 branch, so my best guess is that something changed on the Qt side in terms of how quits are handled. (Fortunately, this PR fixes things on both platforms)

PR #3058, opened 2018-06-23
add missing qt find_package

ecoughlan:
Simple fix to prevent the following error with Qt 5.11 in isolated environments:
```
CMake Error at /nix/store/1y5n2czmjghiljwjx14ckraxkffd3m35-qtbase-5.11.1-dev/lib/cmake/Qt5Core/Qt5CoreMacros.cmake:329 (message):
  Can not use "QuickWidgets" module which has not yet been found.
Call Stack (most recent call first):
  src/cpp/desktop/CMakeLists.txt:363 (qt5_use_modules)
```


gtritchie:
We are updating to use Qt 5.11.1 for RStudio, and I should be checking that in soon. That will include a fix for this issue. So if it's ok, I'll close this PR in lieu of that larger upcoming change. What do you think?

averelld:
Sure, sounds good to me.

ecoughlan:
Sure, sounds good to me.

PR #3164, opened 2018-07-16
discover required dependencies and prompt user

kevinushey:
NOTE: not quite ready for merge.

I realize this PR is coming in quite late but this is something that we originally wanted to get in for v1.2 and the nonstop deluge of WebEngine bugs was starting to drive me insane :-)

This PR implements automatic R package dependency discovery in R source documents, and prompts the user to install any missing packages with a warning bar when necessary.

screen shot 2018-07-16 at 4 39 20 pm

Clicking the <kbd>Install</kbd> button here opens the typical dependency manager installer, and attempts to install all packages from CRAN.

## TODO

- [x] Need to ensure that we only attempt to install packages from CRAN that actually do live on CRAN.
- [x] Provide some UI to allow users to dismiss dialog permanently (per file? globally?) in case we aren't discovering dependencies directly.

### DIFF
--- a/src/cpp/session/modules/SessionCodeTools.R
+++ b/src/cpp/session/modules/SessionCodeTools.R
@@ -13,6 +13,11 @@
 #
 #
 
+.rs.addJsonRpcHandler("discover_package_dependencies", function(docId, fileType)
+{
+   .rs.discoverPackageDependencies(docId, fileType)
+})
+
 .rs.addFunction("error", function(...)
 {
    list(
@@ -1401,9 +1406,10 @@
 
 .rs.addFunction("recursiveSearch", function(`_node`, fn, ...)
 {
-   if (fn(`_node`, ...)) return(TRUE)
+   if (fn(`_node`, ...))
+      return(TRUE)
    
-   if (is.call(`_node`))
+   if (is.recursive(`_node`))
       for (i in seq_along(`_node`))
          if (.rs.recursiveSearch(`_node`[[i]], fn, ...))
             return(TRUE)
@@ -1415,7 +1421,7 @@
 {
    fn(`_node`, ...)
    
-   if (is.call(`_node`))
+   if (is.recursive(`_node`))
       for (i in seq_along(`_node`))
          .rs.recursiveWalk(`_node`[[i]], fn, ...)
 })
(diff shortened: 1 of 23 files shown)


kevinushey:
They are indeed all collected into the same prompt:

screen shot 2018-07-17 at 8 42 12 am

Of course since these aren't actual valid CRAN package names there's still work to be done to ensure that the packages we attempt to install do truly exist in the current repositories

kevinushey:
Now with a <kbd>Don't Show Again</kbd> button so that users can dismiss the dialog permanently for a particular file:

screen shot 2018-07-17 at 11 52 44 am

We also look at the result of `available.packages()` to ascertain whether a package is actually available from CRAN before notifying the user that it may be installed.

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTargetWidget.java
```diff
@@ -854,6 +855,32 @@ public void showReadOnlyWarning(final List<String> alternatives)
...
+      {
+         String disableDependencyDiscovery = docUpdateSentinel_.getProperty("disableDependencyDiscovery");
+         if (StringUtil.equals(disableDependencyDiscovery, "1"))
+         {
+            return;
+         }
+      }
+      
+      CommandWithArg<Boolean> onInstallComplete = (Boolean success) -> {
+         hideWarningBar();
```
jmcphers: I think we should keep the warning bar if `success` is false -- that could mean that the user cancelled, or that they couldn't install the packages. In either case we should let them try again.

src/gwt/src/org/rstudio/core/client/widget/InfoBar.java
```diff
@@ -93,6 +95,44 @@ public int getHeight()
...
+      }
+      
+      label_.setText(message);
+      
+      if (labelRight_.getWidgetCount() == 0)
+      {
+         labelRight_.add(label("Install", () -> {
+            RStudioGinjector.INSTANCE.getDependencyManager().installPackages(packages, onPackageInstallCompleted);
+         }));
+
```
jmcphers: I think we need something between "install right now" and "don't ever show this again". Can we add *Hide*? (I can't remember whether the warning bar has dismissal mechanism built in; disregard if it does!)

src/cpp/session/modules/SessionCodeTools.R
```diff
@@ -2071,3 +2077,148 @@
...
+
+.rs.addFunction("discoverPackageDependencies", function(id, type)
+{
+   # check to see if we have available packages -- if none, bail
+   available <- .rs.availablePackages()
+   if (is.null(available$value))
+      return(character())
+   
+   # attempt to read the file
+   contents <- .rs.tryCatch(.rs.readSourceDocument(id))
```
jmcphers: It looks like this is expecting `.rs.readSourceDocument` to raise an error but as far as I can tell it can't do this at present (instead it returns null if it can't read the doc).

src/cpp/session/modules/SessionCodeTools.R
```diff
@@ -2071,3 +2077,148 @@
...
+{
+   contents <- .Call("rs_readSourceDocument", as.character(id), PACKAGE = "(embedding)")
+   Encoding(contents) <- "UTF-8"
+   contents
+})
+
+.rs.addFunction("discoverPackageDependencies", function(id, type)
+{
+   # check to see if we have available packages -- if none, bail
+   available <- .rs.availablePackages()
```
jmcphers: It seems like the very first call to discoverPackageDependencies will trigger the package list to be cached but this call (and any subsequent ones made prior to the completion of available.packages) will be dropped on the floor. Do you think it'd be worth emitting a client event when we've completed the cache population so that we can finish checking for dependencies? 

I'm imagining e.g. a scenario wherein the IDE boots into a session with R files already open (e.g. as part of a first-use project). Will those get package install prompts?

src/gwt/src/org/rstudio/core/client/widget/InfoBar.java
```diff
@@ -93,6 +95,44 @@ public int getHeight()
...
    }
    
+   public void showRequiredPackagesMissingWarning(List<String> packages,
+                                                  CommandWithArg<Boolean> onPackageInstallCompleted,
+                                                  Command onDismiss)
+   {
+      String message;
+      
+      int n = packages.size();
+      if (n == 1)
```
jmcphers: Thanks for taking the time to make this summary nice!

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTargetPackageDependencyHelper.java
```diff
@@ -0,0 +1,98 @@
...
+      docDisplay_ = docDisplay;
+      
+      init();
+   }
+   
+   private void init()
+   {
+      RStudioGinjector.INSTANCE.injectMembers(this);
+      
+      docDisplay_.addSaveCompletedHandler((SaveFileEvent event) -> {
```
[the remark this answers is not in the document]
kevinushey: My thought was that user's might want to see this after e.g. modifying a script and injecting new `library()` calls, only to discover that they're attempting to use a package they haven't installed yet.

I agree 100% that this should only occur during manual saves, though -- I'll fix that up!

src/gwt/src/org/rstudio/core/client/widget/InfoBar.java
kevinushey: The warning bar does indeed have a little red [X] at the far right for dismissal :)

src/cpp/session/modules/SessionSource.cpp
```diff
@@ -1241,6 +1241,18 @@ SEXP rs_requestDocumentSave(SEXP idsSEXP)
    return r::sexp::create(success, &protect);
 }
 
+SEXP rs_readSourceDocument(SEXP idSEXP)
+{
+   std::string id = r::sexp::asString(idSEXP);
+   boost::shared_ptr<SourceDocument> pDoc(new SourceDocument());
+   Error error = source_database::get(id, pDoc);
+   if (error)
```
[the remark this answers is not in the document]
kevinushey: I think we should log it but raising an R error is likely overkill as there is no obvious resolution the user could make if we failed to read the source document here.

kevinushey:
Should we provide a global UI option (e.g. in the Global Options pane) to toggle dependency discovery on / off? What's the most appropriate place for that option to live?

PR #3172, opened 2018-07-18
Add config option to send package build output to package folder

gtritchie:
Normally the `Check Package`, `Build Source Package`, and `Build Binary Package` commands write to the parent folder of the package itself. So for `/home/me/mypackage`, files are written to `/home/me`.

This change supports writing these files inside the package folder itself, for a case where users don't have write access to the folder containing the package project. This was requested by the cloud team (see https://github.com/rstudio/rstudio-pro/issues/482). Implementing in open-source server.

This can be enabled in `rsession.conf` by adding:

`package-output-to-package-folder=1`

Note: this only works if package operations are being done via devtools package; if they are being done directly via R CMD check/install/build, this setting will not change the output location.

In addition to changing the output location, the .gitignore and .Rbuildignore files have additional exclusions dynamically added, covering the files generated by these operations.

These exclusions will be added only when this setting is enabled, and can happen when first creating a project or when opening an existing project.

FYI @ssinnott

### DIFF
--- a/src/cpp/core/include/core/r_util/RPackageInfo.hpp
+++ b/src/cpp/core/include/core/r_util/RPackageInfo.hpp
@@ -1,7 +1,7 @@
 /*
  * RPackageInfo.hpp
  *
- * Copyright (C) 2009-12 by RStudio, Inc.
+ * Copyright (C) 2009-18 by RStudio, Inc.
  *
  * Unless you have received this program directly from RStudio pursuant
  * to the terms of a commercial license agreement with RStudio, then
@@ -71,6 +71,10 @@ private:
 };
 
 bool isPackageDirectory(const FilePath& dir);
+
+// Determine name of a package project in given directory. Returns empty string
+// if not a package or any other error.
+std::string packageNameFromDirectory(const FilePath& dir);
 
 } // namespace r_util
 } // namespace core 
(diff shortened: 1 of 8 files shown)


jjallaire:
Why not for R CMD check and only for devtools? I would guess that many package projects will not have devtools so in that case we'll still get the errors / failures.

gtritchie:
I didn't realize it was a common scenario (not having devtools). Sure, I'll look at fixing that, too.

gtritchie:
Changed reviewer: per our area ownership this should have been Kevin, not Jonathan.

kevinushey:
What about performing the build within a subdirectory of the package root, e.g. `build-<package>`? This would be cleaner in terms of the various augmentations that `.gitignore` and `.Rbuildignore` require right now -- we could then just ignore that single folder in each case and avoid 'polluting' `.Rbuildignore` and `.gitignore`.

kevinushey:
(other than that overarching question the PR looks good to me)

gtritchie:
Received signoff from @kevinushey, going ahead with merge.

jspiewak:
@dfalty please let me know when a build with this fix has been verified.

dfalty:
@jspiewak verified the functionality on 1.2.909-1. Should be good to go with the conf setting.

PR #3178, opened 2018-07-18
warn when attempting to launch before ready

kevinushey:
I suspect some users are having an RStudio failure to launch due to the front-end attempting initialization before Qt is ready, e.g.

https://community.rstudio.com/t/rstudio-daily-build-never-loads/7972/30

This PR increases the initial wait delay to 1 minute, and also logs when we do this sort of un-synchronized launch.


kevinushey:
We do already synchronize the launch in that way, but we have a timeout for the case where Qt seems to have stalled or something.

Perhaps instead of this we should be displaying some front-end UI that more clearly communicates the state of affairs to the user, though.

PR #3209, opened 2018-07-24
implement tri-state toggle

kevinushey:
This PR implements a new tri-state toggle for the chunk options popup panel (hat tip to @nwstephens)

screen shot 2018-07-24 at 9 24 14 am

The intention of this toggle is to make it more clear to the user that three possible states are supported by the toggle, and make it a bit more clear what the 'indeterminate' state is. Compare with the old treatment:

screen shot 2018-07-24 at 9 37 54 am

I think the new UI is overall an improvement, although I'm somewhat torn as to whether we should use color or greyscale here. Example of what that might look like:

screen shot 2018-07-24 at 9 39 31 am

Also wonder if it's worth increasing the contrast a bit on the 'knob' to make it stand out a bit more.

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/widget/Toggle.java
+++ b/src/gwt/src/org/rstudio/core/client/widget/Toggle.java
@@ -62,29 +62,31 @@ public class Toggle
       initWidget(container_);
    }
    
-   public void setState(State state)
+   public void setState(State state, boolean animate)
    {
-      clearToggleStyles();
-      
       if (!indeterminateStateEnabled_ && state == State.INDETERMINATE)
       {
          assert false : "Attempted to set indeterminate state on binary toggle";
       }
       
+      if (animate)
+         knob_.removeStyleName(styles_.transitionDisabled());
+      else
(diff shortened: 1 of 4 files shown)


kevinushey:
Grayscale with contrast:

screen shot 2018-07-24 at 9 46 03 am

jmcphers:
I think grayscale with contrast is the best treatment -- the green/red colors risk giving off too much of a "correct"/"incorrect" vibe. Can we make these a little bigger (at least the height of a capital letter) and give them a little room on the right?

Also somewhat on the fence about whether this is the right treatment. Though it's boring, a simple drop list might be clearer -- it's very difficult to convey "follow defaults". We could use this in combination with a checkbox -- e.g.

    Chunk Options
    [  ] Output [ Code only ]
    [x ] Show warnings [ Always ]
    [  ] Show messages [ Never ]

You would need to check the option in order to enable the drop-list. Unchecked options wouldn't be added to the chunk.

What do you think?

kevinushey:
Grayscale + with height slightly increased:

screen shot 2018-07-24 at 6 02 48 pm

Issue #3217, opened 2018-07-25
Menu bar with black text on dark background in Kubuntu with dark desktop theme


asalarian:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.1.456
    OS Version      : Kubuntu 18.04 LTS
    R Version       : MS R Open 3.5.0

### Describe the problem in detail
I am using a dark theme on Kubuntu so I have also selected a dark theme for the editor (Pastel on Dark). The issue as you can see in the picture below is that the menu bar is rendered with black text on dark background! 

![image](https://user-images.githubusercontent.com/10533408/43204378-f230be42-9020-11e8-8c86-7883d13463ab.png)

Also, the dialog boxes such as the "About RStudio" or "Global options" do no show up correctly as the dialog box is rendered using light theme but some widget are dark with black text:

![image](https://user-images.githubusercontent.com/10533408/43204625-b1051598-9021-11e8-9461-a2d81decf9f6.png)

If I switch to a light theme for the editor while keep using a dark theme for my Kubuntu desktop, in RStudio everything changes to light but the menu bar remains the same (i.e. black text on dark background).

### Describe the behavior you expected
The menu bar should use white for text and when using a dark desktop theme, the dialog boxes should use a dark theme as well.

kevinushey:
What styling is used for other applications on your machine (for menu text and input select boxes)?

asalarian:
I am using the "Modern" desktop theme. The behavior is the same when I use the out of the box "Breeze Dark" theme.
Currently, RStudio is the only application that has this problem. Everywhere else the dark theme is working correctly with the menu text, etc.

sebastian-de:
I have this problem, too, when using the "Breeze Dark" theme in Kubuntu.
The menu bar text always stays black, but the colors of menus look fine:
![screenshot_20181013_204200](https://user-images.githubusercontent.com/7430495/46908821-de4c8b80-cf28-11e8-9c42-350ca5c327cb.png)
The palette for Breeze Dark can be found here:
https://cgit.kde.org/breeze.git/tree/colors/BreezeDark.colors

As a quick fix, I currently start RStudio with the default theme:
`QT_QPA_PLATFORMTHEME=“” rstudio`

kevinushey:
I see the same thing in some other applications (e.g. FireFox) so it seems like elements of the system theme are leaking inappropriately into applications. Fixing this likely implies rolling our own versions of all the 'default' UI elements so that system themes don't get in the way.

[...]

PR #3225, opened 2018-07-25
use custom menu styles for Windows + GNOME 3

### LINKED ISSUE #3052: tweak styling of menus with GNOME 3 / Ubuntu 18.04
See e.g.

screen shot 2018-06-21 at 8 07 15 am

It'd be nice if we could reduce the contrast a little bit -- the menus are a bit too 'bright' by default.

kevinushey:
Before:

screen shot 2018-07-25 at 3 06 40 pm

After:

screen shot 2018-07-25 at 3 05 47 pm

Closes #3052.

@jjallaire, let me know what you think.

Also let me know what you think about the display of mnemonics. It looks like GNOME 3 explicitly removed support for 'auto' mnemonics (e.g. toggling the display of mnemonics by pressing the <kbd>Alt</kbd> key -- why, I have no idea) so we have no choice but to either always display them or never display them.

(Or, we could try to manage their display ourselves -- some applications seem to attempt this, e.g. LibreOffice -- but the experience there is far from perfect and the mnemonics seem to flicker a lot)


jmcphers:
I think people with dark themes might be annoyed that the menus are always bright white (in the case of dark RStudio + dark desktop theme, they could be pretty jarring). Would it be easy to load a reverse-video stylesheet when using a dark theme? 

Does this fix https://github.com/rstudio/rstudio/issues/3217, too? (If not, maybe the `gnome` test needs to be more permissive?)

kevinushey:
Now with support for dark themes:

screen shot 2018-07-25 at 4 31 39 pm

Note to self: take the time to properly bundle + load the CSS styles rather than just embedding them in the C++ source ...

kevinushey:
I think it's probably worth the time to match the main menu theme to the IDE theme. The difference in color between the main menu and the IDE itself is a bit jarring.

kevinushey:
And now with themes adapted for Windows:

screen shot 2018-07-27 at 1 59 39 pm

Note: I wish we could also adjust the title bar colon, but as far as I can tell Windows nor Qt expose any APIs for doing this. That said, it certainly must be possible as e.g. one can enable a dark theme Microsoft Edge which gives a dark-colored title bar.

kevinushey:
Looking at this further, it's not possible to style the title bar on an application-specific basis in Windows. Rather, the application must instead 'fake it' and implement its own menu-titlebar hybrid.

This implies making the window frameless. Unfortunately, making a window frameless implies turning off Windows' default resize + move behaviors for Windows; one must instead explicitly implement the 'regular' window behaviors by hand; that is:

1. Choosing the appropriate cursor based on the cursor position (ie: drawing resize cursors near the window edge),

2. Implementing appropriate window resize behavior (ie: tracking mouse down, mouse move, and mouse up events to implement window resize)

3. Handle cases where the mouse is e.g. near the top border but actually hovering over a menu item,

4. Implement facades for the minimize, maximize, and close icons.

This is all doable, and doable in Qt, it's just rather involved and potentially bug-prone as IIUC we need to install a native event filter for processing these events (ie: implement our own WndProc overrides for certain mouse events)

Worth noting that once we get into the business of rendering our own menu bar, it no longer needs to happen in Qt: we could just implement it directly in GWT (as we do for RStudio Server) but hook up the appropriate desktop events in that case. Not sure if that's truly easier than implementing it in Qt, beyond that we'd get better support for certain things e.g. SVG animations + styling + transition effects for e.g. hover.

jjallaire:
Okay, I would argue that given the effort required we shouldn't implement our own titlebar using a frameless window (too many small things to get right, we'd be chasing this one for a long time).

jmcphers:
Agreed, in my view frameless windows can be made to look very cool but are expensive and hard to do tastefully!

@kevinushey given that we don't want to pursue more aggressive styling or control, is this PR ready to go as-is?

kevinushey:
Yes, I think this PR should be ready to go!

PR #3278, opened 2018-08-07
fixups for setting desktop zoom level

### LINKED ISSUE #3277: adjusting zoom level in Global Options pane doesn't work
### System details

    RStudio Edition : Desktop
    RStudio Version : v1.2.877
    OS Version      : Ubuntu 18.04 64bit (GNOME 3)
    R Version       : 3.5.1

Tools -> Global Options... -> Appearance -> Zoom. Try changing the zoom level, and then hitting apply.

The old zoom level is used.

The new zoom level should be used.

Note that the new zoom level is properly persisted at `~/.config/RStudio/desktop.ini` so it appears that we're caching and reusing an old zoom value.

kevinushey:
Closes #3277.

### DIFF
--- a/src/cpp/desktop/DesktopGwtCallback.cpp
+++ b/src/cpp/desktop/DesktopGwtCallback.cpp
@@ -1274,8 +1274,7 @@ double GwtCallback::getZoomLevel()
 
 void GwtCallback::setZoomLevel(double zoomLevel)
 {
-   options().setZoomLevel(zoomLevel);
-   desktopInfo().setZoomLevel(zoomLevel);
+   pOwner_->setZoomLevel(zoomLevel);
 }
 
 void GwtCallback::zoomIn()
--- a/src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/AppearancePreferencesPane.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/AppearancePreferencesPane.java
@@ -426,19 +426,22 @@ public class AppearancePreferencesPane extends PreferencesPane
       {
          uiPrefs_.theme().setGlobalValue(themeList_.get(theme_.getValue()));
       }
+      
       if (Desktop.isDesktop())
       {
          if (!StringUtil.equals(initialFontFace_, fontFace_.getValue()))
          {
-            Desktop.getFrame().setFixedWidthFont(StringUtil.notNull(fontFace_.getValue()));
+            String fontFace = fontFace_.getValue();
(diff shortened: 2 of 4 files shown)


src/cpp/desktop/DesktopGwtCallback.cpp
```diff
@@ -1274,8 +1274,7 @@ double GwtCallback::getZoomLevel()
 
 void GwtCallback::setZoomLevel(double zoomLevel)
 {
-   options().setZoomLevel(zoomLevel);
-   desktopInfo().setZoomLevel(zoomLevel);
```
kevinushey: Note that this was redundant (setting the preference in options also updates the desktopInfo instance) so it's been removed.

src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/AppearancePreferencesPane.java
```diff
@@ -426,19 +426,22 @@ public boolean onApply(RPrefs rPrefs)
...
+            Desktop.getFrame().setFixedWidthFont(fontFace);
             restartRequired = true;
          }
          
          if (!StringUtil.equals(initialZoomLevel_, zoomLevel_.getValue()))
          {
             double zoomLevel = Double.parseDouble(zoomLevel_.getValue());
+            initialZoomLevel_ = zoomLevel_.getValue();
             Desktop.getFrame().setZoomLevel(zoomLevel);
-            restartRequired = true;
```
kevinushey: Restart is indeed no longer required when setting the zoom level with WebEngine (hurray!)

src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/AppearancePreferencesPane.java
```diff
@@ -426,19 +426,22 @@ public boolean onApply(RPrefs rPrefs)
...
+            String fontFace = fontFace_.getValue();
+            initialFontFace_ = fontFace;
+            Desktop.getFrame().setFixedWidthFont(fontFace);
             restartRequired = true;
          }
          
          if (!StringUtil.equals(initialZoomLevel_, zoomLevel_.getValue()))
          {
             double zoomLevel = Double.parseDouble(zoomLevel_.getValue());
+            initialZoomLevel_ = zoomLevel_.getValue();
```
kevinushey: Note that updating this is necessary as otherwise changing to a zoom level but then attempting to revert to the previous setting would fail (as the UI would interpret this as no change)

PR #3283, opened 2018-08-07
provide ability to change rendering engine

kevinushey:
This PR allows user to select an alternate rendering engine with RStudio Desktop. This is primarily intended to be an escape hatch for users who are bumping into rendering problems or crashes when using the default hardware accelerated renderer.

screen shot 2018-08-07 at 2 57 49 pm

At least in my own local testing, ANGLE + Software are much slower than regular desktop rendering. However, I no longer see crashes or hangs when using the software renderer in my virtual machine, so there is additional evidence that this will help users avoid some classes of issues.

### DIFF
--- a/src/cpp/desktop/DesktopMain.cpp
+++ b/src/cpp/desktop/DesktopMain.cpp
@@ -265,6 +265,42 @@ bool useRemoteDevtoolsDebugging()
 #endif
 }
 
+void initializeRenderingEngine()
+{
+   QString engine = desktop::options().desktopRenderingEngine();
+   if (engine.isEmpty())
+      return;
+   
+   using namespace core::system;
+   if (engine == QStringLiteral("desktop"))
+   {
+      setenv("QT_OPENGL", "desktop");
+   }
+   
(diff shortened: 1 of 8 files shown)


src/cpp/desktop/DesktopMain.cpp
```diff
@@ -265,6 +265,60 @@ bool useRemoteDevtoolsDebugging()
 #endif
 }
 
+void initializeRenderingEngine(std::vector<char*>* pArguments)
+{
+   QString engine = desktop::options().desktopRenderingEngine();
```
jmcphers: If I'm not mistaken then this means that users who are somehow totally unable to launch RStudio could get unblocked by seeding an options file (which is great!)

kevinushey:
Good point on the select box. I'm not exactly sure why I went with radio buttons given that we generally use select boxes everywhere else -- tunnel vision perhaps!

kevinushey:
Now much tidier with a select box:

screen shot 2018-08-07 at 3 33 30 pm

PR #3297, opened 2018-08-10
Remove superfluous whitespaces in shiny template 

dirkschumacher:
Just a minor thing, but I noticed that the shiny template has superfluous whitespaces when I added it to git.

Disclaimer: I used the web interface of Github to make the changes, so I haven't build RStudio and tested it. I also could not figure out how to add multiple files to a single commit, so I created three commits with the same message. Needs squashing :)


gtritchie:
Thanks! Have you signed the contributor agreement per https://github.com/rstudio/rstudio/blob/master/CONTRIBUTING.md?

dirkschumacher:
Yes, years ago in 2014.

PR #3301, opened 2018-08-10
Support for custom source actions with special comment

javierluraschi:
Some packages like [mlflow](https://github.com/rstudio/mlflow/R/mlflow) would like to customize which operation triggers when an R file gets sourced. For `mlflow`, this would enable tracking model parameters, results, artifacts, etc. just by saving their models.

This PR adds support for a `# !source function args` similar to `-- !preview` and `// !preview` that allows R scripts to override the source action.

screen shot 2018-08-10 at 3 07 21 pm

See also [rstudio/mlflow/issues/11](https://github.com/rstudio/mlflow/issues/11)

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTargetWidget.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTargetWidget.java
@@ -694,8 +694,10 @@ public class TextEditingTargetWidget
       
       sourceOnSave_.setVisible(canSourceOnSave);
       srcOnSaveLabel_.setVisible(canSourceOnSave);
-      if (fileType.isRd() || fileType.isJS() || canPreviewFromR || fileType.isSql() )
+      if (fileType.isRd() || fileType.isJS() || canPreviewFromR || fileType.isSql())
          srcOnSaveLabel_.setText(fileType.getPreviewButtonText() + " on Save");
+      else if (hasCustomSource())
+         srcOnSaveLabel_.setText("Custom Action on Save");
       else
          srcOnSaveLabel_.setText("Source on Save");
       codeTransform_.setVisible(
@@ -814,6 +816,11 @@ public class TextEditingTargetWidget
       return extendedType_ != null &&
              extendedType_.startsWith(SourceDocument.XT_TEST_SHINYTEST);
    }
+
+   private boolean hasCustomSource()
+   {
+      return SourceDocument.hasCustomSource(extendedType_);
+   }
    
    @Override
    public void setNotebookUIVisible(boolean visible)
@@ -854,13 +861,20 @@ public class TextEditingTargetWidget
       knitDocumentButton_.setText(width < 450 ? "" : knitCommandText_);
       
       if (editor_.getFileType().isRd() || editor_.getFileType().isJS() || 
-          editor_.getFileType().isSql() ||editor_.getFileType().canPreviewFromR())
+          editor_.getFileType().isSql())
       {
          String preview = editor_.getFileType().getPreviewButtonText();
          srcOnSaveLabel_.setText(width < 450 ? preview : preview + " on Save");
       }
+      else if (hasCustomSource())
+      {
+         srcOnSaveLabel_.setText(width < 450 ? "Custom" : "Custom Action on Save");
+      }
       else
+      {
          srcOnSaveLabel_.setText(width < 450 ? "Source" : "Source on Save");
+      }
+
       sourceButton_.setText(width < 400 ? "" : sourceCommandText_);
    }
    
--- a/src/cpp/session/CMakeLists.txt
+++ b/src/cpp/session/CMakeLists.txt
@@ -168,6 +168,7 @@ set (SESSION_SOURCE_FILES
    modules/connections/ConnectionHistory.cpp
    modules/connections/ConnectionsIndexer.cpp
    modules/connections/SessionConnections.cpp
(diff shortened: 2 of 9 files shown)


src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTargetWidget.java
```diff
@@ -854,13 +861,20 @@ private void manageToolbarSizes()
...
       if (editor_.getFileType().isRd() || editor_.getFileType().isJS() || 
-          editor_.getFileType().isSql() ||editor_.getFileType().canPreviewFromR())
+          editor_.getFileType().isSql())
       {
          String preview = editor_.getFileType().getPreviewButtonText();
          srcOnSaveLabel_.setText(width < 450 ? preview : preview + " on Save");
       }
+      else if (hasCustomSource())
+      {
+         srcOnSaveLabel_.setText(width < 450 ? "Function" : "Function on Save");
```
kevinushey: This text seems odd to me -- why not just keep `Source on Save`?

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTargetWidget.java
```diff
@@ -854,13 +861,20 @@ private void manageToolbarSizes()
       knitDocumentButton_.setText(width < 450 ? "" : knitCommandText_);
       
       if (editor_.getFileType().isRd() || editor_.getFileType().isJS() || 
-          editor_.getFileType().isSql() ||editor_.getFileType().canPreviewFromR())
```
kevinushey: Why are we removing the `canPreviewFromR()` check here? (seems like it's orthogonal to this feature)

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTargetWidget.java
```diff
@@ -694,8 +694,10 @@ public void adaptToFileType(TextFileType fileType)
       
       sourceOnSave_.setVisible(canSourceOnSave);
       srcOnSaveLabel_.setVisible(canSourceOnSave);
-      if (fileType.isRd() || fileType.isJS() || canPreviewFromR || fileType.isSql() )
+      if (fileType.isRd() || fileType.isJS() || canPreviewFromR || fileType.isSql())
          srcOnSaveLabel_.setText(fileType.getPreviewButtonText() + " on Save");
+      else if (hasCustomSource())
+         srcOnSaveLabel_.setText("Function on Save");
```
kevinushey: Not a fan of 'Function on Save'; I think keeping it as 'Source on Save' is fine.

kevinushey:
We were tracking this issue on the RStudio side with https://github.com/rstudio/rstudio/issues/3257.

Note that in particular we wanted to choose a scheme of the form:

    # !source altparsers::src(.file, parser = tidy_parse_file)

That is, the right-hand side of the source request should just be an R function to execute. What do you think about just allowing this?

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTargetWidget.java
javierluraschi: Yeah, we could, we change the `source` message, but I think I agree with you, might be better just to keep the usual `Source on Save`, will change.

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTargetWidget.java
javierluraschi: Yeah, lets keep it the way it is.

javierluraschi:
It should work but this would be a bit different, you would run isntead:

```
# !source altparsers::src .file parser=tidy_parse_file
```

which is consistent with the other preview methods we've been adding.

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTargetWidget.java
jjallaire: +1

jjallaire:
I think we agreed though that we also wanted this mechanism to be able to handle the code of the source file. That discussion also established that it would be better to have this be less "magical" than the `!preview` syntax we established.

As a compromise, how about if we recognize the simple form of just the function name, so:

```r
# !source mlflow::mlflow_run
```

Then if you wanted to add parameters you would need to use `.file` or `.code` directly:

```r
# !source mlflow::mlflow_run(.file, foo = TRUE)
# !source mlflow::mlflow_run(.code, foo = TRUE)
```

I honestly think this is better than what we did for r2d3 (as it's much more clearly R code than some special syntax or form.

javierluraschi:
@jjallaire should we also change SQL and D3 preview to use the function + `.file` and `.code` syntax to be consistent?

R2D3 would need the docs and CRAN update to fix a couple examples but that's about it.

javierluraschi:
Alright, so here is what I changed:
- D3 and SQL supports the old format still; however, one can also use the new `.code`, `.file` with explicit functions.
- Support for `.code` and `.file` using R's `substitute()` to allow for complex substitutions.
- Support for functional calls, as in `# !source print(x = .code)`; however, the D3/SQL calls are also supported for `!source`, as in `# !source print x=.code`.

PR #3302, opened 2018-08-11
don't warn when failing to revert a deleted file

### LINKED ISSUE #3231: rstudio crashes on close if file no longer exists
### System details

    RStudio Edition : Desktop
    RStudio Version : v1.2.846
    R Version       : 3.5.1

Open a file in the IDE, then delete that file from the disk. You'll get a **File Deleted** prompt:

screen shot 2018-07-26 at 6 37 36 pm

Click <kbd>No</kbd>. The file remains open in the IDE, but 'dirty'. Then, attempt to quit the IDE (with <kbd>Cmd + Q</kbd>). You'll be prompted to save changes:

screen shot 2018-07-26 at 6 38 27 pm

Click <kbd>Don't Save</kbd>.

A **Save File** error displays:

screen shot 2018-07-26 at 6 38 52 pm

After clicking <kbd>OK</kbd>, the session 'aborts' and the bomb is shown. (In reality, the process has already exited; the display of the dialog obscures that fact)

RStudio shouldn't crash on close.

kevinushey:
Closes #3231.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/server/ServerErrorCause.java
+++ b/src/gwt/src/org/rstudio/studio/client/server/ServerErrorCause.java
@@ -36,4 +36,6 @@ public class ServerErrorCause
    private final int code_;
    private final String category_;
    private final String message_;
+   
+   public static final int FILE_NOT_FOUND = 2;
 }
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTarget.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTarget.java
@@ -2387,7 +2387,7 @@ public class TextEditingTarget implements
    
    public void revertChanges(Command onCompleted)
    {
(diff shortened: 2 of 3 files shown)


src/gwt/src/org/rstudio/studio/client/server/ServerErrorCause.java
```diff
@@ -36,4 +36,6 @@ public String toString()
    private final int code_;
    private final String category_;
    private final String message_;
+   
+   public static final int FILE_NOT_FOUND = 2;
```
gtritchie: Is this value platform-independent? Guessing it is, but just checking.

kevinushey: Indeed it is, although I think it's mostly just a happy coincidence. On Unix the associated error code ENOENT (no such path or directory) maps to 2 on Linux + macOS; on Windows Boost defines its own 'file not found' error code that just so happens to equal 2 as well. (Perhaps Boost chose the Windows error codes explicitly to match the common Unix error codes?)

Issue #1977, opened 2018-01-15
Is Tools > Shell redundant now that we have Tools > Terminal?


jennybc:
This question is the conclusion of a discussion elsewhere with @gtritchie.

Do the cons of having both (mostly confusion, I think) outweigh the pros? Maybe this issue will attract more concrete examples of pros/cons or a rousing defence of *Tools > Shell*.

rmflight:
My question would be:

Does the terminal **completely** replace a native shell yet?

I know there were issues using keyboard shortcuts to copy and paste from the shell (#1978), and copying and pasting from a Tmux session (#1765).

It is still nice to be able to open a shell in the project directory, IMO. 

But I also see where there could be confusion. Does the menu need to be better, to help avoid it? Or make the *shell* something enabled by users who know what they are doing, and not available in the default config?

gtritchie:
I'm not planning on removing Tools / Shell, so closing this issue.

PR #3314, opened 2018-08-14
run script before activating terminal

kevinushey:
Closes https://github.com/rstudio/rstudio/issues/3313.

### DIFF
--- a/src/cpp/desktop/mac-terminal.in
+++ b/src/cpp/desktop/mac-terminal.in
@@ -2,8 +2,8 @@
 on run argv
   set dir to quoted form of (first item of argv)
   tell app "Terminal"
-     activate
      do script "cd " & dir
+     activate
   end tell
 end run
 


kevinushey:
It seems like this PR may be insufficient -- it looks like I'm still getting duplicate terminal windows on launch.

My suspicion is that this behavior seems to depend on whether the Terminal application has already been launched. If it hasn't, activating it also forces the creation of a terminal window. However, the `do script` section _also_ launches a new terminal window, for execution of that script.

It seems like we need some way of saying, "ensure the Terminal application is launched, but please don't open any Windows just yet", and then "please open a Terminal window and run a script in it for us".

gtritchie:
I played with this quite a bit during 1.1; Visual Studio code has a similar issue (they also use AppleScript to launch the shell). Never found a solution that could handle all the cases (not to say there isn't one, but was turning into something more trouble than it was worth, especially now that we have the built-in terminal). There's also an issue to consider removing the shell command altogether, #1977, so might not be worth spending much time on this.

kevinushey:
I think you're probably right. Given that I was the one who reported the double-terminal-window issue, and this buglet has been in RStudio for a while, I'm guessing most users aren't launching shells in this way.

PR #3339, opened 2018-08-17
Improve spellcheck markdown coverage

adamconroy:
Bugfix for: https://github.com/rstudio/rstudio/issues/1588

Previously spellchecking an R markdown document would only check tokens labeled with the "text" type. Due to how highlighting is set up many tokens that are actually text types have other classes such as "constant" and "numeric". This PR adds the "text" type to existing types as well as adds "string" and "header" as valid types to check.

aejnhiy

### DIFF
--- a/src/gwt/acesupport/acemode/markdown_highlight_rules.js
+++ b/src/gwt/acesupport/acemode/markdown_highlight_rules.js
@@ -91,22 +91,22 @@ var MarkdownHighlightRules = function() {
     // handle highlighting for *abc*, _abc_ separately, as pandoc's
     // parser is a bit more strict about where '_' can appear
     var strongUnderscore = {
-        token: ["text", "constant.numeric"],
+        token: ["text", "constant.numeric.text"],
         regex: "(\\s+|^)(__\\S.+?\\S__)\\b"
     };
 
     var emphasisUnderscore = {
-        token: ["text", "constant.language.boolean"],
+        token: ["text", "constant.language.boolean.text"],
         regex: "(\\s+|^)(_(?=[^\\s_])(?:(?:\\\\.)|(?:[^\\s_\\\\]))*?_)\\b"
     };
 
     var strongStars = {
-        token: ["constant.numeric"],
+        token: ["constant.numeric.text"],
         regex: "([*][*]\\S.+?\\S[*][*])"
     };
 
     var emphasisStars = {
-        token: ["constant.language.boolean"],
+        token: ["constant.language.boolean.text"],
         regex: "([*](?=[^\\s*])(?:(?:\\\\.)|(?:[^\\s*\\\\]))*?[*])"
     };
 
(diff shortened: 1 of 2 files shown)


src/gwt/acesupport/acemode/markdown_highlight_rules.js
```diff
@@ -91,22 +91,22 @@ var MarkdownHighlightRules = function() {
     // handle highlighting for *abc*, _abc_ separately, as pandoc's
     // parser is a bit more strict about where '_' can appear
     var strongUnderscore = {
-        token: ["text", "constant.numeric"],
+        token: ["text", "constant.numeric.text"],
```
adamconroy: I was wondering about the "text" type in the first index of the array here, it doesn't seem to carry over to the spellchecker so I did have to add the ".text" in the second array parameter. If it's ok/preferred to delete the first "text" just let me know.

src/gwt/src/org/rstudio/studio/client/common/filetypes/TextFileType.java
```diff
@@ -450,8 +450,13 @@ public TokenPredicate getTokenPredicate()
          @Override
          public boolean test(Token token, int row, int column)
          {
-            return reTextType_.match(token.getType(), 0) != null
-                  && reNospellType_.match(token.getType(), 0) == null;
+            if (reNospellType_.match(token.getType(), 0) != null) {
+               return false;
+            }
```
adamconroy: I prefer to separate negative and positive invariants when possible, especially when they're so similar as I find that it greatly improves code legibility. If this is strongly discouraged for some reason I'm happy to discuss and revert if necessary.

kevinushey: FWIW I agree 100% (I try to exit early for anything not on the 'main' code path); but I do think some other team members might prefer having only a single point of return in a function (ie, at the end).

We don't have any hard rules here though -- I think whatever keeps the code legible and correct is fine.

src/gwt/acesupport/acemode/markdown_highlight_rules.js
kevinushey: For regular expressions that include captures, Ace will map the types in the array back to the capture (by index) In this case:

```
        token: ["text", "constant.language.boolean.text"],
        regex: "(\\s+|^)(_(?=[^\\s_])(?:(?:\\\\.)|(?:[^\\s_\\\\]))*?_)\\b"
                 ^^^^^^  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                  text    constant.language.boolean.text
```

You'll notice that some of the type names are a bit bogus (why is the text within underscores given a 'boolean' class !?) but this is mostly us just being lazy and using that type because it maps to some CSS style in our existing set of themes that we wanted to apply here.

PR #658, opened 2016-03-28
implement 'quick add next'

kevinushey:
This PR implements 'quick find next' functionality (similar to that of Sublime Text). It allows you to add multiple find targets to the selection, using multiple cursors. I take over the `Meta + D` keybinding for this action (note that this was previously bound to 'delete line', so we may want to reconsider the binding)

Note that I didn't add this to our Find + Replace toolbar yet -- since real estate there is a bit limited, I wasn't sure if we really wanted to stuff it in there; or, if we did we might want to fiddle around to generate some more space.

quick-find

PR #3344, opened 2018-08-20
Sublime Text key bindings

### LINKED ISSUE #3341: sublime text  style keybinds
Although the upstream [ACE editor](https://github.com/ajaxorg/ace/issues/1287) doesn't support sublime mode directly, the [cloud 9 editor](https://github.com/c9/c9.ide.ace.keymaps/tree/39cbe25d2ec8280249840d82a285bc9d82658628) has a plugin to enable Sublime Text mode and the corresponding keybinds are posted [on github](https://github.com/c9/c9.ide.ace.keymaps/blob/39cbe25d2ec8280249840d82a285bc9d82658628/sublime/keymap.js).

It will be great if RStudio is able to make use of it and enable the Sublime Text style keybinds.

randy3k:
I am trying to add Sublime Text key bindings to RStudio and I got some partial success. Just to make sure I am doing things legitimately, I want to get some feedback from RStudio developers before I implement all the features. If they give me a green light, I'd map out the rest of the other keys.

What implemented so far:
- [x] `c+shift+]` and `c+shift+[`: change tab
- [x] `c+shift+l`: split into lines
- [x] `cmd+]` and `cmd+[`: indent and outdent (mac only)
- [x] `ctrl+g`: go to line
- [x] `c+l`: expand selection to line
- [x] `c+j`: join lines
- [x] `c+shift+d`: duplicate line
- [x] `c+d`: add selection and cursor
- [x] `ctrl+shift+m`: expand selection to brackets
- [x] `ctrl+m`: move between bracket pair
- [x] `c+p`: show all files
- [x] `c+shift+r`: show functions
- [x] `ctrl+cmd+up/down` or `ctrl+shift+up`: swap line up/down 
       In mac, a beep sound may occur. Atom has the same issue and it is a known mac issue: https://apple.stackexchange.com/questions/23981/command-control-arrow-beeps-plays-alert-sound-in-lion
- [x] `c+/`: toggle comment
- [x] `ctrl+shift+k`: remove line

screen shot 2018-08-20 at 3 36 27 pm

close #3341

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/command/ShortcutManager.java
+++ b/src/gwt/src/org/rstudio/core/client/command/ShortcutManager.java
@@ -25,6 +25,7 @@ import java.util.Map;
 import java.util.Set;
 
 import org.rstudio.core.client.BrowseCap;
+import org.rstudio.core.client.ElementIds;
 import org.rstudio.core.client.Pair;
 import org.rstudio.core.client.StringUtil;
 import org.rstudio.core.client.command.KeyMap.CommandBinding;
@@ -40,6 +41,7 @@ import org.rstudio.studio.client.events.EditEvent;
 import org.rstudio.studio.client.workbench.addins.AddinsCommandManager;
 import org.rstudio.studio.client.workbench.commands.Commands;
 import org.rstudio.studio.client.workbench.commands.RStudioCommandExecutedFromShortcutEvent;
+import org.rstudio.studio.client.workbench.views.source.editors.text.AceEditor;
 import org.rstudio.studio.client.workbench.views.source.editors.text.ace.AceEditorNative;
 import org.rstudio.studio.client.workbench.views.source.editors.text.ace.AceKeyboardActivityEvent;
 import org.rstudio.studio.client.workbench.views.terminal.xterm.XTermWidget;
@@ -473,6 +475,28 @@ public class ShortcutManager implements NativePreviewHandler,
       
       keyBuffer_.add(keyCombination);
       
+      // we handle `Ctrl+L` for the sublime mode here
+      // because it is conflicted with `consoleClear`
+      if (editorMode_ == KeyboardShortcut.MODE_SUBLIME)
+         {
+            if (keyCombination.getKeyCode() == KeyCodes.KEY_L &&
+                keyCombination.getModifier() == KeyboardShortcut.CTRL)
+         {
+            Element target = Element.as(event.getEventTarget());
+            AceEditor editor = AceEditor.getEditor(target);
+            AceEditorNative nativeEditor = AceEditorNative.getEditor(target);
+            if (editor != null && nativeEditor != null &&
+                  nativeEditor == editor.getWidget().getEditor() &&
+                  editor.getWidget().getElement().getId() ==
+                     ElementIds.getElementId(ElementIds.SOURCE_TEXT_EDITOR))
+            {
+               keyBuffer_.clear();
+               commands_.expandToLine().execute();
+               return true;
+            }
+         }
+      }
+
       // Loop through all active key maps, and attempt to find an active
       // binding. 'pending' is used to indicate whether there are any bindings
       // following the current state of the keybuffer.
--- a/src/gwt/src/org/rstudio/core/client/command/ShortcutManager.java
+++ b/src/gwt/src/org/rstudio/core/client/command/ShortcutManager.java
@@ -300,6 +300,8 @@ public class ShortcutManager implements NativePreviewHandler,
             mode |= KeyboardShortcut.MODE_VIM;
          else if (item == "emacs")
             mode |= KeyboardShortcut.MODE_EMACS;
+         else if (item == "sublime")
+            mode |= KeyboardShortcut.MODE_SUBLIME;
          else
             assert false: "Unrecognized 'disableModes' value '" + item + "'";
       }
--- a/src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml
@@ -1478,6 +1478,30 @@ well as menu structures (for main menu and popup menus).
         menuLabel="Add Cursor Below Current Cursor"
         context="editor"/>
         
+   <cmd id="moveLinesUp"
+        menuLabel="Move Lines Up"
+        context="editor"/>
+      
+   <cmd id="moveLinesDown"
+        menuLabel="Move Lines Down"
+        context="editor"/>
+
+   <cmd id="expandToLine"
+        menuLabel="Expand Selection to Line"
+        context="editor"/>
+
+   <cmd id="copyLinesDown"
+        menuLabel="Copy Lines Down"
+        context="editor"/>
+
+   <cmd id="joinLines"
+        menuLabel="Join Lines"
+        context="editor"/>
+
+   <cmd id="removeLine"
+        menuLabel="Remove Line"
+        context="editor"/>
+
    <cmd id="splitIntoLines"
         menuLabel="Split Into Lines"
         desc="Create a new cursor on each line in current selection"
--- a/src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml
@@ -543,6 +543,9 @@ well as menu structures (for main menu and popup menus).
       <!-- Use spaces for key sequences, e.g. 'Ctrl+X Ctrl+F' -->
       <!-- Separate shortcuts with '|', e.g. 'Ctrl+X Ctrl+F|Cmd+O' -->
       <shortcutgroup name="Source Editor">
+         <shortcut refid="copyDummy" value="Cmd+C"/>
+         <shortcut refid="cutDummy" value="Cmd+X"/>
+         <!-- <shortcut refid="pasteDummy" value="Cmd+V"/> -->
          <shortcut refid="insertChunk" value="Cmd+Alt+I"/>
          <shortcut refid="insertSection" value="Cmd+Shift+R"/>
          <shortcut refid="extractFunction" value="Cmd+Alt+X"/>
--- a/src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml
@@ -577,6 +577,8 @@ well as menu structures (for main menu and popup menus).
          <shortcut refid="expandToLine" value="Meta+L" title="Expand Selection to line" if="org.rstudio.core.client.BrowseCap.isMacintosh()" disableModes="default,vim,emacs" />
          <!-- Ctrl+L to be handled by ShortcutManager -->
          <!-- <shortcut refid="expandToLine" value="Ctrl+L" title="Expand Selection to line" if="!org.rstudio.core.client.BrowseCap.isMacintosh()" disableModes="default,vim,emacs" /> -->
+         <shortcut refid="copyLinesDown" value="Cmd+Shift+D" title="Copy Lines Down" disableModes="default,vim,emacs" />
+         <shortcut refid="joinLines" value="Cmd+J" title="Join Lines" disableModes="default,vim,emacs" />
          <shortcut refid="removeLine" value="Ctrl+Shift+K" title="Delete Line" disableModes="default,vim,emacs" />
          <shortcut refid="yankRegion" value="Ctrl+W" title="Kill Region" disableModes="default,vim,sublime"/>
          <shortcut refid="yankBeforeCursor" value="Ctrl+U" title="Yank Line Up to Cursor" disableModes="vim"/>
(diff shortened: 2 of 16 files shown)


randy3k:
Oh, just notice that there is https://github.com/rstudio/rstudio/pull/658 which implement the `cmd+D/ctrl/D` things.

src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml
```diff
@@ -547,7 +547,10 @@ well as menu structures (for main menu and popup menus).
          <shortcut refid="extractFunction" value="Cmd+Alt+X"/>
          <shortcut refid="extractLocalVariable" value="Cmd+Alt+V"/>
          <shortcut refid="commentUncomment" value="Cmd+Shift+C"/>
-         <shortcut refid="reindent" value="Cmd+I"/>
+         <shortcut refid="blockOutdent" value="Meta+[" if="org.rstudio.core.client.BrowseCap.isMacintosh()" disableModes="default,vim,emacs"/>
+         <shortcut refid="blockIndent" value="Meta+]" if="org.rstudio.core.client.BrowseCap.isMacintosh()" disableModes="default,vim,emacs"/>
+         <shortcut refid="reindent" value="Cmd+I" disableModes="default,vim"/>
+         <shortcut refid="reindent" value="Cmd+I" disableModes="default,vim"/>
```
kevinushey: Looks like this command entry was duplicated. Also not sure why we're disabling in default + vim modes?

randy3k: I am sorry. I was testing things...will fix it.

kevinushey:
Everything looks great so far! We appreciate your taking the time to make this PR.

I think the model you're following is exactly correct. If I understand correctly, the file at https://github.com/c9/c9.ide.ace.keymaps/blob/39cbe25d2ec8280249840d82a285bc9d82658628/sublime/keymap.js simply remaps existing Ace commands to keybindings understood by Sublime Text; that is, we don't need to bring in a supporting `sublime.js` keybinding file. Does that sound correct?

If so, I think the approach you're following here is indeed right: register commands in `command.xml`, and make sure the appropriate handlers are added to the `AceEditor` class as required.

kevinushey:
One thing I should caution: we're getting fairly close to the v1.2 release, and so we're pretty risk averse right now in terms of taking PRs implementing new features beyond what we've already scoped time for. There's a chance that we won't feel that we can take this in time for v1.2, but given that this PR doesn't (at least yet) introduce any new commands it is relatively lower risk.

randy3k:
You are correct, what cloud 9 does is simply remapping the sublime text keys to Ace commands. I am glad that my approach is sound. 

Speaking of the time frame, it will be (of course) up to the developement team to decide the when to merge it. Feel free to merge it whenever you like (once it is already).

randy3k:
@kevinushey 

When there is no selection, `cmd+c` (`cmd+x`)  in Sublime Text will copy (cut) the whole line, do you feel like this behavior also apply to the default mode?

It also relates to https://github.com/rstudio/rstudio/issues/2409 which is about pasting.

kevinushey:
I'm not sure if we would want to make it the default behavior -- while I like the behavior personally, we tend to be conservative about changing any long-standing behavior in RStudio.

Perhaps for now we should just have the behavior be specific to sublime, and then perhaps turn it on for other modes or even provide a UI preference to toggle it on / off for users as requested.

randy3k:
There is a sublime key `ctrl+l` which expands to line, which conflicts with the clear console key `ctrl+l`. So, the command `consoleClear` has to be [disabled](https://github.com/rstudio/rstudio/pull/3344/commits/dfa012676733b06c36c1d60b529f343b675b7de3) in order to make "expand to line" works. Of course, now, the `ctrl+l` doesn't clear the console even if the focus is in the console.

Is there a way to specify a keybind which is only activated in a particular pane, e.g., the Editor, but not the console?

If there is no way to do it, should I write a function `doCtrlL` which dispatchs different code based on the focus?

randy3k:
Similarly, 
```
<shortcut refid="copyDummy" value="Cmd+C"/>
```
also overrides the `Ctrl+C` in terminal to interrupt process.

randy3k:
> Use of Cut / Copy / Paste menu items with RStudio Server has always shown that dialog as web applications indeed cannot access the user's clipboard arbitrarily in a web browser.

Given that, I guess it is incorrect for me to patch [`copyDummy`](https://github.com/rstudio/rstudio/pull/3344/files#diff-66f0a10a75c960b154a80cbfb411fe79R252) like what that to support copying the whole line if no selection.

randy3k:
The set of keybindings implemented is a bit opinionated. Please let me know if there are any missing keybindings which you miss from Sublime Text.

@kevinushey It is now ready for review. I am not pursuing the "copying whole line" feature in this PR because it seems to involve a lot of works.

src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml
```diff
@@ -543,10 +543,14 @@ well as menu structures (for main menu and popup menus).
       <!-- Separate shortcuts with '|', e.g. 'Ctrl+X Ctrl+F|Cmd+O' -->
       <shortcutgroup name="Source Editor">
          <shortcut refid="insertChunk" value="Cmd+Alt+I"/>
-         <shortcut refid="insertSection" value="Cmd+Shift+R"/>
+         <shortcut refid="insertSection" value="Cmd+Shift+R" disableModes="sublime"/>
+         <shortcut refid="insertSection" value="Ctrl+Shift+R" if="org.rstudio.core.client.BrowseCap.isMacintosh()"/>
```
kevinushey: Does <kbd>Cmd + Shift + R</kbd> still work for inserting a section on macOS when Sublime keybindings are disabled?

kevinushey: Sorry, I'm guessing it should work given the command binding immediately above.

src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml
```diff
@@ -562,55 +566,76 @@ well as menu structures (for main menu and popup menus).
...
          <shortcut refid="findPrevious" value="Cmd+Shift+G" if="!org.rstudio.core.client.BrowseCap.isWindows()"/>
          <shortcut refid="findPrevious" value="Shift+F3" if="org.rstudio.core.client.BrowseCap.isWindows()"/>
          <shortcut refid="replaceAndFind" value="Cmd+Shift+J"/>
          <shortcut refid="goToFileFunction" value="Ctrl+."/>
-         <shortcut refid="goToLine" value="Alt+G" disableModes="default,vim"/>
+         <shortcut refid="goToFileFunction" value="Cmd+P" disableModes="default,vim,emacs"/>
+         <shortcut refid="goToFileFunction" value="Meta+Shift+R" if="org.rstudio.core.client.BrowseCap.isMacintosh()" disableModes="default,vim,emacs"/>
+         <shortcut refid="goToFileFunction" value="Ctrl+Shift+R" if="!org.rstudio.core.client.BrowseCap.isMacintosh()" disableModes="default,vim,emacs"/>
+         <shortcut refid="goToLine" value="Alt+G" disableModes="default,vim,sublime"/>
+         <shortcut refid="goToLine" value="Alt+G" disableModes="default,vim,sublime"/>
```
kevinushey: Looks like this command binding was duplicated.

src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml
```diff
@@ -753,11 +785,10 @@ well as menu structures (for main menu and popup menus).
          <shortcut refid="toggleFullScreen" value="Ctrl+Meta+F" if="org.rstudio.core.client.BrowseCap.isMacintoshDesktop()"/>
          <shortcut refid="showOptions" value="Meta+," if="org.rstudio.core.client.BrowseCap.isMacintosh()"/>
          <shortcut refid="projectOptions" value="Shift+Meta+," if="org.rstudio.core.client.BrowseCap.isMacintosh()"/> 
-         <shortcut refid="goToHelp" value="Ctrl+C Ctrl+V" disableModes="default,vim"/>
+         <shortcut refid="goToHelp" value="Ctrl+C Ctrl+V" disableModes="default,vim,sublime"/>
       </shortcutgroup>
       <shortcutgroup name="Console">
          <shortcut refid="consoleClear" value="Ctrl+L" disableModes="emacs"/>
-         <shortcut value="Cmd+Up" title="Popup Command History"/>
```
kevinushey: Why was this command removed?

randy3k: my bad, will revert it

src/gwt/src/org/rstudio/core/client/command/ShortcutManager.java
```diff
@@ -471,6 +475,28 @@ public void execute()
       
       keyBuffer_.add(keyCombination);
       
+      // we handle `Ctrl+L` for the sublime mode here
+      // because it is conflicted with `consoleClear`
+      if (editorMode_ == KeyboardShortcut.MODE_SUBLIME)
+         {
+            if (keyCombination.getKeyCode() == KeyCodes.KEY_L &&
+                keyCombination.getModifier() == KeyboardShortcut.CTRL)
```
jmcphers: Make sure you set `rebindable=false` on this command in `Commands.cmd.xml` since this binding is hard-coded (i.e. no matter what users change the binding to we'll always expand the selection here)

ateucher:
I hope you don't mind me piping in here - Regarding `cmd+D/ctrl+D` (Quick Add Next), Sublime also has `cmd+K, cmd+D/ctrl+K, ctrl+D` ([Quick Skip Next](https://www.sublimetext.com/docs/3/multiple_selection_with_the_keyboard.html)), which I find really handy and I miss in RStudio.

src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml
```diff
@@ -562,55 +566,75 @@ well as menu structures (for main menu and popup menus).
...
+         <shortcut refid="joinLines" value="Cmd+J" title="Join Lines" disableModes="default,vim,emacs" />
+         <shortcut refid="removeLine" value="Ctrl+Shift+K" title="Delete Line" disableModes="default,vim,emacs" />
+         <shortcut refid="yankRegion" value="Ctrl+W" title="Kill Region" disableModes="default,vim,sublime"/>
          <shortcut refid="yankBeforeCursor" value="Ctrl+U" title="Yank Line Up to Cursor" disableModes="vim"/>
          <shortcut refid="yankAfterCursor" value="Ctrl+K" title="Yank Line After Cursor" />
          <shortcut refid="pasteLastYank" value="Ctrl+Y" title="Insert Yanked Text" />
          <shortcut value="Ctrl+T" title="Transpose Letters" if="org.rstudio.core.client.BrowseCap.isMacintosh()"/>
          <shortcut refid="insertAssignmentOperator" value="Alt+-" title="Insert Assignment Operator"/>
-         <shortcut refid="insertPipeOperator" value="Cmd+Shift+M" title="Insert Pipe Operator" />
+         <shortcut refid="insertPipeOperator" value="Meta+Shift+M" title="Insert Pipe Operator" />
```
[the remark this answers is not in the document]
randy3k: this implementation allows macOS users to insert pipe operator with `ctrl+shift+m`

src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.java
```diff
@@ -140,6 +140,12 @@
    public abstract AppCommand expandToMatching();
    public abstract AppCommand addCursorAbove();
    public abstract AppCommand addCursorBelow();
+   public abstract AppCommand moveLinesUp();
```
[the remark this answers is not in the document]
randy3k: done

src/gwt/src/org/rstudio/core/client/command/ShortcutManager.java
randy3k: But it is only activated for the sublime mode, how can I only apply `rebindable` to sublime mode?

src/gwt/src/org/rstudio/core/client/command/ShortcutManager.java
```diff
@@ -471,6 +475,28 @@ public void execute()
...
+         {
+            Element target = Element.as(event.getEventTarget());
```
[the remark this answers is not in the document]
randy3k: We need this because otherwise https://github.com/rstudio/rstudio/blob/5a7c762642ba4f267af29764f27697d9b779d35b/src/gwt/src/org/rstudio/core/client/command/ShortcutManager.java#L515-L532 will steal the `ctrl+L` key combination.

jmcphers:
Yes, you'll need this custom logic to bind <kbd>Ctrl</kbd>+<kbd>L</kbd> specifically for Sublime mode.

All looks well -- very much appreciate the contribution!

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTarget.java
```diff
@@ -3283,6 +3283,42 @@ private boolean isSelectionValueEmpty(String selection)
       return selection == null || selection.trim().length() == 0;
    }
 
+   @Handler
```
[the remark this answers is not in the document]
randy3k: But it doesn't seem that these commands are useful in the console, maybe with exception of `onBlockIndent/onBlockOutdent`?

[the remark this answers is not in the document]
kevinushey: Still, it's preferable to have the commands implemented there so that they work with all Ace instances (not just those that happen to live in the Source pane)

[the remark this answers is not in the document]
randy3k: Should I also move `onReindent`, `onSplitIntoLines` etc while working on this PR?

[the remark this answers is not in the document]
kevinushey: I think that's worth doing.

randy3k:
I have moved the handlers to ace editor and rebased the commits to current master.

kevinushey:
Thanks!

PR #3350, opened 2018-08-21
improve stan editor mode

kevinushey:
This PR brings a few enhancements to the Stan editor mode in RStudio.

### Autocompletion

We now use our own native autocompletion dialog for autocompletion, including our regular 'fuzzy matching' logic.

screen shot 2018-08-21 at 10 49 44 am

### Diagnostics

On save, we run the Stan compiler and extract error messages to display in the IDE.

screen shot 2018-08-21 at 10 50 45 am

### Snippets

... were actually already a feature of our Stan editor mode, but the use of our own autocompletion popup should unify the experience with e.g. R.

screen shot 2018-08-21 at 11 02 27 am

---

This PR also attempts to refactor some of the 'guts' of the R + Python completion managers, into a base class which in theory does the majority of the language-agnostic work, and allows subclasses implement the required behaviors for a specific language.

Because we're getting close to the v1.2 feature-complete window, I opted to perform this refactoring _without_ re-writing the R + Python completion managers to use the new base class; this is something we can and should do in the future but there is some regression potential so I think it's not worth doing right now.

One huge gain is that this PR should make it significantly easier to add new completion managers (and diagnostics) for new languages down the line.


jjallaire:
@kevinushey This looks great! Are there other Stan items we are planning on picking up for the release? Maybe we should talk realtime on Monday re: what features seem like good candidates.

kevinushey:
Sounds good! I didn't have anything else specifically in mind so it'd be good to enumerate some other Stan features we could try to support.

kevinushey:
I'm going to merge this and iterate more features in new PRs.

PR #3355, opened 2018-08-23
AppleScript support

### LINKED ISSUE #3354: RStudio no longer supports AppleScript
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.2.907
    OS Version      : 10.13.4
    R Version       : 3.5.1

The applescript no longer works
```
tell application "RStudio"
    cmd "a=1"
end tell
```

In https://github.com/rstudio/rstudio/pull/527, I have submitted a PR to add support of AppleScript. It was applied to the mac specific app in `desktop-mac`. In RStudio 1.2, it seems that the development has moved back to the cross-platform code located in `desktop`. However, the code related of AppleScript was not migrated to the `desktop` app.

randy3k:
Not the most beautiful way to do it, but it works.

close #3354

### DIFF
--- a/src/cpp/desktop/DesktopMainWindow.hpp
+++ b/src/cpp/desktop/DesktopMainWindow.hpp
@@ -40,11 +40,7 @@ class MainWindow : public GwtWindow
 public:
    explicit MainWindow(QUrl url=QUrl());
 
-private:
-   static MainWindow *instance;
-
 public:
-   static MainWindow *getInstance();
    QString getSumatraPdfExePath();
    void launchSession(bool reload);
    void launchRStudio(const std::vector<std::string>& args = std::vector<std::string>(),
--- a/src/cpp/desktop/DesktopMainWindow.hpp
+++ b/src/cpp/desktop/DesktopMainWindow.hpp
@@ -40,7 +40,11 @@ class MainWindow : public GwtWindow
 public:
    explicit MainWindow(QUrl url=QUrl());
 
+private:
+   static MainWindow *instance;
+
 public:
+   static MainWindow *getInstance();
    QString getSumatraPdfExePath();
    void launchSession(bool reload);
    void launchRStudio(const std::vector<std::string>& args = std::vector<std::string>(),
--- a/src/cpp/desktop/ScriptCommand.h
+++ b/src/cpp/desktop/ScriptCommand.h
@@ -1,0 +1,22 @@
+/*
+ * ScriptCommand.h
+ *
+ * Copyright (C) 2009-17 by RStudio, Inc.
+ *
+ * Unless you have received this program directly from RStudio pursuant
+ * to the terms of a commercial license agreement with RStudio, then
+ * this program is licensed to you under the terms of version 3 of the
+ * GNU Affero General Public License. This program is distributed WITHOUT
+ * ANY EXPRESS OR IMPLIED WARRANTY, INCLUDING THOSE OF NON-INFRINGEMENT,
+ * MERCHANTABILITY OR FITNESS FOR A PARTICULAR PURPOSE. Please refer to the
+ * AGPL (http://www.gnu.org/licenses/agpl-3.0.txt) for more details.
+ *
+ */
+
+#import <Foundation/Foundation.h>
+
+@interface evaluateRScriptCommand : NSScriptCommand
+
+-(id) performDefaultImplementation;
+
+@end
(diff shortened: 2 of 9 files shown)


src/cpp/desktop/DesktopMainWindow.hpp
```diff
@@ -40,7 +40,11 @@ class MainWindow : public GwtWindow
 public:
    explicit MainWindow(QUrl url=QUrl());
 
+private:
+   static MainWindow *instance;
+
 public:
+   static MainWindow *getInstance();
```
kevinushey: I'm not a fan of this. Why not have `ScriptCommand.mm` accept a `MainWindow` as part of its constructor? Or if that's not possible for some reason, have a method that it can call to 'bind' to a window instance.

src/cpp/desktop/ScriptCommand.h
```diff
@@ -0,0 +1,22 @@
+/*
+ * ScriptCommand.h
+ *
+ * Copyright (C) 2009-17 by RStudio, Inc.
```
kevinushey: nit: 2009-18

src/cpp/desktop/ScriptCommand.mm
```diff
@@ -0,0 +1,55 @@
...
+ *
+ */
+
+#import <Foundation/Foundation.h>
+#import "ScriptCommand.h"
+#include "DesktopMainWindow.hpp"
+#include <boost/algorithm/string/replace.hpp>
+
+
+std::string jsEscape(std::string str)
```
kevinushey: We have a couple escape helpers here: do any of them suffice?

https://github.com/rstudio/rstudio/blob/f5ba8af725c845a1097277ad3f812103b8ab090f/src/cpp/core/include/core/StringUtils.hpp#L91-L95

src/cpp/desktop/ScriptCommand.mm
```diff
@@ -0,0 +1,55 @@
...
+   return str;
+}
+
+
+// evaluate R command
+void evaluateRCommand(std::string rCmd)
+{
+   rCmd = jsEscape(rCmd);
+   std::string js = "window.desktopHooks.evaluateRCmd(\"" + rCmd + "\")";
+   rstudio::desktop::MainWindow::getInstance()->webPage()->runJavaScript(QString::fromUtf8(js.c_str()));
```
kevinushey: nit: there is a `QString::fromStdString()` method that you might consider using here

src/cpp/desktop/DesktopMainWindow.hpp
randy3k: I am not sure if it is possible to dynamically initialize `NSScriptCommand`, akaik, it is usually initialized automatically by the application. This implementation just mirrors `desktop-mac`'s: https://github.com/rstudio/rstudio/blob/54caf1f66b724680091623b6eca157870e3a3e6c/src/cpp/desktop-mac/MainFrameController.mm#L37

kevinushey: Perhaps the `evaluateRScriptCommand` class could gain a function e.g. `- initialize(MainWindow* window)` that gets called by the `MainWindow` class when appropriate? And that could take care of setting the `MainWindow` instance so that it can be used.

kevinushey:
Great!

Just to double-check: does this work as expected when there are multiple RStudio windows open (e.g. if you have multiple main windows; perhaps with different projects open)?

randy3k:
it will always send to the newest instance. In order to support multiple instances, we will need communications between the sessions. I may be able to work on it in the future if you could point me to the correct files to look at.

kevinushey:
I think it's okay if we just send to the newest RStudio instance. You might even want to set things up so that commands are sent to the most recently focused RStudio window.

randy3k:
>I think it's okay if we just send to the newest RStudio instance. You might even want to set things up so that commands are sent to the most recently focused RStudio window.

@kevinushey unfortunately it is not possible without a session manager. Ideally, there should be only one instance and a new session is opened in the same instance, like how Safari works.

randy3k:
@kevinushey I hope the RStudio team would consider merging this into v1.2 release as AppleScript was already supported v1.1.

kevinushey:
Yes, we will take this for v1.2. Thanks again for the PR!

randy3k:
Awesome.

PR #3365, opened 2018-08-24
Adaptive macOS Title Bar Color

randy3k:
Demo:
aug-24-2018 02-53-08

[Wombat](https://github.com/randy3k/dotfiles/blob/master/.R/rstudio/themes/Wombat.rstheme) is a customized theme with a different dark background other than the default blue.

### DIFF
--- a/src/cpp/desktop/DesktopGwtCallbackMac.mm
+++ b/src/cpp/desktop/DesktopGwtCallbackMac.mm
@@ -23,6 +23,7 @@
 #import <AppKit/NSSavePanel.h>
 #import <AppKit/NSButton.h>
 #import <AppKit/NSImage.h>
+#import <Cocoa/Cocoa.h>
 
 #include <core/FilePath.hpp>
 #include <core/Macros.hpp>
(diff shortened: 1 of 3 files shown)


src/cpp/desktop/DesktopGwtCallbackMac.mm
```diff
@@ -460,7 +461,26 @@ bool showOfficeDoc(NSString* path, NSString* appName, NSString* formatString)
    }
 }
 
+void GwtCallback::changeTitleBarColor(int red, int green, int blue) {
+    WId winId = pOwner_->asWidget()->effectiveWinId();
+    if (winId == 0) return;
+    NSView* view = (NSView*)winId;
+    NSWindow* window = [view window];
+    if (pow(red, 2.) + pow(green, 2.) + pow(blue, 2.) < 48387) {
```
kevinushey: Where does the magic number 48387 come from? (Just as a curiosity; not necessarily requiring any change in the PR)

src/cpp/desktop/DesktopGwtCallbackMac.mm
```diff
@@ -460,7 +461,26 @@ bool showOfficeDoc(NSString* path, NSString* appName, NSString* formatString)
...
+        window.appearance = [NSAppearance appearanceNamed: NSAppearanceNameVibrantDark];
+    } else {
+        window.appearance = [NSAppearance appearanceNamed: NSAppearanceNameVibrantLight];
+    }
+    if (pow(red, 2.) + pow(green, 2.) + pow(blue, 2.) > 150000) {
```
kevinushey: Same for 150000?

kevinushey:
This is awesome! 😄  Thanks for the PR!

src/cpp/desktop/DesktopGwtCallbackMac.mm
randy3k: 127^2 *3  The sum of average color!?

src/cpp/desktop/DesktopGwtCallbackMac.mm
randy3k: this is just a random number which works.

src/cpp/desktop/DesktopGwtCallbackMac.mm
kevinushey: For clarity, it would be worth encoding this as a constant, e.g. `const int kAverageColor = pow(127.0, 2.0) * 3`.

randy3k: Definitely. It may be easier for you to commit.

kevinushey:
I tried playing with this a bit but I'm seeing something odd: the text color in the title bar changes when I move the window around. Initially, when the window is focused, the color is closer to white; if I move the window around or interact with the window a bit; the text changes to a more grey color.

Are you by any chance seeing the same thing? Any idea what might be going on?

randy3k:
Ya, I could reproduce it. It seems that the titlebar reacts like as if the window is unfocused if the window is moved. Let me see if there are any similar bug reports.

kevinushey:
It seems really finicky since even trying to take a screenshot of the lighter-colored text causes that color to be lost; if I use the Digital Color Meter to try and look at the color it eventually snaps from light to grey as well (even without actually interacting with the window directly).

randy3k:
What's even weird is when I do `Cmd + Shift + 4` on Mac to capture the screen and select half of the word "RStudio". Only the part when is captured turns to grey after capturing...

randy3k:
I got the idea from here: https://forums.xamarin.com/discussion/99677/nswindow-titlebarappearstransparent-titlebar-corner-does-not-change-color
For reasons which I still don't understand, we need to set `wantsLayer` property of `NSView`.

I tested it and it works. Will submit an PR for it together with the `kAverageColor` thing.

PR #3369, opened 2018-08-24
set wantsLayer properpty of NSView

randy3k:
Fix a bug spotted by @kevinushey 
> I tried playing with this a bit but I'm seeing something odd: the text color in the title bar changes when I move the window around. Initially, when the window is focused, the color is closer to white; if I move the window around or interact with the window a bit; the text changes to a more grey color.

Solution inspired by https://forums.xamarin.com/discussion/99677/nswindow-titlebarappearstransparent-titlebar-corner-does-not-change-color

Additionally, use white background instead of `clearColor` because otherwise, the title bar would become transparent.

### DIFF
--- a/src/cpp/desktop/DesktopGwtCallbackMac.mm
+++ b/src/cpp/desktop/DesktopGwtCallbackMac.mm
@@ -462,19 +462,25 @@ void GwtCallback::showPptPresentation(QString qPath)
 }
 
 void GwtCallback::changeTitleBarColor(int red, int green, int blue) {
+    const int kAverageColor = pow(127.0, 2.0) * 3;
+    const int kBrightColor = pow(217.0, 2.0) * 3;
+
     WId winId = pOwner_->asWidget()->effectiveWinId();
     if (winId == 0) return;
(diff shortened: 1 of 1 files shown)


kevinushey:
Thanks!

hrbrmstr:
@randy3k The next time I go to visit offspring #3 (he's at U Maine) you're definitely getting a thank you in person, too!

randy3k:
Thank you for your word. By the way, why did you end up at this commit?

hrbrmstr:
https://kevinushey.github.io/blog/2019/03/09/the-macos-rendering-bug/

randy3k:
Thanks for the link. Otherwise I would never know that my commits was causing trouble.

kevinushey:
@randy3k sorry, this definitely wasn't an attempt to call you out or anything like that! Like I said; I'm the one who merged the PR, and even though with the benefit of hindsight we might've been able to point at that code as a potential culprit, I don't think anyone would've expected it to cause an issue like this.

randy3k:
No worries. I have also learned something from it.

PR #3383, opened 2018-08-28
Enable editing of GitLab CI files in RStudio by not hiding them from the file pane

hrbrmstr:
GitHub seems to be a first-class citizen, relegating GitLab into the void. This PR adds an `else` section to at least enable selecting and editing of gitlab CI yaml files without having to go the route of opening a file dialog manually, un-hiding dot files and selecting the gitlab CI yml file.

### DIFF
--- a/src/cpp/session/SessionModuleContext.cpp
+++ b/src/cpp/session/SessionModuleContext.cpp
@@ -1805,6 +1805,10 @@ bool fileListingFilter(const core::FileInfo& fileInfo)
    {
       return true;
    }
+   else if (name == ".gitlab-ci.yml")
+   {
+      return true;
+   }
    else if (userSettings().hideObjectFiles() &&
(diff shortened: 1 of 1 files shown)


kevinushey:
LGTM -- thanks for the PR!

(we really need to make the file filter here user-configurable in some way ...)

kevinushey:
@hrbrmstr: do we already have a [CLA](https://www.rstudio.com/wp-content/uploads/2014/06/RStudioIndividualContributorAgreement.pdf) from you?

hrbrmstr:
Hrm. I think I filled one out ages ago but never got a confirmation. I'll do another one now.

Agreed on having it be user-configurable in some way. Perhaps a config file under `~/.R/rstudio/` (vs having to add another edit pane that needs UX automated testing).

hrbrmstr:
just sent it to jj

kevinushey:
Thanks for the PR!

PR #3396, opened 2018-08-29
allow interrupt of sub-processes launched by R

kevinushey:
Currently, in RStudio, it is not possible to interrupt R when a sub-process is being executed, e.g. via a `system()` call. This is because the R session receives the interrupt RPC request, enqueues it, and then only processes it later when R is ready to process events -- and this doesn't occur until the sub-process has finished executing. Furthermore, for interrupts, we only set the interrupt flag on the R session itself, whereas in cases where R is run in a terminal, pressing <kbd>Ctrl + C</kbd> would signal all processes in that (terminal) process group.

This PR alleviates this issue by responding to the interrupt RPC requests directly, rather than enqueueing them. To make sure that the interrupt is properly handled and propagated, we use two different solutions:

1) On POSIX systems, we use `::killpg()` to send the interrupt to all processes in the `rsession`'s process group. We register a signal handler for `SIGINT` on the `rsession` process, and we simply set the interrupts pending flag when it does receive the interrupt.

2) On Windows systems, we accomplish the same with the use of a `CTRL_C_EVENT` console event + a console control handler.

There's an open question about whether we really want to signal _all_ processes in the process group. Sometimes this is too broad: for example, we might have a daemonized process in the same process group that we would want to survive an interrupt. However, some daemons we _will_ want to interrupt: for example, the slave processes launched by the parallel package. I'm also not sure whether we can control the granularity for interrupt signaling on Windows -- everything I've read thus far indicates that the whole process group receives the signal. That said, in each case the daemonized process itself could decide what it wants to do when it receives an interrupt, so we could argue that such daemons should simply implement an appropriate interrupt handler.

This PR deserves extra scrutiny so all feedback is appreciated!

### DIFF
--- a/src/cpp/desktop/DesktopSessionLauncher.cpp
+++ b/src/cpp/desktop/DesktopSessionLauncher.cpp
@@ -51,13 +51,21 @@ void launchProcess(const std::string& absPath,
                    const QStringList& argList,
                    QProcess** ppProc)
 {
-   auto* pProcess = new QProcess();
+   QProcess* process = new QProcess();
+   process->setProgram(QString::fromStdString(absPath));
+   process->setArguments(argList);
    if (options().runDiagnostics())
-      pProcess->setProcessChannelMode(QProcess::ForwardedChannels);
-   else
-      pProcess->setProcessChannelMode(QProcess::SeparateChannels);
-   pProcess->start(QString::fromUtf8(absPath.c_str()), argList);
-   *ppProc = pProcess;
+      process->setProcessChannelMode(QProcess::ForwardedChannels);
+
+#ifdef _WIN32
+   process->setCreateProcessArgumentsModifier([](QProcess::CreateProcessArguments* cpa)
+   {
+      cpa->flags |= CREATE_NEW_PROCESS_GROUP;
+   });
+#endif
+
+   process->start();
+   *ppProc = process;
 }
 
 FilePath abendLogPath()
--- a/src/cpp/core/CMakeLists.txt
+++ b/src/cpp/core/CMakeLists.txt
@@ -240,6 +240,7 @@ if (UNIX)
       system/PosixChildProcessTracker.cpp
       system/PosixEnvironment.cpp
       system/PosixFileScanner.cpp
+      system/PosixInterrupts.cpp
       system/PosixLibraryLoader.cpp
       system/PosixNfs.cpp
       system/PosixParentProcessMonitor.cpp
@@ -288,6 +289,7 @@ else()
       system/Win32FileScanner.cpp
       system/RegistryKey.cpp
       system/Win32Environment.cpp
+      system/Win32Interrupts.cpp
       system/Win32LibraryLoader.cpp
       system/Win32ParentProcessMonitor.cpp
       system/Win32Pty.cpp
--- a/src/cpp/core/include/core/system/Interrupts.hpp
+++ b/src/cpp/core/include/core/system/Interrupts.hpp
@@ -1,0 +1,29 @@
+/*
+ * Interrupts.hpp
+ *
+ * Copyright (C) 2009-18 by RStudio, Inc.
+ *
+ * Unless you have received this program directly from RStudio pursuant
+ * to the terms of a commercial license agreement with RStudio, then
+ * this program is licensed to you under the terms of version 3 of the
+ * GNU Affero General Public License. This program is distributed WITHOUT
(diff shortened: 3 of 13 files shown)


kevinushey:
We could also consider something a bit more direct: we might iterate over all processes in the process group, and only kill processes that:

1. Are direct children of the `rsession` process,
2. Are PSOCK 'daemon' processes.

On Windows, we might be able to accomplish this by sending a `WM_CLOSE` message to the child R sessions (or even `::TerminateProcess()`, but that's usually a last resort IIUC)

gtritchie:
Given terminal shells run as children of the rsession, will they react to the interrupt as well? If there's something running in a terminal and you hit "Stop" in the console, is that going to kill the terminal or at least interrupt anything running in the terminal?

kevinushey:
Fortunately, I believe because terminal processes request a pty and so get their own session + process group, they don't receive these interrupts (at least on POSIX).

Will have to double-check this is true for non-interactive console processes + the processes we launch on Windows, though.

gtritchie:
When you try this out on Windows, keep an eye on winpty-agent.exe; since Windows doesn't have the concept of pty's we launch this helper process which creates a hidden console windows and screen-scrapes them to mimic a pty. Whee!!

src/cpp/desktop/DesktopSessionLauncher.cpp
```diff
@@ -51,13 +51,21 @@ void launchProcess(const std::string& absPath,
...
-   else
-      pProcess->setProcessChannelMode(QProcess::SeparateChannels);
-   pProcess->start(QString::fromUtf8(absPath.c_str()), argList);
-   *ppProc = pProcess;
+      process->setProcessChannelMode(QProcess::ForwardedChannels);
+
+#ifdef _WIN32
+   process->setCreateProcessArgumentsModifier([](QProcess::CreateProcessArguments* cpa)
+   {
+      cpa->flags |= CREATE_NEW_PROCESS_GROUP;
```
jjallaire: We should figure out how this interacts with the Jobs functionality we already have (which is used to ensure that when the rdesktop app is terminated the rsession and it's children go along with it). If this ends up "breaking out" of the job then it will lead to many undesirable orphaned processes.

jjallaire:
I think it's definitely debatable whether we'd want to kill all children in the process group vs. doing some sort of enumeration with filtering. We have some longer running children (like git tasks, or completion related R sessions) but it's not clear to me that we really need to keep them running. I think it would be worth looking at all the child processes we create and ensuring that it's okay for them to die.

src/cpp/desktop/DesktopSessionLauncher.cpp
jjallaire: In doing some cursory scanning of documentation it looks like these mechanisms might be orthogonal so there is no issue with job objects. Here's where we create the job object: https://github.com/rstudio/rstudio/blob/8af730409bb6d651cc8f6816d136bea91441e7a4/src/cpp/core/system/Win32System.cpp#L71

It should be easy to test whether this is a problem by just using task manager to kill rdesktop and seeing whether rsession is also automatically terminated (may also want to do the same with rsession children).

jjallaire: Some related SO discussion: https://stackoverflow.com/questions/1798771/send-ctrlc-to-subprocess-tree-on-windows

kfeinauer: I added the `core::system::getChildProcesses` method which walks the child processes for a given pid (regardless of their group ids), and there is an `rsession.conf` option `session-quit-child-processes-on-exit` which should kill all children on exit via `core::system::terminateChildProcesses`.

kevinushey: @kfeinauer does this work for background processes? (E.g. if someone has run `system("sleep 100", wait = FALSE)` from R -- does that process also get cleaned up on exit?)

kfeinauer: Yes, it should. The methods scan the child processes recursively for a given pid, so even if it changes its process group it should still get killed.

kevinushey:
I'm going to merge this PR now just so we can get some more early exposure (I'd like to hear ASAP if we find out it breaks any particular workflows).

src/cpp/desktop/DesktopSessionLauncher.cpp
kevinushey: My point is that such processes will now have the `init` process as the parent, even though they still live in the same process group.

PR #3401, opened 2018-08-30
Use a shinyapp in connection pane without having sparklyr on the system

### LINKED ISSUE #3400: Using shiny application for advanced connection interfaces always required sparklyr to be installed.
### System details

    RStudio Edition : Server Pro
    RStudio Version : 1.1.453
    OS Version      :  Redhat 7.4
    R Version       :  3.4.3

About connection extension and contract, as mentioned in the [https://rstudio.github.io/rstudio-extensions/](https://rstudio.github.io/rstudio-extensions/rstudio-connections.html#shiny-application) and following the example of [sparklyr](https://github.com/rstudio/sparklyr/blob/master/inst/rstudio/connections.dcf), it is possible to create a shiny app and use `ShinyApp` field in `inst/rstudio/connection.dcf` in a 📦, I believe even in a package that don't need sparklyr.
```yaml
Name: MyApp
ShinyApp: lauch_shinyapp
Icon: icons/myicon.png
```

Then when your  📦 is installed (package that do not depend on sparklyr), you go in the connection pane, see you custom connection entry, but when you click and the shinyapps should load, there is an error.
> Error in packageVersion("sparklyr") : package ‘sparklyr’ not found

The `traceback()`
```r
> traceback()
3: stop(gettextf("package %s not found", sQuote(pkg)), domain = NA)
2: packageVersion("sparklyr")
1: (function (package, name) 
   {
       if (package == "sparklyr" & packageVersion("sparklyr") <= 
           "0.5.4") {
           return(.rs.error("sparklyr ", packageVersion("sparklyr"), 
               " does not support this functionality. ", "Please upgrade to sparklyr 0.5.5 or newer."))
       }
       connectionContext <- .rs.rpc.get_new_connection_context()$connectionsList
       connectionInfo <- Filter(function(e) identical(as.character(e$package), 
           as.character(package)) & identical(as.character(e$name), 
           as.character(name)), connectionContext)
       if (length(connectionInfo) != 1) {
           return(.rs.error("Connection for package ", package, 
               " and name ", name, " is not registered"))
       }
       connectionInfo <- connectionInfo[[1]]
       consoleCommand <- paste("shiny::runGadget(", connectionInfo$newConnection, 
           ", viewer = .rs.embeddedViewer)", sep = "")
       .rs.enqueClientEvent("send_to_console", list(code = .rs.scalar(consoleCommand), 
           execute = .rs.scalar(TRUE), focus = .rs.scalar(FALSE), 
           animate = .rs.scalar(FALSE)))
       .rs.success()
   })("MyPackage", "MyPackage")
```

Apparently, the first function call when clicking in a button in the connection pane > New Connecion is an internal RStudio function

https://github.com/rstudio/rstudio/blob/4694bef8d5594bc46b02c390a5cf7e643e491233/src/cpp/session/modules/SessionConnections.R#L700-L735

And the first thing it does is check that sparklyr is installed on my system, which I don't have. 
If I install it, everything goes as expected. 

Currently, in `if (package == "sparklyr" & packageVersion("sparklyr")`, it runs the second test even if the first is FALSE, and the second testt through an error if sparklyr is not installed. I believe we want to run the content of the `if` only when `sparklyr` is used in connection. 

By using a `&&` instead of the `&`, it should solve the issue as the second test will not run if the first is FALSE.

cderv:
fixes #3400 

Don't apply the second test if the first if FALSE because it throws an error if sparklyr not installed in the environment .

---
This change is [Reviewable](https://reviewable.io/reviews/rstudio/rstudio/3401)

### DIFF
--- a/src/cpp/session/modules/SessionConnections.R
+++ b/src/cpp/session/modules/SessionConnections.R
@@ -699,7 +699,7 @@ options(connectionObserver = list(
 
 .rs.addJsonRpcHandler("launch_embedded_shiny_connection_ui", function(package, name)
 {
-   if (package == "sparklyr" & packageVersion("sparklyr") <= "0.5.4") {
+   if (package == "sparklyr" && packageVersion("sparklyr") <= "0.5.4") {
       return(.rs.error(
          "sparklyr ", packageVersion("sparklyr"), " does not support this functionality. ",
          "Please upgrade to sparklyr 0.5.5 or newer."


kevinushey:
Thank you for the PR! As far as I can tell, we don't yet have a [CLA](https://www.rstudio.com/wp-content/uploads/2014/06/RStudioIndividualContributorAgreement.pdf) from you -- could you please sign a copy and send one over to JJ?

cderv:
Yes I don't think you have one on this project. I think I have done it for yihui projects and contributes also on other packages but not sure if they ask me for a CLA. 
I'll do it next week where I can print and scan back.

cderv:
CLA sent today.

kevinushey:
Awesome! Thank you for the PR.

PR #3411, opened 2018-08-31
add API functions for build tool installation

kevinushey:
See also: https://github.com/rstudio/rstudioapi/pull/110


kevinushey:
We're making some of these hooks available primarily for the `rstan` package. A Stan script defines a model that gets compiled into C++ code, which is itself then compiled into a model object that can be used in the R session.

For `rstan`, a lot of their trouble (especially on Windows) is getting development environments set up (especially on Windows); we're hoping these hooks solve at least part of the problem.

PR #3429, opened 2018-09-05
improve completions for SQL chunks / code

kevinushey:
NOTE: Not quite ready for merge (still needs refinement) but putting up for some discussion.

---

This PR implements a completion engine for SQL code:

screen shot 2018-09-05 at 2 32 48 pm

In particular, we provide completions for:

1. SQL keywords,
2. Tables associated with the current connection,
3. Fields associated with all tables on the current connection.

Note: I'm guessing that (2) and (3) could be quite expensive, so I'm curious if there's a way we could more intelligently populate and / or cache these -- suggestions are welcome. Some thoughts:

1. We could enumerate the context wherein a table name is expected, and only provide the table names there;

2. In other contexts, we could draw field names from the set of tables that have been specified thus far.

### DIFF
--- a/src/cpp/session/CMakeLists.txt
+++ b/src/cpp/session/CMakeLists.txt
@@ -217,6 +217,7 @@ set (SESSION_SOURCE_FILES
    modules/rmarkdown/RMarkdownPresentation.cpp
    modules/rmarkdown/RMarkdownTemplates.cpp
    modules/shiny/SessionShiny.cpp
+   modules/sql/SessionSql.cpp
    modules/stan/SessionStan.cpp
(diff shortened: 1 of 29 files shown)


kevinushey:
Now context-aware:

screen shot 2018-09-05 at 4 34 59 pm

screen shot 2018-09-05 at 4 35 22 pm

jjallaire:
Does this also work within a standalone .sql file?

kevinushey:
Indeed it does :-)

kevinushey:
I'm going to merge this now and iterate a bit more on master, but (time permitting) reviews are appreciated.

PR #3445, opened 2018-09-08
Improvements to Windows bootstrapper and dockerfile

gtritchie:
When our Windows build machine needs to recreate the docker image, it will install the version of MSVC obtained by downloading and running: `https://aka.ms/vs/15/release/vs_buildtools.exe`

The version this installs will change over time as Microsoft revs things, and we currently have some PATH info hardcoded with specific version numbers, breaking our build in the process.

Even if we fixed that, we don't want our official build using different tools as time goes on, without us choosing to do so.

The PR pins specific versions in both the dev machine bootstrapper script (`Install-RStudio-Prereqs.ps1`), and the Windows Dockerfile.

Switched to using chocolatey to install toolchain in both scenarios (there were some prior issues with this in Docker but have been fixed). Also made the PowerShell downloading code more robust, and faster (Invoke-WebRequest reads everything into memory before saving it, and is painfully slow). Also some issues with PowerShell using TLS 1 by default which caused problems on some download sites.

For Docker only, switched back to using an archive of the Qt files pulled from S3, instead of using the interactive installer. I could not get the latter to work in containers running on Server Core (or, they were so slow that waiting two hours was not long enough for this single step). Updated my notes on the Pro wiki for how to update this in the future. Basically, install with the bootstrapper, then create an archive from that.

Tested the following:

- Clean Windows-10 Pro VM and clean Windows Server 2016 VM, used bootstrapper script to install dev environment, built and ran RStudio
- Clean `Windows Server Datacenter Core for Containers 1709` (catchy name) running on Google Compute, able to build the Windows docker image, then build the RStudio setup package in a container; downloaded that to a clean machine and installed RStudio successfully.

Wrote up (hopefully reproducible) details on how I tested the Dockerfile and sent out via Slack.


kevinushey:
Thanks for drudging through this! This sort of work is painful but is absolutely required -- we certainly don't want to risk our toolchain changing out from under our feet and causing any regressions down the line.

PR #3470, opened 2018-09-12
Exposes GWT CSS classes as requested.

### LINKED ISSUE #3368: Add stable CSS class names for themeable elements

MariaSemple:
This PR exposes the GWT CSS classes as requested in #3368. (See https://github.com/rstudio/rstudio/issues/3368#issuecomment-418568294).
Closes #3368.

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/theme/DocTabLayoutPanel.java
+++ b/src/gwt/src/org/rstudio/core/client/theme/DocTabLayoutPanel.java
@@ -1085,7 +1085,7 @@ public class DocTabLayoutPanel
 
          contentPanel_ = new HorizontalPanel();
          contentPanel_.setTitle(tooltip);
-         contentPanel_.setStylePrimaryName(styles_.tabLayoutCenter());
+         contentPanel_.setStylePrimaryName(styles_.rstheme_tabLayoutCenter());
          contentPanel_.setVerticalAlignment(HasVerticalAlignment.ALIGN_MIDDLE);
 
          if (icon != null) {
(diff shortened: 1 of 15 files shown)


kevinushey:
Do we need a mechanism that helps us enforce the fact that these CSS classes should not change (to support custom themes)?

@randy3k would you be willing to confirm that this PR gives you enough for your own dark theme with stable class names?

randy3k:
One more class and all is good.
```
.org-rstudio-core-client-theme-res-ThemeStyles-minimizedWindowObject .org-rstudio-core-client-theme-res-ThemeStyles-center
```
screen shot 2018-09-13 at 12 12 35 am

jmcphers:
I'd suggest two small improvements:

1. We should have some kind of decoration at the selector that indicates that we've exposed it externally and shouldn't change its name without consideration. I don't think it's worth building a system to mechanically enforce this (just a comment at the declaration site is fine)

2. I think that all selectors exposed in this way should have a common prefix (such as `rstheme`) to make them possible to locate mechanically should we need to do so in the future.

randy3k:
I actually need to override some other classes with prefix `gwt-`, not sure if they are stable or not.

jmcphers:
@randy3k Those are stable unless GWT changes them (very unlikely!)

Issue #3477, opened 2018-09-13
Missing PDB warnings when building Debug RStudio IDE on Windows


gtritchie:
When a developer builds a debug configuration of RStudio on Windows a large number of link warnings of the following forms are seen. Build still works, but this generates a lot of noise making it harder to see actual issues during development.

```
libeay32.lib(cryptlib.obj) : warning LNK4099: PDB 'lib.pdb' was not found with 'libeay32.lib(cryptlib.obj)' or at 'C:\Users\gtritchie\rstudio\src\build-cpp-Desktop_Qt_5_11_1_MSVC2017_64bit-Debug\session\lib.pdb'; linking object as if no debug info

ssleay32.lib(t1_ext.obj):-1: warning: LNK4099: PDB 'lib.pdb' was not found with 'ssleay32.lib(t1_ext.obj)' or at 'C:\Users\gtritchie\rstudio\src\build-cpp-Desktop_Qt_5_11_1_MSVC2017_64bit-Debug\desktop\synctex\rsinverse\lib.pdb'; linking object as if no debug info
```

gtritchie:
Fixed via https://github.com/rstudio/rstudio/pull/4545

PR #3480, opened 2018-09-13
rebuild boost to fix warnings with newer MSVC

### LINKED ISSUE #3140: Warnings when building Windows desktop with newer MSVC and Boost
Since the update to Qt 5.11.1 and corresponding update to use MSVC 2017, there are a lot of warnings emitted related to Boost not recognizing the compiler:
```
Unknown compiler version - please run the configure tests and report the results
```
These are harmless but highly annoying and can mask real warnings. Looking at newer Boost builds, they have commented out the `pragma message` that does this with a comment that MSVC builds change too frequently for them to keep up, and that there isn't currently any value to showing this warning. 

Also tracking another warning we get many times building with the newer MSVC / older Boost:

```
warning STL4019: The member std::fpos::seekpos() is non-Standard, and is preserved only for compatibility with workarounds for old versions of Visual C++. It will be removed in a future release, and in this release always returns 0. Please use standards-conforming mechanisms to manipulate fpos, such as conversions to and from streamoff, or an integral type, instead. If you are receiving this message while compiling Boost.IOStreams, a fix has been submitted upstream to make Boost use standards-conforming mechanisms, as it does for other compilers. You can define _SILENCE_FPOS_SEEKPOS_DEPRECATION_WARNING to acknowledge that you have received this warning, or define _REMOVE_FPOS_SEEKPOS to remove std::fpos::seekpos entirely.
```

This was fixed in newer Boost via: https://github.com/boostorg/iostreams/pull/57.

gtritchie:
Modified scripts to patch the two issues that are throwing several hundred warnings during build of IDE on Windows. I chose to make these small tweaks on the current version of Boost we are using, and not upgrading. Safer to upgrade Boost early in 1.3.

One fix is to comment out pragma message complaining about unknown compiler version; this is what newer Boost does (literally comments this line out rather than continuing to track MSVC versions).

Other fix ia a pre-proc tweak to prevent use of a non-standard STL method in iostreams that recent MSVC warns about.

Between these, on my machine, a release build of the IDE goes from 816 warnings to 402. The remainder are more significant (size_t vs. int, that sort of thing, and will be easier to look at them when we decide to work on this).

Note that debug builds still throw a bunch more warnings due to missing PDBs, there's a separate issue open on that #3477.

This also removed the 32-bit boost builds, so archive is half the size it used to be. The script was changed to do this a while ago but didn't rebuild boost at that time.

Built and uploaded boost archive to S3, using same name as previous archive (boost-1.65.1-win-msvc141.zip); the previous archive is renamed to boost-1.65.1-win-msvc141_warnings.zip in case there's a reason to put it back.

### Dev Action Needed

To get this updated boost on a dev box, need to blow away the boost folders under dependencies\windows and rerun install-dependencies.cmd. Likely need to do same on Jenkins build environment (I _think_ we preserve the dependencies\windows between builds). Not urgent in either case. No expected change in the build output, just the compile-time warnings.

Fixes #3140

### DIFF
--- a/dependencies/windows/tools.R
+++ b/dependencies/windows/tools.R
@@ -119,3 +119,27 @@ ensure_dir <- function(...) {
             fatal("Failed to create directory '%s'\n", dir)
    }))
 }
+
+replace_one_line = function(filepath, orig_line, new_line) {
+   new_file <- tempfile()
+   unlink(new_file)
+   con <- file(filepath, "r")
+   found_line <- FALSE
+   while (TRUE) {
+      line = readLines(con, n = 1)
+      if (length(line) == 0) {
+         break
+      }
+      if (!found_line && line == orig_line) {
+         line <- new_line
+         found_line <- TRUE
+      }
+      write(line, file=new_file, append=TRUE)
+   }
+   close(con)
+   if (!found_line) {
+      fatal("Failed to locate expected line '%s' in '%s'.", orig_line, filepath)
+   }
+   unlink(filepath)
+   file.rename(new_file, filepath)
+}
--- a/dependencies/windows/tools.R
+++ b/dependencies/windows/tools.R
@@ -121,25 +121,8 @@ ensure_dir <- function(...) {
 }
 
 replace_one_line = function(filepath, orig_line, new_line) {
-   new_file <- tempfile()
-   unlink(new_file)
-   con <- file(filepath, "r")
-   found_line <- FALSE
-   while (TRUE) {
-      line = readLines(con, n = 1)
-      if (length(line) == 0) {
-         break
-      }
-      if (!found_line && line == orig_line) {
-         line <- new_line
-         found_line <- TRUE
-      }
-      write(line, file=new_file, append=TRUE)
-   }
-   close(con)
-   if (!found_line) {
-      fatal("Failed to locate expected line '%s' in '%s'.", orig_line, filepath)
-   }
-   unlink(filepath)
-   file.rename(new_file, filepath)
+   contents <- readLines(filepath)
+   replaced <- gsub(contents, orig_line, new_line)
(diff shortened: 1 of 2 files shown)


dependencies/windows/tools.R
```diff
@@ -119,3 +119,27 @@ ensure_dir <- function(...) {
             fatal("Failed to create directory '%s'\n", dir)
    }))
 }
+
+replace_one_line = function(filepath, orig_line, new_line) {
```
kevinushey: I think this might be more simply expressed as:

```r
contents <- readLines(filepath)
replaced <- gsub(contents, orig_line, new_line)
if (!identical(contents, replaced))
  writeLines(replaced, filepath)
```

dependencies/windows/tools.R
```diff
@@ -119,3 +119,27 @@ ensure_dir <- function(...) {
...
+      }
+      if (!found_line && line == orig_line) {
+         line <- new_line
+         found_line <- TRUE
+      }
+      write(line, file=new_file, append=TRUE)
+   }
+   close(con)
+   if (!found_line) {
+      fatal("Failed to locate expected line '%s' in '%s'.", orig_line, filepath)
```
kevinushey: Does this really need to be fatal?

dependencies/windows/tools.R
gtritchie: Neato, will change it to that.

dependencies/windows/tools.R
gtritchie: I think so. This is something we run manually, very infrequently, so I'd rather this explicitly barf and stop to call out attention that we should check and likely get rid of this altogether if our new version of boost already has this.

gtritchie:
Merging, script has already been used to generate new boost download package.

PR #3507, opened 2018-09-19
avoid over-zealous anchor detach (closes #3504)

### LINKED ISSUE #3504: signature tooltip is 'sticky'
### System details

    RStudio Edition : Desktop
    RStudio Version : v1.2.998
    R Version       : 3.5.1

Try typing the following:

```R
identity(identity())
```

The signature tooltip pops up, but is 'sticky' and does not go away even after navigating the cursor outside of the inner `identity()`'s parentheses:

screen shot 2018-09-19 at 11 06 55 am

The tooltip should be dismissed after the cursor leaves the associated function's parentheses.

Marking as v1.2 milestone as this issue is supremely annoying when it occurs.

kevinushey:
This PR fixes an issue where a signature tooltip could get stuck on the screen. The issue here was in this sequence of events:

1. Signature tooltip is currently showing;
2. User types a new function call (e.g. `identity(`), triggering a new tooltip to show,
3. A new anchor is set for the newly-inserted function call,
4. The old tooltip is hidden, and erroneously detaches the newly-attached anchor (!)
5. The new tooltip is shown, but with no accompanying anchor and so is not closed on regular cursor navigation.

The fix here is to relax where anchor detach occurs.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/r/SignatureToolTipManager.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/r/SignatureToolTipManager.java
@@ -48,7 +48,6 @@ import com.google.gwt.event.dom.client.BlurHandler;
 import com.google.gwt.event.dom.client.FocusEvent;
 import com.google.gwt.event.dom.client.FocusHandler;
 import com.google.gwt.event.dom.client.KeyCodes;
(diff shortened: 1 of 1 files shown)


gtritchie:
Was this a regression in 1.2? Do we know what caused it?

kevinushey:
It looks like this is indeed a regression in v1.2 -- I can't reproduce the same behavior in v1.1 or v1.0.

PR #3522, opened 2018-09-21
update to Ace 1.4.1 (for RStudio v1.3)

kevinushey:
This PR updates to Ace 1.4.1, and ports all relevant Ace patches as well.

The previous method by which we generated patches for Ace and merged them in was somewhat at odds with a typical Git workflow, and made the process of patching Ace rather cumbersome. I think we can simplify the system simply by maintaining separate branches of Ace for each RStudio release we seek to produce. To that end, I've opened a new fork of Ace here:

https://github.com/rstudio/rstudio-ace/tree/rstudio/v1.3

and each patch is just a set of commits on top of the old Ace repository. It should be easy enough to merge / cherry-pick these patches to new branches as required.

Note that there are still code change requirements to accommodate some changes in Ace's own APIs; I'm planning to prepare those as part of a separate PR just to make that PR more reviewable.

(As a side note, I was quite happy to see that nearly all of our bugfix patches has made their way into Ace trunk -- the only things that remain are explicit features that we've built on top of Ace)


kevinushey:
Merging into v1.3 (with patches to arrive after)

PR #3546, opened 2018-09-26
check 'rstheme' UI pref for active theme

### LINKED ISSUE #3532: builtin dark themes don't report themselves as dark (?)
### System details

    RStudio Edition : Desktop
    RStudio Version : v1.2.997
    OS Version      : macOS 10.14
    R Version       : 3.5.1

### Description

Compare the output of `library(tidyverse)` between RStudio v1.1 and RStudio v1.2, with the builtin 'Tomorrow Night Bright' theme selected:

#### RStudio v1.1

screen shot 2018-09-24 at 11 47 14 am

#### RStudio v1.2

screen shot 2018-09-24 at 11 46 36 am

Note that the text is no longer white in RStudio v1.2.

(originally filed at https://github.com/tidyverse/tidyverse/issues/156)

kevinushey:
Closes #3532.

### DIFF
--- a/src/cpp/r/R/Api.R
+++ b/src/cpp/r/R/Api.R
@@ -706,12 +706,21 @@ options(terminal.manager = list(terminalActivate = .rs.api.terminalActivate,
 })
 
 .rs.addApiFunction("getThemeInfo", function() {
-   editor <- .Call("rs_readUiPref", "theme")
-   global <- .Call("rs_readUiPref", "flat_theme")
-
+   
+   # read theme preferences
+   editor <- .rs.readUiPref("theme")
+   global <- .rs.readUiPref("flat_theme")
+   
+   # if 'theme' is not set, try pulling from 'rstheme'
    if (is.null(editor)) {
-      editor <- "TextMate"
+      rstheme <- .rs.readUiPref("rstheme")
+      if (!is.null(rstheme))
+         editor <- rstheme[["name"]]
    }
+   
+   # if we still don't have a theme, default to textmate
+   if (is.null(editor))
+      editor <- "TextMate"
 
    global <- switch(
       if (is.null(global)) "" else global,


src/cpp/r/R/Api.R
```diff
@@ -706,12 +706,21 @@ options(terminal.manager = list(terminalActivate = .rs.api.terminalActivate,
...
-
+   
+   # read theme preferences
+   editor <- .rs.readUiPref("theme")
+   global <- .rs.readUiPref("flat_theme")
+   
+   # if 'theme' is not set, try pulling from 'rstheme'
    if (is.null(editor)) {
-      editor <- "TextMate"
+      rstheme <- .rs.readUiPref("rstheme")
```
MariaSemple: `theme` is deprecated, it should always be `rstheme` now. Good catch.

kevinushey: I wonder if we should instead read `rstheme` before `theme` then? Could these variables get out of sync?

MariaSemple: `theme` is unused everywhere else, we can just stop reading it altogether. For 1.2 people will only have a `theme` value in their prefs file if they upgraded from an older version. Unfortunately this means users will have to re-set their theme preference after upgrade, but the storage structure changed so we really had to do this.

kevinushey: Got it. I'll remove the old `"theme"` usage altogether.

src/cpp/r/R/Api.R
```diff
@@ -707,16 +707,14 @@ options(terminal.manager = list(terminalActivate = .rs.api.terminalActivate,
 
 .rs.addApiFunction("getThemeInfo", function() {
    
-   # read theme preferences
+   # read theme preferences (note: 'theme' is deprecated)
    editor <- .rs.readUiPref("theme")
```
MariaSemple: If you check this value, it may show a stale value for a user who just upgraded and hasn't set their theme yet. I.e. before upgrade they were using Tomorrow Night. After upgrade, it looks like they have Textmate set, but really there's no `rstheme` value in the preferences. Then if they call this function, it will return Tomorrow Night instead of Textmate.

kevinushey: Ah, got it. Fixed that up here: https://github.com/rstudio/rstudio/commit/75a3891d744f48a6cba5b228da6d99331163b0d9

PR #3588, opened 2018-10-03
implement commands for moving multiselect cursor up, down

kevinushey:
This PR makes it possible to move a multi-select cursor up or down using <kbd>Ctrl+Alt+Shift+{Up/Down}</kbd>. This can be useful when the generated cursor does not lie on the line you wanted it to lie on.

skip-cursor


kevinushey:
Looks like this conflicts with the bindings we have to expanding / shrinking the selection. Will need to think about this a bit more.

PR #3599, opened 2018-10-05
Make title window color to match menu color not backgorund color

javierluraschi:
Follow up to PR (pointed out by @kevinushey): https://github.com/rstudio/rstudio/pull/3365

I just upgraded to Mojave, my first impression felt like the dark theme was a bit broken, since the color cuts between the window and the menu:

screen shot 2018-10-04 at 5 18 58 pm

I experimented with setting the menu bar to be the same as the background to match the title, but feels a bit weird since there is no color distinction. @kevinushey suggested matching the title with the menu, which I think it works well, but certainly open to other suggestions:

screen shot 2018-10-04 at 5 18 42 pm

Default theme is left unchanged, even though the change is also applied. Seems to me like Mojave only applies title colors that are dark?

Before:

screen shot 2018-10-04 at 5 24 15 pm

After:
screen shot 2018-10-04 at 5 24 41 pm

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/application/DesktopFrame.java
+++ b/src/gwt/src/org/rstudio/studio/client/application/DesktopFrame.java
@@ -160,6 +160,7 @@ public interface DesktopFrame extends JavaScriptPassthrough
    void zoomActualSize();
    
    void setBackgroundColor(JsArrayInteger rgbColor);
+   void changeTitleBarColor(int r, int g, int b);
    void syncToEditorTheme(boolean isDark);
    
    void getEnableAccessibility(CommandWithArg<Boolean> callback);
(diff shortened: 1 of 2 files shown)


kevinushey:
I wonder how it might look with the unified background color, but a horizontal line to separate the title bar (similar to the one used to separate the main menu bar from the lower content)

jmcphers:
A lot of apps use to create a simpler/flatter look. Here's Chrome:

image

Here's WhatsApp -- again, window controls just get mixed together with app controls.

image

ITunes is similar:

image

And here's another app:

image

Arguments can be made either way for whether we should visually distinguish the title bar from the body of the app, but aesthetically I think that most apps have chosen to let them be seamless, so it won't hurt for us to follow suit.

PR #3606, opened 2018-10-05
Consider upgrading openssl to 1.0.2p

javierluraschi:
Mitigates https://github.com/rstudio/rstudio/issues/3572 by upgrading OpenSSL from November 2017 to August 2018 release.


dependencies/windows/install-openssl/install-openssl.R
```diff
@@ -1,5 +1,5 @@
 OWD <- getwd()
-URL <- "https://www.openssl.org/source/old/1.0.2/openssl-1.0.2m.tar.gz"
+URL <- "https://www.openssl.org/source/openssl-1.0.2p.tar.gz"
```
gtritchie: Wonder if this URL will stop working when there's another release of openssl 1.0 (i.e. 1.0.2p becomes `old`). I got bitten by this with scripts / Dockerfile for R. Not sure what you can do about it until then.

javierluraschi: Ah, yes, I did try to use the old URL; however the binary does not exist there. The good news, the old binaries do exist in the same path, so for instance, this URL still works: https://www.openssl.org/source/openssl-1.0.2m.tar.gz

Issue #2330, opened 2018-03-06
Too many open scripts to display in drop-down doesn't allow scrolling down to hidden tabs


ronblum:
### System details

    RStudio Edition : Desktop and Server
    RStudio Version : 1.1.383, 1.1.423, 1.2.380-1
    OS Version      : Ubuntu 16.04, Debian 8, MacOS 10.12.6
    R Version       : 3.4.0, 3.4.3

### Steps to reproduce the problem

1. Open many R scripts, creating many tabs—say, 33.
2. Right-click on the double arrows/greater-than signs (`>>`) next to the rightmost tab.

### Describe the problem in detail

There are too many open scripts for the vertical list to show them all—the bottom of the IDE would have to be extended, and the screen might not be big enough to handle it. In this example, there are 33 open scripts/tabs, but the vertical drop-down menu only has enough room to display 22. There is no scrolling mechanism to show the remaining 11. 

<img width="365" alt="too_many_files" src="https://user-images.githubusercontent.com/31009952/37058375-5555722a-2158-11e8-859f-742388226072.png">

Note that it is possible to mouse down all the way to 33, or mouse up from 1 so that it wraps around to 33 and on, and then select a script that's hidden. You can't see that the script/tab is selected, but you can choose it, and the focus will switch to that tab. That is, you'll move to another tab, but you won't know which one ahead of time.

### Describe the behavior you expected

It would be helpful to be able to scroll through a list that's longer than the UI allows.

gtritchie:
See also #1794 and #1760

PR #3635, opened 2018-10-09
clamp popup menu size and make scrollable

### LINKED ISSUE #1760: Unable to scroll in list of deployment servers
In a project, i've had a lot of servers that i've deployed to. On my screen this has become an issue in that the list is so long that i can't see all of the servers because I can't seem to scroll up or down to see more servers that aren't visible already. I'm running version 1.2.176 of the IDE.

kevinushey:
This PR limits the size of menu bars, and makes them scrollable by default.

screen shot 2018-10-09 at 10 15 36 am

screen shot 2018-10-09 at 10 15 18 am

Closes #1760, #1794, #2330.

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/command/BaseMenuBar.java
+++ b/src/gwt/src/org/rstudio/core/client/command/BaseMenuBar.java
@@ -14,6 +14,7 @@
  */
 package org.rstudio.core.client.command;
 
+import com.google.gwt.core.client.Scheduler;
 import com.google.gwt.dom.client.Element;
 import com.google.gwt.event.dom.client.KeyCodes;
 import com.google.gwt.user.client.DOM;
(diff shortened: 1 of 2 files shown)


jjallaire:
I'm assuming if they don't overflow then they don't show scrolling indicators?

kevinushey:
That's right (we use `overflow-y: auto` so the browser only draws them when needed)

kevinushey:
I spoke too soon. `overflow-y: auto` seems to force a scrollbar even when un-needed. I'll tweak the PR.

gtritchie:
What's the first screenshot showing? The shorter one? Is the main window really short, so the menu is drawing off the bottom?

kevinushey:
That's correct (the IDE has been resized to be very small in height so the screenshot shows the menu has clamped itself to the height of the IDE window)

PR #3638, opened 2018-10-10
update win32 manifests to state support for 8, 8.1, 10

### LINKED ISSUE #2899: win.version() under windows 10
### System details

    RStudio Edition : Desktop
    RStudio Version :  1.2.672 Build 695 (25de3d38) 
    OS Version      :  Windows 10, build 17134.48 (as reported by Window Settings)
    R Version       :  3.5.0 (2018-04-23) Platform: x86_64-w64-mingw32/x64 (64-bit)

In RStudio on Windows 10, at the very beginning of a session, run `win.version()`. Note that it shows a build number (`9200`) [corresponding to Windows 8.0](https://en.wikipedia.org/wiki/List_of_Microsoft_Windows_versions).
```
"Windows >= 8 x64 (build 9200)"
```

In RGUI (same R version) on Windows 10, at the very beginning of a session, run `win.version()`. The OS and build numbers are now correct.
```
 "Windows 10 x64 (build 17134)"
```

On Windows, when running R under RStudio, `win.version()` does not report the correct OS release and version number. However, the correct values appear in RGUI (using the same R version) - please see "Steps to reproduce the problem". The issue has been reported before, on the [old RStudio support site](https://support.rstudio.com/hc/en-us/community/posts/207830207-utils-win-version-gives-incorrect-results-for-Windows-10).

This behavior is [documented in the help for `win.version()`](https://www.rdocumentation.org/packages/utils/versions/3.4.1/topics/winextras). However, it has been reported as being [fixed in R 3.2.3](https://www.r-statistics.com/2015/12/r-3-2-3-is-released-with-improvements-for-windows-users-and-general-bug-fixes/)

> Compatibility information has been added to the manifests for ‘Rgui.exe’, ‘Rterm.exe’ and ‘Rscript.exe’. This should allow win.version() and Sys.info() to report the actual Windows version up to Windows 10.

I expect the output of `win.version()` not to depend on whether R is run within RStudio or in a stand-alone console. (I believe the solution might be related to the Windows application manifest.)

gtritchie:
fixes #2899 

Use same Windows version list in the application manifest as used by R GUI (adding 8, 8.1, and 10 in addition to existing Vista and Win-7 GUIDs).

On Windows 8.1+ this causes the RStudio processes (including rstudio.exe and rsession.exe) to run in the actual operating system context, instead of the downlevel context.

More specifically, causes `win.version()` to return the actual Windows build instead of being stuck on showing build 9200.

This seems worth taking for 1.2 because it makes RStudio behavior match R GUI in this regard, and is "only" a metadata update (i.e. we can easily remove this if we discover any issues). That said, if we'd rather wait for 1.3, I'd understand.

https://docs.microsoft.com/en-us/windows/desktop/sbscs/application-manifests

### DIFF
--- a/src/cpp/desktop/rstudio.exe.manifest
+++ b/src/cpp/desktop/rstudio.exe.manifest
@@ -14,10 +14,16 @@
    </trustInfo>
    <compatibility xmlns="urn:schemas-microsoft-com:compatibility.v1">
       <application>
-         <!--The ID below indicates application support for Windows Vista -->
+         <!-- Windows 10 -->
+         <supportedOS Id="{8e0f7a12-bfb3-4fe8-b9a5-48fd50a15a9a}"/>
+         <!-- Windows 8.1 -->
+         <supportedOS Id="{1f676c76-80e1-4239-95bb-83d0f6d0da78}"/>
+         <!-- Windows Vista -->
          <supportedOS Id="{e2011457-1546-43c5-a5fe-008deee3d3f0}"/>
-         <!--The ID below indicates application support for Windows 7 -->
+         <!-- Windows 7 -->
          <supportedOS Id="{35138b9a-5d96-4fbd-8e2d-a2440225f93a}"/>
+         <!-- Windows 8 -->
+         <supportedOS Id="{4a2f28e3-53b9-4441-ba9c-d69d4a4a6e38}"/>
       </application>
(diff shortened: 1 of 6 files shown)


kevinushey:
LGTM! IMHO this is worth taking this for v1.2.

PR #3677, opened 2018-10-15
can't safely get address of empty vector

gtritchie:
Ran into this doing Pro Server work on an Ubuntu Bionic (18) VM, with a Debug build of the product.

Server would assert and terminate on startup. This is same issue that httpuv project was hitting here:

https://github.com/rstudio/httpuv/issues/133

Debug build uses _GLIBCXX_ASSERTIONS and which asserts on usage of this pattern:

&v[0], where "v" is a vector

Take address of first element of a potentially empty vector is undefined behavior in C++.

Also can't just use vector::data(), because the return value of data() may or may not be nullptr for an empty vector, per C++ docs. (D'oh)

So added `safe_vec_addr` as used by httuv. I only fixed the one case that was preventing me from starting the server with a Debug build. This case isn't an actual runtime problem, but other cases could be, at least in theory.

I will open an issue to do a comprehensive survey of the codebase to find other potential problems of this sort and/or fix them as I hit them.

### DIFF
--- a/src/cpp/core/include/core/SafeConvert.hpp
+++ b/src/cpp/core/include/core/SafeConvert.hpp
@@ -1,7 +1,7 @@
 /*
  * SafeConvert.hpp
  *
- * Copyright (C) 2009-12 by RStudio, Inc.
+ * Copyright (C) 2009-18 by RStudio, Inc.
  *
  * Unless you have received this program directly from RStudio pursuant
  * to the terms of a commercial license agreement with RStudio, then
@@ -137,10 +137,18 @@ boost::optional<TOutput> numberTo(TInput input)
    return result;
 }
 
+// Indexing into an empty vector causes assertion failures on some platforms.
+// Cannot rely on vector::data() as it is not guaranteed to be nullptr for
+// empty vectors.
+template <typename T>
+T* safe_vec_addr(std::vector<T>& vec)
+{
+   return vec.size() ? &vec[0] : nullptr;
+}
+
 } // namespace safe_convert
 } // namespace core 
 } // namespace rstudio
-
 
 #endif // CORE_SAFE_CONVERT_HPP
 
--- a/src/cpp/core/system/PosixGroup.cpp
+++ b/src/cpp/core/system/PosixGroup.cpp
@@ -1,7 +1,7 @@
 /*
  * PosixGroup.cpp
  *
- * Copyright (C) 2009-16 by RStudio, Inc.
+ * Copyright (C) 2009-18 by RStudio, Inc.
  *
  * Unless you have received this program directly from RStudio pursuant
  * to the terms of a commercial license agreement with RStudio, then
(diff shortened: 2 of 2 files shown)


src/cpp/core/system/PosixGroup.cpp
```diff
@@ -55,10 +55,10 @@ Error groupFrom(const boost::function<int(
    int result = 0;
    do
    {
-      buffer.reserve(buffSize);
+      buffer.resize(buffSize);
 
       // attempt the read
-      result = getGroup(value, &grp, &(buffer[0]), buffSize, &temp);
+      result = getGroup(value, &grp, safe_convert::safe_vec_addr(buffer), buffer.size(), &temp);
```
kfeinauer: Isn't the `safe_vec_addr` unnecessary here after the call to `buffer.resize` above?

src/cpp/core/include/core/SafeConvert.hpp
```diff
@@ -137,10 +137,18 @@ boost::optional<TOutput> numberTo(TInput input)
    return result;
 }
 
+// Indexing into an empty vector causes assertion failures on some platforms.
+// Cannot rely on vector::data() as it is not guaranteed to be nullptr for
+// empty vectors.
+template <typename T>
+T* safe_vec_addr(std::vector<T>& vec)
```
kfeinauer: I don't feel like this really belongs in a utility method in `safe_convert` myself, but @jmcphers probably knows better. I'd be tempted to say we should create our own wrapper container for `std::vector` and require us to use that instead of the regular vector (perhaps via compiler flags to spit out deprecation warnings or errors?)

gtritchie: So thinking wrapping vector in all cases might be a bigger change than necessary vs. just a selective tweak. Still, that would require extra thinking versus a gigantic search and replace.

kevinushey:
> Also can't just use vector::data(), because the return value of data() may or may not be nullptr for an empty vector, per C++ docs. (D'oh)

Shouldn't this still be fine? The only problem here was the attempt to access the first element of either a 0-length buffer or a null pointer; as long as we're using `vector.data()` and declaring that `bufSize` is 0 then we should avoid that issue entirely. Or am I misunderstanding something?

(FWIW I saw similar errors with MSVC debug builds and fixes those by moving to the `.data()` method)

src/cpp/core/include/core/SafeConvert.hpp
kfeinauer: I agree completely, but the downside with this approach is we have to remember this pattern everywhere going forward. But this small issue is likely not worth my suggestion, like you suggest.

kfeinauer: If I had to suggest where else we might move such a method, I'd say we'd put it in a collection utils namespace.

gtritchie:
Yes, agree it should be fine with .data() in any sane behavior of the function being invoked. Got to this pattern in a circuitous way.

I first fixed with .data(), and it worked in this case (where bufSize > 0).

Switched to `safe-vec_addr` when I was thinking I'd do a broader code sweep, and wanted to use a consistent calling pattern in all cases. Initially that failed because `safe-vec-addr` uses .size() to determine if vector is empty, and since we'd done a `reserve()`, it was size zero, thus returned nullptr and crashed. So switched to `resize()` and left in `safe-vec-addr`. Then decided to defer a broader search for "later".

So question is really whether there is a case where having a non-null .data() for an empty vector is dangerous and it would be better to have a guaranteed nullptr to identify pathological usage of the results.

Of course this all assumes there are actual implementations that don't return nullptr for empty vector's .data().

I suspect it is unlikely, and would be find just using .data(). :-)

PR #3742, opened 2018-10-25
avoid special number formatting when serializing JSON

kevinushey:
Candidate fix for #3734.


kevinushey:
Will explore alternative fix that doesn't touch such a hot codepath.

PR #3782, opened 2018-10-29
generate config.h for core project, even on Windows

gtritchie:
A change was made in pro source that introduced need to have `config.h` generated in `core` in order for Pro desktop to build on Windows.

Checking in here to keep files in sync (i.e. merge open source to pro after this is merged).


gtritchie:
Got a linker error later in the (very slow) Windows Pro build. Pulling back this PR until we confirm if it's a right thing to do.

PR #3785, opened 2018-10-30
Fix Windows build

gtritchie:
Fixes to Windows builds; sync pro repo after merging this into open source.

(Karl was having difficulties getting his Windows dev environment happy so pushing on his behalf)


gtritchie:
Nope, still not right. Sigh!

Issue #3809, opened 2018-11-02
Ctrl-W doesn't close the preview window in Windows


ronblum:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.1.463 and 1.2.1086
    OS Version      : Windows 7 and 10
    R Version       : 3.4.4 and 3.5.1

### Steps to reproduce the problem

1. Create a trivial Rmd. Just a title in the YAML and a single paragraph is sufficient (or the default Rmd).
2. Make sure that "Preview in Window" is selected.
3. Knit.
4. Ctrl-w (expecting to close the window). This doesn't work.
5. Click within the document (giving it focus).
6. Command-w still doesn't close the preview window.

### Describe the problem in detail

Ctrl-w does not close the preview window.

### Describe the behavior you expected

Ctrl-w should close the preview window. Note that it successfully closes tabs in the Editor pane.

This is an existing issue in 1.1.463.

[Also, I more-or-less copied the nice repro from #3753.]

dfalty:
Verified fixed on version 1.2.1099.

PR #3814, opened 2018-11-02
bind Ctrl+W to close window for satellites

kevinushey:
Closes https://github.com/rstudio/rstudio/issues/3809.


kevinushey:
Perhaps worth noting that our WebView class also implements handling for window close shortcuts:

https://github.com/rstudio/rstudio/blob/6974095254a196f3df888d49c8b2bce6cf5d14a9/src/cpp/desktop/DesktopWebView.cpp#L142-L161

but that only works once the document itself has focus, which isn't true when the window is initially opened.

Issue #1889, opened 2017-12-29
Opening Global Options takes ages


Thermi:
Whenever I open the Global Options, Rstudio takes ages to open the dialog. It seems to take a long time in "Reading options...".
Processor load ist low while it does that, so it seems it is waiting for a lock or something. What's up with that?

kevinushey:
Can you also let us know what version of RStudio + what operating system you're running on? You might also try downloading and installing the preview release, as we have a number of fixes in there for issues of this form.

https://www.rstudio.com/products/rstudio/download/preview/

Thermi:
It's 1.1.383 on Arch Linux.  1.1.408 takes a long time, too.

kevinushey:
Some other information that might be useful...

Can you try attaching a debugger to the `rsession` process? From the terminal, you can run:

    sudo gdb -p `pidof rsession`

(assuming you only have one `rsession` instance running and have the `pidof` utility installed; if not, just use `Sys.getpid()` from the R session to the the associated PID)

Then, run the following to ask GDB to log any generated output to a file:

    set logging file rsession-gdb.txt
    set logging on
    continue

Then, allow the process to continue running, then try opening the Global Options dialog. While the R session is 'stuck', try interrupting the process a few times, to see what the threads are up to (from the gdb REPL, use <kbd>Ctrl + C</kbd> to interrupt the process, and enter `bt` to print the backtrace).

After the Global Options dialog has successfully opened, you can detach the debugger and copy the `rsession-gdb.txt` file to e.g https://gist.github.com and provide us a link.

Thermi:
GDB debug output is attached. Disregard the fancy binary character lines they're because I got a fancy .gdbinit file.

Opening the global options afterwards (after they were already opened once) is instant.

[rsession-gdb.txt](https://github.com/rstudio/rstudio/files/1598404/rsession-gdb.txt)

kevinushey:
Strange -- it seems like the `rsession` process itself is mostly just waiting around as well.

Is the RStudio IDE itself responsive while waiting for the Global Options pane to open? (Can you open other panes, or type in the editor, or interact with the menu bars -- that sort of thing?)

You might also look at the request log, using <kbd>Ctrl + `</kbd>, after trying to open the Global Options dialog. In my case, I see:

[...]

PR #3833, opened 2018-11-06
maintain cache of installed monospace fonts

kevinushey:
This PR seeks to further improve startup performance, especially in the case when there are a very large number of fonts installed on the system.

The QFontDatabase object is unfortunately quite slow -- it takes a long time (especially when many fonts are installed) to read and parse installed fonts on the system. This PR tries to mitigate some of that cost by maintaining a local cache of monospace fonts.

The font cache is rebuilt eagerly when it doesn't exist, and in a separate thread (anticipating that new fonts might have been added) when it does. One must still restart RStudio to have it pick up any newly-installed fonts, and while that's not great I don't think it's too detrimental since font installation is done fairly infrequently.

Closes https://github.com/rstudio/rstudio/issues/1889.

### DIFF
--- a/src/cpp/desktop/DesktopGwtCallback.cpp
+++ b/src/cpp/desktop/DesktopGwtCallback.cpp
@@ -1262,14 +1262,7 @@ bool isProportionalFont(const QString& fontFamily)
 
 QString GwtCallback::getFixedWidthFontList()
 {
-   QFontDatabase& db = desktop::fontDatabase();
-   QStringList families = db.families();
-
-   QStringList::iterator it = std::remove_if(
-            families.begin(), families.end(), isProportionalFont);
-   families.erase(it, families.end());
-
-   return families.join(QString::fromUtf8("\n"));
+   return desktop::getFixedWidthFontList();
 }
 
 QString GwtCallback::getFixedWidthFont()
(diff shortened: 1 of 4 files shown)


kevinushey:
That's a good question, and you're right, it does mean the first launch will be quite slow.

The fixed-width font list is only used by the front-end in one place, though:

https://github.com/rstudio/rstudio/blob/master/src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/AppearancePreferencesPane.java

and that shouldn't be getting initialized eagerly. Perhaps this has something to do with the Q_PROPERTY attributes being initialized eagerly in the DesktopInfo class:

https://github.com/rstudio/rstudio/blob/55831e0ba586d0157c1da1d30d69cc25b3e31ea0/src/cpp/desktop/DesktopInfo.hpp#L50-L55

I should take the time to figure out why the font list is being built on startup, and perhaps see if we can instead just do that in a separate thread or something similar (since the font list is only needed in the Global Options pane)

kevinushey:
Superceded by https://github.com/rstudio/rstudio/pull/3833.

josswright:
> Superceded by #3833.

(I think you meant #3842 -- thanks so much for working on this!)

kevinushey:
Oops, yes :-)

PR #3840, opened 2018-11-07
support compilation with OpenSSL 1.1

kevinushey:
Fairly minimal change to allow compilation against OpenSSL 1.1. If I'm reading the documentation at https://www.openssl.org/docs/man1.1.0/crypto/ERR_load_crypto_strings.html correctly, explicit initialization here is no longer necessary.

### DIFF
--- a/src/cpp/core/system/Crypto.cpp
+++ b/src/cpp/core/system/Crypto.cpp
@@ -115,8 +115,10 @@ private:
    
 void initialize()
 {
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
    // load global error string table
    ::ERR_load_crypto_strings();
+#endif
 }
    
 Error HMAC_SHA2(const std::string& data,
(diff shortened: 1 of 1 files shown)


src/cpp/core/system/Crypto.cpp
```diff
@@ -163,7 +165,7 @@ Error HMAC_SHA2(const std::string& data,
 Error sha256(const std::string& message,
              std::string* pHash)
 {
-   SHA256_CTX shaCtx = {0};
+   SHA256_CTX shaCtx;
```
kevinushey: This just silences a compiler warning (AFAIK one does not need to initialize the object on construction; `SHA256_init()` does everything we need)

kevinushey:
Note that this PR doesn't intend to solve the more general issue of allowing RStudio to bind to and use whatever version of OpenSSL is available at runtime; this just tries to get us compiling against OpenSSL 1.1+.

src/cpp/core/system/Crypto.cpp
```diff
@@ -115,8 +115,10 @@ class BIOFreeAllScope : boost::noncopyable
    
 void initialize()
 {
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
    // load global error string table
    ::ERR_load_crypto_strings();
```
kfeinauer: So this function was removed in 1.1?

PR #3852, opened 2018-11-08
provide color for whitespace characters (closes #3718)

### LINKED ISSUE #3718: ace_invisible not being generated by tmTheme conversion
When converting from tmTheme it doesn't appear that a rule for `ace_invisible` is being generated, which results in invisibles always appearing as the same colour as the default text colour for the theme. This can be quite bright for invisible characters.

kevinushey:


kevinushey:
Note that this would require users who want highlighting of whitespace characters to re-import a `.tmTheme` but that's probably fine (no need to patch existing themes since we haven't made a release yet)

src/cpp/session/resources/themes/compile-themes.R
```diff
@@ -1067,6 +1067,18 @@
       replace = operatorBgColor
    )
    
+   ## Provide a color for invisible characters if none is defined.
```
MariaSemple: The tmTheme -> ace conversion code should always generate an invisible color, it's just not being used right now (there's a default if one isn't set, but it seems like most themes have one set). I think to fix this we should either add 
```
.%cssclass% {
   color: %invisible%
}
```
to the ace_theme_template.css file (here: https://github.com/rstudio/rstudio/blob/master/src/cpp/session/resources/templates/ace_theme_template.css)

kevinushey:
Superseded by https://github.com/rstudio/rstudio/pull/3908.

PR #3904, opened 2018-11-15
support file.choose(new = TRUE)

### LINKED ISSUE #3903: file.choose() not honoring new=TRUE on macOS
It's me again. Latest RStudio dev as usual (but this issue is from SO so likely impacting 1.1 as well).

https://stackoverflow.com/a/53323980/1457051

    file.choose(new=TRUE)

is supposed to have a filename prompt. it does not. 

vis -a-vis

screen shot 2018-11-15 at 11 19 49

screen shot 2018-11-15 at 11 20 08

screen shot 2018-11-15 at 11 20 58

screen shot 2018-11-15 at 11 38 12

no filename prompt

kevinushey:
Closes #3903.


hrbrmstr:
OMGosh. The # of 🍻 I owe you two is going to break #4's college fund

PR #3952, opened 2018-11-27
allow users to customize rendering engine in choose R version dialog

kevinushey:
This PR makes it possible for users to customize the rendering engine used by RStudio in the choose R version dialog:

screen shot 2018-11-27 at 10 14 23 am

The intention is that this could be used as an escape hatch for users with rendering issues on Windows that make the IDE unusable; e.g. as described in https://community.rstudio.com/t/rstudio-preview-ide-rstudio-1-2-1139-screen-is-black/18819/3. The intention is that, with this feature, we could instruct such users to launch RStudio by holding down <kbd>Ctrl</kbd> when RStudio is launched, and then they could try changing the rendering engine used in this dialog.

Note that the user will technically still need to restart the IDE once more (since these settings are applied too late; ie, after GPU-related things are initialized in the QApplication object). Is there a clean way we could communicate this requirement? Or should we just display a dialog requesting the user restart RStudio if they change the rendering engine here?

### DIFF
--- a/src/cpp/desktop/DesktopChooseRHome.cpp
+++ b/src/cpp/desktop/DesktopChooseRHome.cpp
@@ -32,6 +32,18 @@ using namespace rstudio::desktop;
 
 namespace {
 
+static std::map<std::string, std::string> s_engineValueTolabel = {
+   {"auto", "Auto-detect (recommended)"},
+   {"desktop", "Desktop OpenGL"},
+   {"software", "Software"}
+};
+
+static std::map<std::string, std::string> s_engineLabelToValue = {
+   {"Auto-detect (recommended)", "auto"},
+   {"Desktop OpenGL", "desktop"},
+   {"Software", "software"}
+};
+
 QListWidgetItem* toItem(const RVersion& version)
 {
    QListWidgetItem* pItem = new QListWidgetItem();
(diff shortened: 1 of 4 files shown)


jmcphers:
It's definitely going to be confusing if you select a rendering engine before the main window appears, but the main window doesn't use the engine you _just selected_. This dialog mixes things that affect the pending launch (the R version) and things that affect the launch _after_ that (the renderer). That's an awkward combination that we probably can't explain away.

However I don't think we need to over-optimize this UX since changing the rendering engine is rare and most people will never need to do it. We could just show a dialog if you change the rendering engine, which says something like:

> Rendering engine and R version updated successfully. Start RStudio again to use your new settings.

... and we quit right away. A behind the scenes restart would be ideal, but probably more trouble than it's worth (let me know if you disagree).

kevinushey:
I think quitting and requesting a manual user restart is good enough here (I mainly wanted to confirm that we felt this was okay and didn't want to sink too much time into UX here).

Ultimately, this is just an escape hatch for users who cannot use RStudio due to some kind of rendering issue.

kevinushey:
Now with a message box informing the user a restart is required:

screen shot 2018-11-27 at 12 51 56 pm

kevinushey:
screen shot 2018-11-27 at 12 51 56 pm

jmcphers:
Nice!

I think we still need to quit (abort the launch) after showing the dialog (it's weird to be asked to restart and then to see the app open up). What do you think?

kevinushey:
Agreed, and that's what happens now :-) (The IDE quits in response to an 'invalid' R version detected, and we journey on that codepath by returning an empty string here when we see the rendering engine has changed)

jmcphers:
Ah, missed that codepath. LGTM!

PR #3957, opened 2018-11-27
Update dracula.rstheme

meztez:
Changed the console input color #de935f for Dracula theme to match executed console input color #ff79c6.
So when your type code in the console it doesn't change color when executed.

### DIFF
--- a/src/cpp/session/resources/themes/dracula.rstheme
+++ b/src/cpp/session/resources/themes/dracula.rstheme
@@ -205,7 +205,7 @@
   color: #cc6666 !important;
 }
 .nocolor.ace_editor .ace_line span {
-  color: #de935f !important;
+  color: #ff79c6 !important;
 }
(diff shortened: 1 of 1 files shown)


MariaSemple:
Hi @meztez, thanks for submitting this PR. The rule modified in this PR is fairly general and could impact more aspects of the theme than the color of the text currently being typed in the console. Rather than change the pre-existing theme included with RStudio, you could define your own custom theme as a fork.

In the RStudio 1.2 preview you can add custom themes to RStudio. Custom themes with the same name as pre-existing themes will override the pre-existing theme, so you can even hide the pre-existing version of the theme with the one you prefer.

For more information about adding and removing custom themes, please check out our blog post about custom themes: 
https://blog.rstudio.com/2018/10/29/rstudio-ide-custom-theme-support/

PR #4015, opened 2018-12-07
upgrade to Qt 5.12.0

kevinushey:
This PR upgrades RStudio to Qt 5.12.0. Some notable changes:

1. We now explicitly specify a platform target on macOS; targeting macOS Sierra (10.12). This appears necessary to work around some Qt bugs that seem to appear only when no macOS target is specified; at least as seen in my local development builds on Mojave.

2. Some of the workarounds we had included for Qt 5.11.1 are no longer necessary.

### DIFF
--- a/dependencies/common/install-boost
+++ b/dependencies/common/install-boost
@@ -87,8 +87,8 @@ then
    if [ "$PLATFORM" == "Darwin" ]
    then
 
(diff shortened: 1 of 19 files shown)


kevinushey:
Note: not planning to merge right away (will wait for QA to give a test pass next week)

gtritchie:
Groovy! I will fire off a test build of the Windows docker image on my google compute environment.

PR #4016, opened 2018-12-07
Fix output pane when moving between jobs with and without output

### LINKED ISSUE #4014: Jobs Pane doesn't always show "Job has not emitted output" when it should
### System details

    RStudio Version : 1.2.1182
    OS Version      : any
    R Version       : any

- In Jobs pane, run a local job that generates some output.
- Now run a local job that doesn't generate any output

The message "Job has not emitted output" is not displayed in the pane.

Should have seen this:

This only happens when you run or view a job that didn't emit output AFTER running/viewing a job that had output.

Secondarily, this bug helped helped reveal a bug on Pro launcher-jobs where that code is errantly sending an empty output event which trigger hiding this label when it shouldn't.

gtritchie:
Fixes #4014

Also ignore showOutput calls with empty output. Launcher jobs are doing this (that will be fixed separately), but might as well handle it, otherwise the "Job has not emitted output" message is prematurely hidden.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/jobs/view/JobOutputPanel.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/jobs/view/JobOutputPanel.java
@@ -41,13 +41,13 @@ public class JobOutputPanel extends Composite
       initWidget(uiBinder.createAndBindUi(this));
       
       // initially empty
-      output_.setVisible(false);
-      empty_.setVisible(true);
+      clearOutput();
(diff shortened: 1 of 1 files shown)


kfeinauer:
FYI the empty string output calls are sent to indicate the end of the stream itself, so I don't think it would be advisable to change that particular behavior. It's nice to have a way to determine when the stream has ended.

gtritchie:
@kfeinauer Thanks. Was wondering about that.

PR #4050, opened 2018-12-18
Update bootstrap scripts to use Qt 3.0.6 online installer

gtritchie:
Qt released a new build of their installer framework and thus their online installer. Supposed to have some performance improvements (less metadata download) and other assorted stuff.

https://blog.qt.io/blog/2018/12/17/qt-installer-framework-3-0-6-released/

Tested on Mac, Windows, Linux by uninstall Qt and reinstalling with these updated dependency scripts after pushing new bits to S3.

No compelling reason to take this for 1.2 (3.0.5 continues to work AFAIK), so suggest just merging into 1.3.

See also updated notes at https://github.com/rstudio/rstudio-pro/wiki/Updating-Qt-SDK on what gets pushed to S3.

### DIFF
--- a/dependencies/linux/install-qt-sdk
+++ b/dependencies/linux/install-qt-sdk
@@ -18,7 +18,7 @@ if [ -z "$QT_VERSION" ]; then
     QT_VERSION=5.11.1
 fi
 if [ -z "$QT_SDK_BINARY" ]; then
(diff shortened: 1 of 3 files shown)


kevinushey:
LGTM! Nice to see that our existing install scripts still work.

PR #4056, opened 2018-12-18
make dictionary paths user-accessible

kevinushey:
To support https://github.com/rstudio/rstudioapi/issues/123.


kevinushey:
Yes -- I'll add these.

Issue #2574, opened 2018-04-03
Messages should not print in red text


hadley:
<img width="306" alt="screen shot 2018-04-03 at 8 25 29 am" src="https://user-images.githubusercontent.com/4196/38258923-973505cc-3718-11e8-9f7b-9d67a0f59611.png">

Display with red text makes errors look like they are problems, when they are often just informative. It also makes it difficult to style messages with crayon because we don't know what the default text colour will be.

jimhester:
Alternatives to coloring in red would be using bold, as is done by R by default in the macOS terminal, or doing nothing as is done on linux systems.

kevinushey:
Note that the issue is ultimately that all `stderr()` output is colored like that:

![screen shot 2018-04-03 at 8 46 32 am](https://user-images.githubusercontent.com/1976582/38260082-86d7a63c-371b-11e8-8332-d7cbec6f1488.png)

Do you think we should attempt out color `message()` output distinctly from `stderr()` output?

hadley:
I think `stderr` and `message()` should be formatted consistently - `stderr` is generally for diagnostic messages, and not just errors.

Warnings and error should continue to print in red however.

jimhester:
Differentiating between `message()` `warning()` `stop()` and `stderr()` would be a great thing to have especially if you don't' have to change the text formatting. E.g. maybe a colored flag or something similar

![screenshot 2018-04-03 11 55 22](https://user-images.githubusercontent.com/205275/38260902-c91cc6c4-3736-11e8-82e1-586cc97d658d.png)

Note we will have the same issues with generating colored warnings and errors if they are continued to be colored red, so it is something to keep in mind.

jjallaire:
I believe that right now R only tags output as "normal" or "error", and stderr, warnings, and R errors all get the error treatment.

If we wanted to be more granular than that I think we'd need to explicitly overwrite the `message()` and `warning()` functions in the base package (even then we wouldn't catch warnings generated from native code so there would be some inconsistency of presentation).

I think the point about red being a bit too harsh for all these conditions is a good one. While I agree that it would be great to have 4 different colors, if we can't achieve that (and I don't think we can without some major hacking) then softening the default red presentation seems like a good idea.

randomgambit:
amazing idea guys. You cant imagine how many times I missed that my code chunk actually didnt do anything because I didnt see that error line pop up among the output. Having errors in RED would be really useful!

breckuh:
[...]

Issue #1959, opened 2018-01-12
Xterm system colors (0-15) are not adjusted in all custom themes


jimhester:
These colors instead should be adjusted based on the theme, so the output matches what you get in other terminals with color theme (e.g. iTerm2) This screenshot is Solarized Light in iTerm2 (left) and RStudio (right). The difference is apparent particularly in the green and magenta colors.

![screenshot 2018-01-12 10 56 11](https://user-images.githubusercontent.com/205275/34883350-68ff997c-f787-11e7-960c-7676a5f1507f.png)

It looks like the `.xtermColor` and `.xtermBgColor` properties for numbers 0-15 were not adjusted for all the themes.

e.g. the solorized colors https://github.com/rstudio/rstudio/blob/0d9d49aa8671527370c6c261fc031ae8a8081c45/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/themes/solarized_light.css#L176-L206 colors 

are the same as the xcode colors
https://github.com/rstudio/rstudio/blob/b07f6cc97f8d14758702f72d1b391bc5b7e77286/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/themes/xcode.css#L172-L203

One theme where this _does_ seem to have been adjusted is dracula
https://github.com/rstudio/rstudio/blob/b07f6cc97f8d14758702f72d1b391bc5b7e77286/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/themes/dracula.css#L273-L304

jimhester:
This is still an outstanding issue

jimhester:
This is still an outstanding issue

PR #4069, opened 2018-12-20
Print Messages and Warnings differently than Errors

jimhester:
This is a simple proof of concept to address https://github.com/rstudio/rstudio/issues/2574

There are some open issues it would be nice to discuss.

Warnings seem to consist of 2 separate error calls resulting in two
separate span blocks. The first is just the text 'Warning:' the second
is the actual warning message. This causes the current code to only color the
text of 'Warning:', which I think actually looks pretty nice, maybe we
could do the same for errors?

Errors (at least short ones) have the entire error in one span, so the
whole thing is colored.

The search string assumes an English locale, it would likely have to be
localized, as the 'Error' text changes when other languages are used.

This only changes the behavior in the R shell, it seems
`getThemeErrorClass()` is called a few more places, they could possibly
need to be changed as well.

We would probably also need to change `consoleWriteExtendedError()`, but I
was unsure when this is called so did not know how to test it.

Fixes https://github.com/rstudio/rstudio/issues/2574

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/common/shell/ShellWidget.java
+++ b/src/gwt/src/org/rstudio/studio/client/common/shell/ShellWidget.java
@@ -263,7 +263,17 @@ public class ShellWidget extends Composite implements ShellDisplay,
    public void consoleWriteError(final String error)
    {
       clearPendingInput();
-      output(error, getErrorClass(), true /*isError*/, false /*ignoreLineCount*/);
+
+      String errClass;
+      if (error.startsWith("Error:")) {
+        errClass = getErrorClass();
+      } else if (error.startsWith("Warning:")) {
+        errClass = getWarningClass();
+      } else {
+        errClass = styles_.output();
+      }
+
+      output(error, errClass, true /*isError*/, false /*ignoreLineCount*/);
 
       // Pick up the elements emitted to the console by this call. If we get 
       // extended information for this error, we'll need to swap out the simple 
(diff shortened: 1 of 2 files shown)


jimhester:
This is how it looks with these changes and the Solarized dark theme.

screen shot 2018-12-20 at 4 56 41 pm

Also I don't know if this is intentional or not, but at least for the solarized themes the css for `ace_constant ace_numeric` gives you orange (`#cb4b16`) rather than red (`#dc322f`) https://github.com/brantb/solarized#the-values. It doesn't seem the red color is actually present in the css file for these themes (this is somewhat related to https://github.com/rstudio/rstudio/issues/1959)

jmcphers:
> Warnings seem to consist of 2 separate error calls resulting in two separate span blocks. The first is just the text 'Warning:' the second is the actual warning message. 

I think this is on the case for short warnings. With `warn=0`, I see `Warning message` instead of `Warning:` which we should probably treat identically.

image


> This only changes the behavior in the R shell, it seems getThemeErrorClass() is called a few more places, they could possibly need to be changed as well.

Yes, there are a few other places where we'd want this. We display R output in other panes too (the build pane, job output pane, etc.) and to the extent that we can we'd like output to have the same treatment in all of them.

> We would probably also need to change consoleWriteExtendedError(), but I was unsure when this is called so did not know how to test it.

The extended error includes a traceback and debugging options. You can trigger it as follows: first, put the two lines below in an R script:

    foo <- function(x) { print(x) }
    bar <- function(y) { foo(y) }

 Save the script, source it, then run `bar()`. You should get an extended error:

image

I think the only way we're going to be able to treat these things reliably is to either (a) provide our own warning() and message() functions which override the defaults, or (b) establish top-level condition handlers. We have a somewhat underhanded way of doing the latter which we use in R Notebooks:

https://github.com/rstudio/rstudio/blob/96fb791d197e0147589aa30625322a846263f246/src/cpp/session/modules/rmarkdown/NotebookConditions.cpp#L52-L74

jmcphers:
Closing this since it's been almost a year since it had any activity (feel free to re-open if there's progress!)

PR #4327, opened 2019-02-16
use RSTUDIO_CRASHPAD_DISABLED to disable crashpad integration

kevinushey:
This PR allows one to set `RSTUDIO_CRASHPAD_DISABLED` to disable Crashpad integration for a build. This is primarily for local development builds using release with debug info where we don't want to strip binaries.


kevinushey:
(alternatively maybe we only want to strip binaries when `RSTUDIO_PACKAGE_BUILD` is set?)

Issue #4408, opened 2019-03-06
IDE not finding spelling dictionaries


gtritchie:
Using latest 1.3 sources, in a dev server configuration. Suspect it may also happen in a desktop build installed normally but didn't confirm.

When starting the IDE, see several errors in JavaScript console. If you do a spell-check it doesn't know any words.

![2019-03-06_12-00-23](https://user-images.githubusercontent.com/10569626/53912184-389bd500-400d-11e9-8c93-c9c5b696d990.png)

adamconroy:
fixed in commit https://github.com/rstudio/rstudio/commit/9160e9e8d06105c9d973234af9a584e9269d4ce1

That commit will make it to v1.3 within a week.

gtritchie:
We don’t close issues until the commit makes it to master (or 1.3 in this case, which will swap places with master soonish). Closing an issue is a signal to QA that they can regress the issue with an official build. So would suggest doing a PR with just that commit, against v1.3, and include a fixes tag in the PR to auto-close the issue when the PR is accepted.

gtritchie:
Fixed in v1.3 branch.

PR #4415, opened 2019-03-07
fix realtime spelling working with non bundled dictionaries bug #4408

adamconroy:
Bugfix for issue #4408


gtritchie:
This does look correct, but we're still confirming that this is working as intended. Seeing some quirky stuff even with this change applied.

PR #4419, opened 2019-03-08
only attempt to change titlebar color on Mojave

kevinushey:
This should (🙏 ) fix the rendering issues seen on Sierra / High Sierra when using Qt 5.12.1 on macOS.

Can confirm at least it fixed things on my macOS High Sierra VM.

Closes https://github.com/rstudio/rstudio/issues/4409.


randy3k:
Awesome, thank you for the fix and I feel sorry about introducing the bug. (Though, I have never expected it has an effect on older macs).

PR #4470, opened 2019-03-18
Bionic dev bootstrapper script

gtritchie:
Setting up a new Ubuntu Bionic dev VM, and discovered that `install-dependencies-debian` script doesn't quite work for it. Openssl, among other things.

Maybe we could/should fix our debian script to handle all permutations and the new 1.3 dependencies, but instead I created a new bionic-specific script, based on installing the same stuff as the bionic Dockerfile.

I didn't try to separate out Pro-only stuff, so this script can be used to configure a machine for both open source and Pro development. It doesn't seem worth the effort or the minor disk-space savings to figure out what's Pro-only.

Unlike our other scripts, this also installs R (if not already installed), and the java-8 openjdk.

You can get a Linux dev environment up and running with the following:

- Install Ubuntu Bionic desktop (18.04.2)
- Install and configure git
- Clone open-source and/or pro repo
- `cd rstudio(-pro)/dependencies/linux`
- `sudo ./install-dependencies-bionic`
- for for lunch, then you can build RStudio using usual techniques

Need to "sudo" the install-dependencies script because the `install-crashpad` script that it invokes doesn't do its own "sudo-ing", unlike the other scripts. Not sure this is on purpose, or just an omission.

### DIFF
--- a/dependencies/linux/install-dependencies-bionic
+++ b/dependencies/linux/install-dependencies-bionic
@@ -36,14 +36,12 @@ sudo apt-get -y install \
   libclang-6.0-dev \
   libclang-dev \
   libcurl4-openssl-dev \
-  libdbi1 \
   libegl1-mesa \
   libfuse2 \
   libgl1-mesa-dev \
   libgtk-3-0 \
   libpam-dev \
   libpango1.0-dev \
-  librrd8 \
   libssl-dev \
   libuser1-dev \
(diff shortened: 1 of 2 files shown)


kfeinauer:
I would add some platform detection code to `install-dependencies-debian` and invoke this script if that platform is found.

PR #4475, opened 2019-03-19
right click context menu for real time spell checking

adamconroy:
A right click context menu for convenient mouse driven spell checking.
WBPbrhX


src/gwt/src/org/rstudio/studio/client/common/spelling/TypoNativeSettings.java
```diff
@@ -0,0 +1,19 @@
+package org.rstudio.studio.client.common.spelling;
```
jmcphers: Add copyright header.

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/AceEditor.java
```diff
@@ -39,18 +40,13 @@
 import com.google.gwt.event.shared.HandlerRegistration;
 import com.google.gwt.user.client.Command;
 import com.google.gwt.user.client.Timer;
-import com.google.gwt.user.client.ui.RootPanel;
-import com.google.gwt.user.client.ui.ScrollPanel;
-import com.google.gwt.user.client.ui.Widget;
+import com.google.gwt.user.client.Window;
+import com.google.gwt.user.client.ui.*;
```
jmcphers: Did your IDE do this? We generally avoid wildcard imports (See https://github.com/rstudio/rstudio-pro/wiki/Coding-Standards)

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTargetSpelling.java
```diff
@@ -39,6 +45,12 @@
 
 public class TextEditingTargetSpelling implements TypoSpellChecker.Context
 {
+   interface Resources extends ClientBundle
+   {
+      @Source("../../../../commands/goToWorkingDir.png")
```
jmcphers: Since these icons are semantically different I think they should be different files (even if they are visually identical to begin with!). Would recommend making a copy of the PNG with an appropriate name and placing it in this folder so you don't need to reach outside the directory to form the bundle.

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTargetSpelling.java
```diff
@@ -168,6 +182,78 @@ void onDismiss()
...
+            true,
+            () -> {
+               typoSpellChecker_.addToUserDictionary(word);
+               docDisplay_.removeMarkersAtCursorPosition();
+            });
+
+         menu.addItem(addToDictionaryItem);
+
+
+         menu.setPopupPositionAndShow((offWidth, offHeight) -> {
```
jmcphers: Do we need to account for the boundaries of the editor here?  (thinking e.g. source windows where going outside the boundaries of the editor probably means getting clipped)

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/AceEditor.java
adamconroy: Yeah, IntelliJ changes it to wildcard if you import everything in the package. It's smart about not importing too much/too little. I can change the setting though.

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTargetSpelling.java
```diff
@@ -168,6 +182,78 @@ void onDismiss()
          releaseOnDismiss_.remove(0).removeHandler();
    }
 
+   private void injectContextMenuHandler()
+   {
+      docDisplay_.addContextMenuHandler((event) ->
+      {
+         // If we have a selection, just return as the user likely wants to cut/copy/paste
```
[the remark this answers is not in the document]
kevinushey: There's also other actions that we make available (Reload, Inspect Element) that some users would find frustrating to lose.

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTargetSpelling.java
adamconroy: I tried scrolling a lot left/right and it always felt alright to me. I'll play around with it more to confirm.

kevinushey:
Enabling context menus in source editors opens a bit of a can of worms for us. In the past, we've relied on the Qt side / browser to present the user with the appropriate set of commands (e.g. cut / copy / paste). I think that if we're going to present the user with a context menu within the editor then we need to think about (1) how can we make sure this will extend cleanly to other editor modes / requested actions, and (2) how we can preserve the regular set of Cut / Copy / Paste actions.

In RStudio Server, can a user still trigger a Paste action from a GWT context menu? If we can't I think this will be a problem for users.

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTargetSpelling.java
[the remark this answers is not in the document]
adamconroy: Pasting when nothing is selected is a good point. I wanted to keep this as simple as possible but maybe that isn't possible. Cut/copy/paste should be viable to get back in.

Reload/Inspect Element are still accessible when clicking anywhere that isn't the AceEditor.

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTargetSpelling.java
```diff
@@ -168,6 +182,78 @@ void onDismiss()
          releaseOnDismiss_.remove(0).removeHandler();
    }
 
+   private void injectContextMenuHandler()
+   {
+      docDisplay_.addContextMenuHandler((event) ->
```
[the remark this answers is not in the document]
adamconroy: My instinct is that, while we're opening the door now and this a bit overly tied to the Spelling target, at the moment, it's a first pass that can be changed as soon as we know that this system will be used more. I'm open to planning it for additional features if we think we'll be adding them sooner than later though!

kevinushey:
Also wonder if we want to place the spelling commands within a sub-menu, e.g.

```
---------------------------------
| Cut                           |
| Copy                          |
| Paste                         |
| Select All                    |
| ---                           |
| Spelling -> word1             |
|             word2             |
|             word3             |
|             -----             |
|             Add to Dictionary |
| ---                           |
| Reload                        |
| Inspect Element               |
---------------------------------
```

adamconroy:
A quick google isn't encouraging for keeping inspect element in the right click menu. Reload is probably doable since it's just a Window.reload() action but I'm seeing a lot of negative answers when it comes to calling inspect element from JS.

adamconroy:
For quick spell checking I really like having the actions be top level. Since a suggestion would only show up in the situation where we think we want to be spell checking (much of markdown, only the comments in an R code file) I'd like to keep the current flat hierarchy at the moment.
jI2XKXA

kevinushey:
Fair enough! I agree we would want a 'More...' entry and limit the number of suggestions to 3-5 or so.

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTargetSpelling.java
adamconroy: Nvm I didn't account for it cutting off due to being constrained by the chromium window. Got it under control.

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/AceEditor.java
gtritchie: In IntelliJ prefs: Editor/Code Style/Java/Imports, and increase number for "Class count to use import with '*':" to something like 99 to prevent this altogether.

PR #4489, opened 2019-03-20
more fixes for new behavior of String.charAt in GWT 2.8.2

gtritchie:
GWT 2.8.2 added range check + exception to String.charAt (standard Java behavior, but was previously missing).

Per advice from @kevinushey in prior round of fixes, added a "safe" StringUtil.charAt that can be used in places where it isn't trivial to accommodate the new behavior.

This particular instance was being hit by autocomplete code when typing near first column of the R Console.


kevinushey:
LGTM -- thanks for making that change!

PR #4506, opened 2019-03-25
Fix an issue caused by old versions of RStudio saving themes with a leading slash.

### LINKED ISSUE #4488: Rstudio desktop free preview 1.2.1329 - can't select text in editor window
I just upgraded from 1.2.1327 to 1.2.1329 and can no longer select text in (at least) the editor window. Click-drag does not select text. Double-click does not select text. Editing is almost unusable.

The behaviour survived a restart of Rstudio and a reboot of the computer. Other programs were fine.

I reverted to 1.2.1327 and the problem disappeared.

I am running the Ubuntu 18+ versions of Rstudio.

My computer is running Ubuntu 18.04 4-bit.

MariaSemple:
Closes #4488.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/themes/AceTheme.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/themes/AceTheme.java
@@ -59,7 +59,10 @@ public class AceTheme extends JavaScriptObject
    
    public native final String getUrl()
    /*-{
-      return this.url;
+      if (this.url.startsWith("/"))
+         return this.url.substr(1);
+      else
+         return this.url;
    }-*/;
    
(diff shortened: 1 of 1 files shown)


src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/themes/AceTheme.java
```diff
@@ -59,7 +59,10 @@ public native final String getName()
    
    public native final String getUrl()
    /*-{
-      return this.url;
+      if (this.url.startsWith("/"))
```
kevinushey: `startsWith()` is a relatively new addition; it became part of JavaScript with ES6: https://www.w3schools.com/jsref/jsref_startswith.asp and may be missing for users with RStudio Server on older browsers.

PR #4587, opened 2019-04-05
prevent creation of crashpad console window

### LINKED ISSUE #4499: An empty console window with crashpad_handler.com opens with RStudio
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.3.66
    OS Version      : Windows 10
    R Version       : 3.5.3

This is what it looks like:

Not much more to say really. The window appears right away when opening RStudio. Actually it appears a tick before the RStudio window.

kevinushey:
Closes #4499. It looks like `crashpad_handler.exe` launches without issue in package builds, so we now direct RStudio to use that version in those cases.

I also noticed that our package builds were selecting `/subsystem:console` when linking `rstudio.exe`, but (to my surprise) avoiding that and enforcing `/subsystem:windows` didn't seem to make a difference. Fortunately, this does and `crashpad_handler.exe` does seem to launch without issue.

### DIFF
--- a/src/cpp/core/CrashHandler.cpp
+++ b/src/cpp/core/CrashHandler.cpp
@@ -244,7 +244,7 @@ Error initialize(ProgramMode programMode)
             handlerPath = googleFilePath(exePath.parent().childPath("crashpad_handler").absolutePath());
          #endif
       #else
-         handlerPath = googleFilePath(exePath.parent().childPath("crashpad_handler.com").absolutePath());
+         handlerPath = googleFilePath(exePath.parent().childPath("crashpad_handler.exe").absolutePath());
 #endif
    }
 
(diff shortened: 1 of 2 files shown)


kfeinauer:
Have you tested to make sure you can actually generate a crash with the exe? I don't remember exactly why I had to use the com file, but I think the exe file wasn't doing what I wanted.

Also, I think we should just use the exe in both the debug and package builds if we can.

kevinushey:
I just tried forcing an rsession crash and can confirm a minidump was written to the crashpad database, so I think it's working as intended. Is there anything else I can do to test?

gtritchie:
Wow, is that really a .com file? Those are 16-bit DOS executables! Haven't seen one of those since I had hair!

kfeinauer:
@kevinushey I think that should all be working now as intended. I'd just go ahead and use the exe for all windows configurations, and I call this PR good to go.

PR #4601, opened 2019-04-09
support for building with address-sanitizer (ASAN)

gtritchie:
Provide support for building RStudio with Address Sanitizer support on Mac and Linux (works with clang, including Xcode's, or recent gcc's). Diagnostic tool for runtime detection of various C++ memory misuses. To use, run CMake with RSTUDIO_ASAN defined and rebuild. For example, in CLion I do my debug generation via:

In a perfect world we'd have massive automated code coverage with fuzzing, and would run it in conjunction with an address-sanitizer build. For now, I'm just doing my development with it enabled to see if anything crops up in routine usage.

I did confirm this was working by putting in some temporary badness and observing the output (writes oodles of details to stderr and terminates).

To get more details on a problem, will be necessary to have `llvm-symbolizer` in the path or pointed to via `ASAN_SYMBOLIZER_PATH` environment variable, but even without that it gives quite useful information.

There are various other sanitizers in this Google suite; haven't tinkered with them yet.

Good overview here:
https://en.wikipedia.org/wiki/AddressSanitizer

More details here:
https://clang.llvm.org/docs/AddressSanitizer.html

CLion has some integration of this and other Google sanitizers, though I haven't played with it:
https://www.jetbrains.com/help/clion/google-sanitizers.html

### DIFF
--- a/src/cpp/CMakeLists.txt
+++ b/src/cpp/CMakeLists.txt
@@ -14,7 +14,7 @@
 #
 
 # set minimum version
-cmake_minimum_required(VERSION 2.6)
+cmake_minimum_required(VERSION 3.4.3)
 
 project (RSTUDIO_CPP)
 
@@ -68,6 +68,12 @@ if(UNIX)
 
    # compiler flags
    add_definitions(-Wall -pthread)
+
+   if(RSTUDIO_ASAN)
+      message(STATUS "Compiling with Address Sanitizer (ASAN) support")
+      add_definitions(-fsanitize=address -fno-omit-frame-pointer -fno-optimize-sibling-calls -O1)
+      set(CMAKE_EXE_LINKER_FLAGS  "${CMAKE_EXE_LINKER_FLAGS} -fsanitize=address")
+   endif()
 
    if(APPLE)
       add_definitions(-Wno-unknown-warning-option -Wsign-compare -Wno-unused-local-typedefs


kfeinauer:
Do we want to just do this implicitly for debug builds?

gtritchie:
> Do we want to just do this implicitly for debug builds?

We could. My main concern was the overhead; according to the official page, typically slows things down 2x. Wikipedia says "increases processing time by about 73% and memory usage by 340%".

kevinushey:
I put together some scripts a long time ago for building RStudio with ASAN / UBSAN in the past:

https://github.com/rstudio/rstudio/tree/master/src/tools

Those are pretty stale now though -- maybe we can just remove those and make `RSTUDIO_ASAN` the right way to request sanitizer builds?

What about also allowing e.g. `RSTUDIO_SANITIZERS=address,undefined` to give the user more control over which sanitizers are enabled?

gtritchie:
> What about also allowing e.g. `RSTUDIO_SANITIZERS=address,undefined` to give the user more control over which sanitizers are enabled? (and then appending the necessary compiler flags to ensure that we get good stack traces etc.)

I was going to do something along those lines (allow specifying multiple sanitizers) but don't have time right now to try them out so figured I'd get this one in as-is.

Didn't know about your scripts, but I think having this done via cmake is probably the way to go?

kevinushey:
Sounds good. I think I'm bikeshedding here so ultimately LGTM :-)

gtritchie:
The real question is how many espresso machines are needed per-bikeshed.

PR #4614, opened 2019-04-10
add bug report helper

kevinushey:
This PR adds a (currently internal) helper for bug reports. This is primarily to save us (and, if we make this an official API, our users) the time in pre-populating some of the fields we care about when forming a bug report.

One can call `.rs.bugReport()` to generate the template. When `clipr` is available, we write the rendered template to the clipboard; otherwise, we just write it to the console and let the user copy + paste.


kevinushey:
We might also want to just implement this in `rstudioapi` directly since that would imply it could work with older versions of RStudio as well.

Issue #2364, opened 2018-03-09
ANSI-colored text in Notebook preview rendered incorrectly


kevinushey:
### System details

    RStudio Edition : Desktop
    RStudio Version : v1.1.423
    OS Version      : Ubuntu 16.04
    R Version       : 3.4.3

### Steps to reproduce the problem

Create an R Notebook with the contents:

`````
---
title: "R Notebook"
output: html_notebook
---

```{r}
library(tidyverse)
```
`````

Then, run the chunk, and attempt to preview output.

### Describe the problem in detail

The ANSI escape codes are not properly interpreted, and so output is not properly displayed in the preview.

### Describe the behavior you expected

RStudio should either strip ANSI escapes it doesn't understand, or figure out a mechanism for translating them into the appropriate HTML for rendering.

### Notes

![screen shot 2018-03-09 at 9 45 28 am](https://user-images.githubusercontent.com/1976582/37221414-b3cac09e-237e-11e8-9734-2a71de3f7816.png)

![screen shot 2018-03-09 at 9 46 00 am](https://user-images.githubusercontent.com/1976582/37221418-b7de0c2c-237e-11e8-8a29-8caa3edd0f43.png)

https://community.rstudio.com/t/character-encoding-issue-when-previewing-notebook/6031/3

GegznaV:
### System details

    RStudio Edition : Desktop 
    RStudio Version : 1.2.1237
    OS Version      : Windows 10
    R Version       : 3.5.2

### Steps to reproduce the problem

1. Create a new RStudio notebook with one chunk. Chunk options `paged.print=FALSE`, code `tibble::as.tibble(head(iris))`. 
2. Run code in the chunk.
3. Push "Preview" button ![image](https://user-images.githubusercontent.com/12725868/51610800-58c47880-1f26-11e9-9ac0-2734dac561c6.png)

![image](https://user-images.githubusercontent.com/12725868/51610944-be186980-1f26-11e9-91ef-e9aef4df3440.png)


### Describe the problem in detail
Output in preview window is distorted, e.g., `[38;5;250m1[39m` appears:

![image](https://user-images.githubusercontent.com/12725868/51611015-dee0bf00-1f26-11e9-841b-eeb000f1a981.png)


### Describe the behavior you expected

In preview window, I expect the same not distorted contents as in RMarkdown window.

GegznaV:
Is there any work-around possible before RStudio 1.3 is released?

kevinushey:
[...]

PR #4707, opened 2019-04-25
remove ANSI escapes during notebook preview

kevinushey:
Closes https://github.com/rstudio/rstudio/issues/2364.

### DIFF
--- a/src/cpp/session/modules/SessionRmdNotebook.R
+++ b/src/cpp/session/modules/SessionRmdNotebook.R
@@ -557,6 +557,9 @@ assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
    if (length(csvData) == 0 || nrow(csvData) == 0)
       return(list())
    
+   # remove ANSI escapes used for color from text
+   csvData$text <- gsub("\033\\[\\d+m", "", csvData$text)
+   
    # split on type
    cutpoints <- .rs.cutpoints(csvData$type)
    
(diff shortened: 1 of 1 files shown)


src/cpp/session/modules/SessionRmdNotebook.R
```diff
@@ -557,6 +557,9 @@ assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
    if (length(csvData) == 0 || nrow(csvData) == 0)
       return(list())
    
+   # remove ANSI escapes used for color from text
+   csvData$text <- gsub("\033\\[\\d+m", "", csvData$text)
```
gtritchie: Don't know if it's possible here, but it would be valid ANSI escape to include multiple numbers (2 or more) separated by semicolons, e.g.: ESC[1;32m is "bold green".

gtritchie: Also, it is valid to have **no** digits, ESC[m means "reset to defaults".

src/cpp/session/modules/SessionRmdNotebook.R
```diff
@@ -557,6 +557,9 @@ assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
    if (length(csvData) == 0 || nrow(csvData) == 0)
       return(list())
    
+   # remove ANSI escapes used for color from text
+   csvData$text <- gsub("\033\\[\\d*(?:;\\d*)?m", "", csvData$text)
```
gtritchie: Per my comment "valid ANSI escape to include multiple numbers (2 or more)" the final "?" should be a "*". The "bold" example was just that, there are multiple attributes that can be specified in a single one of these sequences. "Bold italic foreground background" and so forth.

kevinushey: Thanks! Sorry I missed that in your original comment.

PR #4748, opened 2019-05-03
reduce unnecessary inclusion of websocketpp headers

gtritchie:
The websocketpp headers spit out quite a few build-warnings, noticed during Windows build that source files that shouldn't care less about websockets were generating these warnings.

SessionConsoleProcess.hpp was pulling these in via SessionConsoleProcessSocket.hpp, just to get at the ConsoleProcessSocketConnectionCallbacks struct. Refactored that struct into a separate header.

This takes the Windows build warning count (debug) from 144 to a mere 96. Almost respectable. Kinda a hobby at this point.

Also unpinned the git version number in the bootstrapper script; the dockerfile doesn't specify a version, so neither should this script.

### DIFF
--- a/dependencies/windows/Install-RStudio-Prereqs.ps1
+++ b/dependencies/windows/Install-RStudio-Prereqs.ps1
@@ -68,7 +68,7 @@ refreshenv
 choco install -y jdk8
 choco install -y ant --version 1.10.5
 choco install -y 7zip --version 18.5.0.20180730
-choco install -y git --version 2.18.0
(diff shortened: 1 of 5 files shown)


kfeinauer:
Is there a pragma to ignore warnings in specific headers?

gtritchie:
> Is there a pragma to ignore warnings in specific headers?

Yes, that will be my final stage in this. This seems an easy and justifying tweak first.

PR #4769, opened 2019-05-07
fix failure installing ant in windows dockerfile, registing msdia140.dll

gtritchie:
Ant was failing to install the JRE (which is already installed via
the JDK). Use -i to have ant not worry about dependencies, and
stop requesting specific version of ant and 7zip (while we're at it)
because having latest version of these tools is fine.

Still having problems getting msdia140.dll to regsvr32 inside the
container so moved it to the install-dependencies script.


gtritchie:
Committing this to give it a try in the spot instance.

PR #4789, opened 2019-05-12
VirtualConsole bug when rewriting same line with certain color range pattern

### LINKED ISSUE #4777: Garbled text / buffering issue with fast updates and colored text
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.12.1335
    OS Version      :  macOS Mojave
    R Version       : 3.6

Run the following code in the RStudio console pane. The first two `show_progress()` calls should work as expected, the third generates garbled text.

```r
library(crayon)
library(progress)

show_progress <- function(format, size, width = 80) {
  pb <- progress_bar$new(total = size, format = format, width = width)
  for (i in 1:size) {
    pb$tick()
    Sys.sleep(1 / size)
  }
}

# Works with colors and 1 ms updates
show_progress(":bar " %+% green(":rate"), 1000)

# Works without colors and .1 ms updates
show_progress(":bar " %+% ":rate", 10000)

# Doesn't work with colors and .1 ms updates
show_progress(":bar " %+% green(":rate"), 10000)
```

The same code works fine in a terminal outside of RStudio and even in a R session started in the RStudio terminal pane.

My guess is there is some sort of buffering going on in the RStudio console that is messing up the updates when they happen quickly enough.

gtritchie:
Fixes #4777

I think this is reasonably safe to take for 1.2-patch. This code is finicky but does have good unit test coverage.

### DIFF
--- a/NEWS-1.2-patch.md
+++ b/NEWS-1.2-patch.md
@@ -24,7 +24,7 @@
 - Fixed an issue where the Files pane occasionally would fail to scroll to bottom (#4662)
 - Fixed an issue where RStudio would always use the discrete GPU on macOS (#4672)
 - Fix startup failure when using multiple CRAN repos (#4751)
-
+- Fix console display issue with certain mixed color output patterns (#4777)
 
 ### Miscellaneous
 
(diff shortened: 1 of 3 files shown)


jimhester:
Thanks for working on this @gtritchie!

PR #4845, opened 2019-05-25
fix incorrect return value for some rs_ functions

kevinushey:
This PR fixes the return types for a couple R functions we define. In particular, this fixes a crash that occurs when:

```
.Call("rs_packageLibraryMutated")
```

is invoked directly. (Miraculously, this still worked somehow where we made calls as e.g. `on.exit(.Call("rs_packageLibraryMutated"))`, but that seems like it was just dumb luck?

Candidate for v1.2 backport.


kevinushey:
I agree this is worth backporting (will do after merge)

Issue #3420, opened 2018-09-03
Margin doesn't show on dark themes


EdwinTh:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.1.453
    OS Version      :  macOS 10.13.6
    R Version       : 3.5.0

### Steps to reproduce the problem

Switch back and forth between a light and a dark theme as described below.

### Describe the problem in detail

The margin line that can be enabled with Tools > Global Options > Code > Display tab > show margin (Margine column 80) does only show when a light theme is set. When switching to a dark theme, for instance "Material" the line is invisible. When switching back to a light theme, such as "Eclipse", it shows again. 

Most likely the line is applied in dark within the dark theme. The resolve would then be to give the line a lighter colour when a dark theme is applied.

### Describe the behavior you expected

The margin line is visible in a dark theme.

ronblum:
@EdwinTh I can reproduce this on RStudio Desktop (MacOS) and Server (SUSE), both 1.1.456 and 1.2.957. However, it appears to be theme-specific, only an issue with the Material theme. Other dark themes, such as Dracula and Twilight, work fine. Let me know if you see something different.

EdwinTh:
Confirming that the other dark themes do have the margin line visible. I incorrectly assumed it was a dart theme thing, it is Material specific.

DavisVaughan:
Confirming that the `.ace_print-margin` color is not showing up for Material.

This color specification seems to look good (slightly noticeable, not outlandish), but I'm sure you guys have a material palette to pick from

```
.ace_print-margin {
  width: 1px;
  background: rgb(47, 57, 65)
}
```

mjsteinbaugh:
I'm noticing a visual issue with the [Dracula](https://draculatheme.com/) color theme that may be related to dark mode. On my RStudio install (v.1.1.423), the margin guide for the Dracula is jarringly bright. In don't have such an issue with the other dark themes (e.g. Tomorrow Night).

When looking at the [dracula.rstheme](https://github.com/rstudio/rstudio/blob/master/src/cpp/session/resources/themes/dracula.rstheme) file, here's the current config:

```
.ace_print-margin {
  width: 1px;
  background: #e8e8e8;
}
```

`#e8e8e8` seems to be the culprit here, and should be darkened to a different color in the [recommended palette](https://github.com/dracula/dracula-theme#color-palette).

rovyko:
[...]

PR #4737, opened 2019-05-01
Improve margin and highlight color settings for Dracula theme.

mjsteinbaugh:
Related to issue #3420, where margin isn't set correctly for some dark
themes. Highlight color is based on [Dracula BBEdit theme](https://draculatheme.com/bbedit/).

PR #4863, opened 2019-05-30
Modify material theme editor margin color to make it visible

rovyko:
The material theme's editor max-width column's color is modified to subtly distinguish it from the background and makes it visible, where it previously was not.

partially addresses #3420


rovyko:
Thank you Maria and Jonathan for your quick responses. This (along with PR #4737) fixes #3420

PR #4889, opened 2019-06-03
provide default RStudio HTTP user agent

### LINKED ISSUE #4826: Set the HTTPUserAgent Header Option to Include the R Version
Beginning in R 3.6, there are a number of cases where the http user agent header no longer includes the R version. 

https://github.com/wch/r-source/commit/def075db88ae87104c38437a57e9327b078bb804#diff-b499547952f75d41253bb11901257375R252

Instead, R recommends setting the header through the `HTTPUserAgent` option. 

It would (re)-enable quite a few use cases if RStudio were to set this option by default, including access to RStudio Package Manager package binaries.

RStudio could allow users to disable this behavior to prevent the case that motivated R to remove the header.

kevinushey:
Closes #4826.

With this change, the user agent is reported as:

```
> getOption("HTTPUserAgent")
[1] "RStudio Desktop (99.9.9); R (3.6.0 x86_64-apple-darwin15.6.0 x86_64 darwin15.6.0)"
```

but only if the user hasn't explicitly configured their own user agent.

IIUC, this could be a candidate for backport to v1.2 since we might need this change for RStudio Cloud.

@jmcphers let me know if you think we should be including other fields.

### DIFF
--- a/src/cpp/r/R/Api.R
+++ b/src/cpp/r/R/Api.R
@@ -84,19 +84,11 @@
 
 .rs.addApiFunction("versionInfo", function() {
   info <- list()
-  info$citation <- .Call(getNativeSymbolInfo("rs_rstudioCitation",
-                                             PACKAGE=""))
-
-  info$mode <- .Call(getNativeSymbolInfo("rs_rstudioProgramMode",
-                                         PACKAGE=""))
-
-  info$edition <- .Call(getNativeSymbolInfo("rs_rstudioEdition",
-                                            PACKAGE=""))
-
-  info$version <- package_version(
-    .Call(getNativeSymbolInfo("rs_rstudioVersion", PACKAGE=""))
-  )
-
+  info$citation <- .Call("rs_rstudioCitation", PACKAGE = "(embedding)")
+  info$mode <- .Call("rs_rstudioProgramMode", PACKAGE = "(embedding)")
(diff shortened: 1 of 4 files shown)


jspiewak:
It would be awesome if `lsb_release` were present if this included distributor id and release from:

```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.4 LTS
Release:	16.04
Codename:	xenial
```

kevinushey:
Rather than rely on existence of `lsb_release` command line tool, we'd probably want to read fields directly from `/etc/os-release` or `/etc/redhat-release`.

Which fields in particular do you want to see in the user agent?

jspiewak:
Even better. In the case of:

```
jss@builder2:~$ cat /etc/os-release
NAME="Ubuntu"
VERSION="16.04.4 LTS (Xenial Xerus)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 16.04.4 LTS"
VERSION_ID="16.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
VERSION_CODENAME=xenial
UBUNTU_CODENAME=xenial
```

ID and VERSION_ID.

jspiewak:
Code name might be nice too.

Issue #4894, opened 2019-06-04
Page Down no longer scrolls down in Console history


kevinushey:
### System details

    RStudio Edition : Desktop [Open Source]
    RStudio Version : 1.2.1522
    OS Version      : macOS Mojave 10.14.5
    R Version       : R version 3.6.0 Patched (2019-05-29 r76623)

### Steps to reproduce the problem

Click the Console pane, and then submit some code (or just hold enter) to fill the Console history enough that a scrollbar is shown.

Then, try pressing <kbd>Page Up</kbd> to scroll back in the history. This should work as expected. Then, press <kbd>Page Down</kbd> to try and scroll back down.

### Describe the problem in detail

Nothing happens in response to <kbd>Page Down</kbd>.

### Describe the behavior you expected

We should scroll down.

ronblum:
Testing note: cf <kbd>Fn + Down</kbd>.

ronblum:
Verified in RStudio Desktop 1.2.1539 on MacOS 10.14.6.

mpr1255:
FWIW this is still a problem as of 

```
sysname           release           version             machine            
Windows"          "10 x64"     "build 19042"    "x86-64"              
```

mpr1255:
Note: It's actually possible to page down in the console window (`Juliet Rose" (2570ac7f, 2021-02-22) for Windows`) but only after *clicking* in it. If you don't click in the console window, it won't page down, only page up.

mpr1255:
I suggest this be re-opened. I just got R Studio server going on WSL1 Ubuntu 20.04 in part to avoid this issue -- and get native support for UTF-8 -- but the matter persists! So it cannot be a Windows problem. The console needs to be given a mouse click in order for page-down to work. I am now figuring out how to generate a mouse click with a keyboard shortcut.

PR #4900, opened 2019-06-05
handle PageDown when console has focus

kevinushey:
Closes https://github.com/rstudio/rstudio/issues/4894.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/common/shell/ShellWidget.java
+++ b/src/gwt/src/org/rstudio/studio/client/common/shell/ShellWidget.java
@@ -132,15 +132,16 @@ public class ShellWidget extends Composite implements ShellDisplay,
             if (input_.isPopupVisible())
                return;
             
-            // If the user hits Page-Up from inside the console input, we need
-            // to simulate pageup because focus is not contained in the scroll
-            // panel (it's in the hidden textarea that Ace uses under the
-            // covers).
+            // If the user hits PageUp or PageDown from inside the console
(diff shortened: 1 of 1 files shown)


src/gwt/src/org/rstudio/studio/client/common/shell/ShellWidget.java
```diff
@@ -132,27 +132,50 @@ public void onKeyDown(KeyDownEvent event)
...
+                  
+                  if (scrollPanel_.isScrolledToBottom())
+                     return;
+                  
+                  int newScrollTop =
+                        scrollPanel_.getVerticalScrollPosition() +
+                        scrollPanel_.getOffsetHeight() -
+                        40;
+                  
+                  scrollPanel_.focus();
```
gtritchie: Does this take away user's ability to continuing typing into the console? Anytime I see an explicit `focus()` call there's an automatic keyboard accessibility concern.

gtritchie: FWIW, I haven't dug into accessibility of the console pane yet (it's a minefield of potential problems).

kevinushey: This appears to work transparently -- when we get a key event while the console history has focus, we re-route focus back to the console input (so that it's "as though" the console input had focus all along)

There's probably further tweaks required to make this properly accessible, but I think this should suffice for now (and at least for v1.2-patch).

Issue #4916, opened 2019-06-07
Connect deployments fail with R 3.4.4


kevinushey:
### System details

    RStudio Edition : Desktop [Open Source]
    RStudio Version : 1.2.1522
    OS Version      : macOS  10.14.5
    R Version       : R version 3.4.4 (2018-03-15)

### Steps to reproduce the problem

Compare:

### R 3.4.4

```
> .rs.CRANDownloadOptionsString()
[1] "structure(list(repos = structure('https://cran.rstudio.com/', .Names = 'CRAN', RStudio = TRUE), download.file.method = 'libcurl'), .Names = c('repos', 'download.file.method'))"
```

### R 3.6.0

```
> .rs.CRANDownloadOptionsString()
[1] "options(repos = c(CRAN = 'https://cran.rstudio.com'), download.file.method = 'curl')"
```

We're incorrectly constructing the options object in the first case. This implies deployments using custom CRAN repositories could fail with this version of R.

ncalies:
Customer request - 36082

PR #4919, opened 2019-06-07
use deparse more carefully when forming repos string

kevinushey:
Closes https://github.com/rstudio/rstudio/issues/4916.

### R 3.4.4

```
> .rs.CRANDownloadOptionsString()
[1] "options(repos = structure('https://cran.rstudio.com', .Names = 'CRAN'), download.file.method = 'libcurl')"
```

### R 3.6.0

```
> .rs.CRANDownloadOptionsString()
[1] "options(repos = c(CRAN = 'https://cran.rstudio.com'), download.file.method = 'libcurl')"
```

The output still differs, but is now at least correct in both cases.

### DIFF
--- a/src/cpp/session/modules/SessionCodeTools.R
+++ b/src/cpp/session/modules/SessionCodeTools.R
@@ -1905,17 +1905,23 @@
    # drop NULL entries
    options <- Filter(Negate(is.null), options)
    
-   # deparse to generate an R string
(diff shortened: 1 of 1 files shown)


kevinushey:
@jmcphers can you sanity-check? This fixes an issue where attempts to deploy projects when using R 3.4.4 can fail so I think it's a candidate for v1.2-patch.

PR #4921, opened 2019-06-08
Annotate checkbox and radio menuitems so a11y tools can see the state

### LINKED ISSUE #4911: Menu items with checkboxes not properly annotated for a11y tools
### System details

    RStudio Version : 1.3.230
    OS Version      : any
    R Version       : any

On RStudio Server, main menu items with checkmarks (e.g. Debug / On Error / ..., View / Panes / Show All Panes) do not have markup that allows the checked/unchecked state to be seen by screen readers.

On RStudio Server and Desktop, pop-up menus with checkmarks such as the Workspace Panes toolbar button do not have markup that allows the checked/unchecked state to be seen by screen readers.

Screen reader users are unable to tell that these items are checkable, and what the current state is.

Also, for mutually exclusive items (radio-button like, such as those under Debug / On Error, there is no annotation to indicate they behave this way. This is also the case for sighted users, who kinda just have to figure it out.

At a minimum, screen readers need to tell user a given menu item is checkable, and indicate the current checked state, via `role=menuitemcheckbox` and `aria-checked=true`.

It would be nice if there was also indication of mutually exclusive items via `role=menuitemradio`.

gtritchie:
Fixes #4911 

Menu items for checkable commands are now marked with `role=menuitemcheckbox` and the `aria-checked` property is set to match.

Menu items that are checkable and part of a mutually exclusive set (only one can be selected) are now marked with `role=menuitemradio` and `aria-checked` is set to match state.

For testing, recommend NVDA on Windows (VoiceOver doesn't currently distinguish between regular and radio-variations).

Try out checkable items on the main menu (RStudio server only; the main menu on desktop is native code and not impacted by these changes), including regular checkable items and, for example, this menuitemradio set under Debug / On Error:

Also test non-main-menu checkable items, as those involve different code paths and are on both server and desktop. For example, in the File pane More dropdown, the `Show Hidden Files` item:

### DIFF
--- a/src/gwt/src/com/google/gwt/updatesources.sh
+++ b/src/gwt/src/com/google/gwt/updatesources.sh
@@ -32,7 +32,7 @@ patch SplitPanel.java < SplitPanel.java.diff
 patch SplitLayoutPanel.java < SplitLayoutPanel.java.diff
 
 cd ${SRCBASE}/com/google/gwt/core/ext/linker/impl
-patch XinstallLocationIframe.js < installLocationIframe.js.diff
+patch installLocationIframe.js < installLocationIframe.js.diff
 
 cd ${SRCBASE}/com/google/gwt/core/linker
-patch XCrossSiteIframeLinker.java < CrossSiteIframeLinker.java.diff
+patch CrossSiteIframeLinker.java < CrossSiteIframeLinker.java.diff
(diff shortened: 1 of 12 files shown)


gtritchie:
> I wonder if we should visually distinguish radio menuitems from checkable ones too (re-using the check was a lazy move on my part when I added these!).

I considered that; on server could easily swap in a different graphic (round bullet, perhaps), but on desktop, would require some additional plumbing to create a QActionGroup to hold the mutually exclusive menu items. Doable, but given how few of these we have didn't think it was worth the effort.

gtritchie:
Opened #4923 to track potential additional improvements with mutually exclusive menu items.

Issue #4677, opened 2019-04-19
View / Show Launcher menu appearing on Open Source


gtritchie:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.2.1335 and 1.2.1502
    OS Version      : macOS 10.14.4
    R Version       : 3.5.1

### Steps to reproduce the problem
Open View menu.

### Describe the problem in detail
The bottom menu item is "Show Launcher" and does nothing when you click it.

### Describe the behavior you expected
That menu item is not supposed to be there on open-source; it's for a Pro-only feature.

kevinushey:
Spent some time digging into this. If I understand correctly, the code that we use to manage commands is actually running _before_ we've even registered the commands on the desktop side (!?!).

Here's some debug logging I added:

```
15:54:56: Starting /Users/kevin/rstudio/src/build-cpp-Desktop_Qt_5_12_1_clang_64bit-Debug/desktop/RStudio.app/Contents/MacOS/RStudio ...
[28154:265950:20190605,155456.477054:FATAL double_fork_and_exec.cc:126] execvp /Users/kevin/rstudio/src/build-cpp-Desktop_Qt_5_12_1_clang_64bit-Debug/desktop/RStudio.app/Contents/MacOS/crashpad_handler: No such file or directory (2)
[28151:265942:20190605,155456.477504:ERROR file_io.cc:89] ReadExactly: expected 8, observed 0
void rstudio::desktop::MenuCallback::setCommandLabel(QString, QString) failed:  "setWorkingDir"
void rstudio::desktop::MenuCallback::setCommandLabel(QString, QString) failed:  "reformatCode"
void rstudio::desktop::MenuCallback::setCommandLabel(QString, QString) failed:  "newSourceDoc"
void rstudio::desktop::MenuCallback::setCommandLabel(QString, QString) failed:  "newRNotebook"
void rstudio::desktop::MenuCallback::setCommandLabel(QString, QString) failed:  "newTextDoc"
void rstudio::desktop::MenuCallback::setCommandLabel(QString, QString) failed:  "newCppDoc"
void rstudio::desktop::MenuCallback::setCommandLabel(QString, QString) failed:  "newPythonDoc"
void rstudio::desktop::MenuCallback::setCommandLabel(QString, QString) failed:  "newStanDoc"
void rstudio::desktop::MenuCallback::setCommandLabel(QString, QString) failed:  "newD3Doc"
[...]

PR #4926, opened 2019-06-10
properly toggle duplicated commands in menus

kevinushey:
Closes https://github.com/rstudio/rstudio/issues/4677.

The previous version of the PR failed to account for actions associated with previous RStudio session (e.g. if the page had been reloaded, or similar). This PR ensure the QAction pointers we work with are non-null, and also ensures that null actions get pruned after a page finishes loading.

There may well be a better place to handle the 'pruning' of commands -- let me know if you have an idea, but this overall seemed the safest way forward.


kevinushey:
Nothing that I can think of, either. Fortunately the v1.2-patch runway is still long enough that we'll have an opportunity to shake out these sorts of issues if they pop up.

PR #4935, opened 2019-06-12
disable synthetic tablet events

kevinushey:
This PR fixes an issue where single-clicks with a Tablet pen would actually be registered as double-clicks.

See e.g. https://community.rstudio.com/t/rstudio-1-2-1335-on-ubuntu-18-04-2-with-wacom-intuos-small-tablet


kevinushey:
I think this should be safe for v1.2-patch as well. Fortunately the runway is long enough that if this breaks something I'm sure we'll hear about it!

PR #4941, opened 2019-06-12
make IDE inert during modal dialogs

### LINKED ISSUE #4912: Modal dialogs don't prevent tabbing into disabled UI from outside RStudio
### System details

    RStudio Edition : Server
    RStudio Version : 1.3.230
    OS Version      : any
    R Version       : any

In RStudio Server, bring up a modal dialog which has been fixed to keep tabbing inside the dialog. Currently About dialog and Tools / Global Options.

Now put focus outside RStudio altogether, such as on browser address bar. Then use tab key to get focus back to the RStudio IDE.

Focus goes to the main RStudio UI (e.g. menubar, toolbars, etc.) instead of to the modal dialog. You can even manipulate things that are "under" the modal dialog, type in the text editor or console, etc.

Be definition, modal dialogs MUST prevent interaction with the rest of the UI, even if you move focus outside RStudio and tab back in with the keyboard.

### Other Symptoms
With a modal dialog displayed, invoke VoiceOver rotor on Mac (VO+U) and view form controls. Would expect to only see things in the modal dialog, but see everything under the modal dialog, and can operate these controls via VoiceOver.

gtritchie:
- when modal dialog is displayed, apply inert attribute to the top-level containers of the IDE
- use Google's inert polyfill since no browsers support this directly
- also some tweaks to prevent dialog progressbars being seen by a11y tools when they are hidden
- fixes #4912 
- this may be very sensitive to different combinations of browser and screen-reader, see this article for more background: https://developer.paciellogroup.com/blog/2018/06/the-current-state-of-modal-dialog-accessibility/
- polyfill from: https://github.com/GoogleChrome/inert-polyfill


adamconroy:
TIL I learned about inert!

PR #4956, opened 2019-06-14
No longer stop processes with ::abort, instead use std::_Exit

kfeinauer:
This will perform a similar exit (no destructors run) without generating an exception. 

Fixes IDE-BACKEND-13, Fixes IDE-BACKEND-V, Fixes IDE-BACKEND-3C, Fixes IDE-BACKEND-2W, IDE-BACKEND-1Q, IDE-BACKEND-25

### DIFF
--- a/src/cpp/session/SessionMain.cpp
+++ b/src/cpp/session/SessionMain.cpp
@@ -1338,7 +1338,11 @@ void detectParentTermination()
    if (result == ParentTerminationAbnormal)
    {
       LOG_ERROR_MESSAGE("Parent terminated");
-      core::system::abort();
+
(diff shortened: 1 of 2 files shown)


kevinushey:
We also call abort() here:

https://github.com/rstudio/rstudio/blob/a6ccdbfcdc9d1ce787b5b4d75272c7a282578f27/src/cpp/core/include/core/http/AsyncServerImpl.hpp#L502

Any thoughts on whether this should be `std::_Exit()` as well?

kfeinauer:
I left that one invocation of `abort` on purpose because it seems that we are in a legitimately bad spot and having a crash stack in this instance could be beneficial.

PR #4992, opened 2019-06-21
patch GWT directly to deal with scrollbar issue

### LINKED ISSUE #4662: Lists in the File view don't display all contents at bottom

kevinushey:
Closes #4662.

This PR takes a more direct approach to solving the scrollbar issue. Basically, within DataGrid objects, GWT uses these CustomScrollPanel widgets to draw scrollers, and they try to manually compute their width / height as needed. It looks like these computations can occasionally be off on high DPI displays, and so will cause scrollbars to be drawn when they may not really be needed.

The fix here is to just provide a 1 pixel fuzz factor. In other words, the content needs to be at least 2 pixels larger than its container for a scrollbar to be drawn.

In theory, on other platforms one could end up with a scrollbar not being draw even when there was 1 pixel of overflow... but I think we can live with that over scrollbars showing when we don't want them to.

Candidate for v1.2-patch backport.

### DIFF
--- a/src/gwt/src/com/google/gwt/updatepatches.sh
+++ b/src/gwt/src/com/google/gwt/updatepatches.sh
@@ -11,6 +11,7 @@ cd tmp
 TEMPDIR=`pwd`
 
 jar xvf ${SRCBASE}/../lib/gwt/${GWTVER}/gwt-user.jar \
+  com/google/gwt/user/client/ui/CustomScrollPanel \
   com/google/gwt/user/client/ui/DecoratorPanel.java \
   com/google/gwt/user/client/ui/HorizontalPanel.java \
   com/google/gwt/user/client/ui/Label.java \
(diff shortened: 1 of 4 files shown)


kevinushey:
Looks like this doesn't quite work as expected. Let me explore a bit more.

PR #5001, opened 2019-06-24
refactor: create and use DomUtils.disableSpellcheck()

gtritchie:
Created DomUtils.disableSpellcheck() and replaced all setAttribute("spellcheck", "false") calls with it. Has both element and widget variations.


adamconroy:
Much cleaner 👍

PR #5071, opened 2019-07-05
New Project wizard a11y improvements

gtritchie:
This adds keyboard and screen-reader support for the New Project wizard dialog.

The controls for choosing the next page (first two screenshots below) were previously only usable via mouse; now they are treated as buttons, and can get keyboard focus and be activated via spacebar.

The keyboard focus indicator is still not visible in most cases: as usual that work is being deferred to a separate UI-wide work item.

On other panes (not shown here), ensured form controls are associated with labels, marked decorative images as such.

Also added unique element IDs for the list items on the "Project Type" list as requested by QA for automation purposes.


src/gwt/src/org/rstudio/core/client/widget/ModalDialogBase.java
```diff
@@ -718,16 +713,53 @@ public void refreshFocusableElements()
...
+      {
+         // only include items taking up space
+         if (potentiallyFocusable.getItem(i).getOffsetWidth() > 0 && 
+               potentiallyFocusable.getItem(i).getOffsetHeight() > 0)
+         {
+            focusable.add(potentiallyFocusable.getItem(i));
+         }
+      }
+      return focusable;
+   }
```
adamconroy: Years away from element queries being implementable in our app probably but they'd sure be nice here.

src/gwt/src/org/rstudio/core/client/widget/events/ButtonClickManager.java
```diff
@@ -0,0 +1,55 @@
...
+      addHandler(ClickEvent.getType(), clickHandler);
+      widget.addDomHandler(event -> {
+            if (event.getNativeKeyCode() == KeyCodes.KEY_SPACE)
+            {
+               event.stopPropagation();
+               event.preventDefault();
+               click();
+            }
+         }, KeyDownEvent.getType()); 
+   }
```
adamconroy: Since our app is single-paged this will likely never come up but I typically the standard in web pages is space and shift+space is page down/up and enter is for activating a selected item. How does that align with a11y standards?

gtritchie: Good point, ARIA recommends both space and enter for clicking a button: https://www.w3.org/TR/wai-aria-practices/#button

Note that when focus is in a native HTML button, spacebar doesn't scroll, it only clicks the button.

I'm actually used to using spacebar to click buttons; enter, traditionally, activates the default button in a dialog (which isn't always the same button that has focus), but this model seems to be less common these days even in desktop apps.

Short answer, I'll look at adding support for both, but in a separate change. Thanks for the review!

gtritchie: Looking elsewhere, we do have the "enter activates dialog" behavior in our code. For example, bring up the Global Options dialog, get focus on the "Browse" button (no visual indicator currently) and hitting Enter closes the dialog (equivalent to OK button), whereas spacebar activates the Browse button.

PR #5091, opened 2019-07-11
set HOME when using git-bash via tools/shell on Windows

### LINKED ISSUE #3663: 'Tools -> Shell...' reports incorrect HOME on subsequent launches
https://community.rstudio.com/t/problem-with-shell-in-rstudio-ide-1-1-456-windows/16212

---

### System details

    RStudio Edition : Desktop
    RStudio Version : v1.2.1050
    R Version       : 3.5.1

Run <kbd>Tools -> Shell...</kbd>, and then execute `echo $HOME`. You should see something like:

```sh
$ echo $HOME
/c/Users/kevin
```

All is well. Now, close and re-open the shell window, and try running `echo $HOME` again. I see:

```sh
$ echo $HOME
/c/Users/kevin/scratch/v11/=
```

Note that this is the active project directory; also note that a trailing `=` has been appended to the directory. We should preserve the original, 'true' `HOME` environment variable.

This reproduces in v1.1 and v1.0 so this is either a relatively old bug or perhaps something has changed in recent versions of Git for Windows.

gtritchie:
Fixes #3663 

Explicitly set HOME when launching external shell via Tools/Shell on Windows for all types except WSL (which lives in a separate Windows subsystem and takes care of setting home by itself).

### DIFF
--- a/src/cpp/desktop/DesktopGwtCallback.cpp
+++ b/src/cpp/desktop/DesktopGwtCallback.cpp
@@ -1238,19 +1238,17 @@ void GwtCallback::openTerminal(QString terminalPath,
    QStringList args;
    std::string previousHome = core::system::getenv("HOME");
 
-   switch (shellType)
+   if (shellType == GitBash || shellType == WSLBash)
    {
-   case GitBash:
-   case WSLBash:
       args.append(QString::fromUtf8("--login"));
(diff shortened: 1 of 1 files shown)


kevinushey:
Thanks for fixing that up!

Issue #5108, opened 2019-07-16
Escape key restarts Rstudio Server


raubreywhite:
### System details

    RStudio Edition : Server
    RStudio Version : 1.2.1335
    OS Version      : Rocker (Debian inside of Docker container)
    R Version       : 3.6.1

### Steps to reproduce the problem

https://github.com/rocker-org/rocker/issues/316#issuecomment-511802074

I am using a Rocker image (i.e. RStudio Server inside a docker container)

I enter this in the terminal (i.e. starting up RStudio Server):

```
docker run --rm -p 8789:8787 -e PASSWORD=rstudio1 rocker/rstudio
```

I then type `docker ps` to get the name of my container (blissful_benz). I then enter the container as a root user:

```
docker exec -it blissful_benz bash
```

And I install the linux package `tk`

```
root@5caea91f596a:/# apt-get update && apt-get install -y tk
```

I then go to RStudio Server and install `sm`:

```
install.packages("sm")
```

And get the following output:

```
> install.packages("sm")
Installing package into ‘/usr/local/lib/R/site-library’
(as ‘lib’ is unspecified)
trying URL 'https://mran.microsoft.com/snapshot/2019-07-15/src/contrib/sm_2.2-5.6.tar.gz'
Content type 'application/octet-stream' length 256355 bytes (250 KB)
==================================================
downloaded 250 KB

* installing *source* package ‘sm’ ...
** package ‘sm’ successfully unpacked and MD5 sums checked
** using staged installation
** libs
gcc -I"/usr/local/lib/R/include" -DNDEBUG   -I/usr/local/include  -fpic  -g -O2 -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -g  -c fgamma.c -o fgamma.o
gcc -I"/usr/local/lib/R/include" -DNDEBUG   -I/usr/local/include  -fpic  -g -O2 -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -g  -c init.c -o init.o
gfortran  -fpic  -g -O2  -c  variogram.f90 -o variogram.o
gcc -shared -L/usr/local/lib/R/lib -L/usr/local/lib -o sm.so fgamma.o init.o variogram.o -lgfortran -lm -lquadmath -L/usr/local/lib/R/lib -lR
installing to /usr/local/lib/R/site-library/00LOCK-sm/00new/sm/libs
** R
** data
[...]

PR #5111, opened 2019-07-16
call Rf_onintr directly when handling interrupts

kevinushey:
Closes https://github.com/rstudio/rstudio/issues/5108. See inline comments for rationale.

This may be worth considering for backporting, but it is a very low-level change so we'll have to think through whether we really did rely on `R_ProcessEvents()` being run here.

### DIFF
--- a/src/cpp/r/session/RStdCallbacks.cpp
+++ b/src/cpp/r/session/RStdCallbacks.cpp
@@ -375,10 +375,22 @@ int RReadConsole (const char *pmt,
       // set interrupts pending
       r::exec::setInterruptsPending(true);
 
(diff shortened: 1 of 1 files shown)


kevinushey:
Note: potentially a candidate for v1.2-patch, since this buglet means anyone who uses `tcltk` (or a package that tries to load `tcltk` under the hood) will see this unexpected 'crash' when the Escape key is pressed.

kevinushey:
This unfortunately is a regression; I believe it's likely due to changes in how we handle interrupts (these changes were required to support interruption of background processes, and was mainly driven by the Stan-related work)

PR #5134, opened 2019-07-22
fix for malformed Qt mouseevent

### LINKED ISSUE #5107: Cursor switches to typeover, gets stuck, glitches around
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.2.1335
    OS Version      : macos 10.14.4 
    R Version       : 3.6.0

This problem is completely baffling to me and I'm unable to retrace the steps apart from 
- you have to keep editing a script for a while (but "while" varies a lot)
- you have to keep moving the cursor around (maybe!)

The only way to stop this is to restart RStudio. I have now seen this with two different OS versions on two different machines with macos.

The cursor starts glitching. Behaviours include:

- selecting everything up to $ (so that the next thing you type replaces the line)
- overwriting text instead of inserting
- cursor disappearing

- typing along

kevinushey:
Closes #5107. See diff for comments + rationale.

### DIFF
--- a/src/cpp/desktop/DesktopPosixApplicationLaunch.cpp
+++ b/src/cpp/desktop/DesktopPosixApplicationLaunch.cpp
@@ -21,6 +21,7 @@
 #include <core/r_util/RUserData.hpp>
 
 #include <QKeyEvent>
+#include <QMouseEvent>
 
 namespace rstudio {
 namespace desktop {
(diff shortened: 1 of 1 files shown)


kevinushey:
Note: this is a candidate for v1.2-patch. Even though the issue is not observed as often on RStudio v1.3, I can confirm it does still happen.

PR #5186, opened 2019-08-07
Accept `/.Rproj.user` in .gitignore

flying-sheep:
If the user is specific in their .gitignore file and uses repo-absolute paths, they’ll add `/.Rproj.user` instead of `.Rproj.user`.

RStudio shouldn’t try to add the latter if the former exists.


kevinushey:
LGTM! I'll let @jmcphers merge if all looks well.

PR #5188, opened 2019-08-07
Only disable system boost if we're not using system boost.

MariaSemple:
This is a follow up to the discussion this commit: 4193329


kevinushey:

PR #5246, opened 2019-08-15
support dismissing single-button Mac desktop alert via ESC or Enter

### LINKED ISSUE #5115: Escape key should close the find/replace modal
On Mac

 - `Cmd+F` to search for text  
 - enter something that doesn't exist in the file
 - Press enter to search

You get a modal titled `Find/Replace` with the subtext `No more occurrences` and the only option is `OK`. My instinct every time I see this is to press escape to get out of the modal which doesn't work. Users have to press enter to close the modal with the keyboard.

(I'm in vim mode on RStudio Desktop, if that matters.)

gtritchie:
Fixes #5115

- Mac desktop specific
- also unrelated change of NULL to nullptr since I was touching the file anyway

For example, this can now be dismissing via Enter or ESC:

### DIFF
--- a/src/cpp/desktop/DesktopGwtCallbackMac.mm
+++ b/src/cpp/desktop/DesktopGwtCallbackMac.mm
@@ -1,7 +1,7 @@
 /*
  * DesktopGwtCallbackMac.mm
  *
- * Copyright (C) 2009-18 by RStudio, Inc.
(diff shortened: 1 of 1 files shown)


MichaelChirico:
Awesome! Was just coming to file the same issue. Thanks!

PR #5307, opened 2019-08-27
use DYLD_INSERT_LIBRARIES to force requested libR

kevinushey:
Closes https://github.com/rstudio/rstudio/issues/5238.

With this change, `rsession` no longer explicitly links to a particular `libR.dylib`; instead, the RStudio frontend uses `DYLD_INSERT_LIBRARIES` to inject the discovered version of R when RStudio is launched.

### DIFF
--- a/cmake/modules/FindLibR.cmake
+++ b/cmake/modules/FindLibR.cmake
@@ -21,12 +21,14 @@
 if(APPLE)
 
    find_library(LIBR_LIBRARIES R)
-
+   
    if(LIBR_LIBRARIES MATCHES ".*\\.framework")
-      set(LIBR_HOME "${LIBR_LIBRARIES}/Resources" CACHE PATH "R home directory")
+      set(_LIBR_RESOURCES "${LIBR_LIBRARIES}/Resources")
+      get_filename_component(_LIBR_HOME "${_LIBR_RESOURCES}" REALPATH)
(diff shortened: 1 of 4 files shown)


rundel:
Can confirm that as of nightly 1.3.405 I am no longer seeing issue mentioned in #5286 (R installed via Homebrew)

FilipeamTeixeira:
Unfortunately I keep on having this problem. I tried to remove and add the packages again still with the same issue.

`Error: package or namespace load failed for ‘tidyverse’ in dyn.load(file, DLLpath = DLLpath, ...):
 unable to load shared object '/Library/Frameworks/R.framework/Versions/3.6/Resources/library/glue/libs/glue.so':
  dlopen(/Library/Frameworks/R.framework/Versions/3.6/Resources/library/glue/libs/glue.so, 6): Library not loaded: /Library/Frameworks/R.framework/Versions/3.5/Resources/lib/libR.dylib
  Referenced from: /Library/Frameworks/R.framework/Versions/3.6/Resources/library/glue/libs/glue.so
  Reason: image not found`

I'm using RStudio 1.2.5019, and R 3.6.1 and I'm on a macOS 10.14.6.

kevinushey:
Apparently there's some version cross-talk in your library: `glue` is attempting to reference R 3.5, but you're using R 3.6.

You might want to try reinstalling `glue` if you haven't already.

Issue #5218, opened 2019-08-11
Vertical cursor offset after chunk output


clanfear:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.3.363 (and 1.2.1335) 
    OS Version      :  Win 10 Pro 1903 18362.267
    R Version       : 3.6.1

### Steps to reproduce the problem

As it is always present on my machines, I have no idea how to reproduce it on a machine not experiencing it. This is what triggers it:

(1) Create or open any R Markdown document.
(2) Run a chunk with output.
(3) Highlight text after the chunk; there is now an arbitrary vertical offset between visible the cursor position and the highlight location.
(4) Change the zoom level to arbitrarily change the offset some more.

### Describe the problem in detail

[Symptoms similar to this post.](https://community.rstudio.com/t/vertical-mouse-position-offset-in-rmarkdown-documents/29861) [Here is a crude recording.](https://www.youtube.com/watch?v=gP7-F5-qkB4&feature=youtu.be) Running chunks in markdown docs which produce any output results in a varying level of vertical cursor offset after the chunk. There is no horizontal offset. Changing RStudio zoom alters the offset but does not fix it; once changing zoom, however, the offset is usually gone until a chunk is run (then it returns). The degree of offset is different for different chunks but not linearly related to the length of output, though it is worse when the output is large images or widgets. It does not appear to be additive across multiple chunks. It is unaffected by swapping between software and hardware rendering and checking/unchecking any rendering options. It occurs on all monitors and when moving RStudio across monitors. Problem exists on all three of my machines:

(1) Intel i5-9600K desktop with nVidia GTX 1070 video.
(2) Dell XPS 9570 Intel i7-8750H laptop with nVidia GTX1050 video.
(3) Lenovo Yoga 2 Pro Intel i7-4510U with onboard video.

It also exists on both the current daily build and the release version, and persists on fresh reinstalls.  This problem is also not recent--I've been experiencing it for about a year, I think, but I can't be sure when it appeared. I've also experienced it on a lab computer on my campus (unfortunately while live coding for students) but have not checked recently to see if it is still an issue.

### Describe the behavior you expected

I expect the visible cursor to align with the position clicks and selection are resolved.

kevinushey:
I was able to reproduce as well with the following document:

`````
---
title: "Untitled"
output: html_document
---

```{r}
summary(mtcars)
```

Here is some text.
Here is some more text.
`````

To reproduce:

1. Run the chunk of code;
[...]

PR #5322, opened 2019-08-29
round line-widget row height computations

kevinushey:
Fixes an issue where the computed document row target would not match
the actual clicked row -- that is, the row selected would often be the
row _above_ the one that was actually clicked by the user.

Closes https://github.com/rstudio/rstudio/issues/5218.


src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/ChunkOutputWidget.ui.xml
```diff
@@ -117,7 +117,7 @@
    
    .outer
    {
-      padding-bottom: 10px;
+      padding-bottom: 7px;
```
kevinushey: Note: without this change, the new rounding behavior in the line widget row height computation causes the lower padding for the widget to become much larger than expected. 7px is still just enough breathing room (IMHO).

kevinushey:
Before:

Screen Shot 2019-08-29 at 1 06 08 PM

After:

Screen Shot 2019-08-29 at 1 05 30 PM

PR #5336, opened 2019-08-31
Avoid _R_CHECK_LENGTH_1_LOGIC2_ warning

krlmlr:
Reprex (interactive, tested on RStudio Desktop):

```r
Sys.setenv("_R_CHECK_LENGTH_1_LOGIC2_" = "verbose")
shiny::runGadget(div(), function(input, output){})
```

What's a good way to hot-patch my RStudio installation to include this? Can't seem to make it work from `.Rprofile`.


kevinushey:
LGTM!

One way would be to override the function with e.g.

```
setHook("rstudio.sessionInit", function(...) {
    # overrides here
})
```

Alternatively, you could just directly call `.rs.addFunction()` with the appropriate name + override in the session.

PR #5383, opened 2019-09-09
guard against an occasional null-access exception during terminal start

gtritchie:
- seeing this occasionally, haven't been able to nail down the exact timing that results in this happening
- appears benign from user standpoint, but want to avoid and perhaps expose any other (perhaps more helpful) issues lurking after it

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/terminal/TerminalPane.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/terminal/TerminalPane.java
@@ -32,6 +32,7 @@ import org.rstudio.studio.client.application.events.SessionSerializationEvent;
 import org.rstudio.studio.client.application.events.SessionSerializationHandler;
 import org.rstudio.studio.client.application.model.SessionSerializationAction;
 import org.rstudio.studio.client.common.GlobalDisplay;
+import org.rstudio.studio.client.common.Timers;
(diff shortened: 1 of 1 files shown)


src/gwt/src/org/rstudio/studio/client/workbench/views/terminal/TerminalPane.java
```diff
@@ -949,7 +949,30 @@ private void setFocusOnVisible()
...
+               // before xtermjs has initialized, and the setFocus call throws a null exception.
+               // Nothing is obviously broken when that happens, but guard against it, wait
+               // a tiny bit and try once more.
+               if (visibleTerminal.terminalEmulatorLoaded())
+               {
+                  visibleTerminal.setFocus(true);
+               }
+               else
+               {
+                  new Timer()
```
kevinushey: FYI, the recently-added `Timers.singleShot(200, () -> { ... })` would be a more succinct way of expressing this. https://github.com/rstudio/rstudio/blob/c74f6f35996dca9e1572e3f647549e96bf0c0e19/src/gwt/src/org/rstudio/studio/client/common/Timers.java#L22-L32

gtritchie: Nice!

PR #5390, opened 2019-09-10
fix issue when generating docker images from developer machine

gtritchie:
- when building images using docker-compile.sh, the Dockerfile command to COPY scripts under dependencies/common into the image was pulling in content of subfolders (i.e. dependencies installed on the host machine as part of dev build environment)
- the use of trailing asterisk caused subfolder contents to be flattened into the destination directory instead of preserving the folder structure
- in the case of the renv package dependency, this left files in a state where the installation script would fail, preventing image creation

Fix by not using the trailing asterisk. This will be a no-op in the official builds, since those run against a clean repo where dependencies/common has no subfolders, and on a dev machine, preserves the folder structure of already-installed dependencies.

Discussion of trailing "*" versus not: https://stackoverflow.com/questions/30215830/dockerfile-copy-keep-subdirectory-structure

### DIFF
--- a/docker/jenkins/Dockerfile.bionic-amd64
+++ b/docker/jenkins/Dockerfile.bionic-amd64
@@ -82,7 +82,7 @@ RUN /bin/bash /tmp/install-dependencies
 
 # install common dependencies
 RUN mkdir -p /opt/rstudio-tools/dependencies/common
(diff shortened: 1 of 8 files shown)


kevinushey:
🎉 🎉 🎉

PR #5409, opened 2019-09-15
ensure main window web control gets focus on RStudio startup

### LINKED ISSUE #3711: Main window doesn't get focus and menus don't work
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.2.1065 (but think I've seen this with earlier 1.2 builds)
    OS Version      : macOS 10.14 (but think I saw this with 10.13 on same machine)
    R Version       : 3.5.1

* Start RStudio, don't click on the main window when it appears.
* Try to use any menu items, such as Help / About

* Menus don't do anything
* Console doesn't have focus until you click on it
* Once you click on the main window, the menus start working

* Console has focus on startup, menus work

### Other details
I have only been able to repro on my Macbook Pro, 15", mid-2014, but it happens 100% of the time. Same version of RStudio and R don't exhibit this problem on my iMac.

Does not happen with 1.1.456.

gtritchie:
Fixes #3711

This Mac-desktop-only issue started happening when we upgraded to Qt 5.12.x.

Without this `setFocus` no keyboard messages (e.g. shortcut keys or typing into console) get routed to the RStudio web application until you click on the page and/or Cmd+Tab away and back to the application. Both an accessibility issue, and a generally annoying behavior (and a regression from RStudio 1.1).

After some testing in 1.3, should consider as a candidate for a future 1.2 bug fix release, as it is irritating for users who are used to launching RStudio and being able to use shortcut keys without having to first click on the window.


src/cpp/desktop/DesktopMainWindow.cpp
```diff
@@ -234,6 +234,7 @@ void MainWindow::onWorkbenchInitialized()
 
       avoidMoveCursorIfNecessary();
    });
+   webView()->setFocus();
```
kevinushey: Could it be possible for the user to explicitly focus something in the IDE frame before this function is called (before the workbench is initialized)?

I wonder if it would be worth being extra cautious and only doing this if we detect that the main window has focus, but nothing else in the IDE frame.

gtritchie: Not sure it's possible, but if it is, where might user put focus outside the web view such that this call would potentially yank it back into the web view? Main menus, I suppose?

kevinushey: That's what I was thinking -- it's unlikely, but there might be a case where the user opens a menu but loses focus after the workbench finishes initialization. Probably too unlikely to worth worrying about, though...

gtritchie: Famous last words: "Too unlikely to worth worrying about".

I can repro this with 1.2.1578 on Mac (prior to my change):

- launch app
- quickly get focus onto the not-yet-fully-built RStudio menu (with mouse or standard Ctrl+F2 shortcut on a "real" keyboard, doesn't seem to always work on touchbar)
- wait until the UI appears and renders
- keyboard focus is no longer on the menu

So I don't think I'm making it any worse here. Haven't tried Windows, though, so maybe prudent to scope this down to Mac since that's where the problem is.

Issue #4663, opened 2019-04-18
No 'escape' key on iPad Pro keyboard for RStudio Server


raubreywhite:
### System details

    RStudio Edition : Server
    RStudio Version : 1.2.x
    OS Version      : iOS13
    R Version       : 

### Steps to reproduce the problem

Load RStudio server on iPad (Safari), with a physical keyboard that does not have an ESC key (e.g. Apple smart keyboard). Cannot use ESC key, because it doesn't exist.

### Describe the problem in detail

Cannot use the ESC key like we normally would in RStudio. This affects:

- VI mode
- Cancelling long running code
- Breaking out of code that has unmatched "

### Describe the behavior you expected

It would be nice to have some way to use ESC on a keyboard that doesn't have this button.

gtritchie:
Still a problem with iPadOS 13.1. You can use Ctrl+[ as ESC when using (for example) VIM in the terminal, but that doesn't work anywhere else that I tried.

gtritchie:
Opened an issue on Ace, this behavior repros on the Ace kitchen sink so not RStudio specific.
https://github.com/ajaxorg/ace/issues/4077

ronblum:
Verified using iPad 13.3.1  connecting to RStudio Server 1.3.929, both with the <kbd>Esc</kbd> key and <kbd>Ctrl+[</kbd>.

PR #5486, opened 2019-09-30
allow use of Ctrl+[ as ESC key in vim mode on iPadOS

gtritchie:
- addresses vim part of https://github.com/rstudio/rstudio/issues/4663
- Ace change: https://github.com/ajaxorg/ace/issues/4077


kevinushey:
I've been maintaining Ace patches in a separate fork of Ace here:

https://github.com/rstudio/rstudio-ace

Can you try porting the patch to that repository, on the `rstudio/v1.3` branch? Then, if you run the script here:

https://github.com/rstudio/rstudio/blob/master/src/gwt/tools/sync-ace-commits

it should install and rebuild Ace (and hopefully look exactly the same as this PR does).

kevinushey:
Oops. I see you did exactly that. Thanks :-)

PR #5500, opened 2019-10-03
accessibility preferences, pane and menu

gtritchie:
# Overview
Added accessibility-related preferences, including new pref-pane, and Help/Accessibility submenu with a command for enabling/disabling screen-reader support and to get directly to accessibility prefs.

The menu command is to make it possible to enable screen reader support on Desktop apps, where the native menus are inherently accessible, but our web UI (including the Options dialog) are invisible until screen reader support is enabled. Without this, a screen-reader user could not actually enable accessibility in the desktop product without getting help from a sighted user.

For server, the "enable screen reader" flag is used to control screen reader support in xterm, but will likely be used for addition upcoming work such as gating creation of a shadow accessibility tree for the console. On desktop it also controls whether we disable DOM accessibility when configuring Chromium.

Added setting for "reducing motion" for people with vestibular disorders, and also because our animations can confuse screen-readers. This disables UI animations such as the sliding when zooming panes, moving between wizard pages, and such. I probably missed some cases. Some users may prefer this even if they don't need it as an accommodation.

Added control over wrapping the whole web page in role=application (significant to screen reader keyboard behavior, will be described in support article) and controlling the delay needed after typing before live-regions are updated.

Created a more generalized mechanism in the preferences dialog/pane framework for having panes specify if they want the UI refreshed (GWT page reloaded), and/or a full reload of the desktop frame.

# Visuals


adamconroy:
Looks good! You might already be planning this but a similar help link to the docs like we have in the Terminal Options pane will be great for explaining more precisely what these options do.

I also wonder if we want to implement both `aria-label` and `title` for our checkbox/other form elements across the board for easier detail access. It looks like we don't have that at the moment, at least in our checkboxes in the options dialog.

gtritchie:
Thanks for the review!

> ... but a similar help link to the docs like we have in the Terminal Options pane will be great for explaining more precisely what these options do.

Yes, plan to do that once we are closer to feature-complete and I have written the support article.
 
> I also wonder if we want to implement both `aria-label` and `title` for our checkbox/other form elements across the board for easier detail access.

Per our conversation, I think this would be a good idea on some options where coming up with a sufficiently meaningful label is unrealistic but redirecting to a support article is overkill. The downside of doing this too often is making navigation with screen readers very verbose as it will read this info when tabbing around (behavior varies by screen reader, some may read both the label and the title/aria-label, some may read both, etc.).

Issue #5387, opened 2019-09-10
Tab key not inserting tab (or spaces), tried to auto complete when not needed


dracodoc:
This seem to be a long standing issue.

[2013](https://support.rstudio.com/hc/en-us/community/posts/200658573-Tab-behaviour-in-script-window)

[2016 and 2017](https://support.rstudio.com/hc/en-us/community/posts/115008079068-Cannot-type-TAB-or-reassign-autocomplete-shortcut-key-in-RStudio-Editor)

2018: #2766. My problem is more than the issue described in #2766, so I'm raising new issue and included reference to that one.

My current environment:
RStudio 1.2.1335
```r
R version 3.6.1 (2019-07-05)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 7 x64 (build 7601) Service Pack 1
```

1. In R script, console code chunk of RMarkdown. Pressing `Tab` after ending of code (or other empty places) will first generate a warning tool top "no matches", then next press take effect.
2. In RMarkdown text, **pressing Tab have no effect whatsoever**. This is very annoying when user try to indent something or align text. Interestingly, RStudio server 1.1.453 don't have this problem (but still have problem 1 in R script).

ronblum:
@dracodoc Thank you for raising this. For (2), I can reproduce it; in addition, continuing to click on <kbd>Tab</> repeatedly goes through the same process, despite it now being preceded blank text. 

Also: I checked a couple of other file types. The same problem occurs with Stan files, but not C++.

kevinushey:
How can I reproduce (2)?

dracodoc:
Just input text in RMarkdown, pressing tab don't have effect for me, either at a new line or after some text(or it don't have effect anywhere).

kevinushey:
Thanks -- I can reproduce only if the `[ ] Use tab for multiline autocompletions` completion preference is checked. (Tools -> Global Options... -> Code -> Completion).

You can try disabling that to fix the issue. I'll see if we can figure out what's going on.

dracodoc:
Yes, problem #2 no longer appears once that option was disabled. Thanks for verifying!

dracodoc:
Update:
I found even with the tab for multiline option disabled, tab still only have effect in a new line. With some text, either right after text, or after inserted some space after text, the tab still doesn't take effect.

tmcduff:
Verified: 1.3.749

raffaem:
[...]

PR #5522, opened 2019-10-08
Support the ability to disable tab-based autocompleted in user preferences.

### LINKED ISSUE #2766: Can't Disable AutoComplete via Tab
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.1.442
    OS Version      : Windows 7 Pro
    R Version       : 3.4.4

```
library(tidyverse)
treatment <- tribble(
  ~ person,           ~ treatment, ~response,
  "Derrick Whitmore", 1,           7,
  NA,                 2,           10,
  NA,                 3,           9,
  "Katherine Burke",  1,           4
)

```
Enter text above into console. Place cursor in between `NA` and `2`. Press <kbd>Tab</kbd> key. Autocomplete is triggered no matter the settings that I've chosen in `> Settings > Code > Completion`. 

I'd like autocomplete disabled. When I enter a <kbd>Tab</kbd> I'd like a <kbd>Tab</kbd> entered, instead of receiving an autocomplete dialog. I have a vague recollection that it's bad to use tabs in R code. Maybe this is intentional?

MariaSemple:
Some users find our auto-complete a little over-eager and wish they could disable tab-based auto-completion. This PR adds a user preference that lets user disable tab for auto-complete so that pressing tab will always insert a tab character. Unless the user in emacs mode, pressing `ctrl-space` will still initiate auto-completion, and pressing `shift-tab` after the start of a snippet will still initiate snippet completion.

Screen Shot 2019-10-08 at 3 13 41 PM

Fixes #2766 and part 1 of #5387.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/source/Source.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/Source.java
@@ -46,7 +46,12 @@ import com.google.inject.Inject;
 import com.google.inject.Provider;
 
 import org.rstudio.core.client.*;
-import org.rstudio.core.client.command.*;
+import org.rstudio.core.client.command.AppCommand;
+import org.rstudio.core.client.command.Handler;
+import org.rstudio.core.client.command.KeyCombination;
+import org.rstudio.core.client.command.KeyboardShortcut;
+import org.rstudio.core.client.command.KeySequence;
+import org.rstudio.core.client.command.ShortcutManager;
 import org.rstudio.core.client.dom.WindowEx;
 import org.rstudio.core.client.events.*;
 import org.rstudio.core.client.files.FileSystemItem;
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/source/Source.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/Source.java
@@ -46,10 +46,7 @@ import com.google.inject.Inject;
 import com.google.inject.Provider;
 
 import org.rstudio.core.client.*;
-import org.rstudio.core.client.command.AppCommand;
-import org.rstudio.core.client.command.Handler;
-import org.rstudio.core.client.command.KeyboardShortcut;
-import org.rstudio.core.client.command.ShortcutManager;
+import org.rstudio.core.client.command.*;
 import org.rstudio.core.client.dom.WindowEx;
 import org.rstudio.core.client.events.*;
 import org.rstudio.core.client.files.FileSystemItem;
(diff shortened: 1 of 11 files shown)


MariaSemple:
The default value is true so as not to change the existing behaviour.

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/AceEditor.java
```diff
@@ -191,25 +191,29 @@ public String getType()
...
-         // Never complete if there's an active selection
+         // Never complete if there's an active selection.
          Range range = getSession().getSelection().getRange();
          if (!range.isEmpty())
             return false;
 
-         // Don't consider Tab to be a completion if we're at the start of a
-         // line (e.g. only zero or more whitespace characters between the
-         // beginning of the line and the cursor)
+         // If the user explicitly requested an auto-complete by pressing 'ctrl-space' instead of 'tab', always attempt auto-complete.
```
kevinushey: nit: line is too long (maximum is 100 characters; note that this flows off-screen in the GitHub PR UI)

kevinushey:
Do we need to also manage the shortcut binding we use for autocomplete? Note that we register a mapping here:

https://github.com/rstudio/rstudio/blob/42cdf0835be539d493ce9dc49dde5425b8aa21e4/src/gwt/src/org/rstudio/studio/client/workbench/views/source/Source.java#L401

MariaSemple:
No, I think we still want to trap `tab` and then handle it differently based on modifier (for example we always want to at least think about doing snippet completion if the modifier is `shift`).

kevinushey:
Sorry, I wasn't clear -- the shortcut there is mainly used for the keybinding that gets displayed in menus; e.g.

Screen Shot 2019-10-09 at 9 43 55 AM

My point is that we don't want to display the Tab key here if we've disabled that shortcut (we should instead show `Ctrl + Space`).

src/gwt/src/org/rstudio/studio/client/workbench/views/source/Source.java
```diff
@@ -46,10 +46,7 @@
 import com.google.inject.Provider;
 
 import org.rstudio.core.client.*;
-import org.rstudio.core.client.command.AppCommand;
-import org.rstudio.core.client.command.Handler;
-import org.rstudio.core.client.command.KeyboardShortcut;
-import org.rstudio.core.client.command.ShortcutManager;
+import org.rstudio.core.client.command.*;
```
gtritchie: We generally don't use wildcard imports (there are plenty scattered around in the code so not a huge deal). IntelliJ does them by default but there's a setting to override.

MariaSemple: Oh yes, I think IntelliJ changed this on me. Let me change it back, and I'll update that setting too.

PR #5523, opened 2019-10-08
add unique ids to several toolbar popup menu buttons for automation

### LINKED ISSUE #5520: Automation - unique ID for terminal dropdown
Requesting an unique ID for the terminal dropdown button. See screenshot.

Screen Shot 2019-10-08 at 12 24 47 PM
 
Thanks!

gtritchie:
- terminal pane dropdown
- main toolbar, (new file, mru, vcs, pane layout, projects)
- build pane (build bookdown, more...)
- fixed bug in regex for generating unique IDs based on labels; "terminal #" was being stripped of the "#" instead of including it
- fixes #5520

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/ElementIds.java
+++ b/src/gwt/src/org/rstudio/core/client/ElementIds.java
@@ -48,7 +48,7 @@ public class ElementIds
    public static String idSafeString(String text)
    {
       // replace all non-alphanumerics with underscores
-      String id = text.replaceAll("[^a-zA-Z0-0]", "_");
+      String id = text.replaceAll("[^a-zA-Z0-9]", "_");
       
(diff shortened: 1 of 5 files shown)


src/gwt/src/org/rstudio/core/client/ElementIds.java
```diff
@@ -48,7 +48,7 @@ public static String getElementId(String id)
    public static String idSafeString(String text)
    {
       // replace all non-alphanumerics with underscores
-      String id = text.replaceAll("[^a-zA-Z0-0]", "_");
+      String id = text.replaceAll("[^a-zA-Z0-9]", "_");
```
adamconroy: It wasn't so safe after all 🐙

PR #5526, opened 2019-10-09
fix debian dependency script to pass along exclude-qt-sdk arg

### LINKED ISSUE #5471: ./install-dependencies-debian --exclude-qt-sdk doesn't actually exclude Qt installation
### System details

    RStudio Edition : Server
    RStudio Version : Compiled
    OS Version      : Debian Stretch
    R Version       : 3.6.1

* Run `./install-dependencies-debian --exclude-qt-sdk` on terminal

--exclude-qt-sdk flag is not being passed to install-dependencies-stretch

Qt not being installed

gtritchie:
Fixes #5471


gtritchie:
> Does this need to be done for any other dependency scripts?

No, this is the only dependency script that delegates to other scripts in this particular fashion.

Issue #5525, opened 2019-10-09
Wrong document outline font after macOS Catalina update


msaltieri:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.2.5001
    OS Version      : macOS 10.15 (Catalina)
    R Version       : 3.6.1

### Describe the problem in detail

After updating to macOS 10.15 (Catalina) there is a very ugly font in the document outline applied to every custom separator of this kind:
```
# SECTION #1 ----
```
Instead, the font applied to the functions is the same as always.

### Describe the behavior you expected

It would be great if the font in the document outline would be the same for both separators and functions.

ronblum:
@msaltieri Thank you for filing this issue. Can you please provide a screenshot so that we can compare the output?

msaltieri:
Hello @ronblum! Thank you for the very quick feedback.

Here is the screenshot of how the document outline appear on my IDE:
<img width="242" alt="Schermata 2019-10-09 alle 16 13 42" src="https://user-images.githubusercontent.com/26568338/66489628-37a1b900-eab0-11e9-9617-8e467a68b2d9.png">

ronblum:
@msaltieri You're welcome! Do you mean the switch between the serifed and non-serifed fonts (which are two different fonts altogether? I'm unable to reproduce this—where the functions appear, I get the serifed font but italicized. Would you mind providing the code for the above, or at least a small subset, so that I can try to reproduce exactly the same results?

msaltieri:
Exactly @ronblum, there is an inexplicable switch between Serif/Sans-serif fonts that has never existed before the os update.

You can try yourself with this chunk of code:
```
Schedule <- R6Class(

    # Class Schedule ----
    classname = 'Schedule',

    # * Public Members ----
    public = list(

        # * * Basic Methods ----

        initialize = function(config_file) {
            private$config <- config_file
            private$tab <- NULL
            private$is_refine <- FALSE
            invisible(self)
        }

    ),

    # * Private Members ----
    private = list(

        tab = NULL,
        config = NULL,
        is_refine = NULL

    )

)
```

The document outline appears like this:
<img width="229" alt="Schermata 2019-10-09 alle 16 57 03" src="https://user-images.githubusercontent.com/26568338/66493190-e5639680-eab5-11e9-84b9-323ed4d03825.png">

I'm working with `R6` classes, but the same Serif font appears every time I try to insert a section separator like:
```
# This is a new section ----
```

[...]

PR #5527, opened 2019-10-09
fix sans-serif font usage in doc outline

kevinushey:
Closes https://github.com/rstudio/rstudio/issues/5525.

Candidate for v1.2-patch backport.


jmcphers:
Do you think it's possible to fix this more holistically? As noted in #5525 this affects dozens of places, everywhere in the IDE. Another obvious one is R Markdown output chunks. :-(

image

jmcphers:
Maybe something like this?

https://doc.qt.io/qt-5/qwebenginesettings.html#resetFontFamily

kevinushey:
Aha! That seems like it will do the trick.

jmcphers:
Backported to 1.2-patch

PR #5529, opened 2019-10-09
build full gwt dist to get closer to official dist

gtritchie:
- now includes `gwt-api-checker.jar`
- builds samples (which helps find if I've changed any public GWT signatures with my accessibility work)
- then deletes samples and JavaDoc to avoid checking another 5000+ files into git


src/gwt/tools/build-gwt
```diff
@@ -15,7 +15,7 @@ fi
 
 # Build GWT disto with custom version
 cd ${RUN_DIR}/gwt/gwt
-ant clean elemental dist-dev -Dgwt.version="${RSTUDIO_VER}"
+ant clean elemental dist -Dgwt.version="${RSTUDIO_VER}"
```
kevinushey: Is there a reason why we use `${RSTUDIO_VER}` rather than the GWT version here? (I think this might actually be the setting that confuses Eclipse into believing this version of GWT is too old)

gtritchie: Not really. If you think renaming to something like 2.8.2-rstudio would help, or even just use 2.8.2 (other than a 1-time bit of cleanup needed if you had previously installed the stock 2.8.2), easy to do.

gtritchie: Actually, I think I did it this way when I planned to build a custom package and install it as a dependency; wanted name to be rstudio version specific in case a future rstudio version had its own custom gwt build based on same gwt build, and didn't want to have to reinstall the gwt dependency when switching between version branches. Once I just put the files in the repo, that concern was no longer relevant.

kevinushey: Gotcha. In that case, I'd advocate using the same version as declared by GWT itself. (I don't think we need to rename the folder name, though)

gtritchie: Roger. I'll make that change and update this PR.

PR #5534, opened 2019-10-09
Remove 'not' selectors and add important to focus-visible outlines

adamconroy:
The GWT CSS parser is old and glitchy. Its support for CSS3 selectors in particular is very, very shaky. The story can be followed by poor dankurka's adventures trying to get it to work 4 years ago and it seems like they still haven't fixed it: 
https://github.com/gwtproject/gwt/issues/5770

Currently running `ant unittest` will output a barrage of `[WARN] Line 1315 column 25: encountered ".". Was expecting one of: <IDENT> ` errors when trying to parse the `:not` selectors I added yesterday. The CSS worked just fine in the application itself but the unit test harness very much does not like it.

I tried every permutation and suggested workaround in the myriad threads referencing this problem but couldn't think of any other reliable solution.

I'm removing the `not` selectors and proposing the unfortunate, but limited in scope and functional addition of `!important` to the keyboard navigation CSS. The `focus-visible` functionality can be turned off through the library so, worst case, we can add a flag in accessibility that is even off by default if we really don't like how this works in default cases.

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/files/filedialog/FileDialogStyles.css
+++ b/src/gwt/src/org/rstudio/core/client/files/filedialog/FileDialogStyles.css
@@ -236,7 +236,7 @@ body.windows .breadcrumb table.path span {
    margin-top: 0px;
 }
 
-button.browse:focus:not(.focus-visible) {
+button.browse:focus {
    outline:0;
 }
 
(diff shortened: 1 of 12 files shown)


gtritchie:
Did you try the hack discussed here?

https://stackoverflow.com/questions/3479636/how-to-get-gwt-cssresource-to-parse-not-selectors

adamconroy:
Yeah, I tried that before submitting the first PR yesterday in fact. That hack very much does not work, doesn't even compile at all.

adamconroy:
Sounds good. I'll keep that in mind for more intractable problems!

PR #5538, opened 2019-10-10
Move open-source NOTICE from About dialog to new dialog

### LINKED ISSUE #5533: About dialog's massive license text stalls screen reader
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.3.479
    OS Version      : Windows 10
    R Version       : Any

- Start NVDA screen reader
- Start RStudio
- Help / About

- Takes upwards of 10 seconds before the About dialog appears, the screen reader says anything about the dialog, and the application can be interacted with
- Tabbing around the About dialog once it has come back to life is very sluggish

- Shouldn't take more than a fraction of a second for this dialog to appear, and it should remain responsive when tabbing around it

Assumption is that it is the massive blob of text (3400+ lines) we show documenting open-source components and their licenses.

gtritchie:
- the 3400+ line NOTICE for open-source components was bogging down
  screen readers when opening the About dialog
- moved into a new dialog, with a link to it from the About dialog
- new dialog puts the text into a `<pre>` instead of a `textarea`; this is much faster and does not slow down screen readers
- marked that area as role="document", and made it able to get focus and be selected / copied
- fixes #5533

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -3384,11 +3384,21 @@ RapidJSON
 ------------------------------------------------------------------------
 Copyright (C) 2015 THL A29 Limited, a Tencent company, and Milo Yip.  All rights reserved.
 
-Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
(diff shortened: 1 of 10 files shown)


jmcphers:
I don't think that text styled like a web link should open a dialog.

Would recommend adding a button instead, either where the link is right now or in the lower left corner of the dialog.

gtritchie:
> I don't think that text styled like a web link should open a dialog.

I actually had it as a button originally, but thought it looked awkward visually. We do use these links in other places to open dialogs (though I agree it's a bit weird). Here, for example:

PR #5578, opened 2019-10-17
remove reference to removed initialPath; support curproc

### LINKED ISSUE #5574: FilePath::initialPath() is used but is missing
```
--- src/cpp/core/CMakeFiles/rstudio-core.dir/system/PosixSystem.cpp.o ---
/wrkdirs/usr/ports/devel/RStudio/work/rstudio-1.2.5001/src/cpp/core/system/PosixSystem.cpp:938:37: error: no member named 'initialPath' in 'rstudio::core::FilePath'
   FilePath initialPath = FilePath::initialPath();
                          ~~~~~~~~~~^
1 error generated.
```

src/cpp/core/include/core/FilePath.hpp doesn't declare it.

This happens when /proc/self is missing.

The FreeBSD port update is pending the fix of this.

kevinushey:
Closes #5574.

### DIFF
--- a/src/cpp/core/CMakeLists.txt
+++ b/src/cpp/core/CMakeLists.txt
@@ -162,8 +162,13 @@ if (UNIX)
    check_function_exists(inotify_init1 HAVE_INOTIFY_INIT1)
    check_function_exists(getpeereid HAVE_GETPEEREID)
    check_function_exists(setresuid HAVE_SETRESUID)
+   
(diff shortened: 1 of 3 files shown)


kevinushey:
Hrm, after testing it looks like `/proc/curproc` is not a feature that's on by default and is normally enabled / disabled by the user. So checking its presence at build time is incorrect...

PR #5579, opened 2019-10-17
Fix tabstop wrapping in preferences dialogs when section selection changes

### LINKED ISSUE #5573: Global Options dialog tab trap breaks when tab other than General selected
### System details

    RStudio Version : 1.3.
    OS Version      : Any
    R Version       : Any

- Open global preferences
- With General selected, use Tab key to move forward until Apply button is reached, another Tab moves back to General, as expected (same if you go in reverse with Shift+Tab)
- Use down-arrow or mouse to move to Code tab, or any tab other than the first one
- Tab forward to Apply key, then hit Tab again

- Focus wraps back to General, hitting Tab again goes to currently selected tab
- Going in reverse is worse, Shift+Tab from (for example) Code causes focus to go to the `<body>` element

- Focus should wrap to currently selected tab, not always first to General

gtritchie:
- Fixes #5573
- When a control with a roving tabstop is first (or last) in a modal dialog,
  it must notify the dialog when the selection moves to or from the first
  (or last) tab, otherwise the tab wrapping is broken
- Fixes the two tests marked as xfail in this automation PR: https://github.com/rstudio/rstudio-ide-automation/pull/25

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/prefs/PreferencesDialogBase.java
+++ b/src/gwt/src/org/rstudio/core/client/prefs/PreferencesDialogBase.java
@@ -92,6 +92,12 @@ public abstract class PreferencesDialogBase<T> extends ModalDialogBase
       {
          Integer index = selectionEvent.getSelectedItem();
 
+         // SectionChooser is first focusable control in the modal dialog, and it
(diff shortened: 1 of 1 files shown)


gtritchie:
With the upcoming freeze, and simplicity of this change, I'm going to merge without waiting for the review.

PR #5588, opened 2019-10-21
Add autoConstrain fuctionality to popup panels

adamconroy:
Added a flag (defaulted to on) to ThemedPopupPanel such that it constrains itself to the app window after a position and size is set. I tried to get it as high as I comfortably could in the hierarchy with an option for subclasses to turn it off so I could reach as many panels as possible that may need this treatment but it's possible I missed something, would want a moderate smoke test by QA after merging to find some possible exceptions.

Tiny laptop users can rejoice!
WDeYarO
jRaOUAK
oWNkJNM

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/widget/ScrollableToolbarPopupMenu.java
+++ b/src/gwt/src/org/rstudio/core/client/widget/ScrollableToolbarPopupMenu.java
@@ -48,28 +48,28 @@ public class ScrollableToolbarPopupMenu extends ToolbarPopupMenu
    
    public HandlerRegistration addSelectionHandler(SelectionHandler<MenuItem> handler)
    {
-      return ((ScrollableToolbarMenuBar)menuBar_).addSelectionHandler(handler);
+      return getMenuBar().addSelectionHandler(handler);
    }
 
    public void ensureSelectedIsVisible()
    {
-      if (menuBar_.getSelectedItem() != null)
+      if (getMenuBar().getSelectedItem() != null)
       {
          DomUtils.ensureVisibleVert(scrollPanel_.getElement(),
-                                    menuBar_.getSelectedItem().getElement(),
+                                    getMenuBar().getSelectedItem().getElement(),
                                     0);
       }
    }
    
    public int getSelectedIndex()
    {
-      return ((ScrollableToolbarMenuBar)menuBar_).getSelectedIndex();
+      return getMenuBar().getSelectedIndex();
    }
 
    @Override
    protected Widget createMainWidget()
    {
-      scrollPanel_ = new ScrollPanel(menuBar_);
+      scrollPanel_ = new ScrollPanel(getMenuBar());
       scrollPanel_.addStyleName(ThemeStyles.INSTANCE.scrollableMenuBar());
       scrollPanel_.getElement().getStyle().setOverflowY(Overflow.AUTO);
       scrollPanel_.getElement().getStyle().setOverflowX(Overflow.HIDDEN);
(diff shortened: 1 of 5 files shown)


adamconroy:
Totally agree re: keyboard navigation. Spent a good hour trying different things and, sadly, due to us (GWT) using tables so heavily and a full Panel rework being out of the scope of this issue @gtritchie 
 and I have decided to let this pass for now as another a11y exception. I will create an issue to track the problem.

src/gwt/src/org/rstudio/studio/client/workbench/views/terminal/TerminalPopupMenu.java
```diff
@@ -15,6 +15,7 @@
 
 package org.rstudio.studio.client.workbench.views.terminal;
 
+import com.google.gwt.dom.client.Style;
```
gtritchie: Why add this? Don't see any related changes requiring it.

src/gwt/src/org/rstudio/studio/client/workbench/views/terminal/TerminalPopupMenu.java
```diff
@@ -44,6 +45,7 @@
 {
    public TerminalPopupMenu(TerminalList terminals)
    {
+      super();
```
gtritchie: A Java constructor automatically does a call to no-argument `super()` so this isn't needed.

src/gwt/src/org/rstudio/core/client/widget/ScrollableToolbarPopupMenu.java
```diff
@@ -59,7 +63,7 @@ public void ensureSelectedIsVisible()
    
    public int getSelectedIndex()
    {
-      return menuBar_.getSelectedIndex();
+      return ((ScrollableToolbarMenuBar)menuBar_).getSelectedIndex();
```
gtritchie: Rather than directly accessing the superclass `menuBar_` member, and casting it, consider making it private in `ThemedPopupPanel`, and have a protected method that returns it. Then you can override by return-type in the subclasses and have those return the appropriate type with no need for any casts.

src/gwt/src/org/rstudio/core/client/widget/ThemedPopupPanel.java
```diff
@@ -76,15 +81,44 @@ public ThemedPopupPanel(boolean autoHide, boolean modal, Resources res)
...
+      super.setPopupPosition(left, top);
+
+      if (autoConstrain_)
+         sizeToWindow(top, Style.Overflow.AUTO);
+   }
+
+   // Size the table to the window
+   private void sizeToWindow(int top, Style.Overflow overflowY)
+   {
+      NodeList<Element> e = this.menuBar_.getElement().getElementsByTagName("table");
```
gtritchie: Don't need `this` here.

src/gwt/src/org/rstudio/core/client/widget/ToolbarPopupMenu.java
```diff
@@ -82,10 +82,10 @@ protected void onUnload()
       super.onUnload();
       menuBar_.selectItem(null);
    }
-   
+
    public void selectFirst()
    {
-      menuBar_.selectFirst();
+      ((ToolbarMenuBar)menuBar_).selectFirst();
```
gtritchie: Ditto on using an override getMenuBar instead of using a cast.

gtritchie:
Mentioned on Slack, but I'm still trying to come up with a cleaner structure to recommend here. Don't like that the `ThemedPopupPanel` class has a protected member `menuBar_` that it uses but doesn't set itself (via construction), therefore creating an implicit and unenforced contract on subclasses. Also the need for casts still feels like there's another abstraction needed here (interface or something simpler). Stand by while I ponder.

gtritchie:
I think it boils down to this:
https://github.com/rstudio/rstudio/commit/21b8c213f1d26d0bced322f90a5126736444a44a

adamconroy:
Prefer #5597 simple version.

PR #5603, opened 2019-10-22
Log syslog messages to stderr if there's a TTY attached

MariaSemple:
Also moves the special formatting for syslog from Logger.cpp to SyslogDestination.cpp.

### DIFF
--- a/src/cpp/shared_core/system/SyslogDestination.cpp
+++ b/src/cpp/shared_core/system/SyslogDestination.cpp
@@ -24,7 +24,11 @@
 #include <shared_core/system/SyslogDestination.hpp>
 
 #include <cassert>
+#include <iostream>
 #include <syslog.h>
+
+#include <boost/algorithm/string.hpp>
+#include <boost/regex.hpp>
 
 #include <shared_core/Logger.hpp>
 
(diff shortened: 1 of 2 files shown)


src/cpp/shared_core/system/SyslogDestination.cpp
```diff
@@ -90,7 +94,17 @@ void SyslogDestination::writeLog(
...
+   std::string forSyslog = boost::algorithm::trim_right_copy(in_message);
+   boost::algorithm::replace_all_copy(forSyslog, "\n", "|||");
+
+   // Also remove the leading date and program ID, since those are set by syslog directly.
+   forSyslog = boost::regex_replace(forSyslog, boost::regex("^[^\\]]*\\]\\s"), "");
+   ::syslog(logLevelToLogPriority(in_logLevel), "%s", forSyslog.c_str());
+
+   // Also log to stderr if there is a tty attached.
+   if (::isatty(STDERR_FILENO) == 1)
+      std::cout << in_message;
```
kfeinauer: This should be `cerr`.

MariaSemple: Nice catch, thanks!

Issue #5607, opened 2019-10-23
Error.asString() should use OCCURRED AT and CAUSED BY


MariaSemple:
Here: https://github.com/rstudio/rstudio/blob/master/src/cpp/shared_core/Error.cpp#L316-L333

PR #5626, opened 2019-10-24
remove unused string constants (compiler warnings)

gtritchie:
- these were triggering warning messages during compilation


MariaSemple:
Actually I need to move those to Error.cpp and use them in `asString()`. I have a bug open for it here: https://github.com/rstudio/rstudio/issues/5607, I just haven't gotten to it yet

gtritchie:
Ah, ok. I'll close this.

PR #5644, opened 2019-10-28
save + restore Help Pane scroll position

### LINKED ISSUE #5629: Request: remember scrollbar position in panes
If I browse for the help of some function (e.g. `help(heat.colors)`), check the documentation in the Help pane, scroll down on it, and then switch to a different pane in the same space (say, the Plots pane), then when I return to the Help pane, the scrolling will be reset to the top.

Would be nice if the scrolling position at which it was would be remembered, as sometimes one wants to check some plot against the documentation of the function that produced it, or look up some file from where the documentation came as the documentation is being read.

kevinushey:
Closes #5629.

See comments for rationale on the somewhat surprising gymnastics done to make this work.


dfalty:
@jmcphers @kevinushey The bug is in the 1.2-patch build. Is this going to be backported there?

PR #5671, opened 2019-10-30
Make implementation of Error.asString match the old implementation

### LINKED ISSUE #5607: Error.asString() should use OCCURRED AT and CAUSED BY

MariaSemple:
Fixes #5607


MariaSemple:
Basically by checking here: https://github.com/rstudio/rstudio/blob/v1.2-patch/src/cpp/core/Log.cpp#L47-L71 and having a look at the logs to see that it looks correct. The significance is only that our log format stays pretty similar.

gtritchie:
Thanks!

PR #5679, opened 2019-10-31
Resize and set overflow of ModalDialogBase

### LINKED ISSUE #1605: Global Options Conflict with Zoom
I was really hoping this issue would have been solved with the new release, although admittedly I don't need to access global options with great frequency

Basically if you're zoomed in the text editor, and you access global options, you won't be able to see the whole global options dialogue/pane. So, if you need to access something toward the bottom of options, you must exit options, zoom out, open options (and then exit options and re-zoom in, to continue working). Don't expect a quick fix on this, but I do want to make sure devs are aware of it, as I find the default text size too small

adamconroy:
Currently ModalDialogBase objects don't think too much about where and how they're set. Similarly to what was done in PR #5597 this adds logic to a base UI object class to get a little more intelligent about displaying the object.

This fix will apply to a huge number of modals in our application so I spent a lot of time testing making sure it applied as generically and unobtrusively as possible. Some dialogs aren't compatible with it but in the case where it isn't there are no side effects (that I've been able to find). If any bugs fall out of this fixing them will be as easy as special casing it just like the other PR.

Note: Ideally only the central content would scroll and the header and buttons would be pinned but due to how our modals are laid out and differ this isn't generically feasible and this way we get much more generic functionality for much less code.

iSHHJOq
uSO1p1O

fixes #1605

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/widget/ModalDialogBase.java
+++ b/src/gwt/src/org/rstudio/core/client/widget/ModalDialogBase.java
@@ -39,6 +39,7 @@ import com.google.gwt.user.client.ui.SimplePanel;
 import com.google.gwt.user.client.ui.VerticalPanel;
 import com.google.gwt.user.client.ui.Widget;
 
+import elemental.client.Browser;
 import org.rstudio.core.client.Debug;
 import org.rstudio.core.client.ElementIds;
 import org.rstudio.core.client.Point;
@@ -48,7 +49,6 @@ import org.rstudio.core.client.command.ShortcutManager;
 import org.rstudio.core.client.command.ShortcutManager.Handle;
 import org.rstudio.core.client.dom.DomUtils;
 import org.rstudio.core.client.dom.NativeWindow;
-import org.rstudio.core.client.js.JsWindow;
 import org.rstudio.core.client.theme.res.ThemeStyles;
 import org.rstudio.studio.client.RStudioGinjector;
 import org.rstudio.studio.client.application.ui.RStudioThemes;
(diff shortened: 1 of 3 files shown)


src/gwt/src/org/rstudio/core/client/widget/ModalDialogBase.java
```diff
@@ -205,6 +206,38 @@ protected void positionAndShowDialog(Command onCompleted)
...
       super.center();
       onCompleted.execute();
+
+      // Force the contents of the modal to be vertically scrollable
+      // if the height of the modal is larger than the app window
+      Element e = this.getElement();
+      Element child = e.getFirstChildElement();
+      if (child != null)
+      {
+         int windowInnerHeight = Integer.parseInt(JsWindow.getProp("innerHeight").toString());
```
gtritchie: Instead of adding JsWindow.java and using JSNI, we _could_ do this via 
```
import elemental.client.Browser;

int windowInnerHeight = Browser.getWindow().getInnerHeight();
```
You'd also need to add this to Rstudio.gwt.xml:
```
   <!-- Elemental module dependencies                              -->
   <inherits name='elemental.Elemental'/>
```

Background: `elemental` is a Google project to provide type-safe access to browser APIs from Java with very little overhead. This project has a life beyond GWT (2.x), and is still under active development.

The version bundled with GWT is ancient, but sufficient for basic stuff like this, and if we hit anything it can't do (probably in 1.4), we could switch to the actively maintained version (elemental2: https://github.com/google/elemental2).

The post-GWT 2.x world (in terms of Java-based JavaScript) looks like it will consist of J2CL (https://github.com/google/j2cl), JsInterop (https://github.com/google/j2cl/blob/master/docs/jsinterop-by-example.md), and elemental2, and stuff built on these.

kevinushey: I think this is worth a conversation (maybe on #rstudio-dev). I'd definitely be on board adding `elemental` or `elemental2` but want to make sure there isn't some historical reason why we chose not to do so.

adamconroy:
As far as I can tell `elemental` is just an auto generated sub-project of GWT so no NOTICE entry should be necessary.

PR #5691, opened 2019-11-04
always place postback dir on PATH

kevinushey:
From @jcheng5:

> jcheng 6:38 PM
> Maybe that change was there for a reason... I just upgraded to 1.3.565 on my Windows machine, and now I got this when trying to create a new project from a GitHub repo:
> 
> >>> C:/Program Files/Git/bin/git.exe clone --progress git@github.com:jeroen/curl curl
> Cloning into 'curl'...
> ssh_askpass: exec(C:/Program Files/RStudio/bin/postback/rpostback): No such file or directory
> git@github.com: Permission denied (publickey).
> fatal: Could not read from remote repository.
> 
> Please make sure you have the correct access rights
> and the repository exists.
> And sure enough, adding C:/Program Files/RStudio/bin/postback/ to the path and using rpostback-askpass as the GIT_ASKPASS and SSH_ASKPASS fixed it.

So it looks like the PATH munging is required for Git on Windows. The mirrored solution, then, is to make sure the `rpostback` utilities are always on the PATH.

### DIFF
--- a/src/cpp/core/include/core/system/Environment.hpp
+++ b/src/cpp/core/include/core/system/Environment.hpp
@@ -65,6 +65,10 @@ void unsetenv(Options* pEnvironment,
 
 void getModifiedEnv(const Options& extraVars, Options* pEnv);
 
+// add to the PATH
+void addToPath(const std::string& filePath,
+               bool prepend = false);
+
 // add to the PATH within a string
 void addToPath(std::string* pPath,
                const std::string& filePath,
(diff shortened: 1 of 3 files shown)


kevinushey:
> It's not clear to me whether this reintroduces the issues we had before (wherein we could wind up with the askpass program set to our own executable but no corresponding PATH setting). Does it?

It shouldn't. Previously, we only placed our postback utilities on the PATH when we were explicitly invoking Git ourselves; now, we place it on the PATH by default so that it's always seen (including user invocations of git with e.g. `system("git <...>")`.

PR #5692, opened 2019-11-04
remove trailing slashes from HOME / R_USER

kevinushey:
Re: https://community.rstudio.com/t/mouse-click-return-error-navigating-to/43908

I still don't know why some users are reporting HOME / R_USER with a trailing slash, but the issue is becoming prevalent enough that it's worth working around.


kevinushey:
(it may also be more appropriate for us to find the place where the incorrect path constructions are being made, and fix those)

PR #5702, opened 2019-11-05
Make the new logger implementation threadsafe.

### LINKED ISSUE #5634: Thread-safe Logger Reload on SIGHUP
Keep the logger in a pointer, compare and swap on replacement to avoid locking.

MariaSemple:
Using a reader-writer lock because most of the logging work will be "read" only (writing logs is read only, adding/removing log destinations is write). The R/W lock implementation is write-preferring because writing should happen so much less often than reading. Some cursory profiling on Windows and OSX show there's no appreciable impact on performance by the logging code.

Fixes #5634

### DIFF
--- a/src/.gitignore
+++ b/src/.gitignore
@@ -4,3 +4,6 @@ CMakeLists.txt.user
 cpp-build*/
 cmake-build-*/
 session32/*
+dbg/*
+rel/*
+rwd/*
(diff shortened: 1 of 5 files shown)


MariaSemple:
Boost's reader/writer lock is not re-entrant. 

The RW_LOCK_END macro takes a boolean value which determines whether log writing will be attempted. Within Logger.cpp, the value is always set to false. See [here](https://github.com/rstudio/rstudio/pull/5702/files#diff-3b09ec2f94095a4a79e767c08e42a6dfR156-R174) and [here](https://github.com/rstudio/rstudio/pull/5702/files#diff-e6444c857ea962c6841c1bb152243313R507)

PR #5715, opened 2019-11-07
Fix C++ compiler warning about initialization order

gtritchie:
- On macOS was getting:

```
MonitorClient.cpp:33:7: warning: field 'programIdentity_' will be initialized after base 'rstudio::core::log::ILogDestination' [-Wreorder]
      programIdentity_(programIdentity),
      ^
```

### DIFF
--- a/src/cpp/monitor/MonitorClient.cpp
+++ b/src/cpp/monitor/MonitorClient.cpp
@@ -30,8 +30,8 @@ class MonitorLogDestination : public core::log::ILogDestination
 {
 public:
    MonitorLogDestination(core::log::LogLevel logLevel, const std::string& programIdentity) :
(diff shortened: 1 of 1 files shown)


MariaSemple:
Interesting! I swapped this order because I was getting an initialization warning on Ubuntu when I had it this way.

gtritchie:
> Interesting! I swapped this order because I was getting an initialization warning on Ubuntu when I had it this way.

Well, that's... annoying. :-)

gtritchie:
I just compiled on Ubuntu (RSP), and get the same warning. It is also flagged as a warning on Windows (in Qt Creator's editor, at least, didn't try a build).

MariaSemple:
In that case, I think I must have misremembered where I saw that warning before.

Issue #4058, opened 2018-12-18
git GUI uses password style text box for asking for github user when you push


andresrcs:
### System details

    RStudio Edition : Server
    RStudio Version : 1.2.1194
    OS Version      : Ubuntu Server 16.04.5 LTS
    R Version       : 3.5.1

### Steps to reproduce the problem
Try tu push changes in your git enabled project to github ussing the push button

### Describe the problem in detail
git gui uses a password style text box for your github user when you try to push to github, this prevents chrome credential manager from work properly.

![image](https://user-images.githubusercontent.com/14981085/50179905-365f8f80-02d6-11e9-9533-a767e942ac73.png)

### Describe the behavior you expected
I expect to see a normal text box asking for my user, and with this being able to use chrome credential manager to autofill this text box every time I do a push to github.

PR #5721, opened 2019-11-08
avoid hiding input for username in askpass

kevinushey:
Closes https://github.com/rstudio/rstudio/issues/4058.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/common/vcs/AskPassManager.java
+++ b/src/gwt/src/org/rstudio/studio/client/common/vcs/AskPassManager.java
@@ -15,6 +15,7 @@
 package org.rstudio.studio.client.common.vcs;
 
 import org.rstudio.core.client.Debug;
+import org.rstudio.core.client.MessageDisplay;
 import org.rstudio.core.client.StringUtil;
 import org.rstudio.core.client.MessageDisplay.PromptWithOptionResult;
 import org.rstudio.core.client.widget.Operation;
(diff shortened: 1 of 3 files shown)


src/gwt/src/org/rstudio/studio/client/common/vcs/AskPassManager.java
```diff
@@ -78,10 +79,21 @@ public void onAskPass(final AskPassEvent e)
...
             askpassPending_ = true;
             
-            globalDisplay.promptForPassword(
-                  "Password",
-                  e.getPrompt(),
+            String prompt = e.getPrompt();
+            
+            String title = "Password";
+            int dialogType = MessageDisplay.INPUT_PASSWORD;
+            if (prompt.toLowerCase().indexOf("username") != -1)
```
gtritchie: Could just use `prompt.toLowerCase().contains("username"))`

src/gwt/src/org/rstudio/studio/client/common/vcs/AskPassManager.java
```diff
@@ -78,10 +79,21 @@ public void onAskPass(final AskPassEvent e)
             
             askpassPending_ = true;
             
-            globalDisplay.promptForPassword(
-                  "Password",
-                  e.getPrompt(),
+            String prompt = e.getPrompt();
```
gtritchie: Keying off of a visible UI string to change fundamental behavior makes me... uneasy. That sort of thing will bite us if we every try to localize the UI, and makes the code less self-documenting.

Plus the fact this class is called `AskPassManager` and can now prompt for something that isn't a password seems unexpected (though, to be fair, it was already kinda doing that).

I think this means that rstudioapi::askForPassword will now show a regular text field instead of a password field if the prompt contains "username"?

If separating this behavior entirely (e.g. separate messages and handler class) is overkill, might want to document this quirk in `rstudioapi` and also in SessionAskPass.cpp / askPassword().

kevinushey: > Plus the fact this class is called AskPassManager and can now prompt for something that isn't a password seems unexpected (though, to be fair, it was already kinda doing that).

I definitely agree it's weird, but ultimately that's what a non-interactive `GIT_ASKPASS` program has to do. That is, they are programs that read a prompt from `stdin`, and then return an appropriate value (username or password) based on that prompt. It definitely feels unnatural to me, but ... it is what it is. There may be cleaner ways to handle this, but this is at least a straightforward solution.

Perhaps it would be cleaner if we made that decision on the session side; e.g. here:

https://github.com/rstudio/rstudio/blob/0691592b8743cbe291317af331a0adc5526f1ca0/src/cpp/session/modules/SessionAskPass.cpp#L100-L104

and then just had the UI respond based appropriately based on whether it "looks like" we're prompting for a username or not.

gtritchie: Got it, makes sense.

PR #5755, opened 2019-11-18
Use home dir odbcinst.ini file in linux since users don't need sudo

javierluraschi:
When installing ODBC drivers in Linux, seems like a better option to use `~/odbcinst.ini` to avoid permissions issues since it would otherwise require `sudo` to be modified.

### DIFF
--- a/src/cpp/session/modules/SessionConnectionsInstaller.R
+++ b/src/cpp/session/modules/SessionConnectionsInstaller.R
@@ -340,7 +340,7 @@
 })
 
 .rs.addFunction("odbcBundleOdbcinstPathUseHome", function() {
-   normalizePath("~/odbcinst.ini", mustWork = FALSE)
+   normalizePath("~/.odbcinst.ini", mustWork = FALSE)
 })
 
 .rs.addFunction("odbcBundleOdbcinstPath", function() {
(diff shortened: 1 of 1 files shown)


src/cpp/session/modules/SessionConnectionsInstaller.R
```diff
@@ -333,12 +333,26 @@
...
 
-.rs.addFunction("odbcBundleOdbcinstPath", function() {
+.rs.addFunction("odbcBundleOdbcinstPathWithOdbcinst", function() {
    config <- system2("odbcinst", "-j", stdout = TRUE)
    odbciniEntry <- config[grepl("odbcinst.ini", config)]
    gsub("^[^/\\\\]*", "", odbciniEntry)
 })
 
+.rs.addFunction("odbcBundleOdbcinstPathUseHome", function() {
+   normalizePath("~/odbcinst.ini", mustWork = FALSE)
```
kevinushey:
```suggestion
   normalizePath("~/.odbcinst.ini", mustWork = FALSE)
```

javierluraschi: Thanks! James confirmed this works in his environment as well.

PR #5756, opened 2019-11-18
log warning and attempt to cast as double when casting to int fails

melissa-barca:


src/cpp/r/RSexp.cpp
```diff
@@ -913,14 +913,24 @@ SEXP create(SEXP valueSEXP, Protect* pProtect)
...
 {
    try
    {
       int casted = boost::numeric_cast<int>(value.getValue<T>());
       return create(casted, pProtect);
    }
+   catch(const boost::bad_numeric_cast& e)
+   {
+      LOG_WARNING_MESSAGE("Failed to cast from " + type + " to int: " + e.what());
+      try {
```
kevinushey: nit: brace on its own line

src/cpp/r/RSexp.cpp
```diff
@@ -913,14 +913,24 @@ SEXP create(SEXP valueSEXP, Protect* pProtect)
...
+SEXP createInteger(const core::json::Value& value, const std::string& type, Protect* pProtect)
 {
    try
    {
       int casted = boost::numeric_cast<int>(value.getValue<T>());
       return create(casted, pProtect);
    }
+   catch(const boost::bad_numeric_cast& e)
+   {
+      LOG_WARNING_MESSAGE("Failed to cast from " + type + " to int: " + e.what());
```
kevinushey: Thinking again, I wonder if this should really be a warning message? I worry that this is a warning that might get fired pretty frequently in some environments without any real resolution possible by the user.

I don't think we have any explicit contact that a JSON integer must be mapped to an R integer vector, so it's probably okay to just silently fall back to double in those cases.

gtritchie: I suppose it could use LOG_INFO_MESSAGE? If I recall, logging levels, from least verbose / most serious to most verbose / least serious are: ERROR, WARNING, INFO?

MariaSemple: That's right. There's also DEBUG, which is the most verbose.

gtritchie: Ah right, so maybe LOG_DEBUG_MESSAGE would be most appropriate here. If you enable that level of logging, presumably you're willing to slog through a lot of results!

kevinushey: That sounds good to me as well.

melissa-barca: Updated to use LOG_DEBUG_MESSAGE.

src/cpp/r/RSexp.cpp
melissa-barca: Fixed.

PR #5757, opened 2019-11-18
Set the server name indication so we can work with TLS v 1.3.

MariaSemple:
TLS v 1.3 now requires SNI on requests.

This PR needs to be cherry picked to v1.2-patch.

### DIFF
--- a/src/cpp/core/include/core/http/TcpIpAsyncClientSsl.hpp
+++ b/src/cpp/core/include/core/http/TcpIpAsyncClientSsl.hpp
@@ -34,8 +34,8 @@ class TcpIpAsyncClientSsl
 {
 public:
    TcpIpAsyncClientSsl(boost::asio::io_service& ioService,
-                       std::string address,
-                       std::string port,
+                       const std::string& address,
+                       const std::string& port,
                        bool verify,
                        const std::string& certificateAuthority = std::string(),
                        const boost::posix_time::time_duration& connectionTimeout =
@@ -43,8 +43,8 @@ public:
                        const std::string& hostname = std::string() )
      : AsyncClient<boost::asio::ssl::stream<boost::asio::ip::tcp::socket> >(ioService),
        sslContext_(boost::asio::ssl::context::sslv23_client),
-       address_(std::move(address)),
-       port_(std::move(port)),
+       address_(address),
+       port_(port),
        verify_(verify),
        certificateAuthority_(certificateAuthority),
        connectionTimeout_(connectionTimeout)
--- a/src/cpp/core/include/core/http/TcpIpAsyncClientSsl.hpp
+++ b/src/cpp/core/include/core/http/TcpIpAsyncClientSsl.hpp
@@ -34,16 +34,17 @@ class TcpIpAsyncClientSsl
 {
 public:
    TcpIpAsyncClientSsl(boost::asio::io_service& ioService,
-                       const std::string& address,
-                       const std::string& port,
+                       std::string address,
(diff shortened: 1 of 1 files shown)


src/cpp/core/include/core/http/TcpIpAsyncClientSsl.hpp
```diff
@@ -34,16 +34,17 @@ class TcpIpAsyncClientSsl
 {
 public:
    TcpIpAsyncClientSsl(boost::asio::io_service& ioService,
-                       const std::string& address,
-                       const std::string& port,
+                       std::string address,
```
kfeinauer: Curious why the change to the signature here? Will the compiler not do an implicit move when constructing `address_` and `port_` anyway? In which case, these signatures are essentially the same. Plus, this looks strange as it doesn't adhere to the rest of our codebase.

src/cpp/core/include/core/http/TcpIpAsyncClientSsl.hpp
```diff
@@ -70,6 +71,15 @@ class TcpIpAsyncClientSsl
       // use scoped ptr so we can call the constructor after we've configured
       // the ssl::context (immediately above)
       ptrSslStream_.reset(new boost::asio::ssl::stream<boost::asio::ip::tcp::socket>(ioService, sslContext_));
+
+      // Set the SNI so we still work with TLS v1.3
```
kfeinauer: I think we should note that the TLS v1.3 standard requires the SNI to be set instead of just saying that we will continue to work (the context is important, and very helpful here).

src/cpp/core/include/core/http/TcpIpAsyncClientSsl.hpp
MariaSemple: I was getting clang tidy warnings about it. I think clang tidy prefers pass-by-copy with std::move because if the calling code passes an r-value the compiler can optimize to use move construction rather than copy construction which should be slightly more performant. Here's the clang-tidy page about it: https://clang.llvm.org/extra/clang-tidy/checks/modernize-pass-by-value.html. 

Either way I only really updated it because it's such a minor change to appease the clang-tidy warnings, but if it's something we really don't want to do I don't believe the performance benefits will be very substantial so I can disable the clang-tidy warning.

src/cpp/core/include/core/http/TcpIpAsyncClientSsl.hpp
MariaSemple: Sure thing.

src/cpp/core/include/core/http/TcpIpAsyncClientSsl.hpp
kevinushey: See also: https://abseil.io/tips/117

FWIW, I'd prefer disabling that since it's sort of a micro-optimization that we likely don't need, and the technique does come with caveats.

MariaSemple: I'm fine with removing it, but I generally only get warnings about `std::strings` references that are immediately being copied into a member variable in the constructor (and only used once), so I think the clang-tidy warning is only poking us about obvious improvements that don't really have downsides. That being said, I'd agree that it's a micro-optimization and I'd hope that the compiler would optimize the case when we pass an r-value by const reference and only copy it once anyways.

MariaSemple: Done.

PR #5798, opened 2019-11-27
remove some unused Java classes

gtritchie:
Because sometimes it's better to take away than to give. Or something like that.


src/gwt/src/org/rstudio/core/client/patch/DiffMatchPatch.java
```diff
@@ -1,58 +0,0 @@
...
-import com.google.gwt.dom.client.HeadElement;
-import com.google.gwt.dom.client.ScriptElement;
-import com.google.gwt.resources.client.ClientBundle;
-import com.google.gwt.resources.client.TextResource;
-
-public class DiffMatchPatch
-{
-   interface Resources extends ClientBundle
-   {
-      @Source("diff_match_patch.js")
```
kevinushey: I wonder if this means we no longer require `diff_match_patch.js`?

gtritchie: Ah maybe, didn't notice that part. I'll take a peek when I get a chance.

Issue #4368, opened 2019-02-27
[FR] Be able to control how often the git scan takes place


shrektan:
Hi,

I'm currently maintaining a large project with multiple git submodules inside. However, when using RStudio on this project, the git and file control panels respond very slow. 

I suspect it's because RStudio is somehow running git re-scan whenever a file gets touched. After I closed the version control function, RStudio works in "normal" speed now on this project.

It's a pain to tolerate the slow respond. So I'm asking if __RStudio can add an option regarding to the git scan frequencies__ . To my case, I will be happy even if I have to trigger the git scan manually (that's what the native git GUI does). Otherwise, I have to turn off RStudio version control and run the git GUI there... I prefer stays everything inside of RStudio...

Thanks!

kevinushey:
Are you using a preview release of RStudio? If so, you can execute:

```
Sys.setenv(RSTUDIO_GIT_LOG = "1")
```

to have RStudio dump all the git-related actions it's taking into the console. Do you see the lag occurring with any action in particular?

Note that RStudio doesn't poll the filesystem for changes; it registers file monitors to request a refresh when files within the project change, and also when the RStudio window gains focus.

shrektan:
@kevinushey Thanks for the help. 

Yes, I can see the the action log of git, when I modify the file, click the folder, click the git history... And whenever RStudio GUI gets the focus...

After a few trials, `C:/app/Git/bin/git.exe status -z --porcelain -- C:/my-work-directory`  seems quite slow. It gets called a lot and every time it takes about 3 seconds, even if I hasn't touched any file. 

Maybe it's due to the multiple submodules because `git status` runs quite slow on the terminal as well...

shrektan:
Actually, it runs very fast on my Mac. Highly probably due to the anti-virus software on the computer in the office (the Windows 10 OS may also be the cause). Anyway, since I cannot turn off the anti-virus, being able to close the Git auto-scan will help a lot on my case... Thanks.

kevinushey:
https://community.rstudio.com/t/rstudio-suddenly-slow-processing/4959/9 might be relevant: some users found that adding certain processes to the set of Windows Defender Exclusions helps.

shrektan:
Thanks for the tips but it doesn't help. I bet my antivirus (TrendMicro) is still scanning those processes. Unluckily, I don't have the authorization to add processes into the TrendMicro's white lists so can't test that...

kevinushey:
[...]

PR #5824, opened 2019-12-04
add pref for vcs auto-refresh

kevinushey:
This PR adds a preference + small bit of UI for controlling whether RStudio attempts to do background VCS refreshes, e.g. in response to file changes and window focus.

Screen Shot 2019-12-03 at 4 07 59 PM

Note that, unlike the Environment pane refresh state, this is a user preference as opposed to so-called persistent state. The logic here is that most users who need to disable VCS are likely working in an environment where all projects can suffer from lag etc. when git actions are taken.

Let me know if you feel this is better expressed as a separate type of preference, or perhaps as a project-specific preference.

Closes https://github.com/rstudio/rstudio/issues/4368.


kevinushey:
I've double-checked and indeed the VCS file decoration is used only to determine if an immediate full `git status` is required. (It looks like specifically this is done to respond to changes in e.g. `.gitignore`)

akersting:
@kevinushey Thanks for this feature. Highly appreciated! I would like to request two minor enhancements:

1. There should be a keyboard shortcut for the manual refresh.
2. The refresh button (and actually no other button) should go to the right of the branch name. It will not be visible when working on a branch with a somewhat long name.

kevinushey:
@akersting Thanks for the request, but it's likely to be missed if made just as a comment on a PR. Can you file a separate issue for those enhancements?

PR #5830, opened 2019-12-04
change background of read-only textboxes

### LINKED ISSUE #5811: Cannot use text box for a new project path
Reproduce:

1. click "new project"
2. click "existing directory"
3. you should be in the window called "create project from existing directory"
4. there is a text box having label "project working directory"
5. trying to type text into this text box doesn't work, one has to use "browse" button and pick directory using graphical interface

If the text box is meant to be not writable it should be have grey background, otherwise it is confusing.
Using RStudio Build 93 (7b3fe265, 2019-09-18) on Xubuntu 18.04.

gtritchie:
- Fixes #5811 (at least partially)

In places where we show read-only text boxes, most commonly in file/directory pickers, make background the same as the dialog color to give a hint of its read-only-ness. Can still get keyboard focus allowing selection and copying (and screen reader ... reading).

Making these editable is deemed too much work at this stage due to need to revisit each usage and potentially relocate validation code, etc.

Here's an example of an editable picker, for contrast.

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/theme/res/themeStyles.css
+++ b/src/gwt/src/org/rstudio/core/client/theme/res/themeStyles.css
@@ -28,6 +28,7 @@
 @external gwt-TabLayoutPanelTabs;
 @external gwt-DecoratedPopupPanel;
 @external editor_dark;
+@external gwt-TextBox-readonly;
 
 /* These are exposed to allow users to customize RStudio further with their rsthemes. */
 @external rstheme_tabLayoutCenter;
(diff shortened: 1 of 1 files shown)


kevinushey:
LGTM. My only concern is that the textbox seems to blend in a bit too much with the modal's background, but I'm not sure if there's a great way around that.

gtritchie:
> the textbox seems to blend in a bit too much with the modal's background, but I'm not sure if there's a great way around that.

I've tried other shades, all seemed a bit cluttered or drew too much attention (to my eyes, at least). Also considered removing the border, making this look more like just a plain label (albeit one that can be tabbed to and selected). Examples:

gtritchie:
Another comparison, the Upload Files dialog uses a plain label to show selected file (with button coming before the label); was planning to switch this to use same control as other locations (e.g. button on the right after a read-only textbox).

PR #5838, opened 2019-12-05
provide string overloads when setting error

kevinushey:
Otherwise, calls like `pResponse->setError(error, "a message")` will
have the const char* string implicitly converted to bool, unexpectedly
selecting the overload `setError(Error, bool)`.

Also fix an issue where RStudio would fail to open certain JSON files --
specifically, those detected as 'application/json' as opposed to
'text/json'. renv.lock is one such file that is sometimes detected in
this way.

### DIFF
--- a/src/cpp/core/include/core/json/JsonRpc.hpp
+++ b/src/cpp/core/include/core/json/JsonRpc.hpp
@@ -489,6 +489,14 @@ public:
    
    void setError(const core::Error& error,
                  bool includeErrorProperties = false);
+   
+   void setError(const core::Error& error,
+                 const char* message,
+                 bool includeErrorProperties = false);
+   
+   void setError(const core::Error& error,
+                 const std::string& message,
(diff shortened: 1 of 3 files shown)


kevinushey:
An example of where we do this:

https://github.com/rstudio/rstudio/blob/5796b8aacbe9ff4b881594ce2923d173ae7c0ba5/src/cpp/session/modules/SessionSource.cpp#L239-L240

It looks like other call-sites explicitly force a JSON Value:

https://github.com/rstudio/rstudio/blob/5796b8aacbe9ff4b881594ce2923d173ae7c0ba5/src/cpp/session/modules/presentation/SessionPresentation.cpp#L350-L352

but these overloads should overall be safer + convenient.

kevinushey:
If we do really want to force the user to create a JSON object at the call site, then I think we need to remove or fix the `setError(Error, bool)` overload.

PR #5841, opened 2019-12-05
Object explorer: Remove expando from displaying with empty objects

### LINKED ISSUE #4925: Object Explorer: empty classed objects still get expando
### System details

    RStudio Version : 1.2.1522
    OS Version      : macOS  10.14.5
    R Version       : R version 3.4.4 (2018-03-15)

```r
object <- list()
class(object) <- "foo"
View(list(object))
```

Screen Shot 2019-06-10 at 9 54 02 AM

No expand arrow on the empty classed list.

melissa-barca:
Fixes #4925 

When determining if an object is expandable, check if it is recursive the current context.


melissa-barca:
@kevinushey I updated the code to preserve the hasRelevantAttribute check and also check object's length before marking expandable as true.

PR #5867, opened 2019-12-10
add API for highlighting arbitrary UI in RStudio

kevinushey:
This PR adds an API function, `.rs.api.highlight()`, that can be used to highlight arbitrary UI within the RStudio IDE. Example:

Screen Shot 2019-12-10 at 9 49 03 AM

The highlighter accepts a list of highlight "queries" -- e.g. the aforementioned screenshot is produced by executing:

```
.rs.api.highlight(
   list(
      list(query = "#rstudio_label_tools_menu", parent = 1L),
      list(query = "#rstudio_label_terminal_command", parent = 1L),
      list(query = "#rstudio_label_new_terminal_command", parent = 1L)
   )
)
```

Currently, the implementation relies on `querySelectorAll()`, and accepts arbitrary query selectors from the user. We could limit this scope if desirable, but I could imagine cases where one might want to highlight things that are not explicitly IDed or classed.

The `parent` parameter is used to request highlight of a parent element for a particular matched element.

This PR also implements a small utility for discovering the ID of elements underneath the cursor:

Screen Shot 2019-12-10 at 9 53 15 AM

to help facilitate the process of figuring out how to identify the ID (if any) associated with a particular bit of UI. (It will also help us find pieces of UI which have no ID, but reasonably could.)


kevinushey:
Some tweaks to the element highlighter (mouse cursor over the Diff button)

Screen Shot 2019-12-10 at 10 08 49 AM

gtritchie:
Just FYI, as time goes by, a combination of the accessibility work and Selenium automation work will result in more and more elements having predictable unique IDs.

Some may remain variable in that they will have a suffix which depends which order various UI was invoked (e.g. a shared control with an element-id that gets _1, _2, etc. appended if there are more than one instance showing.) You see that in your screenshot with `rstudio_source_text_editor_2`. Happens here: https://github.com/rstudio/rstudio/blob/4e3ce3b14c7b361f81a0558a67c98691893500ff/src/gwt/src/org/rstudio/core/client/ElementIds.java#L31-L44

Not crazy about this, or the ever-growing list of IDs at the end of this source file, but at the moment this is how we're doing it. Might move it into more of a namespace model, so at least the IDs can be defined near where they are used.

gtritchie:
On the accessibility front, be sure the highlighting in default theme (Textmate) is at least 3:1 contrast per https://www.w3.org/TR/WCAG21/#non-text-contrast

The screenshots of the dark theme look sufficient, though I didn't measure them. At the moment, our focus is on behavior with the default theme, but we hope to have a sufficiently contrasted dark theme at some point, too.

gtritchie:
Now the next accessibility question... what's the screen-reader story for this feature? Without knowing the expected use cases, I can't really speculate on if there's a reasonable answer to that.

Perhaps one approach would be to check our "screen reader enabled" flag, and read the accessible text of the selected element by injecting it into a live region.

kevinushey:
> Not crazy about this, or the ever-growing list of IDs at the end of this source file, but at the moment this is how we're doing it. Might move it into more of a namespace model, so at least the IDs can be defined near where they are used.

It might be worth forcing the issue up-front at the widget-construction site, so that any created UI has to request a specific ID. Otherwise I imagine this will be easily forgotten as new widgets and tools are added. Still, I wanted to just follow existing precedent for any new IDs I added.

kevinushey:
> On the accessibility front, be sure the highlighting in default theme (Textmate) is at least 3:1 contrast per https://www.w3.org/TR/WCAG21/#non-text-contrast

Here's the treatment for the light theme:

Screen Shot 2019-12-10 at 1 13 20 PM

kevinushey:
> Now the next accessibility question... what's the screen-reader story for this feature? Without knowing the expected use cases, I can't really speculate on if there's a reasonable answer to that.

For context, the primary goal here is to support `learnr` tutorials, which will be running in a separate Tutorial pane and will use this API to highlight pieces of UI for the user. As an example, a `learnr` tutorial might want to teach users how to create an R Markdown document, and then Knit that document (ostensibly using the Knit button).

Do you have any thoughts on what the best way to communicate that same information to a visually-impaired user might be? Do we want some way of signaling where that UI lives in the IDE and how it can be accessed, or should we consider a different mechanism (e.g. the keyboard shortcut used to invoke the command, if any)?

> Perhaps one approach would be to check our "screen reader enabled" flag, and read the accessible text of the selected element by injecting it into a live region.

Do we have any examples of this in the wild?

> Or another idea is an option to send keyboard focus to the highlighted element, that would trigger a screen reader to do the reading for you.

One downside to this approach is that this doesn't "teach" the user how to access that bit of UI, especially since this is an action that wouldn't be replicable outside the context of a tutorial.

src/gwt/src/org/rstudio/studio/client/workbench/UserInterfaceHighlighter.java
```diff
@@ -0,0 +1,313 @@
...
+      
+   }
+   
+   
+   // Event Handlers ----
+   
+   @Override
+   public void onCommand(AppCommand command)
+   {
+      // TODO: Currently used for debugging + testing.
```
jmcphers: Should strip this out before merging to master.

gtritchie:
> One downside to this approach is that this doesn't "teach" the user how to access that bit of UI, especially since this is an action that wouldn't be replicable outside the context of a tutorial.

Yes, in many ways this is a theoretical conversation since many aspects of the UI remain (effectively) inaccessible even after all the work I've done, and I've not even evaluated packages or the content produced by packages used by the IDE.

To really understand, we'd need to evaluate accessibility of `learnr` and its ability to produce accessible tutorials in general.

So for now, this will more likely be another feature we call out as inaccessible in the 1.3 timeframe. Ultimately, yes, we need to be able to produce tutorials for teaching people to operate RStudio purely via keyboard, and also purely via screen-reader, but still several mountains to climb to reach that point.

PR #5881, opened 2019-12-12
initial wiring for learnr + tutorial pane

kevinushey:
This PR makes it possible to run learnr tutorials as an RStudio job, with the associated URL then hosted within the Tutorial pane.

Screen Shot 2019-12-12 at 10 49 38 AM

The implementation here borrows from a lot of the existing code used to show Shiny applications in the Viewer pane. In this case, however, output is re-routed to the tutorial pane.

Putting this PR up in a relatively early stage just to confirm that this is the right approach to be taking (ie: that re-using the pathways normally routed to the Viewer should be safe overall).


src/cpp/session/resources/schema/user-prefs-schema.json
```diff
@@ -617,7 +617,7 @@
         },
         "shiny_viewer_type": {
             "type": "string",
-            "enum": ["user", "none", "pane", "window", "browser"],
+            "enum": ["user", "none", "pane", "tutorial", "window", "browser"],
```
jmcphers: I don't think this should be part of the pref enum because it's nonsensical for people to prefer to have their Shiny apps open in the Tutorial pane.

src/cpp/r/R/Api.R
```diff
@@ -815,6 +815,53 @@ options(terminal.manager = list(terminalActivate = .rs.api.terminalActivate,
...
+         shiny_args = shiny_args
+      )
+      
+   )
+   
+   deparsed <- deparse(call)
+   path <- tempfile("rstudio-tutorial-", fileext = ".R")
+   writeLines(deparsed, con = path)
+   
+   .rs.api.runScriptJob(
```
jmcphers: We probably don't want this to be a script job because script jobs have e.g. a progress bar which doesn't make sense here. I'm working on a branch which will enable Shiny background jobs that have no progress bar and some other special treatment (https://github.com/rstudio/rstudio/tree/feature/shiny-background-job) -- hopefully will have a PR soon you can build on!

gtritchie:
Doesn't have to be part of this PR, but wonder if this tab should follow the pattern used by the Terminal and Job / Job Launcher tabs, and be closable? Those tabs remember their visibility, so once closed will stay closed (but have View commands to reshow them and will also auto-show if necessary). As we get more and more tabs, nice to be able to close things you don't use and simplify the UI.

kevinushey:
Definitely agree it should be closable.

kevinushey:
Now with a tutorial landing page, built by indexing all installed packages that provide tutorials:

Screen Shot 2019-12-16 at 3 10 18 PM

Clicking a link will load that tutorial, as before.

src/cpp/r/R/Api.R
kevinushey: I'm going to leave this a script job for now until that branch makes its way to master.

PR #5884, opened 2019-12-12
add more control over terminal behavior after shell exits (user pref)

gtritchie:
- previously, there was a boolean option to either keep the pane open, or close it, but behind the scenes was another check that would keep the pane open if shell exited with non-zero code, and the user had no way to override that (this was added in 1.3)

- new option has three choices, always close, never close, close if shell exits cleanly; this gives user control over this new behavior I added in 1.3 (checking exit code)

- I didn't bother migrating the old setting forward, if user had changed it, they will need to do so again on 1.3 upgrade

- Here's what an exited, non-closed terminal pane shows (this isn't new, but just FYI):

- Also simplified setting of another recently added terminal preference in the UI (I was setting state via bruce force if/then/else instead of letting the SelectWidget do it for me)

### DIFF
--- a/src/cpp/session/include/session/SessionConsoleProcessInfo.hpp
+++ b/src/cpp/session/include/session/SessionConsoleProcessInfo.hpp
@@ -53,6 +53,7 @@ enum AutoCloseMode
    DefaultAutoClose = 0, // obey user preference
    AlwaysAutoClose = 1, // always auto-close
    NeverAutoClose = 2, // never auto-close
+   CleanExitAutoClose = 3, // auto-close only if shell returned zero exit code
 };
 
 enum SerializationMode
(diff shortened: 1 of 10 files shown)


src/cpp/session/include/session/SessionConsoleProcessInfo.hpp
```diff
@@ -53,6 +53,7 @@ enum AutoCloseMode
    DefaultAutoClose = 0, // obey user preference
```
MariaSemple: Aren't you also obeying the user preference when AlwaysAutoClose or NeverAutoClose is set?

gtritchie: This pattern is a bit weird, to be sure. Not entirely sure what I was thinking back in the day, but don't want to mess with it too much. There's this "default" setting that gets associated with the ConsoleProcess metadata, then when a process exits, it looks at the user preference and stamps the actual value onto it. So from then on that particular terminal will have that exact setting even if the user preference is changed again.

MariaSemple: So a terminal that's already opened won't be affected by a preference update?

PR #5885, opened 2019-12-12
fix text color of username when toolbar hidden

### LINKED ISSUE #5651: Username barely visible in menubar when toolbar hidden
### System details

    RStudio Edition : Server
    RStudio Version : 1.3.553 (probably been this way a while, but didn't check)
    OS Version      : Linux (or macOS dev server)
    R Version       : Any

- login to RStudio, use default Textmate editor theme and Sky or Modern RStudio theme
- notice username at right end of toolbar

- View / Hide toolbar

- username is just barely visible in menubar (if you squint)

- it is visible if you use Classic RStudio theme

- should always be visible, and should be at least 4.5:1 contrast in default configuration (Textmate/modern)

### Other Info
- also happens in RSP

gtritchie:
- was hardcoded to white, which is almost invisible in light themes
- now it is black in light themes, white in dark themes (same as it is when toolbar is visible)
- Fixes #5651
- also added invisible aria-label of "Username" so screen reader has something to describe this text

Tested on light and dark themes:

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/application/ui/impl/WebApplicationHeader.java
+++ b/src/gwt/src/org/rstudio/studio/client/application/ui/impl/WebApplicationHeader.java
@@ -15,6 +15,7 @@
 
 package org.rstudio.studio.client.application.ui.impl;
 
+import com.google.gwt.aria.client.Roles;
 import com.google.gwt.core.client.GWT;
 import com.google.gwt.core.client.Scheduler;
 import com.google.gwt.dom.client.ImageElement;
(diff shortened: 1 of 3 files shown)


src/gwt/src/org/rstudio/studio/client/application/ui/impl/WebApplicationHeader.java
```diff
@@ -411,6 +412,7 @@ private void initCommandsPanel(final SessionInfo sessionInfo)
          usernameLabel.setTitle(userIdentity);
          userIdentity = userIdentity.split("@")[0];
          usernameLabel.setText(userIdentity);
+         Roles.getPresentationRole().setAriaLabelProperty(usernameLabel.getElement(), "Username");
```
MariaSemple: I know this is duplicated because that's how the code is now, but out of curiosity is there any way that we can refactor some of the code in this file and DesktopApplicationHeader.java so that the common parts aren't duplicated? Not suggesting you do it in this PR, just a thought I had.

gtritchie: Certainly; this code was added quite recently when Karl did the work to support Server Pro sessions with Desktop Pro (prior to that, the desktop never showed username in the header).

MariaSemple: That makes sense, thanks!

PR #5893, opened 2019-12-14
Accessibility command to put focus on main toolbar

### LINKED ISSUE #5458: Need an easier way to get focus to the main toolbar
### System details

    RStudio Version : 1.3.x
    OS Version      : any
    R Version       : any

As a screen reader user, or a keyboard-only user, try to navigate and use the main toolbar, i.e. the area highlighted here. You can operate the toolbar with the keyboard if you get focus to it.

There is no easy way to get focus to the toolbar.

On server, you can use Ctrl+Alt+F to get focus to the File menu (see Keyboard Shortcuts Help for full list), then use tab several times to maneuver to the toolbar, but that is clunky.

On desktop, there's no real way, other than getting focus somewhere without a keyboard trap and using Shift+Tab to work your way up to the toolbar.

A consistent way to get focus to the toolbar that works on both desktop and server.

gtritchie:
New command on accessibility menu for sending keyboard focus to main toolbar.

Fixes #5458

Also adds a framework to the main window so you can cause an aria-live polite announcement by firing an `AriaLiveStatusEvent` with the string you want spoken.

Hooked that up when main toolbar visibility is changed, and also to announce when main toolbar is hidden and thus cannot get focus. None of these are terribly important scenarios, mainly doing to test the new mechanism. Will be using in a bunch of other places in the future.

Also provided this mechanism in modal dialogs so same event can be used to announce things from a modal dialog.

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/a11y/A11y.java
+++ b/src/gwt/src/org/rstudio/core/client/a11y/A11y.java
@@ -128,28 +128,6 @@ public class A11y
       Roles.getAlertRole().setAriaAtomicProperty(el, true);
    }
 
-   /**
-    * Mark a widget as an aria-live status
-    * @param widget
-    * @param assertive
-    */
-   public static void setStatusRole(Widget widget, boolean assertive)
-   {
-      setStatusRole(widget.getElement(), assertive);
-   }
-
-   /**
-    * Mark as widget as an aria-live status
-    * @param el
-    * @param assertive
-    */
-   public static void setStatusRole(Element el, boolean assertive)
-   {
-      Roles.getStatusRole().set(el);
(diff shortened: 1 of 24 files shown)


adamconroy:
@gtritchie did you intend to request review and not assign?

gtritchie:
> @gtritchie did you intend to request review and not assign?

@adamconroy Oops yes, thanks.

ronblum:
Verified in RStudio Server with Chrome, Firefox, and Safari, and RStudio Desktop. There's no visual indicator that the focus is initially on the New File menu, which you can tell by hitting Enter, Tab, Space, and Right Arrow, so I'll file an issue for that.

PR #5900, opened 2019-12-17
Improve accessibility in replace results

melissa-barca:
Added tab indexing, a keyboard shortcut,  and aria-labels.


src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/FindOutputPane.java
```diff
@@ -127,10 +130,10 @@ protected SecondaryToolbar createSecondaryToolbar()
       replaceToolbar_ = new SecondaryToolbar("Replace");
       replaceMode_ = true;
 
-      replaceLabel_ = new Label();
+      replaceLabel_ = new Label("Replace with: ");
```
gtritchie: This needs to be a `FormLabel`. The `Label` class, despite the name, creates a `div` instead of an HTML label, and there is a significant difference for accessibility. You can tell just by clicking on the label in the UI. If focus goes to the associated control (the textbox), then it is properly associated such that a screen reader would understand but that doesn't happen here.

To fix, make this a `FormLabel`, and create it AFTER the `replaceTextBox_` (but before adding either of them to the `replaceToolbar_`). Use the FormLabel constructor that takes both the string and the control to associate with, i.e.:
```
      replaceTextBox_ = new TextBox();
      replaceLabel_ = new FormLabel("Replace with: ", replaceTextBox_);
      replaceToolbar_.addLeftWidget(replaceTextBox_);
      replaceToolbar_.addLeftWidget(replaceLabel_);
```
This will generate a unique element ID for the text box, and then set `for=ID` attribute on the `<label>` thus establishing the connection.

gtritchie:
Oops, still reviewing, hang on.

src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/model/FindResult.java
```diff
@@ -200,28 +200,32 @@ public final SafeHtml getLineHTML()
             {
                if (parts.remove(0).first)
                {
-                  out.appendHtmlConstant("<strong>");
+                  out.appendHtmlConstant("<span aria-label=\"original\"><strong>");
```
gtritchie: I've done some testing with various screen readers, and additional research. This pattern of putting a span inside the `<strong>` and `<em>` does not work as intended and is too awkward for a screen reader user.

The ideal approach is to use the semantically correct HTML: `<del>` element to wrap text being removed (where you currently use `<strong>`) and `<ins>` where you use `<em>`.

These elements reflect what is actually happening, and screen readers are starting to directly support these. For example, NVDA on Firefox speaks "deleted cat", "inserted dog" for `<del>cat</del><ins>dog</ins>`.

VoiceOver on Safari (mac) unfortunately doesn't do this yet. So for coverage there, we'd also want to do the following, note that I'm putting the hidden span INSIDE the `<ins>`, etc. Lots of whitespace added for clarity:

```
<del>
    <span style=visuallyHidden>original</span>
    dog
</del>
<ins>
    <span style=visuallyHidden>replacement</span>
    cat
</ins>
```
To get the `visuallyHidden` style, you have to do it via `ThemeStyles.INSTANCE.visuallyHidden() `.

These elements come with their own styling you'd need to override to meet your needs.

Adding to the complexity, the find results list (when not doing replace) shouldn't use `<del>` or add the embedded `<span>`, it should probably just be `<strong>` on its own.

Not saying you need to do all this now, but I'd definitely NOT do what is in this PR, it actually makes things worse.

src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/FindOutputPane.java
melissa-barca: Fixed!

src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/FindOutputPane.java
```diff
@@ -76,7 +77,8 @@ protected Toolbar createMainToolbar()
       stopSearch_.setVisible(false);
       toolbar.addRightWidget(stopSearch_);
 
-      showFindButton_ = new LeftRightToggleButton("Find", "Replace", false);
+      showFindButton_ = new LeftRightToggleButton("Find", "Replace", true);
+      showFindButton_.setTabIndex(0);
```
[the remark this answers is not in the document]
melissa-barca: Fixed!

src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/model/FindResult.java
melissa-barca: Based on our slack conversation I added the 'del' and 'ins' tags for replace mode - do you want to take another look?

melissa-barca:
> Not specific to this, but noticed while playing with it.
> 
> If you click "replace" with nothing in the replace field, you get this:
> 
> 
> 
> Issues:
> 
> * occurrences is misspelled
> * Screen reader will say "...This will replace 0 occurrences of 'color' with and cannot be undone.", those empty quotes don't generally get spoken, maybe a special-case message when replacing with empty string? Better for sighted users, too.
> * The count of things to be replaced is always zero, at least in all the cases I tried, even though it was going to replace a bunch of things (maybe that's a work in progress?)

I fixed the misspelling issue but couldn't recreate the issue where the replace count is 0. I also updated the message to be more clear when replace is being used to remove a word.

PR #5942, opened 2019-12-30
cleanup static HTML for accessibility, valid HTML

gtritchie:
- remove unnecessary type="text/javascript"
- remove unnecessary type="text/css"
- ensure trailing script is inside body, not after it
- add missing alt to images
- mark cosmetic svg images with aria-hidden=true, also gave them titles though that's redundant in these cases, but for future copy/pasting...
- remove css units on 0 dimensions (0 vs 0px)
- replace obsolete `tt` element with `samp` element
- add missing DOCTYPE html
- add missing charset="utf-8"
- fix color contrast on text
- replace obsolete align="center" with css

### DIFF
--- a/src/cpp/desktop/html/connect.html
+++ b/src/cpp/desktop/html/connect.html
@@ -29,7 +29,7 @@
                 margin-right: auto;
                 background-color: #2268BC;
                 color: white;
-                border: 0px;
+                border: 0;
                 border-radius: 3px;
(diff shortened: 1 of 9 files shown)


gtritchie:
> All looks good as long as IE11 doesn't do any weirdness with more modern CSS.

I tested with IE11 and works as expected.

PR #5943, opened 2019-12-31
option to show only project name instead of full path in desktop window title

### LINKED ISSUE #1817: Replace project folder location with project name on window title
Currently when working with a project in RStudio the window title in Windows 10 displays the location of the folder where the project is, for example "C:/Users/John/Dropbox/R/tom_cruise - RStudio". This is problematic when having several projects open concurrently and the title info gets cut off on the Windows taskbar. The most relevant part of the window title is least likely to be seen being at the end of the title.

Ideally, the folder location info would be removed leaving only the project name, for example "tom_cruise - RStudio".

gtritchie:
Fixes #1817

Added desktop-only option to General / Advanced / OS Integration / "Show full path to project in window title". Off by default, which more closely matches behavior of RStudio Server

On Windows, and some Linux desktops, makes the taskbar labels more useful when working with projects, e.g.:
Annotation 2019-12-30 172152

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -98,6 +98,7 @@
 * RStudio Server now uses 2048 bit RSA keys, for secure communication of encrypted credentials between server / session and client
 * Add support for running Shiny applications as background jobs (#5190)
 * Make maximum lines in R console configurable; was previously fixed at 1000 (#5919)
+* Option to only show project name instead of full path in desktop window title (#1817)
 
 ### Bugfixes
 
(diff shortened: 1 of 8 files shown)


SimonDedman:
Gary/Jonathan,
Are we sure the new behaviour is the intended final result? While the full path is no longer displayed (which meant all tabs looked the same), the window title is Project - branch - RStudio, which doesn't include the object name, which is in a tab on its own, wasting a screen's width of that tab height.
Any chance this could be tweaked to be the object title, and to remove the title tab if it's been popped out and is thus the only tab in the window?
Thanks in advance!
RstudioTitlesStillWrong

PR #5954, opened 2020-01-03
Exclude patterns and files from Find In Files search and replace

melissa-barca:
This PR allows a user to exclude patterns from a Find In Files search by mimicking the current include functionality. It adds a unique option to exclude the “Standard Git exclusions” which uses `git grep` instead of `grep`. Some previous code existed with the understanding this was a “to do” item, most of that has been modified based on the decision to widen the exclude to more than just the gitignore file.

Added unit tests for the regexes used in the find as I was doing quite a bit of messing around with them to figure out what was needed for `git grep`.

image


src/cpp/session/modules/SessionFind.cpp
```diff
@@ -975,20 +977,44 @@ struct ReplaceOptions
...
+
+   core::Error error = runCommand(cmd,
+                                  options,
+                                  &result);
+   if (error)
+      return error;
+
+   *pResultString = result.stdOut;
+   size_t splitAt;
+   while ((splitAt = pResultString->find('\n')) != pResultString->npos)
```
jmcphers: Would it be simpler to use `boost::algorithm::replace_all` here?

src/cpp/session/modules/SessionFind.cpp
```diff
@@ -1060,12 +1086,33 @@ core::Error runGrepOperation(const GrepOptions& grepOptions, const ReplaceOption
...
-   cmd << shell_utils::EscapeFilesOnly << "--" << shell_utils::EscapeAll;
-
-
    // Filepaths received from the client will be UTF-8 encoded;
    // convert to system encoding here.
    FilePath dirPath = module_context::resolveAliasedPath(grepOptions.directory);
+   for (json::Value filePattern : grepOptions.excludeFilePatterns)
+   {
+      // to retrieve the git exclusions, we run a git command and parse the output
+      if (filePattern.getString().compare("gitExclusions") == 0)
```
jmcphers: `getString()` can throw when the `Value` is not a string, so for safety would recommend testing and skipping non-string values in the array (realize this is unlikely but the consequence of an exception here is a full session crash!)

src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/FindInFilesDialog.java
```diff
@@ -104,6 +100,18 @@ private native JsArrayString getFilePatternsNative() /*-{
...
 
+      public final String[] getExcludeFilePatterns()
+      {
+         return JsUtil.toStringArray(getExcludeFilePatternsNative());
+      }
+
+      private native JsArrayString getExcludeFilePatternsNative() /*-{
+
+         if (!this.excludeFilePatterns)
+            this.excludeFilePatterns="";
```
jmcphers: Should this be `[]` since the function returns an array?

src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/FindInFilesDialog.java
```diff
@@ -203,12 +225,26 @@ protected State collectInput()
             list.add(trimmedPattern);
       }
 
+      String excludeFilePatterns =
+         listPresetExcludeFilePatterns_.getValue(
+               listPresetExcludeFilePatterns_.getSelectedIndex());
+      if (excludeFilePatterns == "custom")
```
jmcphers: Nit: for very complicated reasons, we prefer StringUtil.equals (GWT has actually changed the way string comparison works between releases!)

src/cpp/session/modules/SessionFind.cpp
```diff
@@ -1060,12 +1086,33 @@ core::Error runGrepOperation(const GrepOptions& grepOptions, const ReplaceOption
...
+         if (error)
+            return error;
+
+         std::istringstream stream(excludeGitExclusion);
+         std::vector<std::string> results((std::istream_iterator<std::string>(stream)),
+            std::istream_iterator<std::string>());
+         for (std::string filePattern : results)
+         {
+            filePattern.insert(0, dirPath.getAbsolutePath() + "/");
+            cmd << "--exclude=" + filePattern;
```
jmcphers: `ls-files` can return an unbounded amount of output, but most operating systems have a pretty small bound on the size of a shell command, which is a recipe for trouble here since we're trying to add every line to the command. This could cause us to fail when there are a lot of ignored files with long paths.

jmcphers:
I think we need to rework the `ls-files` approach. There are three ways forward I can see:

1. Find a way to pass an unbounded exclusion list or unbounded inclusion list to `grep` (i.e. not in the shell command itself)
2. Filter based on the exclusion list ourselves if it's long; i.e. let grep return everything and then discard the matches in files the user wants to ignore
3. Use `git grep` when using Git exclusions -- possibly the easiest way forward since presumably git grep is installed if you're using git exclusions, and git grep already ignores git exclusions

src/cpp/session/modules/SessionFind.cpp
melissa-barca: Yes, thank you!

src/cpp/session/modules/SessionFind.cpp
melissa-barca: Fixed

src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/FindInFilesDialog.java
melissa-barca: Fixed

src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/FindInFilesDialog.java
melissa-barca: Fixed

melissa-barca:
@jmcphers I like the idea of using `git grep`, do you think we should pass `--untracked --exclude-standard` so that the search contains files the user hasn't committed, but may plan on committing as well?

jmcphers:
> do you think we should pass --untracked --exclude-standard so that the search contains files the user hasn't committed, but may plan on committing as well?

Yes, that's a good idea!

melissa-barca:
In thinking more about this, the git exclusions don't have be _instead of_ the custom filter - so it may make more sense to add the git exclusion as a checkbox. However, this breaks the existing pattern for includes that lets the user choose Common R Source Files, R Scrips, _or_ Custom Filter.

melissa-barca:
@jmcphers @MariaSemple I've updated the code to use `git grep` when a user wants to exclude the standard git files. This took a bit of manipulation of the regexes that parse our results as git handles color display slightly different from grep (I did test these on Windows as well) and we use color to determine the location of the search pattern. Also updated the description for this PR. Please take another look :)

src/cpp/session/modules/SessionFind.cpp
melissa-barca: I remove the `ls-files` code and rewrote `grep` with `git grep`

src/cpp/session/modules/SessionFind.cpp
```diff
@@ -1257,6 +1344,26 @@ core::Error initialize()
    return initBlock.execute();
 }
 
+boost::regex retrieveFileLineNumberLineRegex(bool gitFlag)
```
jmcphers: Naming: `retrieve` usually implies "fetch from somewhere" -- would suggest `getFileLine...` or just `fileLine...`

src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/FindInFilesDialog.ui.xml
```diff
@@ -56,6 +55,20 @@
             <span ui:field="spanPatternExample_" class="{style.example}">Example: *.R, *.r, *.csv. Separate multiple types with commas.</span>
          </div>
       </p>
+
+      <p><rw:FormLabel ui:field="labelExcludeFilePatterns_" text="Exclude these files:"/>
+         <g:ListBox ui:field="listPresetExcludeFilePatterns_"
+                    styleName="{style.presetFilePatterns}">
+            <g:item value="">None</g:item>
+            <g:item value="gitExclusions">Standard Git exclusions</g:item>
```
jmcphers: Because we now require `git` to make this work, selecting this option is always going to result in an error if (a) git is not installed, or (b) the directory to search is not a git repo. How do you feel about hiding or disabling it up front in these cases? (Both are pretty easy to deduce, and (a) is part of SessionInfo so would be an easy change)

melissa-barca: I like the idea of enabling or disabling the option up front and locally implemented the check for if git is installed. Is it possible to determine if a directory is a git repo without a server call? 
Right now I just have it implemented so the back end will alert the user when it starts processing the find request.

src/cpp/session/modules/SessionFind.cpp
melissa-barca: Fixed in latest commit

melissa-barca:
Note: I just added a commit to reorganize how the ReplaceOperationEnded event gets called because there were some outlier cases where the progress bar was still visible after the replace had ended because this wasn't getting called.

src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/FindInFilesDialog.ui.xml
melissa-barca: I added a isGitDirectory RPC to Files Server and call this when the directory changes to check if it's a git repo or not - then enable/disable the git option accordingly. This is ready for another review.

src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/FindOutputPresenter.java
```diff
@@ -509,6 +525,33 @@ public void onResponseReceived(String handle)
...
+      dialog.getDirectoryChooser().addValueChangeHandler(new ValueChangeHandler<String>() {
+         @Override
+         public void onValueChange(ValueChangeEvent<String> event)
+         {
+            fileServer_.isGitDirectory(dialog.getDirectory(),
+                                   new ServerRequestCallback<Boolean>() {
+               @Override
+               public void onResponseReceived(Boolean isGitDirectory)
+               {
+                  if (isGitDirectory)
```
jmcphers: Nit: GWT has automatic unboxing so this could be simplified to `dialog.setGitStatus(isGitDirectory)`

src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/FindOutputPresenter.java
```diff
@@ -509,6 +525,33 @@ public void onResponseReceived(String handle)
          }
       });
 
+      dialog.getDirectoryChooser().addValueChangeHandler(new ValueChangeHandler<String>() {
```
jmcphers: Maybe we should only add this value change handler when `session_.getSessionInfo().isVcsAvailable` is true? (otherwise there's no point as we'd never show git options)

melissa-barca: done 👍

src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/FindOutputPresenter.java
melissa-barca: Fixed

jmcphers:
LGTM!

src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/FindInFilesDialog.java
```diff
@@ -58,15 +62,15 @@ public static native State createState(String query,
                                              String path,
                                              boolean regex,
                                              boolean caseSensitive,
-                                             boolean gitIgnore,
-                                             JsArrayString filePatterns) /*-{
+                                             JsArrayString filePatterns,
```
MariaSemple: Nit: could you rename this `includeFilePatterns` to mirror `excludeFilePatterns`, as well as other occurrences? I think that would make it a bit clearer when just glancing over the code.

src/cpp/session/modules/SessionFind.hpp
```diff
@@ -31,6 +31,10 @@ core::json::Object findInFilesStateAsJson();
 
 core::Error initialize();
 
+boost::regex getFileLineNumberLineRegex(bool gitFlag);
```
MariaSemple: We discussed on slack renaming this to `getGrepOutputRegex`.

src/cpp/session/modules/SessionFind.cpp
```diff
@@ -1055,18 +1130,25 @@ core::Error runGrepOperation(const GrepOptions& grepOptions, const ReplaceOption
...
+      cmd << shell_utils::EscapeFilesOnly << "--" << shell_utils::EscapeAll;
+      cmd << string_utils::utf8ToSystem(dirPath.getAbsolutePath());
+   }
+   else
+   {
+      for (json::Value filePattern : grepOptions.filePatterns)
+      {
+         cmd << shell_utils::EscapeFilesOnly << "--" << shell_utils::EscapeAll;
+         cmd << filePattern.getString();
+      }
```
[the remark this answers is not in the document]
melissa-barca: Done! I left the constructor as is since we need to initialize it to something and in most cases I don't think we'll be using git grep.

src/cpp/session/modules/SessionFind.cpp
```diff
@@ -786,10 +811,16 @@ class GrepOperation : public boost::enable_shared_from_this<GrepOperation>
...
+         if (regex_utils::match(
+               line, match, getFileLineNumberLineRegex(findResults().gitFlag())))
          {
             std::string file = module_context::createAliasedPath(
                   FilePath(string_utils::systemToUtf8(match[1])));
+            if (findResults().gitFlag())
+            {
+               file.insert(0, "/");
+               file.insert(0, findResults().path());
+            }
```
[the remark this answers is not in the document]
MariaSemple: Ah, makes sense - thanks!

src/cpp/session/modules/SessionFind.cpp
```diff
@@ -786,10 +811,16 @@ class GrepOperation : public boost::enable_shared_from_this<GrepOperation>
 
          errorMessage.clear();
          boost::smatch match;
-         if (regex_utils::match(line, match, boost::regex("^((?:[a-zA-Z]:)?[^:]+):(\\d+):(.*)")))
+         if (regex_utils::match(
+               line, match, getFileLineNumberLineRegex(findResults().gitFlag())))
          {
             std::string file = module_context::createAliasedPath(
                   FilePath(string_utils::systemToUtf8(match[1])));
```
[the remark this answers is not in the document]
MariaSemple: The paranoid part of me still thinks a check is a good idea. What if boost lies to us and then the whole IDE crashes?

src/cpp/session/modules/SessionFind.cpp
```diff
@@ -1081,7 +1163,8 @@ core::Error runGrepOperation(const GrepOptions& grepOptions, const ReplaceOption
                              grepOptions.searchPattern,
                              grepOptions.directory,
                              grepOptions.asRegex,
-                             grepOptions.ignoreCase);
+                             grepOptions.ignoreCase,
+                             gitFlag);
```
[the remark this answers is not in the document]
MariaSemple: Which other filePattern information?

src/cpp/session/modules/SessionFind.cpp
```diff
@@ -1181,16 +1267,17 @@ core::Error completeReplace(const json::JsonRpcRequest& request,
...
+                                  &excludeFilePatterns,
                                   &originalFindCount,
-                                  &replacePattern,
-                                  &useGitIgnore);
+                                  &replacePattern);
    if (error)
       return error;
 
    static const int kUpdatePercent = 5;
    LocalProgress* pProgress = new LocalProgress(originalFindCount, kUpdatePercent);
```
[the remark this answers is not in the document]
MariaSemple: Sure. I think either way a comment would clear it up, since 5 was chosen somewhat arbitrarily (presumably because it seemed to have the best balance of feeling responsive but not being overly frequent?). If you did any testing that led to the choice of 5%, I think it would be good to note it in the code! :)

src/cpp/session/modules/SessionFind.cpp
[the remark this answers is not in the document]
melissa-barca: excludeFilePatterns stores other data that gets added to the --exclude argument in the grep command, which is currently handled in processExcludeFilePatterns. We could pretty easily just call this function before initializing GrepOptions since it needs to be called at some point, but it seems a little weird because it is building actual arguments for the command, not just setting up general arguments.

src/cpp/session/modules/SessionFind.cpp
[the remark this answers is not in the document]
melissa-barca: Ok - added 😸

src/cpp/session/modules/SessionFind.cpp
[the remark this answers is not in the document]
melissa-barca: Ok added a comment

src/cpp/session/modules/SessionFind.cpp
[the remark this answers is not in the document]
MariaSemple: Looking at processExcludeFilePatterns, I do think those would make sense to be part of GrepOptions since all of those fields both change the grep options that will be used and impact nothing else, unless I'm misunderstanding and there isn't a unique GrepOptions per search (and optionally replace) request.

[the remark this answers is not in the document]
melissa-barca: Done 🎉

src/cpp/session/modules/SessionFind.cpp
```diff
@@ -987,26 +987,93 @@ class GrepOperation : public boost::enable_shared_from_this<GrepOperation>
...
+   {
+      return includeFilePatterns_;
+   }
+
+   const json::Array& excludeFilePatterns() const
+   {
+      return excludeFilePatterns_;
+   }
+
+   const bool gitFlag() const
```
MariaSemple: Nit: no need to return a const bool since you're returning by-value rather than by-reference.

melissa-barca: fixed

PR #5990, opened 2020-01-09
change keyboard shortcuts for next/prev terminal to use Shift instead of Ctrl

### LINKED ISSUE #4892: Keyboard shortcuts to cycle terminals
Next terminal and previous terminal would be welcome shortcut options. Thanks!

gtritchie:
Fixes #4892

- avoids a clash with commonly used shortcuts on Windows
- issue reported it on Dell XPS, I was also seeing the same problem on my Lenovo IdeaPad
- tested the next shortcuts on those, macOS Desktop, and Server on Mac and Windows in Chrome

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -102,6 +102,7 @@
 * Option to only show project name instead of full path in desktop window title (#1817)
 * New `rstudio --version` command to return the version of RStudio Desktop (#3922)
 * Scan R Markdown YAML header for R packages required to render document (#4779)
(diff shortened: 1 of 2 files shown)


gtritchie:
> Pretty funky key binds either way. I imagine people would customize these if they're used very commonly anyway.

That's a pretty tame combo by keyboard shortcut standards. There are screen reader combos that use 5 simultaneous keys, e.g. Ctrl+Opt+Cmd+Fn+8, or Ctrl+Opt+Cmd+Shift+Space. Not making those up! :-)

PR #6002, opened 2020-01-13
Accessibility improvements to editor pane

gtritchie:
- when document outline is visually hidden also make it invisible to screen readers
- markup document outline splitter as a splitter
- give document outline a label for accessibility tools
- increase contrast of message shown when document outline is empty
- option to read aria-status messages with no delay
- clear prior aria-live status message after 4 seconds (this is to allow a repeated message to be read and so it doesn't show up again when navigating page with screen reader cursor)
- new command "Speak text editor location" (Ctrl+Shift+B) that reads out current row, col, scope, file type and filename via aria-live; command hidden if screen reader support disabled
- annotate the panels for each tab in editor pane with role=tabpanel and a label (had to be done for text editors, data browser, object browser, profile results, code browser, url viewer)

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/widget/AriaLiveStatusWidget.java
+++ b/src/gwt/src/org/rstudio/core/client/widget/AriaLiveStatusWidget.java
@@ -1,7 +1,7 @@
 /*
  * AriaLiveStatusWidget.java
  *
- * Copyright (C) 2019 by RStudio, Inc.
+ * Copyright (C) 2019-20 by RStudio, Inc.
  *
  * Unless you have received this program directly from RStudio pursuant
  * to the terms of a commercial license agreement with RStudio, then
@@ -36,14 +36,15 @@ public class AriaLiveStatusWidget extends Widget
    {
       if (speakDelayMs < 0)
       {
-         if (updateReaderTimer_.isRunning())
-            updateReaderTimer_.cancel();
+         clearMessage();
          return;
       }
 
       resultsMessage_ = message;
       if (updateReaderTimer_.isRunning())
          updateReaderTimer_.cancel();
+      if (clearReaderTimer_.isRunning())
+         clearReaderTimer_.cancel();
       updateReaderTimer_.schedule(speakDelayMs);
    }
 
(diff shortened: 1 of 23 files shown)


src/gwt/src/org/rstudio/core/client/widget/AriaLiveStatusWidget.java
```diff
@@ -63,8 +66,23 @@ public void clearMessage()
...
+    */
+   private Timer clearReaderTimer_ = new Timer()
+   {
+     @Override
+     public void run()
+     {
+        getElement().setInnerText("");
+     }
+   };
+
```
adamconroy: This seems to just be an extension of what's already there but I always get a little nervous when multiple timers start being about. This looks sound but I always ask in this situation "are there any funny conditions we could get in?"

There are some hardcoded times (like the 4000ms on line 71) as well that, at least in the next revision, might want to get pulled out to at least class constants. Not a big deal for this scope but probably something to think about as these things continue to grow in complexity.

If we end up with a handful of timers in flight in the future it may be worth offloading some of these status changes to an event bus system.

gtritchie: All fair points. I hate when I add (or see) timers in the code, but if you look, we have dozens of them, and most have hardcoded values. That doesn't make it a "good thing".

In this case, there are definitely "funny conditions we could get in", and they are unknowable without testing combinations of screen readers, browsers, user controlled settings for screen reader speech rate, and length of the messages being injected. Our code has no idea how long it will take for a given message to be read, or if a screen reader will read the message even after the text has been cleared.

Two reasons I'm doing this clear after a delay:

1) The text in a live-alert region, although invisible to sighted users, is visible to screen readers so if the user is navigating through the UI with the screen reader cursor, they will come across it and it is weird to get whatever the last announcement was (which might have happened hours / days ago) read again. So I want to clear it after a while to reduce the chances of that.

2) If you use the new command I added that reads the current editor status (row, column, etc), it is quite reasonable that a user might issue that command again while on the same line (missed details the first time). A screen reader will not read the same announcement twice in a row, so clearing it increases the chances of them hearing it the second time (but only if they wait 4 seconds, so far from perfect). Another strategy would be to have two live regions, and have the code alternate between them with each announcement, clearing the other one when updating the current one. Would still want the timer to clear each region after some amount of time, but would be more robust in general, I'll open a task to look into that in the future.

PR #6003, opened 2020-01-13
use aria-pressed to make LatchingToolbarButton a toggle button

gtritchie:
- in one use of this control (for showing/hiding document outline), the accessibility text is changed to indicate the state of the button; in that situation it is NOT appropriate to use aria-pressed as the text is sufficient to tell screen readers what is going on
- in the other use, the Filter button in the data viewer, the text does not change, so aria-pressed is required to indicate that this is a toggle button and to indicate what state it is currently in

## Toggle Button example
Filter toggle button shown in the RStudio data viewer

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/widget/LatchingToolbarButton.java
+++ b/src/gwt/src/org/rstudio/core/client/widget/LatchingToolbarButton.java
@@ -1,7 +1,7 @@
 /*
  * LatchingToolbarButton.java
  *
- * Copyright (C) 2009-19 by RStudio, Inc.
+ * Copyright (C) 2009-20 by RStudio, Inc.
  *
  * Unless you have received this program directly from RStudio pursuant
  * to the terms of a commercial license agreement with RStudio, then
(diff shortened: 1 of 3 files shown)


src/gwt/src/org/rstudio/core/client/widget/LatchingToolbarButton.java
```diff
@@ -39,10 +49,19 @@ public void setLatched(boolean latched)
...
       if (latched_)
+      {
+         if (!textIndicatesState_)
+            Roles.getButtonRole().setAriaPressedState(getElement(), PressedValue.TRUE);
          getElement().addClassName(ThemeStyles.INSTANCE.toolbarButtonLatched());
+      }
       else
+      {
+         if (!textIndicatesState_)
+            Roles.getButtonRole().setAriaPressedState(getElement(), PressedValue.FALSE);
```
adamconroy: Could be worth pulling these out into a local helper method that's merely `setAriaPressedState(TRUE/FALSE)` and it does the `if` check and the longer incantation (if you ever get tired of copy/pasting all that across all these!)

PR #6051, opened 2020-01-17
Fixes a bug where some Error constructors are ambiguous.

MariaSemple:
JJ reported this bug earlier this week. Because `boost::system::error_code`s can be implicitly converted to `int`, sometimes `Error(int code, std::string name, ErrorLocation location)` was being invoked instead of `Error(boost::system::error_code, std::string message, ErrorLocation location)`, resulting in the name (category) of the error being set to the intended message. This PR resolves that issue by swapping the name and code parameters. It also revealed the need for `Error(boost::system::error_condition)` constructors.

There's also a minor change to SessionFind errors.


MariaSemple:
The companion Pro branch is also ready for merge.

PR #6100, opened 2020-01-27
Fix some shared_core bugs and comments

MariaSemple:
This PR fixes the following bugs:
* unimplemented overload of `systemError`
* missing deep comparison for `json::Value`
* inability to construct empty `User`
* abiltiy to parse a JSON array into a `json::Object` and vice versa

This PR also fixes some doxygen comments.


src/cpp/core/CMakeLists.txt
```diff
@@ -355,6 +354,7 @@ if (RSTUDIO_UNIT_TESTS_ENABLED)
       ${CORE_HEADER_FILES}
    )
 
+   message("Core Boost: ${Boost_LIBRARIES}")
```
kevinushey: Did you intend to leave this in? (seems likely it was just for debugging?)

MariaSemple: Oops! No, I did not. Thanks for catching that!

PR #6109, opened 2020-01-29
Blacklist for Typo.js unsupported dictionaries

adamconroy:
Reported in issue #6041 there are some dictionaries that Typo.js has severe issues loading. In fact, due to the size of these dictionaries in memory for most tested spellchecking libraries (6GB for Lithuanian in a tested Python package), some serious problem solving will have to go into this issue at some point.

The discussed stopgap to implement was to add a blacklist for the known incompatible dictionaries and not allow them to be loaded for real time spellchecking. Due to how preferences are saved and handled there was a fair amount of funny business making sure that [hopefully] all weird edge cases would be caught to prevent the user from getting into a bad state.

On the plus side I found and fixed a bug with turning real time spellchecking off and then on again.

spellcheck_unavailable

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/common/spelling/TypoSpellChecker.java
+++ b/src/gwt/src/org/rstudio/studio/client/common/spelling/TypoSpellChecker.java
@@ -183,11 +183,12 @@ public class TypoSpellChecker
 
       // subscribe to spelling prefs changes (invalidateAll on changes)
       ValueChangeHandler<Boolean> prefChangedHandler = (event) -> context_.invalidateAllWords();
+      ValueChangeHandler<Boolean> realtimeChangedHandler = (event) -> loadDictionary();
       ValueChangeHandler<String> dictChangedHandler = (event) -> loadDictionary();
-      userPrefs_.realTimeSpellchecking().addValueChangeHandler(prefChangedHandler);
       userPrefs_.ignoreUppercaseWords().addValueChangeHandler(prefChangedHandler);
       userPrefs_.ignoreWordsWithNumbers().addValueChangeHandler(prefChangedHandler);
       userPrefs_.spellingDictionaryLanguage().addValueChangeHandler(dictChangedHandler);
+      userPrefs_.realTimeSpellchecking().addValueChangeHandler(realtimeChangedHandler);
 
       // subscribe to user dictionary changes
       context_.releaseOnDismiss(userDictionary_.addListChangedHandler((ListChangedEvent event) ->
@@ -335,18 +336,28 @@ public class TypoSpellChecker
 
    private void loadDictionary()
    {
-      // don't load the same dictionary again
       final String language = userPrefs_.spellingDictionaryLanguage().getValue();
-      if (typoLoaded_ && loadedDict_ == language)
+
+      if (!userPrefs_.realTimeSpellchecking().getValue() || typoLoaded_ && loadedDict_.equals(language))
          return;
-      
+
+      // See canRealtimeSpellcheckDict comment, temporary stop gap
+      // Disable real time spellchecking if this dictionary is incompatible
+      // with Typo.js. Final invariant check to ensure that we never try to
+      // load a blacklisted dictionary.
+      if (!canRealtimeSpellcheckDict(language))
+      {
+         userPrefs_.realTimeSpellchecking().setGlobalValue(false);
+         return;
+      }
+
       // check for an active request
       if (activeRequest_ != null && activeRequest_.isAlive())
       {
          // if we're already requesting this language's dictionary, bail
          if (StringUtil.equals(activeRequest_.getLanguage(), language))
             return;
-         
+
          // otherwise, cancel that request and start a new one
          activeRequest_.cancel();
          activeRequest_ = null;
@@ -396,6 +407,26 @@ public class TypoSpellChecker
       }
 
       return true;
+   }
+
+   /*
+      Stop gap function to prevent loading dictionaries that Typo.js has
+      severe issues with. This is being tracked to be fixed in issue #6041 as
+      soon as possible so this can then be removed.
+    */
+   private static String[] realtimeDictBlacklist = {"cs_CZ", "de_DE_neu", "lt_LT", "pt_BR"};
+   public static boolean canRealtimeSpellcheckDict(String dict)
+   {
+      boolean exists = false;
+      for (int i = 0; i < realtimeDictBlacklist.length; i++)
+      {
+         if (realtimeDictBlacklist[i].equals(dict))
+         {
+            exists = true;
+            break;
+         }
+      }
+      return !exists;
    }
 
    public static boolean isLoaded() { return typoLoaded_; }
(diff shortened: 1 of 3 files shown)


src/gwt/src/org/rstudio/studio/client/common/spelling/TypoSpellChecker.java
```diff
@@ -183,11 +183,12 @@ public TypoSpellChecker(Context context)
 
       // subscribe to spelling prefs changes (invalidateAll on changes)
       ValueChangeHandler<Boolean> prefChangedHandler = (event) -> context_.invalidateAllWords();
+      ValueChangeHandler<Boolean> realtimeChangedHandler = (event) -> loadDictionary();
       ValueChangeHandler<String> dictChangedHandler = (event) -> loadDictionary();
-      userPrefs_.realTimeSpellchecking().addValueChangeHandler(prefChangedHandler);
       userPrefs_.ignoreUppercaseWords().addValueChangeHandler(prefChangedHandler);
       userPrefs_.ignoreWordsWithNumbers().addValueChangeHandler(prefChangedHandler);
       userPrefs_.spellingDictionaryLanguage().addValueChangeHandler(dictChangedHandler);
+      userPrefs_.realTimeSpellchecking().addValueChangeHandler(realtimeChangedHandler);
```
adamconroy: If the app had started up with real time turned off, turning it on without restarting the app wouldn't then call `loadDictionary()` to ensure that Typo.js was loaded.

src/gwt/src/org/rstudio/studio/client/common/spelling/TypoSpellChecker.java
```diff
@@ -398,6 +409,26 @@ public boolean shouldCheckSpelling(DocDisplay dd, Range r)
       return true;
    }
 
+   /*
+      Stop gap function to prevent loading dictionaries that Typo.js has
+      severe issues with. This is being tracked to be fixed in issue #6041 as
+      soon as possible so this can then be removed.
+    */
+   private static String[] realtimeDictBlacklist = {"cs_CZ", "de_DE_neu", "lt_LT", "pt_BR"};
```
adamconroy: I spot checked a few other dictionaries and they seemed ok, these are the ones listed in the bug. Adding more to this list will be easy and the intention is to not have this problem for too long. If we decide to not spend more time working on this issue, which could be reasonable considering all the variables, I'd move this to a more permanent external home.

By removing `canRealtimeSpellcheckDict` we can easily be pointed to and clear out all of the subsequent cruft that uses it.

src/gwt/src/org/rstudio/studio/client/common/spelling/TypoSpellChecker.java
```diff
@@ -183,11 +183,12 @@ public TypoSpellChecker(Context context)
 
       // subscribe to spelling prefs changes (invalidateAll on changes)
       ValueChangeHandler<Boolean> prefChangedHandler = (event) -> context_.invalidateAllWords();
+      ValueChangeHandler<Boolean> realtimeChangedHandler = (event) -> loadDictionary();
```
jmcphers: Is the dictionary load idempotent? (seems like if it isn't we'd want to make sure it isn't loaded and avoid loading when the pref changes to false)

src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/SpellingPreferencesPane.java
```diff
@@ -54,12 +57,28 @@ public SpellingPreferencesPane(GlobalDisplay globalDisplay,
...
+      blacklistWarning_.setVisible(!canRealtime);
+
+      add(blacklistWarning_);
+
+      languageWidget_.addChangeHandler((event) -> {
+         boolean canRealtimeCheck = TypoSpellChecker.canRealtimeSpellcheckDict(languageWidget_.getSelectedLanguage());
+         blacklistWarning_.setVisible(!canRealtimeCheck);
+         realtimeSpellcheckingCheckbox_.setValue(realtimeSpellcheckingCheckbox_.getValue() && canRealtimeCheck);
+         realtimeSpellcheckingCheckbox_.setEnabled(canRealtimeCheck);
+         realtimeSpellcheckingCheckbox_.getElement().getStyle().setColor(canRealtimeCheck ? "black" : "darkgrey");
```
jmcphers: There are a number of RStudio themes in the wild which attempt to dark-theme the dialogs. This is not technically supported but this will break them. What do you think about using opacity instead?

src/gwt/src/org/rstudio/studio/client/common/spelling/TypoSpellChecker.java
```diff
@@ -335,18 +336,28 @@ private void updateIgnoredWordsIndex()
...
          return;
-      
+
+      // See canRealtimeSpellcheckDict comment, temporary stop gap
+      // Disable real time spellchecking if this dictionary is incompatible
+      // with Typo.js. Final invariant check to ensure that we never try to
+      // load a blacklisted dictionary.
+      if (!canRealtimeSpellcheckDict(language))
+      {
+         userPrefs_.realTimeSpellchecking().setGlobalValue(false);
```
jmcphers: I think we should avoid committing a pref change here as it's extremely implicit -- prefs should generally only change due to a user gesture that indicates intent since users can now diff their prefs. Would just bailing from the check do the job here?

src/gwt/src/org/rstudio/studio/client/common/spelling/TypoSpellChecker.java
adamconroy: I should probably just rename the function. `loadDictionary` checks the values of all relevant preferences now before proceeding.

src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/SpellingPreferencesPane.java
adamconroy: I'll give it a shot and see if it looks ok. Would be curious to play around with those themes!

src/gwt/src/org/rstudio/studio/client/common/spelling/TypoSpellChecker.java
adamconroy: That should be fine. It's a little weird from the code side because the `LintManager` and other relevant bits will proceed as if real time checking is on but if we always no-op out of the calls I don't see the harm as the user will see the disabled message when they check options.

src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/SpellingPreferencesPane.java
adamconroy: It looks good!
ncpqAbB

jmcphers: Here's one such theme: https://github.com/anthonynorth/rscodeio

Issue #6118, opened 2020-01-29
Keyboard shortcuts reference not accessible to screen readers


gtritchie:
### System details

    RStudio Edition : Desktop or Server
    RStudio Version : 1.3.795
    OS Version      : Any
    R Version       : Any

### Steps to reproduce the problem
With a screen reader on, do Help / Keyboard Shortcuts Help to display keyboard shortcuts reference, then try to read it via screen reader.

### Describe the problem in detail
Any attempt to use the screen reader immediately dismisses the popup.

### Describe the behavior you expected
It is important for a screen reader user to become familiar with RStudio shortcuts to use the product efficiently. Though many can be found by browsing the main menu, many of them are only associated with toobar buttons in the various panes, making them even harder to discover. Having a single-place to browse shortcuts would be very helpful.

When screen-reader mode is enabled, bypass the popup-up and directly load the static HTML keyboard shortcut page normally reached via the (also not accessible) link "See all Shortcuts..." in upper-right of popup. This will allow browsing it via normal HTML screen reader commands.

See also #6117 for some issues with that static page.

ronblum:
Confirmed in RStudio Server and Desktop on Mac and Windows. Works as expected with screen reader support on and off.

PR #6120, opened 2020-01-30
Improve screen reader accessibility of "All Keyboard Shortcuts" page

gtritchie:
This static, manually edited page (see #6117) is not very accessible to screen readers. Plan is to expose this page to screen reader users via #6118. Mostly mechanical and lots of search and replace. Still need to check for changes to shortcuts in 1.3, will do that separately.

Sorry, didn't notice editor had reindented the HTML until I had done all the work. Makes very difficult to review, but I have tested the resulting page with all browsers we care about, validated the HTML (strictly compliant), and re-ran accessibility tests (100% passing).

Details:

- use separate tables (with headings) for each section, instead of one table with headings inside the table
- allows screen readers to see individually named tables
- properly associate headings with columns so screen readers can announce
- switch Mac shortcuts to use Mac Unicode characters (these are supported by VoiceOver on Mac, and we use them on our automatically generated shortcuts overlay)

screenshot of first page of keyboards shortcut page

### DIFF
--- a/src/gwt/www/docs/keyboard.htm
+++ b/src/gwt/www/docs/keyboard.htm
@@ -2,854 +2,934 @@
 <html lang="en">
 
 <head>
+  <title>RStudio: Keyboard Shortcuts</title>
+  <link rel="stylesheet" href="../rstudio.css" type="text/css"/>
 
-<title>RStudio: Keyboard Shortcuts</title>
-
-<link rel="stylesheet" href="../rstudio.css" type="text/css"/>
-
-<style type="text/css">
-#banner {
-  margin-bottom: 0;
-}
-.shortcuts th {
-  text-align: left;
-  padding-right: 20px;
-}
-.shortcuts, .shortcuts td {
-  font-size: 10pt;
-  padding-right: 20px;
-}
-</style>
-
+  <style>
+    #banner {
+      margin-bottom: 0;
+    }
+    .shortcuts th[scope=col] {
+      text-align: left;
+      width: 360px;
+    }
+    .shortcuts th[scope=col]+th[scope=col] {
+      width: 320px;
+    }
+    .shortcuts, .shortcuts td {
+      font-size: 10pt;
+      padding-right: 20px;
+    }
+    .shortcuts td, .shortcuts th {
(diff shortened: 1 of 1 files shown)


gtritchie:
> Might be worth moving away from over use of tables to flexbox when we can to modernize a bit...

In this case table is more semantically appropriate. It is showing tabular data and screenreaders have a specific commandset and behaviors for that. They will show a list of tables, by name, and allow navigating by cell, speak the name of a column, read entire rows or columns, tell hou how many rows and columns there are, etc.

Tables purely for layout are “bad” but tables for “tables” are “good”.

adamconroy:
Flex lives in a bit of a middle ground where it can do tabular data and nicely lay things out at the same time. I can't speak to how well screenreaders can navigate it but typically this type of page would be the #1 candidate for converting from a table to Flex.

gtritchie:
Flex produces inaccessible pages when used for tabular data, and also if it is used to reorder visual layout differently than DOM order (since it messes up visual tab order, a serious WCAG violation).

To make a flex-based tabular data “table” usable with screen reader you have to include aria annotations (sometimes multiple ones) on every cell, every row, every column, every header, and so forth.

adamconroy:
That's very unfortunate but good to know.

gtritchie:
Our terminal emulator (xterm.js) uses canvas for rendering. When screen-reader support is enabled it creates and maintains a visually-hidden but aria-visible DOM representation of what's showing in the canvas, and exposes that to accessibility tools. This is separate from the live-region it uses to announce recent output (same model I used for the console).

So there are (almost) always ways to make it accessible, especially if considered up-front.

gtritchie:
Number one principal of accessible HTML is to use semantically correct elements when they exist. So `<button>` instead of a `<div>` trying to act like a button, and so forth. Same with tables, headings, various form controls, landmark elements, etc. Gives maximum screen-reader support, maximum built-in keyboard support and so forth.

The traditional way of thinking about it (and testing it) is whether the page would be usable and understandable (both to a visual user and a screen-reader user) if all custom CSS is disabled. That's not an experiment / goal I'm bothering with given the nature of RStudio's UI.

PR #6131, opened 2020-01-31
Adds convenience push_back and insert functions to Json.cpp

MariaSemple:
This PR allows users to `push_back` and `insert` literal-typed values onto `json::Array`s and `json::Object`s. The ability to do this was removed when the `json::Value` conversion constructors were made explicit. Since the intention to convert from literal type to `json::Value` is obvious when `push_back` or `insert` are called, this adds back the convenience of calling those functions without explicitly calling the `json::Value` constructor without allowing for accidental implicit conversion in other locations.


kevinushey:
Are we sure we need a separate method for each type we want to support? Why not just use a template?

MariaSemple:
I tried a template implementation first, but I got a lot of compile errors about `int&` being an invalid type for `json::Value` (or something like that) that didn't make sense to me (since I would expect the int to just be copied if need be). I thought it would be faster to implement each type than try and resolve the error. I can give it another try if you think it's worth it though.

PR #6132, opened 2020-02-01
Adds changeOwnership and changeFileMode functions to FilePath

MariaSemple:
Also adds some posix system utility functions to shared_core for use in the Launcher Plugin SDK.


src/cpp/server/ServerMain.cpp
```diff
@@ -571,7 +572,12 @@ int main(int argc, char * const argv[])
             return true;
          };
 
-         error = file_utils::changeOwnership(serverDataDir, options.serverUser(), true, shouldChown);
+         system::User serverUser;
```
kevinushey: In what contexts might this fail? I'm mainly wondering because (IIUC) this adds a new way for `rserver` to fail to start and just want to make sure we understand the failure modes (and are convinced they're rare or understandable + easy to diagnose).

Alternatively: what do you think about using `options.serverUser()` by default, and having an error here be a warning rather than fatal? Or would this ultimately just fail somewhere else further down?

src/cpp/server/auth/ServerAuthHandler.cpp
```diff
@@ -454,9 +454,17 @@ Error initialize()
       return error;
    }
 
+   core::system::User serverUser;
+   error = core::system::User::getUserFromIdentifier(options().serverUser(), serverUser);
```
kevinushey: Similar question here.

src/cpp/shared_core/include/shared_core/FilePath.hpp
```diff
@@ -231,6 +260,53 @@ class FilePath
     */
    static Error uniqueFilePath(const std::string& in_basePath, const std::string& in_extension, FilePath& out_filePath);
 
+#ifndef _WIN32
+   /**
+    * @brief Changes the file mode to the specified file mode.
+    *
+    * @param in_fileModeStr     The posix file mode string. e.g. rwxr-xr-x.
```
kevinushey: Is it worth documenting that only a subset of posix file mode strings are supported?

src/cpp/shared_core/include/shared_core/FilePath.hpp
```diff
@@ -231,6 +260,53 @@ class FilePath
...
+    */
+   Error changeFileMode(FileMode in_fileMode, bool in_setStickyBit = false) const;
+
+   /**
+    * @brief Changes the ownership of the file or directory to the specified user.
+    *
+    * @param in_newUser         The user who should own the file.
+    * @param in_recursive       If this FilePath is a directory, whether to recursively change ownership on all files
+    *                           and directories within this directory.
+    * @param in_shouldChown     A recursive iteration function which allows the caller to filter files and directories.
```
kevinushey: minor nit: the lines here are just long enough to force GitHub's PR UI to overflow horizontally a bit; e.g.

Screen Shot 2020-02-03 at 10 38 33 AM

Not sure if this is motivation enough for us to adjust our editor line widths down a bit.

src/cpp/shared_core/system/PosixSystem.cpp
```diff
@@ -0,0 +1,134 @@
...
+#include <shared_core/system/User.hpp>
+
+namespace rstudio {
+namespace core {
+namespace system {
+namespace posix {
+
+Error enableCoreDumps()
+{
+#ifndef __APPLE__
```
kevinushey: Should we return a "not supported" error for `__APPLE__`? (maybe not as part of this PR since I'm guessing this is mostly just moving things into `shared_core`?)

src/cpp/shared_core/system/PosixSystem.cpp
```diff
@@ -0,0 +1,134 @@
...
+{
+   struct sigaction sa;
+   ::memset(&sa, 0, sizeof(sa));
+
+   sa.sa_handler = SIG_IGN;
+   int result = ::sigaction(in_signal, &sa, nullptr);
+   if (result != 0)
+   {
+      Error error = systemError(result, ERROR_LOCATION);
+      error.addProperty("signal", in_signal);
```
kevinushey: Is there any easy way for us to convert the signal number to the signal name here? (just makes debugging a tad easier)

src/cpp/shared_core/system/PosixSystem.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+   // Change the effective user to root.
+   if (::seteuid(0) < 0)
+      return systemError(errno, ERROR_LOCATION);
+   // Verify
+   if (::geteuid() != 0)
+      return systemError(EACCES, ERROR_LOCATION);
+
+   // Get user info to use in group calls
+   struct passwd* pPrivPasswd = ::getpwuid(0);
```
kevinushey: IIUC errors here and below would lead us to still having root, just not having the effective group set. Is it acceptable for us to end up in this state, or should we try to make sure this function is 'atomic'? (not sure atomic is the best word but it feels like the closest analogue here)

src/cpp/shared_core/system/PosixSystem.cpp
```diff
@@ -0,0 +1,134 @@
...
+   errno = 0;
+
+   // init supplemental group list
+   // NOTE: if porting to CYGWIN may need to call getgroups/setgroups
+   // after initgroups -- more research required to confirm
+   if (::initgroups(in_user.getUsername().c_str(), in_user.getGroupId()) < 0)
+      return systemError(errno, ERROR_LOCATION);
+
+   // set group and verify
+   if (::setresgid(-1, in_user.getGroupId(), ::getegid()) < 0)
```
kevinushey: Similar question re: atomicity in this function.

src/cpp/server/ServerMain.cpp
MariaSemple: The work to get the `system::User` and the error check for that were done inside `file_utils::changeOwnership` before. The SDK options class has a `Error error = serverUser(system::User& out_user)` method, so I wanted this function to take the User object rather than the username.

It _should_ only fail if the server user doesn't exist, which really shouldn't be possible at this point, since a misconfigured rserver.conf should have caused the process to fail much earlier than this.

src/cpp/shared_core/include/shared_core/FilePath.hpp
MariaSemple: Yes, that seems worth it to me. I'm not sure which are supported and which aren't off-hand. Do you know where I can find that information? Or, if you happen to know feel free to suggest an edit.

src/cpp/shared_core/system/PosixSystem.cpp
MariaSemple: That's correct, this just moves it from `core` to `shared_core`. Perhaps it's intentional that we silently fail if core dumps aren't supported? Maybe a warning or info message would be good here, rather than an error?

src/cpp/shared_core/system/PosixSystem.cpp
MariaSemple: In the IDE the `system::ignoreSignal(SignalName name)` function is still available from `System.hpp`. I just need this in the SDK to ignore SIG_PIPE and that's it. I actually had an `ignoreSigPipe` function, but I thought it would be better to not duplicate the code.

Also, I didn't want to move the `SignalName` enum because it's part of `System.hpp` as opposed to `PosixSystem.hpp`, so it may be used on other platforms. I'm open to suggestions on how to resolve that though.

src/cpp/shared_core/system/PosixSystem.cpp
MariaSemple: That's an interesting question. I think in the cases I've seen if this function returns an error the process exits. That being said, it would make the function more "safe", I think, which always seems like a good idea.

At any rate, this is the existing behaviour so I doubt it will cause issues if we don't change it.

src/cpp/server/ServerMain.cpp
kevinushey: That makes sense. Thanks!

src/cpp/shared_core/include/shared_core/FilePath.hpp
kevinushey: Wait, I think I may have misunderstood -- I was looking at the code in https://github.com/rstudio/rstudio/blob/91d3851f1248b40da3fafa0ed4d15bcacd6e50b2/src/cpp/core/include/core/system/FileMode.hpp#L131-L175 and thought that denoted the "supported" file modes, but that appears to be something separate. Sorry!

src/cpp/shared_core/system/PosixSystem.cpp
kevinushey: Looks like we exit on failure to enable core dumps in most places it's used so I think the current behavior is probably intentional. (A warning might be appropriate though.)

src/cpp/shared_core/system/PosixSystem.cpp
kevinushey: It looks like some platforms may have the `strsignal()` function for this purpose: https://linux.die.net/man/3/strsignal

src/cpp/shared_core/system/PosixSystem.cpp
kevinushey: That makes sense. If failures here are handled as process exits then it's not worth changing anything here.

src/cpp/shared_core/include/shared_core/FilePath.hpp
MariaSemple: Oh okay!

src/cpp/shared_core/system/PosixSystem.cpp
MariaSemple: Sure I can log a warning here.

src/cpp/shared_core/system/PosixSystem.cpp
MariaSemple: Sure, I'll double check the usages of this function.

src/cpp/shared_core/include/shared_core/FilePath.hpp
jmcphers: 80? Did someone say 80?! ;-)

MariaSemple:
Companion branch in Pro is ready.

src/cpp/server/auth/ServerAuthHandler.cpp
```diff
@@ -454,9 +453,17 @@ Error initialize()
...
+   if (error)
+   {
+      LOG_ERROR_MESSAGE("Could not get server user details");
```
jmcphers: Feels like this should be metadata attached to the error rather than logged separately?

jmcphers: (i.e.  we typically only `LOG_ERROR_MESSAGE` at the point when we abort an operation, not as part of bubbling an error through)

PR #6209, opened 2020-02-12
cleanup bootstrapper scripts for Linux / Qt

gtritchie:
- the script that drives Qt Installer no longer works because you must login to the installer with a Qt account
- the Dockerfiles have been fixed separately to work around this via a minimal Qt SDK install
- for the developer bootstrap scripts (which are not particularly up-to-date to begin with) just stripped out the now broken Qt installation step and added notes to the README (and the wiki) about needing to install Qt yourself if you want to build desktop on Linux
- also added note in README suggesting consulting the Dockerfiles when determining dependencies for a different system, as those are more up-to-date than the bootstrapper scripts
- still work to do on Windows, just breaking things up into chunks


dependencies/linux/README
```diff
@@ -17,29 +17,8 @@ systems. Specific version requirements for various components include:
...
-that already includes Qt 5.12.6 you may wish to prevent installation of the
-Qt 5.12.6 SDK. To do this pass the --exclude-qt-sdk flag to the install 
-dependencies script, for example:
-
-./install-dependencies-debian --exclude-qt-sdk
-
-More information on running the install-dependencies script is included
-in the section below.
-
+To build RStudio Desktop, install Qt 5.12.6 from https:/qt.io. Include the
```
kevinushey:
```suggestion
To build RStudio Desktop, install Qt 5.12.6 from https://qt.io. Include the
```

PR #6264, opened 2020-02-19
overhaul display of Tutorial pane tutorials

kevinushey:
This PR overhauls the display for RStudio Tutorials. In particular:

1. We no longer show the internal tutorial name, nor the name of the package providing the tutorial. I couldn't figure out a nice way to display this information together without looking too cluttered.

2. We now display a Start Tutorial button, used to launch the related tutorial.

3. We display help links at the bottom of the Tutorial list.

Screenshots:

Screen Shot 2020-02-19 at 11 26 00 AM

Screen Shot 2020-02-19 at 11 26 25 AM

Screen Shot 2020-02-19 at 11 29 25 AM


src/cpp/session/modules/SessionTutorial.cpp
```diff
@@ -267,10 +245,12 @@ void handleTutorialHomeRequest(const http::Request& request,
                << tutorial.description
```
jmcphers: The examples make me wonder if we should interpret Markdown here. We already build and include a self-contained Markdown parser (`sundown`, see R Presentations for use) so might be very straightforward to integrate. What do you think?

kevinushey: That is an excellent idea!

jmcphers:
I think it's important to show which package a tutorial came from, because it tells you how it wound up in the pane (and how to get rid of it). It also provides context for the tutorial itself; for example a "join" tutorial means something different in `dplyr` than it does in `odbc`. 

How about showing the name of the package in x-small, non-bold (light) type beneath the tutorial name? 

(we can iterate on this after your PR lands, just want to make sure it gets tracked!)

src/cpp/session/resources/tutorial_resources/index.html
```diff
@@ -7,9 +7,23 @@
    </head>
 
    <body class="rstudio-tutorials" style="visibility: hidden;">
+
       <div class="rstudio-tutorials-container">
```
gtritchie: Instead of a `div` use `<main class="rstudio-tutorials-container">`. This is more semantically correct and provides more context on the page layout to a screen reader user.

If `main` causes styling issues, could instead add `role="main"` to the `div` and achieve the same result.

src/cpp/session/resources/tutorial_resources/index.html
```diff
@@ -7,9 +7,23 @@
...
          #!tutorials#
       </div>
+
+      <div class="rstudio-tutorials-footer">
+
+         <img class="rstudio-tutorials-logo" alt="RStudio Icon" src="./rstudio-logo.png">
```
gtritchie: The icon doesn't add any useful information to a screen reader user, so just use an empty `alt` to mark it as purely a decorative element that a screen-reader can ignore.

``

src/cpp/session/resources/tutorial_resources/index.html
```diff
@@ -7,9 +7,23 @@
    </head>
 
    <body class="rstudio-tutorials" style="visibility: hidden;">
+
       <div class="rstudio-tutorials-container">
          #!tutorials#
       </div>
+
+      <div class="rstudio-tutorials-footer">
```
gtritchie: Instead of a `div` use `<footer>`. Same reasons as for using `main`.

Alternatively can add `role="contentinfo"` to the `div` to achieve the same result.

kevinushey:
Here's an example with a sub-header showing the package + tutorial name:

Screen Shot 2020-02-19 at 2 23 00 PM

src/cpp/session/modules/SessionTutorial.cpp
```diff
@@ -249,8 +225,11 @@ void handleTutorialHomeRequest(const http::Request& request,
             << htmlEscape(tutorial.title)
```
gtritchie: The heading of each tutorial should actually **be** an HTML heading, probably `<h2>`. Don't want to use `<h1>` because there should generally only be a single `<h1>` in a given page/iframe. In this case we have no `h1` but that's life in the big city.

Reason to use actual heading is that screen readers can then present a list of the tutorials for easier browsing and navigation. Otherwise it's just a big text document with no structure.

Instead of using actual `<h2>` can markup the existing `span` like this:

`<span role="heading" aria-level="2" class="rstudio-tutorials-label">`

src/cpp/session/modules/SessionTutorial.cpp
```diff
@@ -249,8 +225,11 @@ void handleTutorialHomeRequest(const http::Request& request,
...
             << "</span>";
          
-         ss << "<span class=\"rstudio-tutorials-name\">"
-            << htmlFormatTutorialName(pkgName, tutorial)
+         ss << "<span class=\"rstudio-tutorials-run-container\">"
+            << "<button class=\"rstudio-tutorials-run-button\" onclick=\"window.parent.tutorialRun('" << htmlEscape(tutorial.name) << "', '" << htmlEscape(pkgName) << "')\">"
+            << "<span class=\"rstudio-tutorials-run-button-label\">Start Tutorial</span>"
```
gtritchie: The purpose of the button is obvious to a sighted user, but to a screen-reader user the buttons look like a list of identical buttons with no context. Example below from VoiceOver.

To fix this, add an `aria-label="foo"` to each button where "foo" is something like "Start Data Basics Tutorial". This will be read by screen readers, but visual users will just see the existing label.

IMG_4076

PR #6273, opened 2020-02-20
provide "Install learnr" link in Tutorial pane

kevinushey:
This PR provides a small, clickable link / button for installing `learnr` from the main landing page when we detect it is not yet installed.

Screen Shot 2020-02-20 at 12 10 38 PM

This makes it easy for users to immediately get started with the Tutorial pane. After clicking the button, `learnr` will be installed, and then the Tutorial pane will automatically refresh and show the newly-installed tutorials.

@gtritchie, let me know if this is an acceptable implementation of a "link that behaves like a button". I took a different approach than what's done in HyperlinkLabel. The element itself is an anchor element, but has its ARIA role set as a button and has the requisite handlers for Space + Enter so that it responds like a normal button would. (Re-using the anchor element just means we get consistent styling with other anchors; e.g. the "active" state and so on)

### DIFF
--- a/src/cpp/session/modules/SessionTutorial.cpp
+++ b/src/cpp/session/modules/SessionTutorial.cpp
@@ -204,12 +204,9 @@ void handleTutorialHomeRequest(const http::Request& request,
          // Button for 'click here' action (styled like a link; behaves like button)
          std::stringstream clickHere;
          clickHere << "<a"
-                   << " aria-role=\"button\""
-                   << " aria-label=\"Install learnr\""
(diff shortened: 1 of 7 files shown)


src/cpp/session/modules/SessionTutorial.cpp
```diff
@@ -204,12 +204,10 @@ void handleTutorialHomeRequest(const http::Request& request,
          // Button for 'click here' action (styled like a link; behaves like button)
```
gtritchie: obsolete comment

PR #6316, opened 2020-02-25
put back option to set global application role (a11y)

gtritchie:
- I had removed this a while back, thinking it was unnecessary, but a screen-reader user who is helping test out 1.3 requested I put it back
- Changed default to false (off)
- When this role is assigned, there will be Axe violations due to some landmark roles being nested inside it
- Also assigned the role on satellite windows, which I had originally missed
- Overall purpose of this is to potentially change some default behavior of certain screen readers on Windows (whether they are in forms/focus mode or regular browsing mode)

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/common/satellite/SatelliteApplication.java
+++ b/src/gwt/src/org/rstudio/studio/client/common/satellite/SatelliteApplication.java
@@ -1,7 +1,7 @@
 /*
  * SatelliteApplication.java
  *
- * Copyright (C) 2009-19 by RStudio, PBC
+ * Copyright (C) 2009-20 by RStudio, PBC
  *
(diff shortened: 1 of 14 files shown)


src/gwt/src/org/rstudio/studio/client/common/satellite/SatelliteApplication.java
```diff
@@ -27,11 +27,13 @@
 import com.google.gwt.user.client.ui.Widget;
 import com.google.inject.Provider;
 import org.rstudio.core.client.CommandWithArg;
+import org.rstudio.core.client.Debug;
```
adamconroy: looks like vestigial debug import

gtritchie: Yup, thanks. Removed!

Issue #4232, opened 2019-01-28
Syntax highlighter for Makefile should be useable for secondary makefiles


brendanf:
### System details
    RStudio Edition : Desktop
    RStudio Version : 1.1.442 
    OS Version      :  Ubuntu 16.04
    R Version       :  3.5.0

### Steps to reproduce the problem
Open a makefile in the editor which is not named "Makefile".  For example, "include_this.make"

### Describe the problem in detail
The syntax highlighter does not recognize that the file is a Makefile.  "Makefile" is not one of the options which can be chosen in the syntax highlighting dropdown menu at the lower right corner of the editor pane.

### Describe the behavior you expected
At the least, I would like to be able to manually choose "Makefile" as the language in the syntax highlighting dropdown list.  It would also be nice if the ".make" extension were automatically detected as a makefile.

ronblum:
Thank you for raising this issue. We will review it as we work continue development of RStudio.

Reproduced on MacOS 10.14 Mojave using RStudio Desktop 1.2.1256-1.

PR #6333, opened 2020-02-26
register default GNU make names as makefiles & address #4232

pearsonca:
For makefiles which are gmake-only, [CRAN package guidance includes](https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#Writing-portable-packages): "If you only need GNU make for parts of the package which are rarely needed (for example to create bibliography files under vignettes), use a file called GNUmakefile rather than Makefile as GNU make (only) will use the former."

Rstudio should recognize the default GNU [make search names](https://www.gnu.org/software/make/manual/html_node/Makefile-Names.html) -- `GNUmakefile`, `makefile`, `Makefile` -- as Makefiles. This already happens for `Makefile`; this PR adds it for the other two filenames, as well `*.make` as requested in #4232.

I have attempted to test this by building Rstudio on my machine, but I'm not able to get all the cmake or docker build dependencies to behave correctly.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/common/filetypes/FileTypeRegistry.java
+++ b/src/gwt/src/org/rstudio/studio/client/common/filetypes/FileTypeRegistry.java
@@ -68,15 +68,15 @@ public class FileTypeRegistry
          new TextFileType("dcf", "DCF", EditorLanguage.LANG_DCF, ".dcf",
                           new ImageResource2x(ICONS.iconDCF2x()), false, false, false, false, false,
                           false, false, false, false, false, false, false, false);
-   
+
    public static final TextFileType INI =
          new TextFileType("ini", "INI", EditorLanguage.LANG_INI, ".ini",
                           new ImageResource2x(ICONS.iconDCF2x()), false, false, false, false, false,
                           false, false, false, false, false, false, false, false);
-   
-   
+
+
    public static final TextFileType STAN = new StanFileType();
-   
+
    public static final TextFileType MERMAID = new MermaidFileType();
    public static final TextFileType GRAPHVIZ = new GraphvizFileType();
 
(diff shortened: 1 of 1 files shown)


MariaSemple:
Hi @pearsonca, thanks for taking the time to make this contribution! The `rstudio:master` branch is currently for version 1.3 right now, which is closed for new features at this time. Would you mind changing the base of this PR to the `v1.4` branch? I think you'll need to merge your local master into your local `v1.4` branch and then recreate the PR from your local `v1.4` branch onto our `v1.4` branch.

For the build dependencies, you don't necessarily need to use docker. Have you seen [this wiki page about getting started building RStudio](https://github.com/rstudio/rstudio/wiki/RStudio-Development)? If that doesn't help, I'd be happy to help resolve any compilation issues you're having, or I can help by testing the change locally. 

Let me know if you have any questions!

pearsonca:
I'm happy to rebase - it's shut down time in GMT, but I'll have a go at it
tomorrow.

I had been working from other parts of the wiki - will have a look at that
page soon.

MariaSemple:
I built this locally and was able to confirm that the new file types are given the same treatment as `Makefile`s are. Folding demonstrated in this simple makefile called "GNUMakefile" below:

Screen Shot 2020-03-03 at 4 59 37 PM

@pearsonca Can you confirm whether that covers all of your intended changes?

pearsonca:
I suspect that if folding works, the following also work?

 - icon in the files browser (looks like it from source tab icons)
 - force tabs instead of spaces insert
 - syntax highlighting

MariaSemple:
Apologies for the delay. Indeed they do all work!

MariaSemple:
Thanks again for the contribution!

PR #6335, opened 2020-02-26
apply visually hidden style in code during boot screen

gtritchie:
The A11y.setVisuallyHidden call relies on a ThemeStyles style that isn't injected until later in the boot sequence. Apply the same styles directly in code, instead.

This is an experiment (but a valid change on its own) related to #6201.


kevinushey:
LGTM, although I do wonder if we could get away with injecting these styles more eagerly. That seems less safe to take for v1.3 though, since the orchestration of how things are loaded when RStudio is initialized is a bit hairy.

Issue #6325, opened 2020-02-25
show indicator while Tutorial is loading


kevinushey:
Currently, when a tutorial is run, the Jobs pane is focused and the associated progress is shown, but the Tutorial pane itself remains blank.

We should try and show more clearly that RStudio is loading a tutorial and instruct the user to wait during that period.

PR #6342, opened 2020-02-26
set content for tutorial loading page

kevinushey:
This PR adds a tutorial loading page, giving an indication in the pane that a tutorial is being loaded after clicking the Start button for a tutorial.

Screen Shot 2020-02-26 at 3 33 51 PM

Closes https://github.com/rstudio/rstudio/issues/6325.


src/cpp/session/resources/tutorial_resources/loading.html
```diff
@@ -0,0 +1,119 @@
...
+}
+
+</style>
+
+   </head>
+
+   <body class="rstudio-tutorials">
+      <div>
+
+         <div role="img" class="lds-spinner">
```
kevinushey: Is there a more appropriate ARIA role to suggest that the "stuff" in here isn't relevant to a screen reader? (Since these are just divs that form an animation)

src/cpp/session/resources/tutorial_resources/loading.html
```diff
@@ -0,0 +1,119 @@
...
+            <div></div><div></div><div></div><div></div>
+            <div></div><div></div><div></div><div></div>
+            <div></div><div></div><div></div><div></div>
+         </div>
+
+         <h2 style="display: inline-block; margin: 12px 4px;">Loading tutorial...</h2>
```
gtritchie: Minor (change if you want): Don't think this should be a heading, just a styled `div` or `p`. A heading gets special treatment by screen readers and implies there is more content nested under the header.

PR #6352, opened 2020-02-27
register default GNU make names as makefiles & address #4232

pearsonca:
Per request, changing PR target of #6333 - seems to not require rebase to target v1.4 branch instead, so just doing it that way.


MariaSemple:
Oh, great! Sometimes there can be differences between the branches that cause a lot of extra changes in the PR, but it looks like that didn't happen here. In that case, I'll edit the original PR to be based on 1.4 and we can close this one. :)

PR #6356, opened 2020-02-27
Feature/database

kfeinauer:
Add the ability to connect to databases. 

There is a `core` database set of functions, and a `server_core` set which is specific to server processes to connect to an `rstudio` server database. The `core` set can be used to create additional SQLite databases to store things like user settings, if desired, whereas the `server_core` is focused on the central database that will be used by all server processes.


src/cpp/CMakeLists.txt
```diff
@@ -355,6 +361,110 @@ else()
...
+include_directories(SYSTEM ${SOCI_INCLUDE_DIR})
+include_directories(SYSTEM ${SOCI_INCLUDE_BUILD_DIR})
+
+# database library includes
+if (UNIX)
+   if (APPLE)
+      # We assume that libpq is available under the default brew install location
+      # since we install postgresql via brew in install-dependencies-osx
+      include_directories(SYSTEM "/usr/local/Cellar/postgresql/12.1/include")
+      include_directories(SYSTEM "/usr/local/Cellar/postgresql/12.1/include/libpq")
```
kevinushey: This will break if the version of `postgresql` installed by Brew happens to change. You can access the current path using:

```
brew --prefix postgresql
```

Or, alternatively, Homebrew normally symlinks the current version of a formula to `/usr/local/opt`, so you could just check:

```
/usr/local/opt/postgresql
```

instead.

Aside: `libpq` and `postgresql` are separate formula; are you sure you want to use the `postgres` includes here or do you want to use `libpq` specifically?

src/cpp/CMakeLists.txt
```diff
@@ -355,6 +361,110 @@ else()
...
+       list(APPEND SOCI_LIBRARIES "${RSTUDIO_WINDOWS_DEPENDENCIES_DIR}/install-soci/postgresql/lib/libpq-release-${SOCI_ARCH}.lib")
+       list(APPEND SOCI_LIBRARIES "${RSTUDIO_WINDOWS_DEPENDENCIES_DIR}/install-soci/postgresql/lib/libpgcommon-release-${SOCI_ARCH}.lib")
+       list(APPEND SOCI_LIBRARIES "${RSTUDIO_WINDOWS_DEPENDENCIES_DIR}/install-soci/postgresql/lib/libpgport-release-${SOCI_ARCH}.lib")
+   endif()
+endif()
+list(LENGTH SOCI_LIBRARIES SOCI_LIB_COUNT)
+
+if (SOCI_LIB_COUNT EQUAL 0)
+   message(FATAL_ERROR "No SOCI libraries found under ${SOCI_LIBRARY_DIR}. Ensure the SOCI dependency is installed and try again.")
+else()
```
kevinushey: Nit: because the above is a fatal error you could just `endif()` and save a level of indentation.

src/cpp/CMakeLists.txt
```diff
@@ -355,6 +361,110 @@ else()
...
+         if(INDEX GREATER "-1")
+            set(POSTGRES_LIB "${LIBRARY}")
+         endif()
+      endforeach(LIBRARY)
+
+      get_filename_component(SQLITE_LIB "${SQLITE_LIB}" REALPATH)
+      get_filename_component(POSTGRES_LIB "${POSTGRES_LIB}" REALPATH)
+      message(STATUS "SQLite lib: ${SQLITE_LIB}")
+      message(STATUS "PostgreSQL lib: ${POSTGRES_LIB}")
+      install(PROGRAMS "${SQLITE_LIB}" DESTINATION ${RSTUDIO_INSTALL_BIN})
```
kevinushey: Do we need to add something to NOTICE for libsqlite?

src/cpp/CMakeLists.txt
```diff
@@ -355,6 +361,110 @@ else()
...
+      install(PROGRAMS "${POSTGRES_LIB}" DESTINATION ${RSTUDIO_INSTALL_BIN})
```
kevinushey: Similarly: NOTICE for libpq?

src/cpp/core/Database.cpp
```diff
@@ -0,0 +1,635 @@
...
+
+Error Connection::executeStr(const std::string& queryStr)
+{
+   try
+   {
+      // SOCI backends do not necessarily support running multiple statements
+      // in one invocation - to work around this, we split any passed in SQL
+      // into one invocation per SQL statement (delimited by ;)
+      std::vector<std::string> queries;
+      boost::split(queries, queryStr, boost::is_any_of(";"));
```
kevinushey: Will we need to worry about queries which include `;` as part of e.g. strings or something similar?

src/cpp/server_core/ServerDatabase.cpp
```diff
@@ -0,0 +1,204 @@
...
+   FilePath optionsFile("/etc/rstudio/database.conf");
+   if (optionsFile.exists())
+   {
+      // the database configuration file can potentially contain sensitive information
+      // log a warning if permissions are too lax
+      FileMode fileMode;
+      Error error = optionsFile.getFileMode(fileMode);
+      if (error)
+      {
+         LOG_ERROR_MESSAGE("Could not determine file permissions for database configuration file: " +
```
kevinushey: This seems useful enough to factor into a helper function somewhere (IIRC we need to ensure 0600 permissions on a variety of files like this)

src/cpp/server_core/ServerDatabase.cpp
```diff
@@ -0,0 +1,204 @@
...
+constexpr const int   kDefaultPostgresqlDatabaseConnectionTimeoutSeconds = 10;
+
+// environment variables
+constexpr const char* kServerTmpDirEnvVar = "RS_SERVER_TMP_DIR";
+constexpr const char* kDatabaseMigrationsPathEnvVar = "RS_DB_MIGRATIONS_PATH";
+
+//misc constants
+constexpr const size_t kDefaultConnectionPoolSize = 4;
+
+boost::shared_ptr<ConnectionPool> s_connectionPool;
```
kevinushey: Does `rserver` still exit gracefully with this static variable?

src/cpp/server/extras/conf/database.conf
```diff
@@ -0,0 +1,34 @@
+# This file contains database configuration.
+#
+# RStudio Server needs an external database for storage of specific configuration and runtime information.
+# At present, SQLite and PostgreSQL databases are supported.
```
kevinushey: I'd remove this and just ask users to reference the admin guide, as this could become stale as new databases are supported.

dependencies/common/install-soci
```diff
@@ -0,0 +1,64 @@
...
+
+   # checkout desired soci branch (version pinned)
+   git checkout "$SOCI_BRANCH"  
+
+   # make build directory
+   mkdir -p $SOCI_BIN_DIR
+   cd $SOCI_BIN_DIR
+
+   # create symlink to our boost datetime library so it is properly discovered by the SOCI build
+   ln -sf "$BOOST_DIR/lib/libboost_date_time.a" "$BOOST_DIR/lib/rstudio_boost_date_time.a"
```
kevinushey: Does this still work even with our copy of Boost? (a bit surprised it does since we do the extra symbol namespacing work for the bundles of Boost we build)

src/cpp/core/Database.cpp
```diff
@@ -0,0 +1,635 @@
...
+      {
+         return DatabaseError(error);
+      }
+   }
+
+   Error operator()(const PostgresqlConnectionOptions& options) const
+   {
+      try
+      {
+         boost::format fmt("host='%1%' port='%2%' dbname='%3%' user='%4%' password='%5%' connect_timeout='%6%'");
```
kevinushey: Do we need to sanitize any of these strings? (I guess in theory these are only ever read from a config file, which should be considered safe / trusted but worth confirming this is true.)

src/cpp/core/Database.cpp
```diff
@@ -0,0 +1,635 @@
...
+
+   std::string currentVersion;
+   error = databaseSchemaVersion(&currentVersion);
+   if (currentVersion < migrationVersion)
+      return updateToVersion(migrationVersion);
+   else
+      return Success();
+}
+
+Error SchemaUpdater::updateToVersion(const std::string& maxVersion)
```
kevinushey: I'm not as familiar on what the best practices are re: database updates + schema versioning. Do you have any recommended resources for getting better acquainted?

kevinushey:
One bigger question: how will database migration work in tandem with upgrades to RStudio Server? E.g. suppose I'm running RStudio v1.4, and I'm considering an upgrade to RStudio v2.0. Suppose some database schema gets updated in that version change. Is there a mechanism to recover the old database if I later wanted to, or needed to, downgrade back to v1.4?

Mainly, I'm thinking about the case where someone might want to (for example) install a preview build of RStudio Server, but later roll back if so required.

CMakeGlobals.txt
```diff
@@ -261,8 +261,15 @@ endfunction()
 # define custom installation macro to strip symbols from the binary
 macro(add_stripped_executable _target)
    add_executable(${_target} ${ARGN})
+
+   # set RPATH=$ORIGIN to ensure that any bundled libraries are found in our binary dir
+   set(CMAKE_BUILD_WITH_INSTALL_RPATH TRUE)
+   set_target_properties(${_target} PROPERTIES INSTALL_RPATH \$ORIGIN)
```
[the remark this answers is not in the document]
kfeinauer: I had trouble getting that working on all the various platforms, but was able to get everything working out of the bin dir, so figured for now to just go with the simple approach.

src/cpp/CMakeLists.txt
kfeinauer: Because of how the includes are set up in SOCI, both of these paths are necessary, so this was intentional.

src/cpp/CMakeLists.txt
kfeinauer: No, it's actually public domain!

src/cpp/CMakeLists.txt
kfeinauer: Good catch, will add!

src/cpp/server_core/ServerDatabase.cpp
kfeinauer: Yes, any thread that could hold a reference to this will be stopped before it exits.

dependencies/common/install-soci
kfeinauer: Yes, this is what that was tested against (whatever boost we install via install-dependencies). This was required to compile with boost support.

src/cpp/core/Database.cpp
kfeinauer: I don't think so. Like you said, this is coming from a trusted source and PostgreSQL is also picky about what strings it accepts in a connection string.

src/cpp/core/Database.cpp
kfeinauer: This was pretty much taken from Connect's implementation: https://github.com/rstudio/connect/blob/master/src/connect/store/version.go#L261-L345

This is similar to how many popular tools do things as well. See https://flywaydb.org/documentation/migrations

kfeinauer:
@kevinushey With our database migration code, there's no real way to roll back changes, assuming a successful migration occurred. Therefore, it would be up to the user to snapshot (and restore) their database if they ever wished to go backwards.

We could either document this in the admin guide, or we could craft our own crude database dump (where we dump the entire database and its schema+data to a SQL query file). However, I'd prefer to leave this in the hands of admins, as each database generally comes with a production-ready tool that does this.

kfeinauer:
@kevinushey And yes, I needed to pull in a newer `wget.exe` to get around antiquated SSL ciphers.

src/cpp/server/extras/conf/database.conf
kfeinauer: I'm not overly worried about adding support for new databases and this becoming stale. We match feature parity with Connect, which has only had support for SQLite and PostgreSQL for some time now. I think at best, we'd add support for ODBC (which essentially adds support for all databases).

Also, it is common to provide such easy-to-change configurations for pro installations, so figured we'd just add this here since we will likely want an install for pro.

kfeinauer: I think I misunderstood your comment - I think you were suggesting we just take out line 4? I can make that change!

src/cpp/core/Database.cpp
kfeinauer: Yes, this is a good point. My next commit will fix this by using `regex_split`

src/cpp/server_core/ServerDatabase.cpp
kfeinauer: I think this really depends on what you want to do with the file / what you want to log, whether or not you want to exit to make this really generic. The real meat of it is in `getFileMode` at any rate. Will leave this unresolved to solicit feedback from others. If there's more desire to make a utility function for this, I can attempt to do so!

src/cpp/server/ServerMain.cpp
```diff
@@ -622,6 +624,11 @@ int main(int argc, char * const argv[])
       if (error)
          return core::system::exitFailure(error, ERROR_LOCATION);
 
+      // initialize database connectivity
+      error = server_core::database::initialize(true);
```
MariaSemple: Do we ever initialize the server database without attempting to update the schema? If not, is the parameter there because we plan to in the future?

dependencies/windows/install-soci/install-soci.R
```diff
@@ -0,0 +1,97 @@
+# some laziness to ensure we move to the 'install-crashpad' folder
```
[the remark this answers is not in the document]
kfeinauer: It seems none of the other dependency install scripts have the copyright header for whatever reason, so this just follows that pattern.

src/cpp/core/Database.cpp
```diff
@@ -0,0 +1,638 @@
...
+
+Error Connection::executeStr(const std::string& queryStr)
+{
+   try
+   {
+      // SOCI backends do not necessarily support running multiple statements
+      // in one invocation - to work around this, we split any passed in SQL
+      // into one invocation per SQL statement (delimited by ;)
+      std::vector<std::string> queries;
+      boost::regex regex(";[ \\t\\r\\f\\v]*\\n");
```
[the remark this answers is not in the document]
kfeinauer: Correct, this explicitly will not split on that, it will split on only the very last `;`, because the regex requires newline after the `;`.

src/cpp/core/Database.cpp
```diff
@@ -0,0 +1,638 @@
...
+
+   // for postgresql, specifically lock the version table in exclusive mode
+   // to ensure that no other connection can use the version table AT ALL
+   // during this schema update
+   if (connection_->driverName() == POSTGRESQL_DRIVER)
+   {
+      Query query = connection_->query(std::string("LOCK ") + SCHEMA_TABLE + " IN ACCESS EXCLUSIVE MODE");
+      Error error = connection_->execute(query);
+      if (error)
+         return error;
```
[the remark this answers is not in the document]
kfeinauer: The transaction automatically rolls back in the destructor of the `Transaction` object.

src/cpp/core/DatabaseTests.cpp
```diff
@@ -0,0 +1,471 @@
...
+   test_that("Test Setup")
+   {
+      // ensure that the test databases do not exist
+      FilePath sqliteDbPath("/tmp/rstudio-test-db");
+      sqliteDbPath.removeIfExists();
+
+      boost::shared_ptr<IConnection> connection;
+      Error error = connect(postgresConnectionOptions(), &connection);
+      if (error)
+         return;
```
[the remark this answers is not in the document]
kfeinauer: Yes, and that is intended - if the test database cannot be connected to, we won't run the tests. This is also okay because these tests are never run automatically.

src/cpp/core/DatabaseTests.cpp
```diff
@@ -0,0 +1,471 @@
...
+      vals[1][1] = ";;;";
+      for (RowsetIterator it = rowset.begin(); it != rowset.end(); ++it)
+      {
+         Row& row = *it;
+         REQUIRE(row.get<std::string>(0) == vals[i][0]);
+         REQUIRE(row.get<std::string>(1) == vals[i][1]);
+         ++i;
+      }
+   }
+}
```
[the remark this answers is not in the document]
kfeinauer: That's actually covered in this particular test case.

src/cpp/core/include/core/Database.hpp
```diff
@@ -0,0 +1,282 @@
...
+{
+public:
+   SchemaUpdater(const boost::shared_ptr<IConnection>& connection,
+                 const FilePath& migrationsPath);
+
+   // updates the database schema to the highest version
+   Error update();
+
+   // updates the database schema to the specified version
+   Error updateToVersion(const std::string& maxVersion);
```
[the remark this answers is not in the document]
kfeinauer: Both - it will update to no higher than the specified version, and if that version exists, that's what it will be updated to.

But that break condition was something I forgot to add to the code, so will update with that in there.

src/cpp/server/extras/conf/database.conf
```diff
@@ -0,0 +1,32 @@
...
+# sqlite configuration
+# =========================================================================================================
+# Specifies the database provider to use
+#provider=sqlite
+# Directory in which the sqlite database will be written
+#directory=/var/run/rstudio-server
+# =========================================================================================================
+# postgresql configuration
+# =========================================================================================================
+# Note: when connecting to a PostgreSQL database, a default empty rstudio database must first be created!
```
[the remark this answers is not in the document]
kfeinauer: This is generally not done for a variety of reasons (from what I understand), and for example is not something done by Connect (which a lot of this is modeled off of). So I will not attempt to automatically create the database.

src/cpp/core/DatabaseTests.cpp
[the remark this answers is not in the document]
MariaSemple: Yeah, that makes sense. I just wasn't sure if a return exits the whole test case or just the test_that section.

src/cpp/core/Database.cpp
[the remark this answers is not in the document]
MariaSemple: Ah, I think that makes sense since we'll never get a full query (or set of multiple queries) from the user directly - just values for filtering or inserting, right? Can it (or does it need to) support `;`  followed by `\\n` within the literal filter data?

src/cpp/core/include/core/Database.hpp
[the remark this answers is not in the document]
MariaSemple: Could you update the comment to reflect that as well?

src/cpp/server/auth/ServerAuthHandler.cpp
```diff
@@ -93,39 +94,110 @@ void updateCredentialsNotSupported(
...
    }
    END_LOCK_MUTEX
 
-   return false;
+   transaction.commit();
+   return Success();
 }
 
-Error readRevocationList(std::vector<std::string>* pEntries)
+Error readRevocationListFromFile(const FilePath& revocationList,
```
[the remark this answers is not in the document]
kfeinauer: This only happens when migrating an existing file collection to the database for the first time. Then after, the file is never used again.

src/cpp/core/Database.cpp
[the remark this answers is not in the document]
kfeinauer: I really don't think we need to support that. This is only for schema update queries - it's not ever going to be 100% perfect (without vastly overcomplicating this code). I think as written is good enough - this isn't user hardened, it only needs to work for our particular crafted queries in our migration scripts, which I think should pretty much have us covered 100%.

src/cpp/server/auth/ServerAuthHandler.cpp
```diff
@@ -93,39 +94,110 @@ void updateCredentialsNotSupported(
...
-         {
-            it = s_revokedCookies.erase(it);
-            continue;
-         }
-         else
-         {
-            ++it;
-         }
+         const RevokedCookie& cookie = *it;
+         Error error = writeRevokedCookieToDatabase(cookie, connection);
```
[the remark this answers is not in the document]
kfeinauer: We're really not handling a lot of volume here (we're not bulk importing a lot of data). I think for ease of code understanding and simplicity, this is good enough.

src/cpp/server/db/20200226141952248_AddRevokedCookie.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE RevokedCookie(
+   Expiration text NOT NULL,
+   CookieData text NOT NULL,
+   PRIMARY KEY (Expiration, CookieData)
```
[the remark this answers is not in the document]
kfeinauer: I want the expiration on its own column so we can use the database to filter for the bulk delete query. Having the composite key here is going to be negligible performance loss (and this is not super performance critical code at any rate), so I'd rather keep both columns.

src/cpp/core/Database.cpp
```diff
@@ -0,0 +1,638 @@
...
+std::string PooledConnection::driverName() const
+{
+   return connection_->driverName();
+}
+
+boost::shared_ptr<IConnection> ConnectionPool::getConnection()
+{
+   // block until a connection is available
+   boost::shared_ptr<Connection> connection;
+   connections_.deque(&connection, boost::posix_time::pos_infin);
```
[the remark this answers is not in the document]
kfeinauer: This will hang until there is a connection available in the pool, which is what I intended. And like you said, they are handed out via shared ptr to help ensure that they are not leaked.

src/cpp/server/ServerMain.cpp
kfeinauer: We don't yet, but we will. I want derivative processes of rserver to not run migrations, even though they may need to connect to the database (like `rserver-monitor`).

src/cpp/server/db/20200226141952248_AddRevokedCookie.sql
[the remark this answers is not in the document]
MariaSemple: Oh, I didn't mean to remove the expiration column, just to remove it from the primary key if it's not required for unique entry identification. You're right about the performance though - that wouldn't be the reason to change it.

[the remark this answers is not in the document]
kfeinauer: That's a good point - since the Expiration is derivative data (it's redundantly stored on `CookieData`), then there's no reason for it to be part of the primary key. It just made sense in my initial thoughts. I'll change this.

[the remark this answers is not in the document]
kfeinauer: Actually, I do want the `Expiration` partially indexed for faster sorting (since we sort by expiration). I'll need to research more on how compound primary keys affect this.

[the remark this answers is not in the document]
MariaSemple: Are you sure it's a partial index that you want, or a full index? At any rate you should be able to create an index on a non-primary-key column using `CREATE INDEX RCExpiryIndex ON RevokedCookies (Expiration);` or something like it. It makes insertion into the table slower though, so if you're going to do that you may want to revise the code that writes all the revoked cookies to the table (assuming that happens with any frequency).

src/cpp/server/extras/conf/database.conf
```diff
@@ -0,0 +1,32 @@
...
+# Specifies the database provider to use
+#provider=postgresql
+# Specifies the host (hostname or IP address) of the database host
+#host=localhost
+# Specifies the TCP port where the database is listening for connections
+#port=5432
+# Specifies the database connection username
+#username=postgres
+# Specifies the database connection password
+#password=postgres
```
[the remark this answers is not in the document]
kfeinauer: I think this isn't really necessary for collection at this stage. Generally, if there is actually a database problem, we'll need to look into data within their database tables, which this file is not going to help, so a further dialog would be required there anyway which makes the hassle of collecting this file and sanitizing it not worth including in this PR.

src/cpp/server/db/20200226141952248_AddRevokedCookie.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE RevokedCookie(
```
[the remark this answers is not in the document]
kfeinauer: Done

src/cpp/server/db/20200226141952248_AddRevokedCookie.sql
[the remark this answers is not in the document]
kfeinauer: The overhead that this index is going to add, especially for such a small table that doesn't see a very high volume, is minimal and not worth reworking any of the insert code IMO.

kfeinauer:
@MariaSemple All feedback addressed.

src/cpp/core/Database.cpp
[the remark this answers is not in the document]
MariaSemple: Right, I know that hanging until one is available is intended. The concern would be around hanging forever if for some reason the connections are all never returned. Do you think it would be worth setting some very long timeout and then logging an error and exiting? That way it will be easier to find such an issue, in the unlikely event that one every occurs?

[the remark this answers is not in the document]
kfeinauer: I think exiting is a little heavy handed, but I think logging that there's a potential hang isn't a bad idea. I'll add this.

kfeinauer:
@MariaSemple Updated

dependencies/osx/install-dependencies-osx
```diff
@@ -22,6 +22,9 @@ then
   ./install-overlay
 fi
 
+# install postgres, needed by SOCI
+brew install postgres
```
jmcphers: We don't formally have a dependency on homebrew for dev machines, although in practice I think we all use it. I'm fine with introducing one, but make sure you also update the wikis if you do, and maybe add check here to see if homebrew is installed (so attempting to set up a machine w/o homebrew will give a more useful error).

src/cpp/core/Database.cpp
```diff
@@ -0,0 +1,665 @@
...
+      // no migration files - we do not consider this an error, but instead
+      // simply consider that this database cannot be migrated past version 0
+      *pVersion = "0";
+      return Success();
+   }
+
+   // sort descending - highest version filename wins
+   auto comparator = [](const FilePath& a, const FilePath& b)
+   {
+      return a.getStem() > b.getStem();
```
jmcphers: Isn't this going to use character ordering ("2" > "11")?

src/cpp/core/Database.cpp
```diff
@@ -0,0 +1,665 @@
...
+
+   std::vector<FilePath> files;
+   error = migrationFiles(&files);
+   if (error)
+      return error;
+
+   // sort ascending
+   auto comparator = [](const FilePath& a, const FilePath& b)
+   {
+      return a.getStem() < b.getStem();
```
jmcphers: As above re: string comparison of versions.

src/cpp/server/db/20200226141952248_AddRevokedCookie.sql
```diff
@@ -0,0 +1,19 @@
+/* Stores revoked auth cookies - cookies that are not yet expired, but
```
jmcphers: Can we add a README to this parent folder that explains how to name files and perform schema upgrades for devs?

src/cpp/core/Database.cpp
jmcphers: Upon further review I see that the filenames are formatted by convention so that character ordering is sufficient. I think we might make this clearer by using an "initial version" that looks like the incremental additional versions, and/or adding some docs stating explicitly how to format the filenames so this constraint is maintained.

src/cpp/server/db/README.md
```diff
@@ -0,0 +1,23 @@
+## Server Database Schemas
+
+This folder contains the database schemas for the RStudio Server database. These schemas are copied to the build output directory, and are applied to the database every time `rserver` is started.
+
+Schema files rely on a file format convention containing the following:
+* The date time at which the script was generated in the format YYYYMMDDHHmmssuuu (where uuu is microseconds)
+* An underscore separating the date time from the friendly name
+* A friendly name for what the script does
+* A file extension, which can be `.sql`, `.sqlite`, or `.postgresql`. `.sql` files will be run for any database type, but the other two extensions are reserved for only running if the database in use corresponds to that file extension.
```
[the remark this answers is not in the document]
kfeinauer: Updated README with answer

src/cpp/server/db/README.md
```diff
@@ -0,0 +1,23 @@
+## Server Database Schemas
+
+This folder contains the database schemas for the RStudio Server database. These schemas are copied to the build output directory, and are applied to the database every time `rserver` is started.
+
+Schema files rely on a file format convention containing the following:
+* The date time at which the script was generated in the format YYYYMMDDHHmmssuuu (where uuu is microseconds)
```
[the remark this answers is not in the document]
kfeinauer: No, but it absolutely ensures no duplicates. Also, I picked a method that was easy to script, so that plays a big part into what I chose to do (see `./make-schema.sh`

PR #6401, opened 2020-03-04
share web profile across different pages

kevinushey:
Candidate fix for https://github.com/rstudio/rstudio/issues/6201.

Note: merely sharing the web profile across different pages (RStudio windows) was surprisingly not sufficient; I was still able to reproduce cache load failures when opening multiple satellite windows at the same time. It looks like forcing HTTP resources to be cached in-memory helps avoid this issue, at least as far as I can see in testing.

### DIFF
--- a/src/cpp/desktop/DesktopWebProfile.cpp
+++ b/src/cpp/desktop/DesktopWebProfile.cpp
@@ -69,6 +69,14 @@ private:
 WebProfile::WebProfile(const QUrl& baseUrl, QObject* parent)
    : QWebEngineProfile(QString::fromUtf8("rstudio-desktop"), parent)
 {
+   // force an in-memory cache of web resources
+   // otherwise, multiple windows / pages attempting to access
+   // or mutate the disk cache at the same time can cause
+   // data corruption to occur
+   //
+   // https://github.com/rstudio/rstudio/issues/6201
+   setHttpCacheType(QWebEngineProfile::MemoryHttpCache);
+   
    sharedSecret_ = core::system::getenv("RS_SHARED_SECRET");
    setBaseUrl(baseUrl);
 }


src/cpp/desktop/DesktopWebProfile.cpp
```diff
@@ -66,9 +66,23 @@ class Interceptor : public QWebEngineUrlRequestInterceptor
 
 } // end anonymous namespace
 
-WebProfile::WebProfile(const QUrl& baseUrl, QObject* parent)
-   : QWebEngineProfile(QString::fromUtf8("rstudio-desktop"), parent)
+WebProfile* WebProfile::getInstance(const QUrl &baseUrl)
 {
+   static WebProfile* instance = new WebProfile(baseUrl);
+   return instance;
```
kfeinauer: Don't we need to ensure that a different WebProfile object is constructed for different base urls to ensure the interceptor code (for setting the shared secret header) works properly? This seems like that would break that.

src/cpp/desktop/DesktopWebProfile.cpp
```diff
@@ -66,9 +66,23 @@ class Interceptor : public QWebEngineUrlRequestInterceptor
...
+WebProfile::WebProfile(const QUrl& baseUrl)
+   : QWebEngineProfile(QStringLiteral("rstudio"))
+{
+   // force an in-memory cache of web resources
+   // otherwise, multiple windows / pages attempting to access
+   // or mutate the disk cache at the same time can cause
+   // data corruption to occur
+   //
+   // https://github.com/rstudio/rstudio/issues/6201
+   setHttpCacheType(QWebEngineProfile::MemoryHttpCache);
```
kfeinauer: Given this code here, can we live without the static one-per-process WebProfile?

kevinushey: Looks like you're correct. I've updated the PR (so now this is the only new code)

PR #6436, opened 2020-03-10
fix regression in update packages dialog

### LINKED ISSUE #6375: Bug while scrolling in 'Update Packages' panel in preview 1.3.842
### System details

    RStudio Edition :  Desktop
    RStudio Version :  preview 1.3.842 ("Middlemist Red" (74333857, 2020-02-17) for Windows) 
    OS Version      :  Windows 10 x64 (build 18362)
    R Version       :  3.6.0 (2019-04-26)

Problem occurs in a clean session when I view the 'Update Packages' panel.

I have a lot of packages installed and from time to time I check the 'Update Packages' panel to see which I would like to update. The first package displayed in the list is currently ' anytime'. I use the scrollbar to see other packages that can be updated. Then I see e.g. the contents in the screenshot below.

My small but annoying issue is the following: when I want to update both 'devtools' and 'digest' the panel is refreshed immediately after I flag 'devtools' (and then showing 'anytime' again as first updateable package) before I have an opportunity to also flag 'digest'. So I have to scroll downwards to 'devtools' and 'digest' again (where I see that 'devtools' is indeed flagged) and now flag 'digest'.

My preferred way of working would of course be to flag all the relevant packages in the current view and then scroll down to the next set of packages and repeat (without seeing the 'anytime' package anytime).

image

This was asked as question in [RStudio Community](https://community.rstudio.com/t/how-to-avoid-scroll-in-update-packages-panel-in-preview-1-3-842/54421) but marked by [kevinushey](https://community.rstudio.com/u/kevinushey) as a possible bug.

- [ ] I have installed the latest version of RStudio and confirmed that the issue still persists.
**NB this is a preview version!**

gtritchie:
Fixes #6375

Regression introduced by accessibility work (early summer 2019); I enabled keyboard-selection in this dialog's cell table, but that alone wasn't sufficient to make this dialog fully keyboard usable, and caused a nasty regression when using via mouse, so reverting. Didn't see any other instances where I used this pattern. Regression was introduced in 8bf99466.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/packages/ui/PackageActionConfirmationDialog.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/packages/ui/PackageActionConfirmationDialog.java
@@ -1,7 +1,7 @@
 /*
  * PackageActionConfirmationDialog.java
  *
- * Copyright (C) 2009-19 by RStudio, PBC
+ * Copyright (C) 2009-20 by RStudio, PBC
  *
  * Unless you have received this program directly from RStudio pursuant
  * to the terms of a commercial license agreement with RStudio, then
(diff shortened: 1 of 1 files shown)


kevinushey:
LGTM. Is it worth tracking whether this could be resolved in v1.4?

gtritchie:
> Is it worth tracking whether this could be resolved in v1.4? (that is, re-enabling keyboard selection)

This is tracked on a much larger scale; all of our uses of GWT grid-like controls are currently not keyboard accessible for the most part.

Issue #6395, opened 2020-03-04
Ctrl+Shift+V in Windows Desktop app pastes output twice, Windows 10, RStudio 1.3 


remlapmot:
I am not sure if `Ctrl+Shift+V` is a supported shortcut. It doesn't appear to be mentioned in the official list of keyboard shortcuts but there was some discussion in #1687 and Windows users would appreciate/expect it in the Terminal pane.

On Windows 10 I find that it produces double output in both the Source and Terminal panes.

### System details

    RStudio Edition : Desktop
    RStudio Version : 1.3.878
    OS Version      : Windows 10 (Version 1909 Build 18363.657)
    R Version       : 3.6.3

### Steps to reproduce the problem

- copy some text
- paste with `Ctrl+Shift+V` in either the Terminal pane or the Source pane.

### Describe the problem in detail

- The output is pasted twice, e.g. I should just be pasting `echo "Hello, World!"`
![windows-terminal-csv](https://user-images.githubusercontent.com/3777473/75877370-75acca00-5e0f-11ea-907e-8107045789da.png)

### Describe the behavior you expected

- The output should only be pasted once.

- [ x] I have read the guide to submitting good bug reports at https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports . 
- [x ] I have installed the latest version of RStudio and confirmed that the issue still persists.
- [x ] If I am reporting a RStudio crash, I have included a diagnostics report. https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report
- [x ] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
@remlapmot Thank you for filing the issue! I can reproduce this RStudio Desktop 1.3.878 with Command Prompt, PowerShell, and Git Bash. It works properly in 1.2.5033, so this is a regression. I'll mark this ticket for review as we continue development of RStudio. Thanks!

gtritchie:
Not just terminal, happens anywhere you can paste text, the console, the editor, even simple textboxes.

Desktop only, can't repro in RStudio Server with Chrome or Firefox.

Also, Ctrl+Shift+V is not a supported shortcut.

gtritchie:
Don't know that this meets the bar for 1.3, but will investigate and see if there's a safe fix.

gtritchie:
Relatively recent regression, started with build 1.3.812 on Feb 5, 2020.

[...]

PR #6451, opened 2020-03-11
remove 'paste with indent' desktop-level binding

kevinushey:
Closes https://github.com/rstudio/rstudio/issues/6395.

I had tried to implement support for Paste vs. Paste with Indent in a previous PR, but backed that out after running into trouble in having it work properly in different contexts. Unfortunately, this part was left in and caused the previously-referenced issue.

### DIFF
--- a/src/cpp/desktop/DesktopWebView.cpp
+++ b/src/cpp/desktop/DesktopWebView.cpp
@@ -169,13 +169,6 @@ void WebView::keyPressEvent(QKeyEvent* pEvent)
    }
 #endif
  
-   // allow Ctrl+Shift+V to act as a 'paste with indent' action
-   if (pEvent->key() == Qt::Key_V && pEvent->modifiers() == Qt::CTRL + Qt::SHIFT)
-   {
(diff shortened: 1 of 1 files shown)


kevinushey:
Note: we do still implement an "indent on paste" behavior when Shift is held down here:

https://github.com/rstudio/rstudio/blob/6a45dc8e1fd19a093a68bd5b0a533c34b10293b0/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/AceEditor.java#L648-L664

But this only comes into play if the user has explicitly disabled 'paste on indent' and is holding down Shift -- it is not an official command. I'm not sure if there's an appropriate place to document this, though. Maybe as a tooltip on the actual checkbox in the Global Options dialog?

gtritchie:
Yah, we've got a handful of unofficial commands (implemented purely in code). Maybe we should include a commented-out pseudo-entry in commands.cmd.xml for these, and also put them in the static keyboard help (keyboard.htm)? There's an example of this with Ctrl+L. But obviously almost guaranteed to be out of date.

PR #6468, opened 2020-03-13
port Ace patch for wrapped line rendering

kevinushey:
Closes https://github.com/rstudio/rstudio/issues/6457.


kevinushey:
Cherry-picking the Ace commit fixing this issue seemed safer than updating to a new version of Ace so close to release.

PR #6491, opened 2020-03-18
Add equality and inequality operators to User

MariaSemple:
Also need this for the Launcher Plugin SDK.

Two users are the same if they have the same user ID. Also implements move constructor and operator for User.

### DIFF
--- a/src/cpp/shared_core/include/shared_core/system/User.hpp
+++ b/src/cpp/shared_core/include/shared_core/system/User.hpp
@@ -132,6 +132,24 @@ public:
    static Error getUserFromIdentifier(UidType in_userId, User& out_user);
 
    /**
+    * @brief Equality operator.
+    *
+    * @param in_other      The user to compare with this user.
+    *
+    * @return True if this user and in_other have the same user ID; false otherwise.
+    */
+   bool operator==(const User& in_other) const;
+
+   /**
+    * @brief Inequality operator.
+    *
+    * @param in_other      The user to compare with this user.
+    *
+    * @return False if this user and in_other have the same user ID; true otherwise.
+    */
+   bool operator!=(const User& in_other) const;
+
+   /**
     * @brief Checks whether the user represented by this object exists.
     *
     * If this is an empty user, or is a user object which represents all users, this method will return false as it does
--- a/src/cpp/shared_core/include/shared_core/system/User.hpp
+++ b/src/cpp/shared_core/include/shared_core/system/User.hpp
@@ -106,7 +106,7 @@ public:
     *
     * @param in_other   The user to move into this User.
     */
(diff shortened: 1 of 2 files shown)


src/cpp/shared_core/system/User.cpp
```diff
@@ -161,6 +166,48 @@ FilePath User::getUserHomePath(const std::string& in_envOverride)
    return FilePath(getEnvVariable("HOME"));
 }
 
+User& User::operator=(const User& in_other)
+{
+   m_impl->Name = in_other.m_impl->Name;
```
kevinushey: Is it intentional that we explicitly copy the members piece-by-piece rather than just relying on `*m_impl = *in_other->m_impl` or something similar? (Main worry is that it could be easy to forget to update this method if we were to add new members to the Impl struct)

src/cpp/shared_core/include/shared_core/system/User.hpp
```diff
@@ -101,6 +101,13 @@ class User
     */
    User(const User& in_other);
 
+   /**
+    * @brief Move constructor.
+    *
+    * @param in_other   The user to move into this User.
+    */
+   User(User&& in_other) noexcept;
```
kevinushey: I wonder if we could just use the default compiler-implemented move constructor here?

src/cpp/shared_core/include/shared_core/system/User.hpp
```diff
@@ -131,6 +138,42 @@ class User
...
+   User& operator=(const User& in_other);
+
+   /**
+    * @brief Overloaded move operator.
+    *
+    * @param in_other   The user to move to this one.
+    *
+    * @return This user.
+    */
+   User& operator=(User&& in_other) noexcept;
```
kevinushey: Similar question re: using compiler default.

MariaSemple: The compiler won't generate a default because `User::User(const User&)` and `User& User::operator=(const User&)` are implemented.

src/cpp/shared_core/system/User.cpp
MariaSemple: I _think_ that should be safe. I was probably just being overly untrusting of the compiler generated code at the time. I'll make that change.

src/cpp/shared_core/include/shared_core/system/User.hpp
kevinushey: Sorry, I meant using `= default` to force the compiler to generate the default version.

MariaSemple: Oh, that's a good point! Thanks!

PR #6523, opened 2020-03-25
Add parameter to EditingTarget.onActivated indicating whether it's being activated for user interaction

jjallaire:
This PR adds an argument to the  `EditingTarget.onActivated()` method that indicates whether the editor tab is being activated for the purpose of user interaction (during startup `onActivated()` is called for each tab being restored irrespective of whether it will actually end up active at the end of startup). 

The motivation is that panmirror does some potentially expensive (depending on the platform) work at startup (calls pandoc twice). On Linux/OSX these calls take sub 10ms however on Windows they can take 100ms+. If a user has dozens of tabs w/ panmirror to be restored, the cost of doing panmirror initialization eagerly could become prohibitive. With this change we can defer this work until the user is actually going to interact with the document.

I could have "fixed" `onActivated()` to not be called during IDE tabset restoration, however I'm sure we have implicit dependencies on this elsewhere and unwinding them seems risky / cost prohibitive.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/source/Source.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/Source.java
@@ -654,7 +654,7 @@ public class Source implements InsertSourceHandler,
 
             if (view_.getTabCount() > 0 && view_.getActiveTabIndex() >= 0)
             {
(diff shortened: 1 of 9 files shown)


jjallaire:
Yeah, I thought of focus but in this case the editor component is created on demand and then focused, so I need to know about the activation a bit earlier in the control flow.

PR #6532, opened 2020-03-26
Update URL constructor to handle paths beginning with file:// or ftp://

### LINKED ISSUE #5724: Cannot add secondary file repository
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.2.5001
    OS Version      :  Windows 10 Enterprise x64
    R Version       :  3.6.1

Adding a secondary package repository that is hosted on a local Windows fileshare fails.

In the Packages pane under the RStudio options I am trying to add a secondary package repository that is hosted on a local Windows fileshare. The url has the form `file://server/local-cran`. This fails, however, with the error "The given URL does not appear to be a valid CRAN repository". The file `src\contrib\PACKAGES.gz` does however exist on the share.

The secondary repository to be added successfully.

melissa-barca:
Fixes #5724

Update the URL constructor to parse out `file://` and `ftp://`. This will allow users to access and download packages from a Windows fileshare system.

### DIFF
--- a/src/cpp/core/http/URL.cpp
+++ b/src/cpp/core/http/URL.cpp
@@ -28,7 +28,7 @@ namespace http {
 URL::URL(const std::string& absoluteURL)
 {
    std::string protocol, host, path;
-   boost::regex re("(http|https|file|ftp)://([^/#?]+)(.*)", boost::regex::icase);
+   boost::regex re("(http|https|file|ftp|ftps)://([^/#?]+)(.*)", boost::regex::icase);
    boost::cmatch matches ;
    if (regex_utils::match(absoluteURL.c_str(), matches, re))
    {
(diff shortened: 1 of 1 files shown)


src/cpp/core/http/URL.cpp
```diff
@@ -28,7 +28,7 @@ namespace http {
 URL::URL(const std::string& absoluteURL)
 {
    std::string protocol, host, path;
-   boost::regex re("(http|https)://([^/#?]+)(.*)", boost::regex::icase);
+   boost::regex re("(http|https|file|ftp)://([^/#?]+)(.*)", boost::regex::icase);
```
kevinushey: Should we also support `ftps` here?

melissa-barca: Good catch, I've included it 👍

PR #6557, opened 2020-03-30
force clang-darwin toolset when building boost on macOS

kevinushey:
cc: @gtritchie


kevinushey:
I believe it should work in all cases, but haven't yet tested.

kevinushey:
I'm going to merge -- if we see build-system breakage I'll refine further.

Issue #5904, opened 2019-12-19
Python parsing issue


2er0:
### System details

    RStudio Edition : Desktop
    RStudio Version :  1.2.1335
    OS Version : Fedora 31  
    R Version : 3.6.1 

### Steps to reproduce the problem

RUN:
```
def foo():
    width = "foo"
    print(width)
    
    height = "bar"
    print(height)

    print(f'{width} -> {height}')
```
OUTPUT:
```
> reticulate::repl_python()
Python 3.7.5 (venv/bin/python)
Reticulate 1.13 REPL -- A Python interpreter in R.
>>> def foo():
...     width = "foo"
...     print(width)
...     
...     height = "bar"
...     print(height)
... 
>>>     print(f'{width} -> {height}')
NameError: name 'width' is not defined
```

### Describe the problem in detail

```
def foo():
    width = "foo"
    print(width)
    !!! <- this line has white whitespace intend
    height = "bar"
    print(height)
!!! <- this line has no whitespace intend and breaks the parsing of the function
    print(f'{width} -> {height}')
```
Our Git automatically removes whitespace which leads to this issue.

### Describe the behavior you expected

If a section contains an empty line and then a code line with the same intend then this belongs to the same section as before.

2er0:
Is there any update?
In which version of RStudio or Package will this be fixed?

kevinushey:
We're planning to spend some more time on Python IDE integration in v1.4 and so hopefully will be able to take a look at this then. However, IDE input is fundamentally newline-delimited so adding support for this may be challenging.

2er0:
Ok
Thanks for the Information - would be nice to have this feature ;)

dfalty:
Hey @kevinushey I'm still seeing the issue with the same repex on 1.4.484.

R-Hannibal:
@2er0, thanks so much for finding and reporting this issue. 

We have a fix in place in our latest daily build of the upcoming 1.4 release (available [HERE](https://dailies.rstudio.com/)). If you get the chance will you let us know if the issue is resolved for you in that build?

melissa-barca:
Verified in 1.14.1017

BradWebb101:
I have observed this issue again with reticulate v1.20 

> reticulate::repl_python()
Python 3.6.13 
Reticulate 1.20 REPL -- A Python interpreter in R.

R studio output:
>>> time_start = time()
>>> for row in df.iterrows():
...   print(row['numbers])
[...]

PR #6588, opened 2020-04-03
improve handling of multiline Python blocks in console

kevinushey:
This PR fixes an issue where attempts to submit blocks of Python code with inconsistent indentation (e.g. blank lines between statements) to the `reticulate` REPL would fail. To fix this, we attempt to fix the indentation of blank lines post-hoc when code is submitted to the console.

This, admittedly, feels like a kludge. However, we have one major constraint: `reticulate` relies on R's `readline()`, which reads a single newline-buffered line at a time. Ultimately, that implies the fix needs to occur before the code is submitted to `reticulate`. And since an empty line submitted to `reticulate` is normally treated as the end of a block, we need to explicitly side-step that behavior by adjusting the code indent.

An alternative approach would be to use some sort of special encoding for blocks of code submitted to the `reticulate` REPL, and teach `reticulate` how to unravel those blocks, but this solution has the nice property of being compatible with the existing version of `reticulate` and ultimately feeling transparent.

Closes https://github.com/rstudio/rstudio/issues/5904.

### DIFF
--- a/src/cpp/session/SessionConsoleInput.cpp
+++ b/src/cpp/session/SessionConsoleInput.cpp
@@ -196,18 +196,6 @@ void fixupPendingConsoleInput()
    // keep track of last line's indent
    std::string indent;
    
-   // pending empty strings whose indent we need to set
-   std::vector<std::string*> pendingStrings;
-   
-   // helper for getting indent of line
-   auto getIndent = [](const std::string& line)
-   {
-      auto index = line.find_first_not_of(" \t");
-      if (index == std::string::npos)
-         return std::string();
-      return line.substr(0, index);
-   };
-   
    // split input into list of commands
    std::vector<std::string> lines = core::algorithm::split(input.text, "\n");
    for (std::size_t i = 0, n = lines.size(); i < n; i++)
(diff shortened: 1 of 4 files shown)


src/cpp/session/SessionConsoleInput.cpp
```diff
@@ -156,11 +157,127 @@ void reissueLastConsolePrompt()
...
+{
+   // get next input
+   auto input = s_consoleInputBuffer.front();
+   
+   // nothing to do if this is a cancel
+   if (input.cancel)
+      return;
+   
+   // if this has no newlines, then nothing to do
+   auto index = input.text.find('\n');
```
gtritchie, line 181: Minor, could simplify to `if (input.text.find('\n') == std::string::npos)` and eliminate `index` variable.

src/cpp/session/SessionConsoleInput.cpp
```diff
@@ -156,11 +157,127 @@ void reissueLastConsolePrompt()
...
+   
+   // if we're about to send code to the Python REPL, then
+   // we need to fix whitespace in the code before sending
+   bool pyReplActive = modules::reticulate::isReplActive();
+   
+   // pop off current input (we're going to split and re-push now)
+   s_consoleInputBuffer.pop();
+   
+   // does this Python line start an indented block?
+   // NOTE: should consider using tokenizer here
```
gtritchie, line 193: If you're considering it, maybe now is the time? :-)

PR #6590, opened 2020-04-03
fix C++ warnings related to raw literals and unused parameters

gtritchie:
The trick for ignoring the leading newline in raw string literal generates below warning so stopped doing it. Also comment out some unused parameters.

```
/Users/gary/rstudio/src/cpp/session/SessionModuleContext.cpp:2633:24: warning: adding 'int' to a string does not append to the string [-Wstring-plus-int]
   const char* msg = 1 + R"EOF(
                     ~~^~~~~~~~
/Users/gary/rstudio/src/cpp/session/SessionModuleContext.cpp:2633:24: note: use array indexing to silence this warning
```

### DIFF
--- a/src/cpp/desktop/DesktopBrowserWindow.cpp
+++ b/src/cpp/desktop/DesktopBrowserWindow.cpp
@@ -1,7 +1,7 @@
 /*
  * DesktopBrowserWindow.cpp
  *
- * Copyright (C) 2009-19 by RStudio, PBC
+ * Copyright (C) 2009-20 by RStudio, PBC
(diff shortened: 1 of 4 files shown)


kevinushey:
As an alternative, we could also do:

```
&R"EOF(...)EOF"[1]
```

to trim the leading newline from the raw string literal, which should avoid tripping the warnings.

This is of course bike-shedding though -- at least aesthetically I like having the relevant inner text formatted and displayed without any other extraneous syntax on those lines :-)

PR #6428, opened 2020-03-08
Hide any tab from view

melissa-barca:
This PR is the first update in a series of enhancements to our panel system addressing many of the requests brought up in #2854 . 
It updates the IDE's quadrant system so that any tab can be hidden from view. The hidden tabs are stored in a user preference string list and can be determined by looking at the contents of `TabSet1` and `TabSet2`. When a user hides all tabs within a quadrant, the quadrant will still exist and can be minimized. A future enhancement will make the quadrant completely hidden after the last remaining tab is removed. Tabs can only be hidden from `Global Options` as there is not an otherwise convenient way to bring tabs back.
This PR removes code previously used to make sure a tabset was never empty and couldn't contain only hideable tabs as now we allow for empty tabsets and every tab is considered hideable.

PR #6634, opened 2020-04-10
Fix "Move Focus to Console"

melissa-barca:
PR #6428 broke 'Move focus to Console' by calling the ActivatePaneEvent for it when this command doesn't work properly for the Console Pane. This fix prevents the event from being fired.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/console/ConsolePane.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/console/ConsolePane.java
@@ -57,7 +57,9 @@ public class ConsolePane extends WorkbenchPane
                       Commands commands,
                       Session session)
    {
-      super("Console", events);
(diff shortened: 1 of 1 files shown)


kevinushey:
Does this also work when focusing something else active in the Console pane; e.g. if you've opened and focused a Terminal tab?

melissa-barca:
> Does this also work when focusing something else active in the Console pane; e.g. if you've opened and focused a Terminal tab?

Yes, the ActivatePane event isn't made for these tabs but since they do not appear here: 
https://github.com/rstudio/rstudio/blob/master/src/gwt/src/org/rstudio/studio/client/workbench/ui/PaneManager.java#L848-L877 the event returns without doing anything. We could do the same thing for these tabs as `Console` where we pass `null` to the constructor or a null check could be added to PaneManager.cpp so switching focus to `Console` doesn't except out. What are your thoughts?

PR #6642, opened 2020-04-11
Make style consistent for named args in chunk opts

Selbosh:
The `include=FALSE` (no spaces) immediately followed by `echo = TRUE` (with spaces) is inconsistent. Should apply to other skeletons as well.


kevinushey:
Knitr chunks often omit the space between arguments with `=`, e.g. as used in the documentation here:

https://yihui.org/knitr/options/

So I don't think we should make this change.

PR #6660, opened 2020-04-14
remove spaces before statement-ending semicolons (older Java style)

gtritchie:


kevinushey:
Should we do the same for our C++ sources?

PR #6708, opened 2020-04-22
Prevent path traversal: Ignore relative URL path referencing past root

ricardofandrade:
Closes #1495

This PR adds a protection against path traversal in some scenarios.
The core of this change is the implementation of `core::http::Path::cleanup()`.
Goal: Pair each occurrence of `../` with a removal of an item from the path.
Before: If there was no matching pair (it was already on the "root") the `../` was preserved.
Now: If there's no matching pair, the `../` is discarded.

The new behavior is similar to how NGINX handles relative paths.
That was detected when comparing open source and pro.

With the change above, `core::http::URL::complete()` will start to discard extraneous `../`.
The `boost::test` gig present in `URL` was included in `core/http/URLTests.cpp` to be sure we keep testing these changes.

The second part of this change is exposing `core::http::URL::cleanupPath()` as a public function.
This function calls `core::http::Path::cleanup()`.

Finally, the `core::http::pathAfterPrefix()` function calls `cleanupPath()` ensuring that we thwart any attempts of path traversal, including `/help` which was the originally affected path.

### DIFF
--- a/src/cpp/core/http/URL.cpp
+++ b/src/cpp/core/http/URL.cpp
@@ -125,7 +125,12 @@ public:
 
          if (dirs_.at(i) == "..")
          {
-            if (!result.empty() && result.back() != "..")
+            if (result.empty())
+            {
+               continue;
+            }
+
+            if (result.back() != "..")
             {
                result.pop_back();
                continue;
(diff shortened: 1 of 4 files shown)


kfeinauer:
I don't know enough about this code to effectively review it, especially considering it is security sensitive. Adding @jmcphers

PR #6720, opened 2020-04-23
patch visual mode changes into source editor

jjallaire:


src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/AceEditor.java
```diff
@@ -1118,6 +1121,84 @@ public void insertCode(String code, boolean blockMode)
...
+      
+      // process changes
+      for (int i = 0; i<changes.length; i++) 
+      {
+         // get change
+         JsdiffChange change = changes[i];
+         
+         // if it's the first change then set the cursor location to the beginning of the file
+         if (i == 0)
+            selection.moveCursorTo(0, 0, false);
```
jcheng5, line 1175: I'd just move this to before the for-loop begins.

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/AceEditor.java
```diff
@@ -1118,6 +1121,84 @@ public void insertCode(String code, boolean blockMode)
...
+            {
+               col = col + charsLeft;
+               break;
+            }
+            else
+            {
+               charsLeft -= (charsLeftInLine+1); // add 1 for newline
+               col = 0;
+            }
+         }
```
jcheng5, line 1160: The logic looks correct to me. I had to focus hard to make sure that a) `charsLeft` can never become less than 0, and b) that the row will never be advanced accidentally.

This suggestion is (supposed to be) equivalent but a little easier for my brain. I changed the for-loop to while, incrementing `row` manually; and moved the newline "+1" from the end of the loop body to the beginning, which I find makes the if/else easier to reason about.

```java
         int row = startPos.getRow();
         int col = startPos.getColumn();
         while (row < session.getLength()) {

            // how many chars left in the current column?
            String line = session.getLine(row);
            // the +1 is for newline
            int charsLeftInLine = line.length() + 1 - col;

            // is the number of chars we still need to consume lte
            // the number of charsLeft?
            if (charsLeft < charsLeftInLine) 
            {
               col += charsLeft;
               break;
            }
            else
            {
               charsLeft -= charsLeftInLine;
               col = 0;
               row++;
            }
         }
```

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/AceEditor.java
jcheng5: Unless you didn't want to check for `changes.length == 0` I guess...

jjallaire: Yeah, I need to check for changes.length and not do anything to the cursor in that case. I agree though it would be more clear outside the loop.

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/AceEditor.java
jjallaire: Yes, your version reach much nicer. Made that change. Thanks very much for the review, this is not a bit of code I'd ever want to get wrong!

Issue #6703, opened 2020-04-21
Console scrolls on paste


GegznaV:
### System details

    RStudio Edition : Desktop 
    RStudio Version : 1.3.944-1
    OS Version      : Windows 10
    R Version       : 3.6.3

### Steps to reproduce the problem

Possibly related issue. Steps to reproduce:

1) Clear console.
2) Make it small enough.
![image](https://user-images.githubusercontent.com/12725868/79906573-b775ec00-8420-11ea-97c9-55de744f939d.png)
3) Run command that result does not fit in the console, e.g.,
```r
head(iris, n = 20)
```
The console looks like this:
![image](https://user-images.githubusercontent.com/12725868/79907107-bc876b00-8421-11ea-98c2-ce394fd99855.png)
The red arrow indicates the scrollbar position.
4) Place the cursor to position 1 (red dot in the fig. above) and paste (<kbd>Ctrl+V</kbd>) a piece of code that does not fit to one line in the current view, e.g.:
```r
# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
5) The console automatically scrolls and the pasted code gets out of view (red line shows scrollbar position):
![image](https://user-images.githubusercontent.com/12725868/79907277-02dcca00-8422-11ea-9d45-4d5f38d70ce0.png)

### Describe the problem in detail

Console inadequately scrolls and the lines of interest get out of view.

### Describe the behavior you expected

 No automatic scrolling on paste.

Possibly related:  #6697

ronblum:
@kevinushey Can this also be related to #6553?

kevinushey:
@ronblum I believe so; at least, it's definitely in the same family of issues.

PR #6733, opened 2020-04-24
fix ensure visible behavior

kevinushey:
Closes https://github.com/rstudio/rstudio/issues/6703.

Also bundles a brown-bag CSS fix for IE11 -- the empty space in a tab panel was not being styled for dark themes.

Screen Shot 2020-04-24 at 3 02 02 PM

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/theme/res/themeStyles.css
+++ b/src/gwt/src/org/rstudio/core/client/theme/res/themeStyles.css
@@ -2591,12 +2591,12 @@ body.rstudio-themes-flat .rstudio-themes-dark-grey {
 
 .rstudio-themes-flat .rstudio-themes-dark-grey .gwt-TabLayoutPanelTabs table.rstheme_tabLayoutCenter,
 .rstudio-themes-flat .rstudio-themes-dark-grey .rstheme_minimizedWindowObject table.rstheme_tabLayoutCenter {
-   background: THEME_DARKGREY_INACTIVE;
+   background: THEME_DARKGREY_INACTIVE !important;
 }
 
 .rstudio-themes-flat .rstudio-themes-dark-grey .gwt-TabLayoutPanelTabs,
 .rstudio-themes-flat .rstudio-themes-dark-grey .rstheme_minimizedWindowObject .rstheme_center {
-   background: THEME_DARKGREY_MOST_INACTIVE;
+   background: THEME_DARKGREY_MOST_INACTIVE !important;
 }
 
 @if user.agent safari {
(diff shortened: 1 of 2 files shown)


kevinushey:
Testing notes: the simplest way to test this is to try pasting a large, multiline block of code into the console; e.g.

```
foo <- function() {
    # code
    # code
    # code
    # code
    # code
    # code
    # code
    # code
    # code
    # code
}
```

Send that to the console, and then try navigating the history by e.g. pressing the up and down arrow keys. If the issue still occurs, you might see the cursor not scrolled into view (ie: the view may be at the very top of the Console history) in some cases.

Screen Shot 2020-04-24 at 3 42 01 PM

Normally though, the cursor should be at the end of the function definition and in view, e.g.

Screen Shot 2020-04-24 at 3 42 15 PM

src/gwt/src/org/rstudio/studio/client/common/shell/ShellWidget.java
```diff
@@ -433,7 +435,15 @@ public void consolePrompt(String prompt, boolean showInput)
...
-      scrollIntoView();
+      // NOTE: we don't scroll immediately as this is normally called
+      // in response to mutations of the console input buffer, and so
+      // we need to wait until Ace has finished rendering in response
+      // to that change.
+      scrollIntoViewPending_ = true;
+      
+      // In case there wasn't an Ace render in-flight, force a check
+      // for pending scroll (which will then force the cursor into view)
+      Scheduler.get().scheduleDeferred(() -> checkForPendingScroll());
```
jmcphers, line 446: Should we skip doing this if `scrollIntoViewPending_` is already true so there's no way for requests to stack up?

kevinushey: Good idea!

src/gwt/src/org/rstudio/core/client/theme/res/themeStyles.css
```diff
@@ -2591,12 +2591,12 @@ body.rstudio-themes-flat .rstudio-themes-dark-grey {
 
 .rstudio-themes-flat .rstudio-themes-dark-grey .gwt-TabLayoutPanelTabs table.rstheme_tabLayoutCenter,
 .rstudio-themes-flat .rstudio-themes-dark-grey .rstheme_minimizedWindowObject table.rstheme_tabLayoutCenter {
-   background: THEME_DARKGREY_INACTIVE;
+   background: THEME_DARKGREY_INACTIVE !important;
```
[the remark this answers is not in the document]
kevinushey: Yes 😞 and I think in this particular case we're likely working around some erroneous handling of CSS rule specificity that only occurs in IE / "old Edge".

kevinushey:
Backported to v1.3-patch: https://github.com/rstudio/rstudio/commit/72427a8eebef018e70e0fd0ba12cd8c3e4d2f09f

PR #6748, opened 2020-04-27
ensure RTOOLS40_HOME is set

### LINKED ISSUE #6746: compilation of packages on Windows can fail if RTOOLS40_HOME unset
### System details

    RStudio Edition : Desktop
    RStudio Version : any
    R Version       : 4.0.0

Use RStudio to install Rtools; e.g. via

```
if (!file.exists("C:/RBuildTools/4.0"))
  .rs.installBuildTools("Installing Rtools")
```

Then, put that on the PATH:

```
path <- Sys.getenv("PATH")
Sys.setenv(PATH = paste("C:\\RBuildTools\\4.0\\usr\\bin", path, sep = ";"))
```

Then, ensure that `RTOOLS40_HOME` is unset; e.g.

```
Sys.unsetenv("RTOOLS40_HOME")
```

Finally, try to install a package from sources that requires compilation, e.g.

```
install.packages("digest", type = "source")
```

Installation fails as the wrong path to Rtools40 is used:

```
> install.packages("digest", type = "source")
Installing package into 'C:/Users/kevinushey/r/win-library/4.0'
(as 'lib' is unspecified)
trying URL 'https://cran.rstudio.com/src/contrib/digest_0.6.25.tar.gz'
Content length 145642 bytes (142 KB)
downloaded 142 KB

* installing *source* package 'digest' ...
** package 'digest' successfully unpacked and MD5 sums checked
** using staged installation
** libs

*** arch - i386
"c:/rtools40/mingw32/bin/"g++  -std=gnu++11 -I"C:/R/R-40~1.0/include" -DNDEBUG     -I"C:/Users/kevinushey/AppData/Local/renv/extsoft/include"     -O2 -Wall  -mfpmath=sse -msse2 -mstackrealign -c SpookyV2.cpp -o SpookyV2.o
sh: c:/rtools40/mingw32/bin/g++: No such file or directory
make: *** [C:/R/R-40~1.0/etc/i386/Makeconf:229: SpookyV2.o] Error 127
ERROR: compilation failed for package 'digest'
* removing 'C:/Users/kevinushey/r/win-library/4.0/digest'
* restoring previous 'C:/Users/kevinushey/r/win-library/4.0/digest'
Warning in install.packages :
  installation of package 'digest' had non-zero exit status

The downloaded source packages are in
	'C:\tmp\RtmpemEXb7\downloaded_packages'
```

The correct path to the compiler tools should be used.

kevinushey:
Closes #6746.

### DIFF
--- a/src/cpp/core/r_util/RToolsInfo.cpp
+++ b/src/cpp/core/r_util/RToolsInfo.cpp
@@ -198,6 +198,11 @@ RToolsInfo::RToolsInfo(const std::string& name,
       // set BINPREF
       environmentVars.push_back({"BINPREF", "/mingw$(WIN)/bin/"});
 
+      // set RTOOLS40_HOME
+      std::string rtoolsPath = installPath.getAbsolutePath();
+      std::replace(rtoolsPath.begin(), rtoolsPath.end(), '\\', '/');
+      environmentVars.push_back({"RTOOLS40_HOME", rtoolsPath});
+
       // set clang args
 #ifdef _WIN64
       std::string baseDir = "mingw64";


kevinushey:
Note: want to confirm with Jeroen that this is the right change before merging, but it does seem to work on my Windows VM.

jeroen:
Note that `RTOOLS40_HOME` should be a windows path. The rtools40 windows installer stores the installation path in the default `RTOOLS40_HOME` and  our documentation recommends this:

```
writeLines('PATH="${RTOOLS40_HOME}\\usr\\bin;${PATH}"', con = "~/.Renviron")
```

This will only work for Windows paths. So I think you need to remove the `std::replace` ?

Also I very much recommend using the preset `RTOOLS40_HOME` if available, which is basically always. Thereby the behavior of rstudio stays consistent with the command line and RGui.

The only reason that `RTOOLS40_HOME` wasn't available in https://github.com/rstudio/rstudio/issues/6746 was that Kevin installed rtools40 in two different locations, and then uninstalled one of them. The uninstaller removes the variable. But this is really an edge case, it should usually not be needed to override `RTOOLS40_HOME`.

PR #6753, opened 2020-04-28
work around IE CSS styling issue

kevinushey:
This is a better version of the previous fix for styling in https://github.com/rstudio/rstudio/pull/6733.

While this seems to be a bug in IE11 / Edge, I haven't been able to distill a reproducible example.


gtritchie:
Can you elaborate on what the problems were with the original implementation? Good to have that written down / searchable as a historical record if nothing else!

kevinushey:
I think the original issue might've just been me adding `!important` on one element which didn't need it; thereby overriding styles set for the "active" tab element in addition to the background color for the whole panel row. :-/  I blame quarantine-brain.

Regardless, this feels like a cleaner work-around.

PR #6758, opened 2020-04-28
Script for local Docker-based Windows builds

gtritchie:
This finishes a script I started ages ago. You can now build RStudio for Windows on a Windows-10 Pro machine using the same Dockerfile we use for official builds. Simply clone the rstudio repo, then go into rstudio\docker folder (in command prompt) and run `win-docker-compile.cmd`.

Obviously you need Docker installed on your Windows machine, and configured to use Windows containers.

For best results, use a clean repo (no installed dependencies or previous build outputs); can achieve this via cloning new repo copy or "git clean -ffdx".

Was previously blocked due to MSVC 2017 issues with building source mounted into the container as a volume when using Hyper-V based Docker (the only mode supported on Windows-10). Windows-Server also supports process-isolation, which doesn't have this bug.

Worked around this by copying the repo into the container, then installing dependencies inside the running container (as is currently done in the real build). At the end, it copies out the zip and exe build products to /rstudio/docker/package.

### DIFF
--- a/docker/win-docker-compile.cmd
+++ b/docker/win-docker-compile.cmd
@@ -1,17 +1,18 @@
 @echo off
 setlocal enableextensions enabledelayedexpansion
 
-REM **************************************************************************
-REM Build Windows docker image containing tools to build RStudio IDE
-REM **************************************************************************
-REM Currently this goes as far as creating an image, but doesn't 
-REM run the build itself. See TODO's at end of this file.
-REM **************************************************************************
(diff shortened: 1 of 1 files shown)


gtritchie:
Note to self as baseline for future build time improvements: did a full build via Dockerfile on laptop (with pre-existing image), took 1:50 (one hour fifty minutes). This is a i7 / 1.80-2.30 GHz CPU, 16GB RAM, SSD.

PR #6774, opened 2020-04-29
Remove old application agreement feature

gtritchie:
- This was an undocumented feature from the early cloud-only RStudio beta period
- Rather than document and support it, taking opportunity to remove it


gtritchie:
> So red, such soothing!

Yes, I did it just for you!

PR #6828, opened 2020-05-07
Fix coexistence of git for windows and rtools4

### LINKED ISSUE #6696: [Windows] terminal tab crashes when msys2 is on the PATH
In order for Windows users to build packages from source with rtools40, we need to put `make`, `bash`, etc on the PATH. We recommend people to do this in their `~/.Renviron` file:

```r
PATH="${RTOOLS40_HOME}\usr\bin;${PATH}"
```

However doing so will crash the terminal tab in rstudio due to dll conflicts. 

The problem is that bash should always have itself and other coreutils in the front of the path, but rstudio invokes bash.exe from git4windows but appends `/bin` all the way at the end of the PATH instead of the front.

As a result, when we invoke e.g. `ls` we get `ls.exe` from rtools40 while running inside bash.exe from git4windows. Sadly, mixing different msys2 / cygwin runtimes in a process will 💣. 

You can confirm by putting `/usr/bin` (or just `/bin`) on the front, then the problem disappears:

```sh
PATH="/usr/bin:$PATH"
```

So RStudio should probably always do automatically to make sure that other utilities are called from the same bash runtime.

### LINKED ISSUE #6809: cygheap mismatch detected when running GNU tools in RStudio terminal
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.3.956
    OS Version      : Windows 10
    R Version       : 4.0.0

Open the Tools/Terminal

I mentioned that problem in #6801 and it persists when, fllowing the suggestion from #6801, I installed the preview version of RStudio. I was getting similar errors earlier, but now with a more informative message:

```
      0 [main] cut (11840) C:\rtools40\usr\bin\cut.exe: *** fatal error - cygheap base mismatch detected - 0x180317408/0x180316408.
This problem is probably due to using incompatible versions of the cygwin DLL.
Search for cygwin1.dll using the Windows Start->Find/Search facility
and delete all but the most recent version.  The most recent version *should*
reside in x:\cygwin\bin, where 'x' is the drive on which you have
installed the cygwin distribution.  Rebooting is also suggested if you
are unable to find another cygwin DLL.
```

Similar errors pop-up when I try to run `locate`, `which` and so on.

Following the instruction however, Windows does not find any instance of `cygwin1.dll` on my system using file search. Rebooting did not help. I can find tools like `cut.exe` both in Rtools and in Git for Windows (I have both in my `PATH`). 

I would like to execute these commands in the RStudio terminal.

gtritchie:
- Fixes #6696
- Fixes #6809
- Stop directly invoking the MSYS2 bash.exe installed with Git for Windows, and instead use the wrapper bash.exe (also installed by Git for Windows) which correctly sets up the environment
- Doing so allows RStudio terminal to use Git-Bash even if RTools4 has been installed and is on the path (as recommended by RTools4 docs); without this fix the path is wrong and can try to use Posix tools from both environments leading to runtime errors

Test notes for QA
-----------------
- Windows-desktop only codepath
- Must have git-for-windows installed
- Confirm that Terminal detects git-bash by default and loads it
- Also double check Tools / Shell (should launch same type of shell as selected for terminal)
- The bug repros only if you install Git for Windows AND RTools4 (including adding to the path in Documents\\.Renviron as described in https://cran.r-project.org/bin/windows/Rtools/

### DIFF
--- a/NEWS-1.3-patch.md
+++ b/NEWS-1.3-patch.md
@@ -10,3 +10,4 @@
 - Allow multiple space-separated domains in `www-frame-origin` for Tutorial API (Pro)
 - Fix dependency installation for untitled buffers (#6762)
 - Add option `www-url-path-prefix` to force a path on auth cookies (Pro #1608)
+- Fix Terminal to works with both Git-Bash and RTools4 MSYS2 installed on Windows (#6696)
--- a/NEWS-1.3-patch.md
+++ b/NEWS-1.3-patch.md
@@ -10,4 +10,4 @@
 - Allow multiple space-separated domains in `www-frame-origin` for Tutorial API (Pro)
 - Fix dependency installation for untitled buffers (#6762)
 - Add option `www-url-path-prefix` to force a path on auth cookies (Pro #1608)
-- Fix Terminal to works with both Git-Bash and RTools4 MSYS2 installed on Windows (#6696)
+- Fix Terminal to work with both Git-Bash and RTools4 MSYS2 installed on Windows (#6696, #6809)
--- a/src/cpp/session/modules/SessionTerminalShell.cpp
+++ b/src/cpp/session/modules/SessionTerminalShell.cpp
@@ -293,8 +293,16 @@ core::FilePath getGitBashShell()
    core::FilePath gitExePath = modules::git::detectedGitExePath();
    if (!gitExePath.isEmpty())
    {
-      core::FilePath gitBashPath =
-            gitExePath.getParent().getParent().completePath("usr/bin/bash.exe");
+      // Prefer git-for-windows bash wrapper; it does some setup before it invokes MSYS2's
+      // usr/bin/bash.exe; not using this wrapper causes problems if other Cygwin-based
+      // Posix environments are on path (such as those in RTools4)
+      core::FilePath gitBashPath = gitExePath.getParent().completePath("bash.exe");
+      if (gitBashPath.exists())
+         return gitBashPath;
+
+      // fall back to old detection behavior in case there's an older version that only
+      // has this installation pattern
+      gitBashPath = gitExePath.getParent().getParent().completePath("usr/bin/bash.exe");
       if (gitBashPath.exists())
          return gitBashPath;
       else


jeroen:
We had some users report that this still doesn't work in the latest preview of rstudio 1.3. Is that the intention?

I think it would be good to get this into 1.3 release if possible, it's going to affect almost all windows users.

gtritchie:
This fix is currently slated for the first 1.3 bug-fix release (aka 1.3-patch) but not the initial 1.3 release.

At the moment, the "preview" builds still reflect the initial 1.3 release candidate. Once that releases (Real Soon Now) then the preview builds will switch to being the 1.3-patch candidates including this fix.

The fix is also available in the 1.4 daily builds; those are potentially quite unstable at this stage so usual warnings about pre-pre-release software.

jeroen:
OK thanks for the update.

This bug is going to affect almost all Windows users that have rtools. I'm not an rstudio developer obviously, but as maintainer of R for Windows I would recommend you really try to get this fix the release to make rstudio 1.3 work with R/rtools 4.0...

gtritchie:
Understood, I will raise the issue internally.

edonnachie:
I'm running the current preview version (2020-07-07) with R 4.0.2 on Windows 2016 Server, and have the same problem. Was the fix added to this version? If so, you may need to look again at this issue.

gtritchie:
Which problem/issue? There were a couple linked to this change. Best to report what you are seeing in one of those:

- Fixes #6696
- Fixes #6809

edonnachie:
I believe this primarily refers to #6809 

The path to RTools40 is inserted in my Renviron.site file.

```
1 [main] make (8064) C:\rtools40\usr\bin\make.exe: *** fatal error - cygheap base mismatch detected - 0x180331408/0x180316408.
This problem is probably due to using incompatible versions of the cygwin DLL.
Search for cygwin1.dll using the Windows Start->Find/Search facility
and delete all but the most recent version.  The most recent version *should*
reside in x:\cygwin\bin, where 'x' is the drive on which you have
installed the cygwin distribution.  Rebooting is also suggested if you
are unable to find another cygwin DLL.
```

PR #6832, opened 2020-05-07
Add unique rstudio_foo classes to some source panel elements

### LINKED ISSUE #6817: DOM selectors for source pane
We'd like to use `rstudioapi::higlightUI` to select some elements inside the source window. We're hoping to be able to highlight the following areas

### 1. The whole source pane

highlight 001

### 2. Maximize/minimize buttons on the source pane

highlights 021

### ~~3. Buttons at the top of an RMD file~~

~~Each of the icons at the top of the Rmd file (only one illustrated here). In each case, it appears that there is a DOM ID associated with each of the items in the drop down list associated with the icon, but not a DOM ID associated with the icon itself.~~

EDIT: will move to a separate issue

highlights 020

### 4. The outline section of an RMD doc

highlights 022

gtritchie:
Fixes #6817

# Overview

You can now discover via Tools / Diagnostics / Show DOM Elements (and thus highlight) the overall source panel, the min/max buttons within the source panel, and the outline, using class-based CSS selectors. For example:

```R
rstudioapi::highlightUi(
  list(
    list(query = ".rstudio_doc_outline_container", parent = 0L),
    list(query = ".rstudio_source_panel", parent = 0L),
    list(query = ".rstudio_panel_min_btn_source", parent = 0L)
  )
)

```
screenshot showing highlighted elements

# Details

Added the `ClassIds.java` file, which conceptually parallels the existing `ElementIds.java` as the place to define unique and stable DOM class names for elements.

Using classes allows elements that can have multiple instances, including most UI in the source editor (because you can open multiple documents), to have the same selector, and by prefixing them with `rstudio_` the `Show DOM Elements` feature reveals them.

ElementIds must be unique (HTML rule and important for accessibility) so our code disambiguates them at runtime, and you end up with IDs with numeric suffixes (_###) whose values depend on the state of the UI and the order things were opened.

For automation purposes such as rstudioapi::highlightUi the unstable ElementIds are not ideal, hence reason for this class-based approach (multiple elements can have the same class applied, of course).

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/ClassIds.java
+++ b/src/gwt/src/org/rstudio/core/client/ClassIds.java
@@ -1,0 +1,63 @@
+/*
+ * ClassIds.java
+ *
+ * Copyright (C) 2020 by RStudio, PBC
+ *
+ * Unless you have received this program directly from RStudio pursuant
+ * to the terms of a commercial license agreement with RStudio, then
+ * this program is licensed to you under the terms of version 3 of the
+ * GNU Affero General Public License. This program is distributed WITHOUT
+ * ANY EXPRESS OR IMPLIED WARRANTY, INCLUDING THOSE OF NON-INFRINGEMENT,
+ * MERCHANTABILITY OR FITNESS FOR A PARTICULAR PURPOSE. Please refer to the
+ * AGPL (http://www.gnu.org/licenses/agpl-3.0.txt) for more details.
+ *
+ */
+package org.rstudio.core.client;
+
+import com.google.gwt.dom.client.Element;
+import com.google.gwt.user.client.ui.Widget;
+
+/**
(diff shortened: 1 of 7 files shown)


trestletech:
This is great! Thanks.

melissa-barca:
Overall LGTM, we should probably move the work I did to `ClassIds.java` at some point but I did add a comment to say they need to remain stable. Thanks for specifying the importance of starting with `rstudio_`, I'll need to update my code with that as well.

I wasn't aware of `ElementIds::idSafeString` so I wrote my own StringUtils class here: https://github.com/rstudio/rstudio/blob/master/src/gwt/src/org/rstudio/core/client/StringUtil.java#L950-L981. While the `StringUtils` function allows a little more variety in class names, the only issue I see with the `ElementIds` call is that it allows the id to lead with a digit.

PR #6862, opened 2020-05-12
support view of Python objects in Environment pane

kevinushey:
This PR makes it possible to view Python objects (from a `reticulate` session) within the Environment Pane.

At its core, the Environment Pane will now switch between R and Python depending on whether the "regular" R console is active, versus the `reticulate` Python console. See demo:

rstudio-python-environment-pane

After the Python session has been initialized by `reticulate`, an option to view Python environments and modules will be displayed. Arbitrary loaded modules can be viewed.

Screen Shot 2020-05-12 at 4 12 03 PM

---

## TODO

- [x] Audit that I haven't broken any pre-existing implicit contracts between the Environment Pane and the "monitored" environment.
- [x] Handle object removals.
- [x] Ensure behavior is consistent around session restarts, page reloads, and so on.
- [x] Consider whether we could make it possible to `View()` such objects with the Data Viewer or Object Explorer.

### DIFF
--- a/src/cpp/session/modules/SessionObjectExplorer.R
+++ b/src/cpp/session/modules/SessionObjectExplorer.R
@@ -47,9 +47,9 @@
 {
    # retrieve object from cache
    object <- .rs.explorer.getCachedObject(
-      id = id,
+      id             = id,
       extractingCode = extractingCode,
-      refresh = FALSE
+      refresh        = FALSE
    )
    
    # construct context
@@ -71,9 +71,9 @@
 {
    # retrieve object from cache
    object <- .rs.explorer.getCachedObject(
-      id = id,
+      id             = id,
       extractingCode = NULL,
-      refresh = TRUE
+      refresh        = TRUE
    )
    
    # construct context
@@ -93,7 +93,7 @@
 
 .rs.addJsonRpcHandler("explorer_end_inspect", function(id)
 {
-   .rs.explorer.removeCachedObject(id)
+   .rs.explorer.removeCacheEntry(id)
 })
 
 .rs.addFunction("objectAddress", function(object)
@@ -147,7 +147,7 @@
       if (inherits(object, "python.builtin.module"))
          return(FALSE)
       
-      return(reticulate::py_has_attr(object, "__dict__"))
+      TRUE
    }
    
    attrib <- attributes(object)
@@ -230,28 +230,67 @@
       return(object)
    
    # otherwise, evaluate expression to retrieve sub-object
-   envir <- new.env(parent = globalenv())
-   envir[["__OBJECT__"]] <- object
-   
-   tryCatch(
-      eval(parse(text = extractingCode), envir = envir),
-      error = warning
-   )
+   if (.rs.reticulate.replIsActive())
+   {
+      pyid <- paste("_rstudio_viewer", id, sep = "_")
+      code <- sub("`__OBJECT__`", pyid, extractingCode, fixed = TRUE)
+      
+      tryCatch(
+         reticulate::py_eval(code, convert = FALSE),
+         error = warning
+      )
+   }
+   else
+   {
+      envir <- new.env(parent = globalenv())
+      envir[["__OBJECT__"]] <- object
+      
+      tryCatch(
+         eval(parse(text = extractingCode), envir = envir),
+         error = warning
+      )
+   }
 })
 
-.rs.addFunction("explorer.setCachedObject", function(object,
-                                                     id = .rs.createUUID())
+.rs.addFunction("explorer.setCacheEntry", function(entry,
+                                                   id = .rs.createUUID())
 {
+   # place entry in cache
    cache <- .rs.explorer.getCache()
-   cache[[id]] <- object
+   cache[[id]] <- entry
+   
+   # for Python objects, store a reference in main module
+   if (inherits(entry$object, "python.builtin.object"))
+   {
+      main <- reticulate::import_main(convert = FALSE)
+      pyid <- paste("_rstudio_viewer", id, sep = "_")
+      reticulate::py_set_attr(main, pyid, entry$object)
+   }
+
+   # return generated id
    id
 })
 
-.rs.addFunction("explorer.removeCachedObject", function(id)
+.rs.addFunction("explorer.removeCacheEntry", function(id)
 {
+   # retrieve entry from cache
    cache <- .rs.explorer.getCache()
+   entry <- cache[[id]]
+   
+   # remove old reference
    if (exists(id, envir = cache))
       rm(list = id, envir = cache)
+   
+   # if this is a Python object, remove it from the main module
+   if (inherits(entry$object, "python.builtin.object"))
+   {
+      main <- reticulate::import_main(convert = FALSE)
+      pyid <- paste("_rstudio_viewer", id, sep = "_")
+      reticulate::py_set_attr(main, pyid, NULL)
+   }
+   
+   # return id of deleted object
+   id
 })
 
 #' @param name The display name, as should be used in UI.
@@ -353,7 +392,7 @@
    )
    
    # put it in cache and retrieve handle
-   id <- .rs.explorer.setCachedObject(entry)
+   id <- .rs.explorer.setCacheEntry(entry)
    
    # return a handle object
    list(
@@ -396,12 +435,7 @@
    n <- length(.$object)
    expandable <- if (inherits(object, "python.builtin.object"))
    {
-      inherits(object, c(
-         "python.builtin.module",
-         "python.builtin.dict",
-         "python.builtin.tuple",
-         "python.builtin.list"
-      ))
+      TRUE
    }
    else
    {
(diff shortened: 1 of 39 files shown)


kevinushey:
It didn't feel right having Python objects in the Environment pane without being able to actually view them, so ...

Screen Shot 2020-05-18 at 2 30 12 PM

We can also use the data viewer to view Pandas DataFrames, although this relies on the `reticulate` machinery to first convert that to an R `data.frame` first so it's not a completely honest representation of the data (although it should still be useful).

src/cpp/session/modules/SessionObjectExplorer.R
```diff
@@ -952,21 +1176,25 @@
...
-{
-   .Call("rs_setRefCount",
-         as.name("object"),
-         environment(),
-         as.integer(count),
-         PACKAGE = "(embedding)")
-})
+   # NOTE: technically, these objects are expandable (they have attributes
+   # and may even have methods of interest) but since they're usually less
+   # interesting than the object's actual value we ignore those byd default.
```
jmcphers, line 1183: Nit: typo (`byd`)

src/cpp/session/modules/SessionReticulate.cpp
```diff
@@ -56,6 +56,27 @@ void onDeferredInit(bool)
 }
 
 } // end anonymous namespace
+
+bool isPythonInitialized()
```
jmcphers, line 60: Is this called a lot? (if so wondering if it'd be better to detect python initialization eagerly instead of recomputing it on every call when uninitialized)

src/cpp/session/modules/environment/SessionEnvironment.cpp
```diff
@@ -485,9 +493,25 @@ Error setEnvironment(boost::shared_ptr<int> pContextDepth,
...
+   else if (s_environmentLanguage == "Python")
+   {
+      if (environmentName != s_monitoredPythonModule)
+      {
+         s_monitoredPythonModule = environmentName;
+         
+         ClientEvent event(client_events::kEnvironmentRefresh);
+         module_context::enqueClientEvent(event);
+      }
+   }
```
jmcphers, line 513: Since this is a plain string value, maybe dump a log entry if it's something other than the values we expect? (realize it's totally impossible for it to get into that state _now_...)

src/cpp/session/modules/environment/SessionEnvironment.cpp
```diff
@@ -720,25 +772,56 @@ Error setContextDepth(boost::shared_ptr<int> pContextDepth,
...
+            r::exec::RFunction(".rs.reticulate.detectChanges")
+            .addParam(s_monitoredPythonModule)
+            .call();
+      
+      if (error)
+         LOG_ERROR(error);
+   }
+   
    // Ignore if not monitoring
    if (!s_monitoring)
```
jmcphers, line 820: Does this imply that the Python environment pane does not respect the manual/auto refresh setting? (Should it?)

src/gwt/src/org/rstudio/studio/client/workbench/views/environment/EnvironmentPane.java
```diff
@@ -150,9 +170,38 @@ protected Toolbar createMainToolbar()
...
+      
+      languageButton_ = new ToolbarMenuButton(
+            "R",
+            ToolbarButton.NoTitle,
+            (ImageResource) null,
+            languageMenu_);
+      
+      // don't show the language button at first; instead,
+      // toggle its visibility after the widget is created
+      // depending on whether Python has been initialized
```
jmcphers, line 198: Not a checkin blocker, but I wonder if it'd be nicer to show this all the time as a sort of advertisement of what's possible. Attempting to switch modes before Python is initialized could show instructions for enabling the feature (by initializing python or installing the appropriate software). What do you think?

src/cpp/session/modules/SessionObjectExplorer.R
kevinushey: Thanks for catching that! Fixed now.

src/cpp/session/modules/SessionReticulate.cpp
kevinushey: It currently doesn't, but since we do get notified when Python is initialized by `reticulate` we can just use that to set and store a flag rather than making this a more expensive call. I've made that change.

src/cpp/session/modules/environment/SessionEnvironment.cpp
kevinushey: Good idea -- I've added a simple warning log message.

src/cpp/session/modules/environment/SessionEnvironment.cpp
kevinushey: You're right! I've made that change.

src/gwt/src/org/rstudio/studio/client/workbench/views/environment/EnvironmentPane.java
kevinushey: At first I was opposed, but now I think this is worth doing -- especially so now because the trigger that makes this button visible is non-obvious (`reticulate` must be loaded _and_ Python must be initialized). I'll make that change.

src/cpp/session/modules/SessionReticulate.R
```diff
@@ -1411,3 +1462,403 @@ html.heading = _heading
...
+      reticulate::py_get_attr(parent, name)
+   
+   # is this object 'data'? consider non-callable, non-module objects as data
+   isData <- !(
+      grepl("^__.*__$", name) ||
+      reticulate:::py_is_callable(object) ||
+      reticulate:::py_is_module(object)
+   )
+   
+   # TODO: there isn't really a distinction between an objects "type"
```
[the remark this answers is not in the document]
kevinushey: I've converted this to a NOTE, since it really doesn't represent an actionable item -- it's ultimately just to help anyone reading that bit of code understand some of the decisions made there.

LuchaoQi:
Thanks to all developers and your contribution! That would be pretty helpful. I can't wait to try this new feature.

markderry:
Thank you for this!!

Mjvolk3:
Sorry for the rookie question, but how do you use this great new tool to work? Is it as simple as making sure my Rstudio version is up to date?

jmcphers:
You need a daily version of RStudio (1.4), plus an up to date version of the Reticulate R package.

https://dailies.rstudio.com/

Mjvolk3:
Great, thank you!

markderry:
Just loaded it up and it works like a charm!!
Screen Shot 2020-06-01 at 7 43 56 PM

edgBR:
Is this available also in RStudio Server Pro?

jmcphers:
Yes, it is, but as it's a very new feature it is currently only available in the daily builds. You can get a daily build of RSP here:

https://dailies.rstudio.com/

Your existing RStudio Server Pro license will work with the daily builds (the usual caveats about daily build stability apply!).

mdelop721:
Hi sorry for the rookie question, but does it work with python 3.8.5?? I'm not able to run any python code, it gives this error:
"Error un py_call_impl(callable, dots$args, dots$keywords) :
TypeError: use() got an unexpected keyword argument 'warn'

Any help would be appreciated

kevinushey:
Works for me:

```
> reticulate::use_python("/opt/local/python/3.8.5/bin/python3", required = TRUE)
> library(reticulate)
> reticulate::py_config()
python:         /opt/local/python/3.8.5/bin/python3
libpython:      /opt/local/python/3.8.5/lib/libpython3.8.dylib
pythonhome:     /opt/local/python/3.8.5:/opt/local/python/3.8.5
version:        3.8.5 (default, Jul 21 2020, 10:18:06)  [Clang 11.0.3 (clang-1103.0.32.62)]
numpy:           [NOT FOUND]

NOTE: Python version was forced by use_python function
> reticulate::py_run_string("1 + 1")
> reticulate::py_eval("1 + 1")
[1] 2
```

If you can provide more information, please provide a reproducible example at https://github.com/rstudio/reticulate/issues.

jpuig005:
Hey, I just installed the latest build of Rstudio and, even though I am able to View() all the variables in the python environment, I can't visualize the ones from R, when I try to view them, this error pops up: 

Error in py_call_impl(callable, dots$args, dots$keywords) : 
  RuntimeError: Evaluation error: ValueError: call stack is not deep enough.

Is this a bug?

kevinushey:
@jpuig005: can you please provide more details / a reproducible example?

jpuig005:
At the beggining of an R script, I call:

.
library(reticulate)
use_python("C:\\Users\\jpuig\\Anaconda3\\envs\\tf_gpu\\python.exe", required = TRUE)
source_python("expICH/00_core/expICH_functions.py")
.

expICH_functions.py looks like:

import pandas as pd
import numpy as np

def function1():
.
def functionN():

As soon as the source_python function gets called, any variable I create in the R environment can't be visualized, I got the following errors:

image

kevinushey:
Thanks; I've filed this here: https://github.com/rstudio/rstudio/issues/8065.

xiaodfeng:
Works well for me. Thank you.

PR #6866, opened 2020-05-13
Fix default open docs and first project template.

### LINKED ISSUE #6865: Default Open Docs don't work
### System details
    RStudio Version : 1.3.x/1.4.x

Load a new project that has `DefaultOpenDocs`. Notice that the listed documents do not show up in the source pane.

kfeinauer:
First project template just needed code similar to what exists in workspaces.

Default open docs were broken because the scratch path is incorrect during `ProjectContext::startup` and is later correct during `ProjectContext::initialize`.

Fixes #6865
Fixes https://github.com/rstudio/rstudio-pro/issues/1668

### DIFF
--- a/src/cpp/session/SessionMain.cpp
+++ b/src/cpp/session/SessionMain.cpp
@@ -1878,10 +1878,10 @@ int main (int argc, char * const argv[])
       FilePath userScratchPath = options.userScratchPath();
       if (userScratchPath.exists())
       {
-         std::vector<FilePath> scratchChildren;
-         userScratchPath.getChildren(scratchChildren);
-
-         if (scratchChildren.size() == 0)
+         // if the lists directory has not yet been created,
+         // this is a new user
+         FilePath listsPath = userScratchPath.completeChildPath("monitored/lists");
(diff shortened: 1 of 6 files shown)


src/cpp/session/SessionMain.cpp
```diff
@@ -1878,10 +1878,10 @@ int main (int argc, char * const argv[])
       FilePath userScratchPath = options.userScratchPath();
       if (userScratchPath.exists())
       {
-         std::vector<FilePath> scratchChildren;
-         userScratchPath.getChildren(scratchChildren);
-
-         if (scratchChildren.size() == 0)
+         // if the lists directory has not yet been created,
+         // this is a new user
+         FilePath listsPath = userScratchPath.completeChildPath("monitored/lists");
```
kevinushey, line 1883: nit: may want to use `monitoredParentPath()` here; although it would require making this function public:

https://github.com/rstudio/rstudio/blob/c65a73b21833d0e5f9a3e25ef85b038dd36d5113/src/cpp/session/SessionModuleContext.cpp#L543-L550

Issue #6895, opened 2020-05-16
Unexpected "Reload Site?" browser prompts with certain file types


gtritchie:
### System details

    RStudio Edition : Server
    RStudio Version : 1.3.958-3
    OS Version      : Any Posix
    R Version       : Any

This is a regression introduced in 1.3.947, possibly by: 1549686f from PR #6714

### Steps to reproduce the problem
- Load RStudio Server (repro'd in Chrome, Firefox, Safari, Edge, IE11)
- File / New File / R Script (and/or open an existing file of certain types, I tried .txt and .sh)
- Quit the R session ("power" button in upper-right or File / Quit Session)
- Get the R Session Ended dialog
- Click the Start New Session button

### Describe the problem in detail
- Prompt from browser asking if you want to leave or reload the page (wording varies by browser and platform):
<img width="777" alt="screenshot of Chrome for Mac prompt dialog" src="https://user-images.githubusercontent.com/10569626/82126928-d25c2780-9764-11ea-86b7-82bce9129fe4.png">

### Describe the behavior you expected
- Reload without a browser prompt

### Other scenarios
- This happens with a new unsaved R script, but also happens if you open an existing file of certain types (I tried .txt files and .sh files). Does not happen if you load an existing R file!
- This impacts any scenario that reload the page, such as changing a preference that triggers page reload (e.g. toggle screen reader support)

gtritchie:
This is quite painful when it happens and we should understand the cause and hopefully fix for 1.3-patch.

rstudio-build:
Yes, I see this all the time in development mode and it drives me bananas.
We should definitely fix this!

kevinushey:
This is where the window closing message is being set, triggering that behavior:

https://github.com/rstudio/rstudio/blob/187ea2b7d8aba7045cfbcacad4416863d7e9824b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/model/DocUpdateSentinel.java#L168-L177

The issue occurs because, when a new document is created, we receive a value changed event associated with the empty document contents for the new document:

https://github.com/rstudio/rstudio/blob/187ea2b7d8aba7045cfbcacad4416863d7e9824b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/model/DocUpdateSentinel.java#L730-L735

Normally, this would trigger the autosaver to run and update the source document in the source database. However, we explicitly disable the autosaver for the document when it's initially created:

[...]

PR #6905, opened 2020-05-18
suspend change detection handlers during document init

kevinushey:
Closes https://github.com/rstudio/rstudio/issues/6895. See associated issue for discussion.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/source/model/DocUpdateSentinel.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/model/DocUpdateSentinel.java
@@ -722,23 +722,37 @@ public class DocUpdateSentinel
       }
    }
    
-   public void suspendAutosave(boolean value)
+   public void withChangeDetectionSuspended(Command command)
    {
-      suspendAutosave_ = value;
+      try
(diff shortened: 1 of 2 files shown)


src/gwt/src/org/rstudio/studio/client/workbench/views/source/model/DocUpdateSentinel.java
```diff
@@ -722,23 +722,37 @@ private void applyProperties(JsObject properties,
       }
    }
    
-   public void suspendAutosave(boolean value)
+   public void withChangeDetectionSuspended(Command command)
    {
-      suspendAutosave_ = value;
+      try
+      {
+         suspendDetectChanges_ += 1;
```
kevinushey, line 729: Note: using an integer flag here just to make this API safe against re-entrant calls to `withChangeDetectionSuspended()`.

kevinushey:
Backports:

v1.3: https://github.com/rstudio/rstudio/commit/3a09be39fd51a8fafa8ae330007937d31924b395
v1.3-patch: https://github.com/rstudio/rstudio/commit/0c980363c5e323abaed9e3274ef49ab7e82f97fa

Issue #6913, opened 2020-05-19
Javascript exception when loading GWT sources relying on Ace


kevinushey:
### System details

    RStudio Edition : Server
    RStudio Version : 1.4.358
    OS Version      : Ubuntu 20.04
    R Version       : 4.0.0

### Steps to reproduce the problem

Launch RStudio with the DevTools inspector open.

### Describe the problem in detail

Exception while attempting to create the AceCommandManager:

<img width="1919" alt="Screen Shot 2020-05-19 at 11 17 14 AM" src="https://user-images.githubusercontent.com/1976582/82363188-53fcc100-99c2-11ea-9a54-f3927147e53a.png">

Normally, `$wnd.require` is a method which should be injected by Ace when it is loaded. However, `window.require` is not defined (and neither is `window.ace`).

### Describe the behavior you expected

No exceptions on startup.

- [x] I have read the guide to submitting good bug reports at https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports . 
- [x] I have installed the latest version of RStudio and confirmed that the issue still persists.
- [x] If I am reporting a RStudio crash, I have included a diagnostics report. https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

gtritchie:
I saw this one time (a few days ago), but couldn't repro on subsequent attempts.

rich-rstudio:
verified on 1.4.768

PR #6914, opened 2020-05-19
ensure Ace loaded before initializing command manager

kevinushey:
Fixes an issue where some GWT components depending on Ace could fail to initialize due to their attempt to run before Ace itself had been loaded.

This is a somewhat hacky fix -- a more appropriate fix, I think, would be to eschew the use of the Ace loaders altogether and instead just include Ace (+ our other polyfills) into a single `.js` bundle that we load on startup. However, I wanted to get something in quickly to at least fix builds (+ releases!)

Closes https://github.com/rstudio/rstudio/issues/6913.


kevinushey:
Going to merge now to get builds happy again. (@jmcphers let me know if you have any concerns)

PR #6916, opened 2020-05-19
always set role=application on entire page and remove user option

### LINKED ISSUE #6863: Set "application" role as default true value and remove the toggle option from accessibility settings
### System details

    RStudio Edition : Server
    RStudio Version : 1.3.957
    OS Version      : any
    R Version       : any

With NVDA or JAWS loaded, compare between on and off of the following option:

> Help / Accessibility / Accessibility Options... / Entire page has application role

The option above was originally suggested to allow screen-reader users to selectively toggle on and off `role = "application"` for the RStudio web interface.

When it is on, the RStudio interface is recognized by screen readers as an application-like environment, meaning that NVDA and JAWS will turn off "virtual cursor" (AKA, browse mode) while switching to "forms mode" (AKA, focus mode). When off, users have to manually switch between two different screen-reader modes/cursor.

After several hours of testing with screen readers in different browsers, it turns out that there is no particular reason that we have to keep the toggle option because the best screen-reader navigation has been verified with the application role set to true.

Furthermore, the "Console Output" region where screen readers need to be switched back to virtual cursor (AKA, browse mode), has been properly marked up with `role = "document"` meaning that users can be automatically given appropriate navigation mode/cursor contextually.

Having the option could mislead users with no technical background.

1. Please remove "Entire page has application role" from Help / Accessibility / Accessibility Options.
2. Fix `role = "application"` as default (true) attribute for the entire RStudio web UI.

gtritchie:
- Additional research and usage testing by @jooyoungseo has confirmed we should always set this, and there is no reason to present this (extremely hard to explain) option to screen reader users.
- Very safe fix, removes a checkbox from Accessibility options pane, and always sets a role that was previously conditional
- Fixes #6863

### DIFF
--- a/NEWS-1.3-patch.md
+++ b/NEWS-1.3-patch.md
@@ -10,6 +10,7 @@
 - Adds class attributed to RMarkdown chunks, their control buttons, and their output based on their given labels. (#6787)
 - Add option `www-url-path-prefix` to force a path on auth cookies (Pro #1608)
 - Add additional keyboard shortcut (Ctrl+`) for Focus Console Output accessibility command (#6850)
+- Always set application role for screen readers and removed related accessibility preference checkbox (#6863)
 
 ### Bugfixes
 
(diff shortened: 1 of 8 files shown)


gtritchie:
master: 72afd595

Issue #6971, opened 2020-05-27
Tab completion + up/down keyboard scrolling is broken


DavisVaughan:
### System details

    RStudio Edition : Desktop
    RStudio Version : Daily 1.4.358
    OS Version      : Mac
    R Version       : R 4.0.0

### Problem

I am a big user of tab complete + up/down on the keyboard to select the correct tab completion option. When working with C code, this is no longer possible. In the attached video, every time you see the autocompletion suggestion box disappearing I am just pressing the Down key on my keypad. In older versions of RStudio this used to move up and down the list of suggestions.

I think this is related to a previous PR / commit that was focused on making the same autocompletion box not "steal" focus from the editor. I imagine this bug may have been introduced there?

https://vimeo.com/user116518286/review/423289880/a4377f38e5

Issue #6547, opened 2020-03-27
Keyboard issues with toolbar dropdown menus


gtritchie:
### System details

    RStudio Edition : Desktop or Server
    RStudio Version : 1.4.102
    OS Version      : any (most likely; I was trying macOS dev server with Safari/VoiceOver)
    R Version       : any

### Steps to reproduce the problem
- open an rmd
- get keyboard focus to the editor toolbar (currently a bit tricky to do, but see #6543)
- activate a dropdown menu such as the dropdown arrow next to Knit, using spacebar or enter
- hit Esc to close the menu
- open the menu again via keyboard
- arrow down to Knit Directory, the right-arrow to move into submenu
- try to move up and down submenu with arrow keys

### Describe the problem in detail
- closing a toolbar menu via Esc moves focus to the html page's body
- in a dropdown that has submenus (e.g. Format / Text), use right arrow or enter to open the submenu, then try to move up and down with arrow keys; moves entry on the parent menu and closes the submenu

### Describe the behavior you expected
- when closing toolbar menu with Esc, should return focus to corresponding menubutton
- be able to navigate toolbar submenus via keyboard same as main menu

### Other considerations
- can't activate dropdown with down arrow, must use spacebar or enter (nice to have, not required by ARIA)
- known issue: can't move between buttons with arrow keys, must use Tab key, covered by #5960

gtritchie:
Not likely to get to this in 1.4, much higher priority accessibility tweaks ahead of this one.

PR #6976, opened 2020-05-27
cpp completion manager needs to explicitly handle keys

kevinushey:
Closes https://github.com/rstudio/rstudio/issues/6971.

~Candidate for v1.3-patch backport.~ Affects v1.4 only; not a candidate for backport.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/cpp/CppCompletionManager.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/cpp/CppCompletionManager.java
@@ -210,19 +210,25 @@ public class CppCompletionManager implements CompletionManager
             case KeyCodes.KEY_N: return popup.selectNext();
             }
          }
-             
+         
+         // allow navigation with arrow keys
+         // (needs to be handled here as the popup will not be explicitly focused)
+         if (modifier == KeyboardShortcut.NONE)
+         {
+            switch (keyCode)
+            {
+            case KeyCodes.KEY_UP:    return popup.selectPrev();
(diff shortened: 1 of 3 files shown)


gtritchie:
Note, likely related to #6547

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/cpp/CppCompletionManager.java
```diff
@@ -210,21 +210,27 @@ else if (event.getKeyCode() == 113 // F2
...
+         {
+            switch (keyCode)
+            {
+            case KeyCodes.KEY_UP:    return popup.selectPrev();
+            case KeyCodes.KEY_DOWN:  return popup.selectNext();
+            case KeyCodes.KEY_ENTER: return popup.acceptSelected();
+            case KeyCodes.KEY_TAB:   return popup.acceptSelected();
+            }
+         }
+
```
gtritchie, line 226: For parity with the old behavior, please also add support for home, end, PgUp and PgDn. This is what got removed, essentially:

```
                     case KeyCodes.KEY_PAGEDOWN:
                        moveSelectionFwd(5);
                        break;
                     case KeyCodes.KEY_PAGEUP:
                        moveSelectionBwd(5);
                        break;
                     case KeyCodes.KEY_HOME:
                        selectFirst();
                        break;
                     case KeyCodes.KEY_END:
                        selectLast();
                        break;

kevinushey:
Done. I also moved some of these helpers to the `ToolbarPopupMenu` class as it felt appropriate to have them there -- let me know if that makes sense.

Issue #6977, opened 2020-05-27
C++ auto-completion doesn't work with R 4.0.0


kevinushey:
### System details

    RStudio Edition : Desktop
    RStudio Version : v1.3.957
    OS Version      : macOS 10.15.4
    R Version       : 4.0.0

### Steps to reproduce the problem

Given this document, with `|` representing the cursor position:

```
#include <vector>

void test() {
  std::vector<int> x;
  x.|
}
```

Save this as a C++ file (e.g. `test.cpp`), and then try to press Tab to request completions.

### Describe the problem in detail

No completions are displayed.

### Describe the behavior you expected

Completions should be displayed.

### Notes

Output with `.rs.setClangDiagnostics(2)`:

```
CLANG INDEXING: /Users/kevinushey/scratch/test.cpp
CLANG REMOVE INDEX: /Users/kevinushey/scratch/test.cpp
COMPILATION ARGUMENTS:
[
    "-isystem",
    "/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include",
    "-I/Library/Developer/CommandLineTools/usr/lib/clang/11.0.0/include",
    "-I/Library/Developer/CommandLineTools/usr/bin/../include/c++/v1",
    "-I/Library/Developer/CommandLineTools/usr/lib/clang/11.0.0/include",
    "-I/Library/Developer/CommandLineTools/usr/include",
    "-I/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include",
    "-std=gnu++11",
    "-I/Library/Frameworks/R.framework/Resources/include",
    "-DNDEBUG",
    "-I/Users/kevinushey/Library/R/4.0/library/Rcpp/include",
    "-I/Users/kevinushey/scratch",
    "-I/usr/local/include",
    "-fPIC",
    "-include-pch",
    "/tmp/RtmpVEfzoq/rstudio/libclang/precompiled/Rcpp/67CB9813/1.0.4.6-x86_64-apple-darwin17.0-78286-11.0.0/Rcpp-std=gnu++11.pch",
    "-v"
]
  (Creating new index)
Apple clang version 11.0.0 (clang-1100.0.33.8)
Target: x86_64-apple-darwin19.4.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
ignoring nonexistent directory "/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/c++/v1"
[...]

PR #6978, opened 2020-05-27
don't include system include paths when using Apple clang

kevinushey:
Including the path for system headers was only necessary when using non-Apple clang, which was normally only true for R < 4.0.0. Now that R >= 4.0.0 is built using Apple clang, we can omit attempts to force the system include path.

Closes https://github.com/rstudio/rstudio/issues/6977.

Candidate (I think) for v1.3-patch backport.

### DIFF
--- a/src/cpp/session/modules/clang/CodeCompletion.cpp
+++ b/src/cpp/session/modules/clang/CodeCompletion.cpp
@@ -192,6 +192,38 @@ void discoverTranslationUnitIncludePaths(const FilePath& filePath,
    }
 }
 
+void prependSystemIncludePaths(std::vector<std::string>* pCompileArgs)
+{
+#ifdef __APPLE__
+   // newer versions of macOS (e.g. Mojave) no longer install system
+   // headers into /usr/include by default. attempt to find the headers
+   // made available as part of the default toolchain instead.
+   //
+   // this should only be necessary when using a custom version of
+   // LLVM / clang (as was done with R < 4.0.0); when using the default
+   // version of clang (Apple Clang) this is unnecessary.
+   if (!FilePath("/usr/include").exists())
+   {
+      // try multiple locations for path to appropriate system headers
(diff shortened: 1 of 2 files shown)


src/cpp/session/modules/clang/CodeCompletion.cpp
```diff
@@ -309,9 +341,13 @@ void discoverSystemIncludePaths(std::vector<std::string>* pIncludePaths)
       else
          includePathOutput = string_utils::trimWhitespace(result.stdOut);
    }
+   
    if (includePathOutput.empty())
       return;
    
+   // check for Apple clang
```
gtritchie, line 348: Should this be wrapped in `#ifdef __APPLE__` and possibly moved to the location later on where it is used (in the existing __APPLE__)?

src/cpp/session/modules/clang/CodeCompletion.cpp
```diff
@@ -339,6 +375,12 @@ void discoverSystemIncludePaths(std::vector<std::string>* pIncludePaths)
    for (std::size_t i = 0, n = includePaths.size(); i < n; ++i)
       includePaths[i] = string_utils::trimWhitespace(includePaths[i]);
    
+#ifdef __APPLE__
+   // prepend system include paths for non-Apple clang
+   if (!isAppleClang)
```
gtritchie, line 380: So the system include path we see at runtime is different if we are using R 4.0 versus < R 4.0? Want to make sure I understand (and that this works as expected if using pre R 4.0).

kevinushey:
Thinking about this a bit more, it seems like this might be the right change, for the wrong reason ...?

Ultimately, we're piping these arguments into a selected version of `libclang`; normally, this will be the version provided by Apple's command line tools, regardless of what toolchain is actually used to build R packages. So whether we include `-isystem` should be a property of which `libclang` implementation we're binding to.

Going to re-visit this to make sure I tie the loose ends up correctly...

Issue #6942, opened 2020-05-22
objects masking functions can inhibit completion


kevinushey:
### System details

    RStudio Edition : Desktop
    RStudio Version : v1.4.x
    OS Version      : macOS 10.15.4
    R Version       : 4.0.0

### Steps to reproduce the problem

With this document, with `|` reflecting the cursor position:

```
embed <- 42
embed(|
```

Execute the first line of code, and then press `<TAB>` with the cursor in the indicated position to request completions.

### Describe the problem in detail

The name `embed` is provided as an (auto-inserted) completion candidate, rather than the arguments to `stats::embed()`.

### Describe the behavior you expected

Non-function objects masking function objects shouldn't inhibit completions.

- [x] I have read the guide to submitting good bug reports at https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports . 
- [x] I have installed the latest version of RStudio and confirmed that the issue still persists.
- [x] If I am reporting a RStudio crash, I have included a diagnostics report. https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

dfalty:
Verified stats::embed() are now shown in this case on version 1.4.874.

PR #6980, opened 2020-05-27
ensure masked function objects are resolved correctly

kevinushey:
Closes https://github.com/rstudio/rstudio/issues/6942.


gtritchie:
Is this a new issue specific to R 4.0, or something that has been misbehaving all along that we're just handling now?

kevinushey:
This has been broken basically forever (I reproduced back in v0.99.903); but I figured it'd be nice to fix now.

PR #7008, opened 2020-05-30
return path to website index, even when output is not the index

kevinushey:
Closes https://github.com/rstudio/rstudio/issues/6945.

This issue occurs because of the way we end up munging the output file path here:

https://github.com/rstudio/rstudio/blob/275964ec5f5c338daf848168e161922fdd14f627/src/cpp/session/modules/rmarkdown/SessionRMarkdown.cpp#L1178-L1179

Because this is occurring, the right fix here is to point the output file path to the website root, e.g. as `file.path(websiteDir, "index.html")`. Which is awkward, since that path is later munged to point at the "correct" currently knitted file, which is _not_ at the root of the website.

tl;dr: There's something going on here that I don't quite understand, so this PR bears some extra scrutiny. There is probably a more appropriate fix.

### DIFF
--- a/src/cpp/session/modules/SessionRMarkdown.R
+++ b/src/cpp/session/modules/SessionRMarkdown.R
@@ -501,18 +501,24 @@
    normalizePath(subbin[[1]], mustWork = TRUE)
 })
 
-.rs.addFunction("bookdown.renderedOutputPath", function(outputPath)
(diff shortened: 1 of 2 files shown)


kevinushey:
Backported to v1.3: https://github.com/rstudio/rstudio/commit/914f19c385e0528c6bafe88459065dc43ffee368

PR #7020, opened 2020-06-02
don't increase size of mathjax popup font on windows

jjallaire:
This was likely needed for QtWebKit but seems unnecessary for QWebEngine

@kevinushey Two other small things I noticed which you may want to look at:

1. The background color of the popup is unthemed (it's always white). Not sure if this was intentional though?

2. There is now a rectangular border around the math which I didn't see previously. This to me doesn't look ideal.


kevinushey:
LGTM! This was indeed a workaround for QtWebKit on Windows that we shouldn't need now.

> The background color of the popup is unthemed (it's always white). Not sure if this was intentional though?

I think this was just an oversight -- it would be worth respecting the IDE's theme here.

> There is now a rectangular border around the math which I didn't see previously. This to me doesn't look ideal.

I agree with that as well. I think we can file these both as a separate issue.

PR #7027, opened 2020-06-02
Rainbow parentheses feature

### LINKED ISSUE #1888: Enhancement: Colour matched parenthesis/braces
In order to enhance the visual parsing of deeply nested parenthesis and braces it could be beneficial to colour code pairs in different colours (cycling through a palette of, say, 6 different colours).

This is akin to the colour highlighting done in formulas in Excel and is also something that has been implemented in the VIM plugin [*Rainbow Parentheses*](https://github.com/kien/rainbow_parentheses.vim) and in several Visual Studio Code plugin, e.g. [*Visafora*](https://marketplace.visualstudio.com/items?itemName=TomasRestrepo.Viasfora)

adamconroy:
Colorize matching parentheses, braces, and brackets in a series of colors. This is customizable in the theme CSS. It can also be turned on/off dynamically per document through the `Code -> Rainbow parentheses` option.

Example in light and dark themes:
Oth1C9K

Closes #1888


src/gwt/acesupport/acemode/rainbow_paren_highlight_rules.js
```diff
@@ -0,0 +1,71 @@
+/*
+ * rainbow_paren_highlight_rules.js
+ *
+ * Copyright (C) 2009-20 by RStudio, PBC
```
gtritchie, line 4: Switch to our new "current-year only" format.

src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/EditingPreferencesPane.java
```diff
@@ -77,6 +77,7 @@ public EditingPreferencesPane(UserPrefs prefs,
             "When enabled, the indentation for documents not part of an RStudio project " +
             "will be automatically detected."));
       editingPanel.add(checkboxPref("Insert matching parens/quotes", prefs_.insertMatching()));
+      editingPanel.add(checkboxPref("Rainbow parentheses", prefs_.rainbowParentheses()));
```
gtritchie, line 80: Does everything in this pane still fit (vertically)? Noticed this pane is getting very full. I ran into a problem adding a checkbox to the Code / Display pane in #7030 and had to remove some other spacing or it was slightly clipped at the bottom

gtritchie: Makes me think it might be time to support scrolling preferences panes...

src/cpp/session/include/session/prefs/UserPrefValues.hpp
```diff
@@ -773,6 +774,12 @@ class UserPrefValues: public Preferences
    bool highlightRFunctionCalls();
    core::Error setHighlightRFunctionCalls(bool val);
 
+   /**
+    * Whether to highlight parantheses (and other bracketed characters) in various colors
```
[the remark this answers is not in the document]
adamconroy: 👍

src/gwt/acesupport/acemode/c_cpp_highlight_rules.js
```diff
@@ -181,24 +182,16 @@ var c_cppHighlightRules = function() {
             token : "keyword.punctuation.operator",
             merge : false,
             regex : "\\?|\\:|\\,|\\;|\\.|\\\\"
-         }, {
```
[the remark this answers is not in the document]
adamconroy: Yeah, the rainbow rules don't work with `<>` and if you see the removed regex it apples to all of the characters so separated it as `<>` doesn't work the same in non-CPP languages. We could add an interface for custom regexes but I think not coloring `<>` differently in CPP looks better anyway as it's not something that you nest.

src/gwt/acesupport/acemode/c_cpp.js
```diff
@@ -157,7 +157,13 @@ oop.inherits(Mode, TextMode);
 
    this.inRLanguageMode = function(state)
    {
-      return state.match(/^r-/);
+     if (!state)
```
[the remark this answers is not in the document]
adamconroy: The type of `state` is variable. In fact it's actually normally an array in custom `onMatch` code (see: `yaml_highlight_rules.js`). Our code here was making an assumption about the type that is naive.

src/gwt/acesupport/acemode/r_highlight_rules.js
```diff
@@ -17,491 +17,473 @@
...
-
 var $colorFunctionCalls = false;
 
 define("mode/r_highlight_rules", ["require", "exports", "module"], function(require, exports, module)
 {
-   function include(rules) {
-      var result = new Array(rules.length);
-      for (var i = 0; i < rules.length; i++) {
-         result[i] = {include: rules[i]};
+  function include(rules) {
```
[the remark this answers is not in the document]
adamconroy: Sorry this was IntelliJ just doing its thing, I didn't even see this change. JS files should be changed to 2 spaces instead of 3 but I didn't intend for it here. For this case here you can just click the gear at the top and click "hide whitespace changes" or append `?w=1` to the URL for easier viewing.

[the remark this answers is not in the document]
gtritchie: Ah, I didn't know about that option to hide whitespace changes. Nifty!

kevinushey:
Also echoing Gary's question here:

> what happens if someone has previously imported a custom theme, then upgrades and turns on rainbow parentheses; just curious how that works in general when we add new css rules to the themes

I think we need to still include the default `paren.keyword.operator` type so that "default" highlight rules can be selected for parentheses, in case a custom theme doesn't define these colors yet. (But we want the rainbow rules to have a higher precedence than the "default" rules.)

adamconroy:
Unfortunately in the themes `paren.keyword.operator` is defined with `!important`. I can add important to all the styles that I just added but that's a pretty bad escalation that leads to more trouble down the line. I think if we can just communicate that custom themes will need to adapt as we add styles that will be fine, not everything can be backwards compatible. I can be thorough about this is the blog post about the feature. Rainbow parens defaults to off and still uses the old style so they don't lose any functionality in that case.

src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/EditingPreferencesPane.java
adamconroy: It does, there's still room for like...one more checkbox. It's definitely cramped though. We are in the situation where we want to start thinking about a different format for our options panel.

src/cpp/session/resources/schema/user-prefs-schema.json
```diff
@@ -515,6 +515,11 @@
             "title": "Highlight R function calls",
             "description": "Whether to highlight R function calls in the code editor."
         },
+        "rainbow_parentheses": {
+            "type": "boolean",
+            "default": false,
```
jmcphers, line 520: Add a `title` to this (and regenerate prefs) so it can be set from the Command palette.

jmcphers: (note you'll probably have to merge master into your branch to get this capability)

adamconroy:
That solution sounds good. I wish we could just remove `!important` from most/all of the styles but there are so many of them who knows what will happen, especially with custom themes.

jmcphers:
If you go the route of adding default styles, it looks like you could even forgo the addition of adding CSS rules to individual themes, since we just have one rainbow set for dark themes and another for light themes (easily scoped with `.editor_dark`). That would make the change much simpler w/o closing the door to custom rainbow colors in custom themes.

adamconroy:
Ok this should be good to go now!

adamconroy:
@kevinushey it this looks good to you it's ready to go!

jmcphers:
🎉

PR #7032, opened 2020-06-02
Refactor editor events

jcheng5:
Modify event system to allow events to (optionally) have additional
data attached. The definitions of the different types of events are
now in event-types.ts; see DispatchEvent for an example of an event
type that needs additional data (a Transaction object in this case).

### DIFF
--- a/src/gwt/panmirror/src/editor/src/api/event-types.ts
+++ b/src/gwt/panmirror/src/editor/src/api/event-types.ts
@@ -1,0 +1,26 @@
+/*
+ * event-types.ts
+ *
+ * Copyright (C) 2020 by RStudio, PBC
+ *
+ * Unless you have received this program directly from RStudio pursuant
+ * to the terms of a commercial license agreement with RStudio, then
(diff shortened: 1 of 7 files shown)


jcheng5:
Note that the compiler won't force you to provide a user data object when emitting an event, even if the EventType says one is expected; nor will it force your event handler to declare a parameter that receives it.

However, the compiler _will_ enforce that if you do provide a data object when you emit an event, or you do take a parameter in your event handler, that they must be of the correct types.

jcheng5:
@jjallaire The only effect on existing code is:

1. `emit`/`emitEvent` can now take an optional additional `detail` argument
2. handlers can now optionally have a corresponding `detail` parameter
3. the `EditorEvent` enum no longer exists--the event types have moved to `event-types.ts` where they're exported individually

jcheng5:
BTW there's nothing editor specific about this anymore; because we're not using an enum, the set of event types is open, not closed. I didn't rename the `EditorEvents` interface to something more generic but we totally could, if we ever need to use this for another context.

PR #7054, opened 2020-06-06
Correctly cancel the preview timer

dragonstyle:
When the component unmounts, cancel the preview timer. This fixes:

unnamed

which is caused by the timer calling back when the component has been closed (reproduces by mousing over the grid, then closing the popup immediately).

### DIFF
--- a/src/gwt/panmirror/src/editor/src/behaviors/insert_symbol/insert_symbol-grid.tsx
+++ b/src/gwt/panmirror/src/editor/src/behaviors/insert_symbol/insert_symbol-grid.tsx
@@ -96,7 +96,7 @@ const SymbolCharacterGrid = React.forwardRef<any, CharacterGridProps>((props, re
   React.useEffect(() => {
     if (mayShowPreview) {
       updatePreviewPosition();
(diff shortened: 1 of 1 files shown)


src/gwt/panmirror/src/editor/src/behaviors/insert_symbol/insert_symbol-grid.tsx
```diff
@@ -116,6 +116,7 @@ const SymbolCharacterGrid = React.forwardRef<any, CharacterGridProps>((props, re
       previewTimer.current = window.setTimeout(() => {
         setShowPreview(true);
       }, kWaitToShowPreviewMs);
+      return () => { window.clearTimeout(previewTimer.current); };
```
jjallaire, line 119: Will this work correctly even if `previewTimer.current` is unset/null?

dragonstyle: Yep - if it isn't initialized It will be undefined (clearTimeout is ok with this).

PR #7055, opened 2020-06-07
fix first ESC key press being ignored

### LINKED ISSUE #7045: First Esc keypress in IDE is not handled
### System details

    RStudio Edition : Server
    RStudio Version : Development (master)
    OS Version      :  OS X
    R Version       :  3.6

Load up the IDE and then do something which triggers a completion (e.g. type "plo" in the console or in a source editor). Once the completion popup is show, press Escape. The popup is not dismissed. Subsequent presses of Escape work as expected.

Note that this also occurs if I do something in visual mode which triggers a visual mode completion popup (e.g. type ":smi") so it's unlikely to be related to completion handling per se, but more likely some global keyboard shortcut handling code.

gtritchie:
- Fixes #7045
- Regression introduced via "add API for highlighting arbitrary UI in RStudio" #5867
- eager singleton causing a hidden MiniPopupPanel to be created at startup and that class had a quirk where it would intercept the first ESC keypress and swallow it
- fix is to only attached its key previewer when the panel has been displayed, and remove it when hidden (following existing pattern for drag handler)

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/widget/MiniPopupPanel.java
+++ b/src/gwt/src/org/rstudio/core/client/widget/MiniPopupPanel.java
@@ -86,6 +86,7 @@ public class MiniPopupPanel extends DecoratedPopupPanel
    public void show()
    {
       addDragHandler();
+      addEscHandler();
       super.show();
    }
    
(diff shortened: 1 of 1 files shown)


kevinushey:
LGTM -- thanks for running this down!

Sounds like this might be a candidate for v1.3-patch?

gtritchie:
> Sounds like this might be a candidate for v1.3-patch?

Agreed, will apply it there first and update NEWS.

gtritchie:
Will create new PR targeting 1.3.

Issue #7072, opened 2020-06-09
Save workspace fails if the imager library is loaded


justacec:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.2.1511
    OS Version      : Windows Server 2008
    R Version       : 3.5.1

### Steps to reproduce the problem

 1. load the imager library
 2. Click on the save worksapce icon
 3. Select file and click on "Ok"

### Describe the problem in detail

Cannot save the workspace because the imager library defines a function called `save.image`.

### Describe the behavior you expected

To save the workspace to the file

### Describe what happened instead

After clicking on "Ok" I received an error with the content:

```
Error in save.image(<FILE LOCATION>) :
  First argument should be an image
```

- [ ] I have read the guide to submitting good bug reports at https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports . 
- [ ] I have installed the latest version of RStudio and confirmed that the issue still persists.
- [ ] If I am reporting a RStudio crash, I have included a diagnostics report. https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

I have submitted PR #7073 to correct the issue.

PR #7073, opened 2020-06-09
Specify the `base` namespace for `save.image`

justacec:
This PR is intended to correct issue #7072.  This has not been tested and the namespace might need to be added to other places.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/environment/EnvironmentPresenter.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/environment/EnvironmentPresenter.java
@@ -438,7 +438,7 @@ public class EnvironmentPresenter extends BasePresenter
       consoleDispatcher_.saveFileAsThenExecuteCommand("Save Workspace As",
                                                       ".RData",
                                                       true,
-                                                      "save.image");
+                                                      "base::save.image");
    }
 
    void onLoadWorkspace()


kevinushey:
While this is the "correct" change, I would still prefer to use a plain `save.image()` in the case where we know the default version provided by `base` would be found on the search path. That is, I think we should only emit `base::save.image()` if we know `save.image()` is masked by something else on the search path.

justacec:
Humm.  Not sure I follow the motivation to only use the namespace when absolutely necessary.  It should always work without ambiguity and does not require additional code to check for function collisions.  

I guess what I am saying is that there I no apparent gain by only using the namespace when necessary.

Can you help me understand what I am missing here?  I know you have been doing this for a long time and I am confident you are holding this position.  I just don’t understand.

kevinushey:
Sorry for the late response.

My main concern is whether this could potentially be confusing to new users -- e.g. imagine following a tutorial that indicated pressing this button would emit `save.image()` to the console, but instead you saw `base::save.image()`. That dissonance could be frustrating to a new user; especially if they're not already familiar with the use of `::` to reference objects directly from a namespace. I realize this is a minor point but IMHO these little things matter.

justacec:
@kevinushey Ahh.  I had not thought about that angle (obviously :) ).  I forget that some people use the R commands behind the RStudio UI as a learning tool.  

While I see that it might be confusing, I would say that most users who would be using RStudio in this way would just use the full string `base::save.image` and not think twice about the colon namespace specifier.  That might actually be safer for them.

As for the name searching function.  I did some looking and did not find a convenient base R solution (I saw the `conflicted` library, but seems a bit overkill to use that for this purpose. No need for another dep).  I drafted the following short function:

```R
name_collision = function(function_name) {
  # Get an array of search items
  search_path = search()
  
  # Filter out unwanted items from the search path
  remove_items = c('tools:rstudio', 'package:base', 'Autoloads')
  remove_indexs = which(sapply(search_path, function(pn) { pn %in% remove_items}))
  if(length(remove_indexs) > 0) search_path = search_path[-remove_indexs]
  
  # Look through each of the loaded libraries to test for the presence of a
  # function of the same name as the provided parameter
  was_included = which(sapply(search_path, function(pn) {
    function_name %in% objects(pn)
  }))
  
  # Return TRUE if there is a function in the search path which has a function of the same name
  length(was_included) > 0
}
```

This can be tested with the following:

```R
library(imager)
name_collision('save.image')
name_collision('something.else')
```

I would go and say that I feel this function is expensive and certainly not something that I would suggest using for functions that are called repeatedly.  But, I don't think that the save workspace is called all that often.

So, I see a few courses of action here:
 1. Decide that emitting the namespace specifier is not detrimental to users in the long run
 2. Integrate the provided function with the `onSaveWorkspace()` function
 3. Integrate the provided function with the `rstudioapi` library for broader use and then use it in the `onSaveWorkspace()` function
 4. Integrate the provided function with the `rstudioapi` library for broader use and then use it in the `ExecuteCommand` family of functions.  In this situation, I would see that the `ExecuteCommand` functions would get a parameter (which defaults to 'base') to track the desired namespace.  Then the `ExecuteCommand` function would use the new provided function to ensure there are no name collisions.

I realize that was a lot of possible COA's.  But I was trying to think of all options.  

Thoughts?

justacec:
@kevinushey Poke.  ;)

PR #7097, opened 2020-06-10
require system libclang on macOS

kevinushey:
Closes https://github.com/rstudio/rstudio/issues/6977.

In theory, we allow the use of an optional installation of LLVM, but in practice that seems to never work (and orchestration of include paths etc. to make that work and compatible with what R does isn't worth the effort when users will normally have macOS command line tools installed anyhow).

Given that we're now always binding to the macOS version of libclang, setting up the extra system include paths seems to be no longer necessary.

Candidate for v1.3 backport.

### DIFF
--- a/src/cpp/core/libclang/LibClang.cpp
+++ b/src/cpp/core/libclang/LibClang.cpp
@@ -20,22 +20,27 @@
 
 #include <boost/regex.hpp>
 
-#include <core/Log.hpp>
 #include <shared_core/FilePath.hpp>
-#include <core/RegexUtils.hpp>
 #include <shared_core/SafeConvert.hpp>
 
+#include <core/Log.hpp>
+#include <core/RegexUtils.hpp>
+#include <core/system/Environment.hpp>
 #include <core/system/LibraryLoader.hpp>
 
-#define LOAD_CLANG_SYMBOL(name) \
-   error = core::system::loadSymbol(pLib_, "clang_" #name, (void**)&name); \
-   if (error) \
-   { \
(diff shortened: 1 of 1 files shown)


src/cpp/core/libclang/LibClang.cpp
```diff
@@ -20,22 +20,27 @@
...
+      Error error =                                                            \
+         core::system::loadSymbol(pLib_, "clang_" #name, (void**) &name);      \
+      if (error)                                                               \
+      {                                                                        \
+         Error unloadError = unload();                                         \
+         if (unloadError)                                                      \
+            LOG_ERROR(unloadError);                                            \
+         return error;                                                         \
+      }                                                                        \
+   } while (0);
```
gtritchie, line 43: Did something change to make it so we now require the do/while(0) idiom for the macro, or just doing as a best practice? Obviously fine to have it just curious.

kevinushey: Nothing changed here -- just clang-format + best practice for hygienic macros.

randy3k:
Does it fix https://github.com/rstudio/rstudio/issues/4685#issuecomment-633316768?

kevinushey:
@randy3k I think fixing this also requires the change in this PR:

https://github.com/rstudio/rstudio/pull/7227

This should also function as a workaround; I'm curious if it works in your case:

```
cat("SDKROOT = /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk", file = "~/.Renviron", append = TRUE)
```

Issue #7107, opened 2020-06-11
improve handling around editing of `$$` boundaries


kevinushey:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.4.458
    OS Version      : macOS 10.15.5
    R Version       : 4.0.1

### Steps to reproduce the problem

Within visual mode, create a Latex chunk with e.g.

`````
$$
\sum_{i=1}^nX_i
$$
`````

Then, try to modify the `$` chunk boundaries (e.g. by deleting a character). Then re-apply the `$`.

### Describe the problem in detail

The chunk boundaries are no longer highlighted as a single token. See e.g.

<img width="145" alt="Screen Shot 2020-06-11 at 1 34 52 PM" src="https://user-images.githubusercontent.com/1976582/84436486-62688200-abe8-11ea-97dc-f055757b4793.png">

You can also get in an awkward state where inner `$`s are tokenized as part of the Latex expression, instead of the bounds.

<img width="158" alt="Screen Shot 2020-06-11 at 1 36 24 PM" src="https://user-images.githubusercontent.com/1976582/84436610-99d72e80-abe8-11ea-99d3-c612aa09a3f0.png">

### Describe the behavior you expected

The chunk boundaries should be highlighted as a single token, and we should be robust to these edits of the chunk boundaries.

(Or perhaps these should get a similar "chunk" treatment as code chunks already do get?)

- [x] I have read the guide to submitting good bug reports at https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports . 
- [x] I have installed the latest version of RStudio and confirmed that the issue still persists.
- [x] If I am reporting a RStudio crash, I have included a diagnostics report. https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

jjallaire:
This gets into one of the toughest issues for the editor: we want to allow certain special character sequences to automatically create "richer" objects. For example, `$...$` should create inline math, `[@cite]` should create a citation, `\@ref(fig:foo)` should create a cross-reference. 

[...]

PR #7111, opened 2020-06-12
don't break display math marks on delimiter edit (wait until save)

jjallaire:
This is a potential fix for #7107 

The proposal is to not be so finicky about removing display math marks when the user edits them directly. This allows them to easily repair the mark after inadvertently removing a delimiter:

Jun-12-2020 07-36-26

Note that there is some feedback that the mark is broken b/c the equation preview now thinks that it's an inline mark.

If the mark continues to be broken when the user saves (or switches to source view) then it will in fact no longer be a display math equation. This is consistent w/ the behavior of source mode (where removing a dollar sign from a display math delimiter causes pandoc to no longer render it as math).


jcheng5:
Sorry I missed your review request--I wouldn't be able to fully understand the implications without a conversation with you anyway.

PR #7140, opened 2020-06-16
Target active column with commands

melissa-barca:
Commands utilizing the Source Panel and Editing Target are now only enabled on one column at a time. These commands most commonly appear under the `Code` menu list. Whenever the active editor changes, we iterate over the commands for each column and enable/disable them as needed.

This PR also fixes an issue where the panel state was not being saved properly between start ups. It also includes some code clean up.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/source/SourceColumnManager.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/SourceColumnManager.java
@@ -288,7 +288,12 @@ public class SourceColumnManager implements CommandPaletteEntrySource,
          }
          return;
       }
-      setActive(findByName(name));
+
+      // If we can't find the column, use the main column. This may happen on start up.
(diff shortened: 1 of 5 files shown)


src/gwt/src/org/rstudio/studio/client/workbench/views/source/SourceColumnManager.java
```diff
@@ -272,35 +280,44 @@ public void setActive(String name)
    {
       if (StringUtil.isNullOrEmpty(name))
       {
-         activeColumn_.setActiveEditor("");
-         activeColumn_ = null;
+         Debug.logWarning("Active column should be set to the main source window instead of null.");
```
gtritchie, line 283: Did you want to leave this in?

melissa-barca: Nope, thank you!

Issue #7089, opened 2020-06-10
Tab-trap in visual editor mode


gtritchie:
### System details

    RStudio Edition : Server
    RStudio Version : 1.4.453
    OS Version      : any
    R Version       : any

### Steps to reproduce the problem
- enable RMD visual editor and open an RMD file
- ensure keyboard focus in editor via <kbd>Ctrl+1</kbd>
- hit Tab key, nothing happens (maybe that's by-design, not sure)
- toggle the "Tab Key Always Moves Focus" accessibility option <kbd>Alt+Shift+[</kbd> on Windows/Linux, <kbd>Ctrl+Opt+[</kbd> on Mac
- try to Tab or Shift+Tab out of the editor

### Describe the problem in detail
- nothing happens

### Describe the behavior you expected
- keyboard focus should to go next/previous focusable item in DOM
- example use might be to get to the editor toolbar quickly via Shift+Tab to use it via keyboard
- note you can see the state of this setting via Help / Accessibility / Focus menu (checkbox)

- [x] I have read the guide to submitting good bug reports at https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports . 
- [x] I have installed the latest version of RStudio and confirmed that the issue still persists.
- [x] If I am reporting a RStudio crash, I have included a diagnostics report. https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

jjallaire:
Yes, tab is currently ignored by default (https://github.com/rstudio/rstudio/blob/master/src/gwt/panmirror/src/editor/src/api/basekeys.ts#L70-L72). We did this because when hosting codemirror instances in the editor the tab key would advance to the codemirror instance, which was quite surprising.

Perhaps though we should do this:

1) Remote any tabindex set by default for codemirror (as they are plenty of other ways to navigate to it w/ the keyboard and tab seems "wrong" to jump around to different regions in the editor).

2) Consider what tab should do by default and/or when the  "Tab Key Always Moves Focus" option is toggled. 

There are some other uses of tab in the editor (indenting list items, moving between table cells) and it does seem surprising when tab switches focus out of the editor. Would it be sufficient to only make it move move focus when that option is toggled?

Issue #5960, opened 2020-01-03
Toolbars don't follow WAI-ARIA keyboard navigation model


gtritchie:
### System details

    RStudio Edition : Desktop or Server
    RStudio Version : 1.3.x
    OS Version      : any
    R Version       : any

### Steps to reproduce the problem
- Get keyboard focus onto a toolbar, e.g. by using Help / Accessibility / Focus Main Toolbar
- See the first button has focus rectangle
- Navigate between toolbar buttons

### Describe the problem in detail
- Navigation between buttons is currently done with Tab / Shift+Tab; the arrow keys, home/end keys don't do anything

### Describe the behavior you expected
- Per WAI-ARIA recommendations, the arrows should move between toolbar buttons, and Tab should move to the next focusable item after the toolbar. Having to Tab through all the toolbar buttons is inefficient.
- https://www.w3.org/TR/wai-aria-practices/#toolbar

PR #7141, opened 2020-06-16
Enable entering/exiting focus using the tab key in visual mode

jjallaire:
WIP to address: https://github.com/rstudio/rstudio/issues/7089

First change removes our explicit ignore of the tab key and prevents codemirror (or it's embedded toolbar buttons, which have other keyboard gestures) from grabbing the tab key.

Focus key now works and moves focus out of the visual editor (when it's not activated for another purpose like creating a nested list)

@gtritchie A couple of questions:

1) What do you think of this as the default behavior? (should we be worried that users will inadvertently press the tab key and end up unexpectedly out of the editor -- i.e. should this only trigger when the "Tab Key Always Moves Focus" preference is active?

2) Does the preference being active mean that we should actually *suppress* the other meanings of tab (i.e. don't use it for list indenting or moving between table cells).


gtritchie:
Fixing this for 1.4 is low priority since we have many other ways to get focus elsewhere via keyboard-only (Ctrl+2 and so forth).

That said, behavior of Tab key moving focus out of the editor should only happen when this preference is active, and when active it must suppress all other meanings of the Tab key (i.e. move focus instead of indenting).

This is what happens in Console and regular editor. Idea came from similar setting in VSCode: https://vscode.readthedocs.io/en/latest/editor/accessibility/#tab-trapping

Your call if you want to mess with this for 1.4; I'd be totally fine punting it. It's arguably not necessary at all even by strict accessibility standards.

gtritchie:
FYI, _eventually_ we'd want the toolbar to be keyboard accessibility following standard pattern for toolbars (where the whole toolbar is a single tab-stop and arrow keys move between controls within the toolbar). We don't quite do this even for our existing toolbars, per: #5960

jjallaire:
Okay, this is now implemented as suggested.

Note that there is a catch-all keyboard handler for the entire insert menu (Cmd+/).

PR #7173, opened 2020-06-18
further refine Python execution semantics

kevinushey:
Closes https://github.com/rstudio/rstudio/issues/5904.


src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/EditingTargetCodeExecution.java
```diff
@@ -163,18 +163,7 @@ public void executeSelection(boolean consoleExecuteWhenNotFocused,
       // advance if there is no current selection
       if (noSelection && moveCursorAfter)
       {
-         if (DocumentMode.isCursorInPythonMode(docDisplay_))
```
kevinushey, line 166: This extra logic is no longer necessary as the REPL provided by `reticulate` is smart enough to handle this correctly now.

PR #7195, opened 2020-06-22
improve handling of chunk toolbars in modified documents

kevinushey:
This PR makes two changes:

1. When updating chunk toolbars, we use the most recent set of edits in the editor to decide which toolbars need to be updated. This is necessary since a document might not necessarily be mutated at the cursor position -- for example, when inserting text via the `rstudioapi` package.

2. We no longer defer adding line widgets according to the next Ace render. It looks like this was responsible for some of the challenging-to-reproduce behaviors around chunk toolbars occasionally failing to show up, or showing up "slowly", or even in some cases sticking around after a chunk was deleted.

Closes https://github.com/rstudio/rstudio/issues/7067.

Note for (2): this behavior was added a while ago, as part of efforts to eliminate "floating" Ace line widgets. However, I believe this particular behavior is essentially made superfluous by https://github.com/rstudio/rstudio/pull/795.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/EditingTargetCodeExecution.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/EditingTargetCodeExecution.java
@@ -310,7 +310,7 @@ public class EditingTargetCodeExecution
       // if we're in a chunk with in-line output, execute it there instead
       if (!onlyUseConsole && docDisplay_.showChunkOutputInline())
       {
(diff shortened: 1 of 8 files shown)


kevinushey:
QA testing notes: please try to test all of the corner cases around chunk creation, mutation, and so on -- especially around editing chunk boundaries. If we see anything like:

1. Chunk toolbars not showing up when they're supposed to,
2. Chunk toolbars showing up in the wrong location,
3. Other issues with inline outputs (e.g. plots, HTML widgets)

then we want to try and tackle those ASAP.

Issue #7223, opened 2020-06-25
rstudioapi::getActiveDocumentContext() doesn't return source editor context


kevinushey:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.4.514
    OS Version      : macOS 10.15.5
    R Version       : 4.0.2

### Steps to reproduce the problem

Create an R document, with the contents:

```
rstudioapi::getActiveEditorContext()
```

Focus the console with `Ctrl + 2`, then focus the Source editor again with `Ctrl + 1`, and execute the code.

### Describe the problem in detail

The console editor context is returned instead:

```
> rstudioapi::getActiveDocumentContext()
Document Context: 
- id:        '#console'
- path:      ''
- contents:  <1 rows>
Document Selection:
- [1, 1] -- [1, 1]: ''
```

### Describe the behavior you expected

`rstudioapi::getActiveDocumentContext()` should return the Source editor context, as the source editor is currently focused at time of execution.

- [x] I have read the guide to submitting good bug reports at https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports . 
- [x] I have installed the latest version of RStudio and confirmed that the issue still persists.
- [x] If I am reporting a RStudio crash, I have included a diagnostics report. https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
@melissa-barca it looks like it might be the result of a flipped conditional here:

https://github.com/rstudio/rstudio/blob/master/src/gwt/src/org/rstudio/studio/client/workbench/views/source/Source.java#L2629

Should that instead be `!columnManager_.hasActiveEditor()`?

kevinushey:
That does indeed fix the issue; I'm going to create a PR.

dfalty:
Verified fixed the context is correct switching back and forth from console and source editor on version 1.4.874.

PR #7225, opened 2020-06-25
fix active document check

kevinushey:
Closes https://github.com/rstudio/rstudio/issues/7223.


kevinushey:
@melissa-barca I'm going to merge now to get this fix into the daily builds but please feel free to follow up after if you think I missed anything.

melissa-barca:
This is the correct fix, thanks for handling it Kevin

PR #7234, opened 2020-06-26
use Double rather than Integer to avoid GWT cast issues

kevinushey:
I'm not exactly sure what the cause is, but it seems like GWT's JSNI gets upset if you try to use a GWT Integer to "box" a JavaScript number (even though this should've worked in the past).

Also switches to using GWT's type inference, rather than just plain Java objects.

Stumbled upon this while looking at https://github.com/rstudio/rstudio/issues/7066 (but doesn't fix that issue)


kevinushey:
I think so. It's worth checking any JSNI we have for usages of `Integer` vs. `Double`.

kevinushey:
Looks like DesktopFrame doesn't have any usages of Integer, so we're safe on that front.

PR #7249, opened 2020-06-29
convert some xterm.js support code from jsni to jsinterop

gtritchie:
- removed extra logging in terminal diagnostics dialog I had added while upgrading xterm.js last year (no real value to showing it)
- no longer checking XTermDimension for NaN (this was fixed in xterm.js a couple years ago; could track down the issue if anybody cares)
- still need to convert the `XTermNative` class but wanted to get first part in


adamconroy:
Yes! I'll convert older stuff to jsinterop as I touch it as well.

PR #7258, opened 2020-07-01
ensure F1 shows additional help in Python completion popup

### LINKED ISSUE #5982: No display of the help topic when pressing the F1 button over a Python symbol
Hello, 

I recently installed Reticulate v1.14 on RStudio v1.2.5033 with Python v3.8.0 (r-miniconda) and R v3.4.2. I run Mac OS 10.14.
Unfortunatetly, I get no display of the help topic when pressing the F1 button over a Python symbol, contrary to what is described [here](https://rstudio.github.io/reticulate/articles/rstudio_ide.html). 
I get the same issue whether coding in R scripts, in Python chunks of R Markdowns, or in Python scripts.

How could I solve this ?
Thanks for your help.

kevinushey:
Closes #5982.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/console/shell/assist/CompletionManagerBase.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/console/shell/assist/CompletionManagerBase.java
@@ -57,6 +57,11 @@ import com.google.inject.Inject;
 public abstract class CompletionManagerBase
       implements CompletionRequestContext.Host
 {
+   public abstract void goToHelp();
+   public abstract void goToDefinition();
+   public abstract void showAdditionalHelp(QualifiedName completion);
+   public abstract boolean getCompletions(String line, CompletionRequestContext context);
+   
    public interface Callback
    {
       public void onToken(TokenIterator it, Token token);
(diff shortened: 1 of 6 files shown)


src/gwt/src/org/rstudio/studio/client/workbench/views/console/shell/assist/CompletionManagerBase.java
```diff
@@ -57,6 +57,11 @@
 public abstract class CompletionManagerBase
       implements CompletionRequestContext.Host
 {
+   public abstract void goToHelp();
```
gtritchie, line 60: Given the empty implementation of these methods in various subclasses (with the NYI comments), would it be more appropriate to have empty implementations here in CompletionManagerBase instead of making these abstract?

kevinushey: My main reason to keep these abstract is just that it's "easier" to implement a single class method, versus overriding an existing class method. E.g. as an author of a subclass, I might need to consider:

1. Which methods should I override to achieve some functionality? (need to potentially look at all overridable methods in base class, versus trusting the base class to "tell" you which methods are abstract and so require implementations)
2. Should I call the superclass method? If so, do I call it first, or last, or something else? (sometimes it's obvious, sometimes it's not; either way, you were forced to think about it)

Those considerations are moot for pure abstract methods, and so I think are then easier to reason about. There's a little bit more code, but IDEs auto-generate that for you so it's not as much of a concern.

kevinushey:
Merged in https://github.com/rstudio/rstudio/commit/de6844cde915ce142f8c8a92edb07085be25df81.

PR #7272, opened 2020-07-05
source history navigation using mouse forward/back buttons

jjallaire:
This PR implements support for binding source navigation (back/forward) to the back and forward mouse buttons.

This appears to only work in the browser though, in Qt-land the events don't seem to make their way to JavaScript. I verified that the events *do* land in the MacEventFilter we have setup to fix some other Qt issues, but I'm not contemporary enough in Qt/QtWebEngine to sort out the right way to forward these on to GWT. @gtritchie Could you help out with this? 

@melissa-barca Part of this is testing whether the mouse is currently over the source pane. I used `Source.asWidget()` to sort this out, but I'm not sure whether the source columns work makes this calculation a bit more complex? Let me know if you think the current implementation will work okay.

cc @yihui


gtritchie:
@jjallaire Will look at this sometime this week. Bit of a backlog at the moment (I'm on support rotation, too).

melissa-barca:
@jjallaire Today `asWidget()` only gives you the display for the active column so this won't work as expected across columns. The columns are added directly to `MainSplitPanel` so there isn't a widget that exclusively holds them. There is a method `ArrayList<Widget> getWidgets(boolean excludeMain)` in `SourceColumnManager` that returns all of the source columns, but I also see the benefit of adding something to make them more contained.

jjallaire:
@melissa-barca I think that the `getWidgets()` method on `SourceColumnManager` will be all we need for this case. I'll update  the PR w/ a call to that method.

melissa-barca:
@jjallaire looks good!

yihui:
I'm just confirming that this works beautifully in the daily build! I'm really excited about the support for these mouse buttons. Now my cursor can be anywhere in the editor. The Forward/Back buttons on the toolbar have probably been the buttons that I click the most frequently in the editor (since the keyboard shortcuts Cmd+F9/F10 are difficult to do).

Not sure if I'm asking for too much, but I'll also be excited if the buttons could be supported in the Viewer. Currently I have to right-click to go Forward/Back:

image

This doesn't bother me as much as the previous issue, though.

Thanks!

Issue #7273, opened 2020-07-05
Wrong indents in document outline when function doc contains macro


petrbouchal:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.4.538
    OS Version      : MacOS 10.15.5
    R Version       : 4.0.0

### Steps to reproduce the problem

place the following in an R script

```r
#' An x function
#'
fnx <- function(x) {
  return(x)
}

#' A y function
#'
#' \lifecycle{deprecated}
fny <- function(y) {
  return(y)
}
```

### Describe the problem in detail

When the Roxygen text in an R file contains a `\macro{command}`, the indentation in the document outline is wrong - the outline looks as if the below function were nested inside the above function. When the macro line is removed, the indentation in the outline is correct. When the bracketed part in the macro is removed, the indentation is incorrect in a different way (different number of tabs/spaces.) 

### Describe the behavior you expected

Functions existing at the same level in a script should appear at the same level in the outline.

- [x] I have read the guide to submitting good bug reports at https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports . 
- [x] I have installed the latest version of RStudio and confirmed that the issue still persists.
- [x] If I am reporting a RStudio crash, I have included a diagnostics report. https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
I'm not able to reproduce:

<img width="293" alt="Screen Shot 2020-07-06 at 9 38 42 AM" src="https://user-images.githubusercontent.com/1976582/86617697-7e9fda80-bf6c-11ea-877f-516fea508d7a.png">

... unless I turn on rainbow parentheses:

<img width="319" alt="Screen Shot 2020-07-06 at 9 40 17 AM" src="https://user-images.githubusercontent.com/1976582/86617820-b4dd5a00-bf6c-11ea-97c4-04123c43567b.png">

Are rainbow parentheses enabled in your case?

petrbouchal:
> Are rainbow parentheses enabled in your case?

Yes, they are - sorry, hadn't tested that eventuality.

kevinushey:
> Yes, they are - sorry, hadn't tested that eventuality.

No worries; there are a lot of knobs and dials in the IDE and it's rarely clear which knobs might affect other parts of the IDE!

kevinushey:
[...]

PR #7281, opened 2020-07-06
fix issue with Ace '#tmp' state from tokenizer

kevinushey:
With the Ace tokenizer, rules that make use of line-specific metadata (via the "stack" parameters) can cause Ace to emit a temporary state `#tmp`:

https://github.com/ajaxorg/ace/blob/2ccda8c1ba350415de0e949abd50888abb6645fe/lib/ace/tokenizer.js#L354-L357

Normally, that temporary state is only resolved later, in response to `getLineTokens()`:

https://github.com/ajaxorg/ace/blob/2ccda8c1ba350415de0e949abd50888abb6645fe/lib/ace/tokenizer.js#L229-L237

It's not completely clear to me why this `#tmp` state is used; the only documentation I can see is the associated commit:

https://github.com/ajaxorg/ace/commit/7a3a84c8583d49f6acf5948ad6a386737fa7ee17

It _seems_ like it's some implicit way of indicating that state management is being done through the `stack` variable, as opposed to other components of the rule.

Regardless, the fix here is to skip `#tmp` when we observe a stack of states.

Closes https://github.com/rstudio/rstudio/issues/7273.

### DIFF
--- a/src/gwt/acesupport/acemode/r_code_model.js
+++ b/src/gwt/acesupport/acemode/r_code_model.js
@@ -813,7 +813,7 @@ var RCodeModel = function(session, tokenizer,
          position = iterator.getCurrentTokenPosition();
 
          // Skip roxygen comments.
-         var state = this.$session.getState(position.row);
+         var state = Utils.getPrimaryState(this.$session, position.row);
          if (state === "rd-start") {
             iterator.moveToEndOfRow();
             continue;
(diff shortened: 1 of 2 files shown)


adamconroy:
Thanks for looking at this Kevin. What an odd bug. The state variable being so flippity floppity is 🤦

PR #7302, opened 2020-07-08
navigate to bookdown xrefs in omni-search

kevinushey:
This PR makes it possible to navigate to sections, figures and tables within a bookdown project via the fuzzy finder.

Screen Shot 2020-07-07 at 4 08 32 PM

Some questions worth answering before final merge:

1. What should we display for sections / figures / tables? The caption? The ID? Some combination?
2. Is it worth refining the search for the associated title string in source-only XRef navigation?


src/cpp/session/modules/SessionCodeSearch.cpp
```diff
@@ -1543,6 +1552,100 @@ SourceItem fromCppDefinition(const clang::CppDefinition& cppDefinition)
...
+               "title", title);
+      
+      if (error)
+      {
+         LOG_ERROR(error);
+         continue;
+      }
+      
+      // figure out appropriate source item type
+      SourceItem::Type sourceType = SourceItem::None;
```
kevinushey, line 1609: @jjallaire: are there any other reference types I should handle here?

src/cpp/session/modules/SessionCodeSearch.cpp
```diff
@@ -1399,13 +1401,16 @@ class SourceItem
...
-      Enum = 4,
+      None      = 0,
+      Function  = 1,
+      Method    = 2,
+      Class     = 3,
+      Enum      = 4,
       EnumValue = 5,
-      Namespace = 6
+      Namespace = 6,
+      Section   = 7,
```
jjallaire, line 1411: There's also equations and theorems (although the latter obscure enough that we probably don't need navigation for it).

jjallaire, line 1411: Do we want to distinguish between heading levels here? (Hadley had asked for distinguishing headings levels in xref completions b/c it's helpful context). Paul could do separate icons for H1-H6.

src/cpp/session/modules/SessionCodeSearch.cpp
jjallaire: All the types are here (save for `eq` which I appear to have forgotten!): https://github.com/rstudio/rstudio/blob/master/src/gwt/panmirror/src/editor/src/api/xref.ts#L99

src/cpp/session/modules/SessionCodeSearch.cpp
```diff
@@ -1543,6 +1552,99 @@ SourceItem fromCppDefinition(const clang::CppDefinition& cppDefinition)
...
+         sourceType = SourceItem::Table;
+      }
+      else if (type == "h1" || type == "h2" || type == "h3" ||
+               type == "h4" || type == "h5" || type == "h6")
+      {
+         sourceType = SourceItem::Section;
+      }
+      
+      // form appropriate text for display
+      std::string displayText = title;
```
jjallaire, line 1625: I'm torn as to whether we should be showing IDs here, at least for figures, tables, and equations -- as that's the 'identifier' that's being searched for (i.e. for functions we don't show the roxygen title if available we show the name of the function). I've also seen that figure and table captions can get quite long (even multiple sentences) and I think the id would be a better mental "bookmark' in the case cases.

For section I can see the argument that we should use the actual title (since the id is typically algorithmically derived from the title).

src/gwt/src/org/rstudio/studio/client/workbench/codesearch/CodeSearchOracle.java
```diff
@@ -192,51 +192,41 @@ public void requestSuggestions(final Request request,
...
-            }
-            
-         }
-      }
+      // nothing to do for non-file targets
+      if (!codeSearchSuggestion.isFileTarget())
+         return target;
+      
+      // nothing to do if the query doesn't have ':'
+      int colonIndex = query.indexOf(":");
```
jjallaire, line 210: Perhaps it's accounted for elsewhere, but do we want to have `fig:`, `tab:`, etc. filter on xref type?

src/cpp/session/modules/SessionCodeSearch.cpp
hadley: Even just h1 vs h2 vs everything else would be useful

src/cpp/session/modules/SessionCodeSearch.cpp
hadley: How about both, like `title (#id)` ? Or is that likely to be too long?

src/gwt/src/org/rstudio/studio/client/workbench/codesearch/CodeSearchOracle.java
hadley: That would be nice but it's not very discoverable, so hard to know if people would ever use it.

src/cpp/session/modules/SessionCodeSearch.cpp
jjallaire: I think it will be too long (a section title would typically be ~ 10-30 characters which is already over the effective limit for the UI).

src/gwt/src/org/rstudio/studio/client/workbench/codesearch/CodeSearchOracle.java
jjallaire: That's true. Probably not worth it then.

src/cpp/session/modules/SessionCodeSearch.cpp
kevinushey: If we think displaying the ID is necessary we could truncate the section title and always append the ID.

hadley: I'd say try it without, and then if it's frustrating we figure out how to add it.

jjallaire: I think no ID for sections/headers to start with is fine, but we definitely should use the IDs for figures, tables, and equations. In those cases that's the actual "symbol" they are navigating to (it's the chunk label).

kevinushey: Sorry, to re-iterate: do we want to be able to search based on the contents of the caption (e.g. for a figure), or only based on the figure ID?

jjallaire: I think we should search based on contents, but we should show the id in the list

kevinushey:
Going to merge this now (to avoid future merge conflicts); we can tweak the display post-merge.

PR #7304, opened 2020-07-08
Fix toolbar buttons being set to visible when they shouldn't be

melissa-barca:
Enable Changed and Visibility Changed events on `CommandSourceColumnToolbarButton` were being passed to the `AppCommand` handlers rather than their own handler causing buttons to appear when they shouldn't. Their visibility is now properly assigned.


src/gwt/src/org/rstudio/studio/client/workbench/views/source/events/PopoutDocEvent.java
```diff
@@ -45,7 +45,6 @@ public PopoutDocEvent(PopoutDocInitiatedEvent originator,
       originator_ = originator;
       sourcePosition_ = sourcePosition;
       column_ = column;
-      Debug.logToConsole("new PopoutDocEvent for: " + originator_.getDocId());
```
gtritchie, line 48: Can the `Debug` import go away now?

melissa-barca: Yep, thank you!

PR #7305, opened 2020-07-08
Assign overflow source tabs to correct columns

melissa-barca:
Selecting the overflow chevron previously only displayed the tabs of the active column, now the tabs for that specific column are displayed and selecting a column to switch to works as expected.

Previously, the drop down menu learned the state of the source tabs from an event triggered by a source column. Now that we have multiple source columns, the menu needs more knowledge about where the change is coming from. I've removed `DocTabsChangedEvent` so that the column communicates directly to its menu. The Source Column's chevron button's click handler needed to be updated to notify the `SourceColumnManager` to set the active column on selection.


src/gwt/src/org/rstudio/studio/client/workbench/views/source/SourcePane.java
```diff
@@ -92,7 +93,11 @@ public void onClose(CloseEvent<PopupPanel> popupPanelCloseEvent)
       chevron_ = new Image(new ImageResource2x(ThemeResources.INSTANCE.chevron2x()));
       chevron_.setAltText("Switch to tab");
       chevron_.getElement().getStyle().setCursor(Cursor.POINTER);
-      chevron_.addClickHandler(event -> tabOverflowPopup_.showRelativeTo(chevron_));
+      chevron_.addClickHandler(event -> {
```
gtritchie, line 96: nit, our coding standard is to put opening brace on next line

Issue #7117, opened 2020-06-12
RStudio Desktop slow to quit on Windows


gtritchie:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.3.1019
    OS Version      : Windows-10 2004
    R Version       : 4.0.0

### Steps to reproduce the problem
- Start IDE on Windows, no source files open, no terminals, no project, etc.
- <kbd>Alt+F4</kbd> to quit (or via File / Quit, etc.)

### Describe the problem in detail
- Consistently takes about 3 seconds for RStudio to actually exit

### Describe the behavior you expected
- Exiting should be relatively instantaneous

- [x] I have read the guide to submitting good bug reports at https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports . 
- [x] I have installed the latest version of RStudio and confirmed that the issue still persists.
- [x] If I am reporting a RStudio crash, I have included a diagnostics report. https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
Thank you for filing this -- I've noticed this as well and it's definitely annoying.

gtritchie:
Marking as v1.3-giant-goldenrod. I haven't investigated cause yet, but suspect this will be a major issue for some workflows.

For example, the same delay seems to be there if you are working on a package and using "Install and Restart". There's a delay during the "Restarting R" phase.

gtritchie:
Caused by this change: #5232 (ensure file monitor stopped for all forms of quit).

I built v1.3-giant-goldenrod with `system::file_monitor::stop();` commented out in SessionMain.cpp, and the delay on closing went away on Windows.

Note that this delay doesn't always repro if you start and exit the IDE quickly. You need to leave the IDE running for some period of time (just a few seconds seems to be enough, usually), for the delay on closing to manifest.

Assigning to @kevinushey to investigate further.

gtritchie:
Closing, fix was merged into 1.3-goldenrod several days ago.

R-Hannibal:
Verified Fixed on: RStudio Version - 1.3.1065, R Version - 4.0.2, Windows 10 x64 - Desktop 

The close time (whether through manually closing window, Alt + F4, File > Quit Session, Session > Quit Session) for the IDE on Windows is now less than a second.

PR #7314, opened 2020-07-09
stop file scanner on exit

kevinushey:
When RStudio is initialized with a project, it registers a file monitor for that project. This entails scanning the files used within that project recursively, which can be slow for projects with many files (e.g. RStudio itself).

When we try to stop the file monitor, we were previously forced to wait for an in-flight registration to complete (or for 3 seconds, after which we just detached the thread and continued trying to exit).

With this change, when we stop the file monitor, we also tell the file scanner to stop (hence "finishing" the registration of the file monitor early), which then gives us a chance to properly close and clean up the file monitor.

As a bonus ... this PR also adds `logToNotepad()`, a simple helper function for logging text to a Notepad window, for cases on Windows where you just need some debug output to show up somewhere easily accessible.

Ultimately, I think this change is probably too risky for v1.3 so I'd say we just take this for v1.4.

Closes https://github.com/rstudio/rstudio/issues/7117.

### DIFF
--- a/src/cpp/core/Debug.cpp
+++ b/src/cpp/core/Debug.cpp
@@ -1,0 +1,55 @@
+/*
+ * Debug.cpp
+ *
+ * Copyright (C) 2020 by RStudio, PBC
+ *
+ * Unless you have received this program directly from RStudio pursuant
+ * to the terms of a commercial license agreement with RStudio, then
+ * this program is licensed to you under the terms of version 3 of the
+ * GNU Affero General Public License. This program is distributed WITHOUT
+ * ANY EXPRESS OR IMPLIED WARRANTY, INCLUDING THOSE OF NON-INFRINGEMENT,
+ * MERCHANTABILITY OR FITNESS FOR A PARTICULAR PURPOSE. Please refer to the
+ * AGPL (http://www.gnu.org/licenses/agpl-3.0.txt) for more details.
+ *
+ */
+
+#include <core/Debug.hpp>
+
+namespace rstudio {
+namespace core {
+namespace debug {
+
+void logToNotepad(const char* fmt, ...)
+{
+#ifdef _WIN32
+
+   // try to find Notepad window
+   HWND hNotepad = ::FindWindow("Notepad", NULL);
+   if (hNotepad == NULL)
+      return;
+
+   // try to find Notepad's edit surface
+   HWND hEdit = ::FindWindowEx(hNotepad, NULL, "EDIT", NULL);
+   if (hEdit == NULL)
+      return;
+
+   // generate log message
+   char buffer[512];
+   va_list args;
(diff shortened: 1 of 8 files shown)


src/cpp/core/Debug.cpp
```diff
@@ -0,0 +1,55 @@
+/*
+ * Debug.cpp
```
kfeinauer, line 2: This needs some scoping to be Windows-only, right?

kfeinauer: (I see the ifdefs, but should we not make the entire function Windows-only?)

kevinushey: You're right -- I'll make that change.

gtritchie:
I haven't looked at the change itself yet, but no hope of a minimal fix we'd be comfortable taking for 1.3? As a user I’d be super-irritated by this, it even impacts doing things like install & restart for working on packages, etc. and having a program take 3 extra seconds to exit, every time, feels broken.

If it was happening on the Mac version I bet we'd fix it. :-)

kevinushey:
> What happens if we just back out the change that introduced this (#5232)?

Then I think we'd reintroduce this:

https://github.com/rstudio/rstudio/issues/5222

which was fairly ugly as well.

Looking again, it looks like the problem here just might be that we're not using boost::thread the "right" way on Windows. This is where we try to stop the thread by interrupting and detaching it:

https://github.com/rstudio/rstudio/blob/de0b176a6afd48792c74109032f13a15586fbdf8/src/cpp/core/system/file_monitor/FileMonitor.cpp#L630-L641

Note the caveats for calling `interrupt()`, from https://www.boost.org/doc/libs/1_54_0/doc/html/thread/thread_management.html#thread.thread_management.tutorial.interruption:

> If *this refers to a thread of execution, request that the thread will be interrupted the next time it enters one of the predefined interruption points with interruption enabled, or if it is currently blocked in a call to one of the predefined interruption points with interruption enabled .

However, the monitor thread on Windows loops with `::SleepEx()` here:

https://github.com/rstudio/rstudio/blob/de0b176a6afd48792c74109032f13a15586fbdf8/src/cpp/core/system/file_monitor/Win32FileMonitor.cpp#L597-L612

So the right fix might be to either:

1. Use Boost's APIs for sleeping the thread;
2. Use `interruption_point()` to explicitly allow the thread to check for an interrupt.

I'll trial one of these -- if it works, I think that's safe enough for v1.3.

gtritchie:
> Then I think we'd reintroduce this:
> 
> #5222
> 
> which was fairly ugly as well.

Quite the tangled web.

kevinushey:
Alternate approach in https://github.com/rstudio/rstudio/pull/7343.

PR #7322, opened 2020-07-10
Fix up toolbar buttons and commands

melissa-barca:
This PR adds some functionality #7305 missed when managing command icons across source columns. Specifically, it allows a button to not be synced with its command and the `SourceColumnManager`. 

`SourceAppCommand` now saves whether the button(s) for the command are visible or not so that the buttons can be properly set on attachment if the command was managed before the event handlers were added.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/source/SourceAppCommand.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/SourceAppCommand.java
@@ -41,6 +41,7 @@ public class SourceAppCommand
                                               String buttonTitle,
                                               ImageResourceProvider imageResourceProvider,
                                               ClickHandler clickHandler,
+                                              /* false when button is managed elsewhere */
                                               boolean synced)
       {
          super(buttonLabel, buttonTitle, imageResourceProvider, clickHandler);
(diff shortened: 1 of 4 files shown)


src/gwt/src/org/rstudio/studio/client/workbench/views/source/SourceAppCommand.java
```diff
@@ -40,24 +40,32 @@ public CommandSourceColumnToolbarButton(SourceAppCommand sourceCmd,
                                               String buttonLabel,
                                               String buttonTitle,
                                               ImageResourceProvider imageResourceProvider,
-                                              ClickHandler clickHandler)
+                                              ClickHandler clickHandler,
+                                              boolean synced)
```
gtritchie, line 44: I'd love a comment briefly describing what "synced" means, even just a single-line `// foo bar` style here.

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTargetWidget.java
```diff
@@ -351,7 +351,7 @@ private Toolbar createToolbar(TextFileType fileType)
       toolbar.addLeftWidget(previewHTMLButton_ =
          mgr.getSourceCommand(commands_.previewHTML(), column_).createToolbarButton());
       knitDocumentButton_ =
-         mgr.getSourceCommand(commands_.knitDocument(), column_).createToolbarButton();
+         mgr.getSourceCommand(commands_.knitDocument(), column_).createToolbarButton(false);
```
gtritchie, line 354: I'm not a big fan of boolean arguments like these (though I know we do it a lot, and I've written a bunch of code that does it, too). So this is absolutely just me thinking out loud, no real need to change anything.

Alternatives include creating a separate method, e.g.

`.createUnsyncedToolbarButton()`

Commenting the argument:

`.createToolbarButton(false/*synced*/)`

Pass an enum, instead (`public enum SyncType { UNSYNCED, SYNCED }`).

melissa-barca: I like `.createUnsyncedToolbarButton` and switched to that.

PR #7334, opened 2020-07-13
onEdit method for EditingTarget

jjallaire:
I am hoping to create a mechanism for delaying potentially expensive `TextEditingTarget` operations until the tab is actually selected for editing (as opposed to selected during the startup sequence). We currently have an `onActivate()` method, but it is always called for each tab during source pane restoration, so isn't suitable as a trigger for deferred work.

This PR introduces a new `onEdit()` method to `TextEditingTarget`. This method is called from within `SourceColumn.onSelection()` but *only if* all of the documents have already been restored (i.e. startup is complete). This means that for the *actually* active tab `onEdit()` will not yet have been called after restoring all of the tabs, so a call to `onEdit()` is also made in the `onBeforeShow()` method  of `SourceColumn`. `onEdit()` is also called unqualified from `SourceColumn.onActivate()`. 

Note that it's entirely fine if `onEdit()` is sometimes called twice when once would have been suitable, so long as these calls occur after restoration of all of the tabs (i.e. not during startup), as the `onEdit()` method already has to defend itself against doing unnecessary work (as the user can switch tabs at any time and that should never be expensive save for the initial activation for editing). It is however *not at all fine* if `onEdit()` isn't called when it should have been, as this might cause the editor tab to essentially not load at all (since the "deferred" expensive operations will never occur). Just wanted to put that lens on the code review as we should err on the side of caution w/r/t when `onEdit()` is invoked.

My current intended use of this is to defer rendering of visual mode until the user actually opens the tab for editing. The main benefit of this is noticeably improved IDE startup time in the case of 10+ visual mode tabs open. It's also somewhat motivated by the observation that Ace editor instances currently take much longer than CodeMirror instances to initialize, so the scenario of 10+ visual mode tabs has gotten materially worse. I'm hopeful that we can resolve the Ace initialization issues, but this mechanism is here in case we can't.

I could also imagine using this mechanism for other tabs that are expensive to load (not sure if we have any of these, I do recall that profiler tabs could take a while depending on the amount of profile data being viewed).


jjallaire:
In taking a closer look at using the `onEdit()` callback to implement deferred loading of visual mode, I don't think this will be practical b/c there are now too many points in the code where having the logical and physical state of visual mode activation diverge will cause incorrect behavior (and/or incredibly complex workarounds to get correct behavior). @jmcphers I think we will just need to figure out how to bring down the Ace initialization costs :-)

Closing this PR for now but we can revive if and when we have another need for it.

jjallaire:
@melissa-barca and @jmcphers I am re-opening this PR just because I think it would be useful to have a reliable `onEdit()` callback in the system (as `onActivate()` is currently called during document tab restoration so isn't a good indicator of user activation for editing). As I said in the comments above, I am not likely to use this for visual mode but it would be nice to have it just in case I need to.

jjallaire:
In doing some more testing, this PR currently doesn't reliably deliver onEdit (at least for initial load). So I'll close again for now.

PR #7374, opened 2020-07-17
resolve shortcuts in file dialogs on Windows

### LINKED ISSUE #2476: RStudio expands symlinks in preferences, breaking on upgrades/changes to target (e.g. git)
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.1.423
    OS Version      : OS X 10.13.3
    R Version       : 3.4.4

Change Preferences->Git/SVN->Git executable to /usr/local/bin/git, which is a symlink to a homebrew-installed git at /usr/local/Cellar/git/[version]/bin/git

When the change is made, RStudio expands the symlink to the real absolute path. Because this expanded path contains a version number, updates to new versions of git via brew break the RStudio<->git configuration. One is required to go into RStudio's preferences after every upgrade and fix the path.

Entering a symlink to the git executable (or anything) should not expand, so that RStudio is tolerant of changes to the symlink's target. Symlinks are use by homebrew to abstract away the underlying versioned installation and RStudio should use this abstraction. While this is an issue raised for homebrew in particular, it's probably a good idea to respect symlinks literally in any case.

### LINKED ISSUE #7327: Can't follow windows shortcut (.lnk)
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.3.959 
    OS Version      : Win 10 v2004 (19041.329)
    R Version       :  4.0.2

- Create a windows shortcut (.lnk file) in the default working directory. The shortcut points to a folder containing R projects.
- In RStudio, File | Open Project. R opens a file browser window in the default working directory.
- Double click on the shortcut
- Error appears
image

### Describe the problem in detail/the behaviour you expected
I use a number shortcuts in my default working directory and Documents folder to redirect apps opening/saving files to various locations.  In other windows apps, double-clicking on these links when opening files moves the file browser window directly to the folder in the shortcut.
RStudio is not recognising the shortcut as a link, but is instead trying to open the .lnk file as a project and failing.

kevinushey:
Closes #7327.

Note: this also re-fixes #2476. It looks like this may have regressed as part of a Qt update? The documentation in https://doc.qt.io/qt-5/qfilesystemmodel.html#resolveSymlinks-prop states, surprisingly:

> This property holds whether the directory model should resolve symbolic links
> 
> This is only relevant on Windows.
> 
> By default, this property is true.

Whereas, no such caveat holds for the `DontResolveSymlinks` flag (https://doc.qt.io/qt-5/qfilesystemmodel.html#Option-enum).

Regardless, it seems like the right solution is to disable resolution of symlinks on non-Windows via the file dialog option.

### DIFF
--- a/src/cpp/desktop/DesktopGwtCallback.cpp
+++ b/src/cpp/desktop/DesktopGwtCallback.cpp
@@ -260,13 +260,15 @@ QString GwtCallback::getOpenFileName(const QString& caption,
 
    dialog.setFileMode(mode);
    dialog.setLabelText(QFileDialog::Accept, label);
-#if QT_VERSION < QT_VERSION_CHECK(5, 13, 0)
-   dialog.setResolveSymlinks(false);
-#else
-   dialog.setOption(QFileDialog::DontResolveSymlinks, true);
-#endif
    dialog.setWindowModality(Qt::WindowModal);
 
+   // don't resolve links on non-Windows platforms
+   // https://github.com/rstudio/rstudio/issues/2476
+   // https://github.com/rstudio/rstudio/issues/7327
+#ifndef _WIN32
+   dialog.setOption(QFileDialog::DontResolveSymlinks, true);
+#endif
+   
    QString result;
    if (dialog.exec() == QDialog::Accepted)
       result = dialog.selectedFiles().value(0);


kevinushey:
Indeed it is -- looks like it was available even back in Qt 4.8!

Issue #7265, opened 2020-07-02
learnr is loaded on startup if installed


kevinushey:
### System details

    RStudio Edition : Desktop
    RStudio Version : v1.3.1053, v1.4.x
    OS Version      : All
    R Version       : All

### Steps to reproduce the problem

Install the `learnr` package, e.g. via:

```
install.packages("learnr")
```

Then, restart the R session, and look at what packages are loaded via `loadedNamespaces()`.

### Describe the problem in detail

`learnr` and its dependencies are loaded on startup:

```
> loadedNamespaces()
 [1] "compiler"  "graphics"  "htmltools" "tools"     "parallel"  "utils"     "yaml"      "grDevices"
 [9] "stats"     "datasets"  "rmarkdown" "knitr"     "methods"   "xfun"      "digest"    "packrat"  
[17] "rlang"     "base"      "evaluate" 
```

### Describe the behavior you expected

We should avoid aggressively loading packages on startup when possible; especially so on Windows where updates of installed packages can fail.

### Notes

We can put the following hook in `~/.Rprofile` to get a traceback from when `rmarkdown` is loaded:

```
  setHook(
    packageEvent("rmarkdown", "onLoad"),
    function(...) print(rlang::trace_back())
  )
```

This shows:

```
 █
  1. └─(function (pkgRoot) ...
  2.   └─base::lapply(tutorialDirs, .rs.tutorial.findTutorialsImpl)
  3.     └─FUN(X[[i]], ...)
  4.       └─rmarkdown:::parse_yaml_front_matter
  5.         ├─base::get(name, envir = asNamespace(pkg), inherits = FALSE)
  6.         └─base::asNamespace(pkg)
  7.           └─base::getNamespace(ns)
  8.             └─base::loadNamespace(name)
  9.               └─base:::runUserHook(package, pkgpath)
 10.                 ├─base::try(fun(pkgname, pkgpath))
 11.                 │ └─base::tryCatch(...)
 12.                 │   └─base:::tryCatchList(expr, classes, parentenv, handlers)
 13.                 │     └─base:::tryCatchOne(expr, names, parentenv, handlers[[1L]])
 14.                 │       └─base:::doTryCatch(return(expr), name, parentenv, handler)
 15.                 └─fun(pkgname, pkgpath)
```

And so we can ascertain that `rmarkdown` is being loaded and used while attempting to enumerate available tutorials. We should see if we can avoid needing to load `rmarkdown` to enumerate tutorials here.

---

[...]

PR #7378, opened 2020-07-17
avoid loading rmarkdown, packrat packages when unnecessary

kevinushey:
Closes https://github.com/rstudio/rstudio/issues/7265.

### DIFF
--- a/src/cpp/session/modules/SessionTutorial.R
+++ b/src/cpp/session/modules/SessionTutorial.R
@@ -249,7 +249,9 @@
    contents <- readLines(tutorialFile, encoding = "UTF-8", warn = FALSE)
    
    # find YAML header (if any). note that we avoid using rmarkdown here
-   # just because we want to avoid 
+   # just because we want to avoid loading the package and all its
+   # recursive dependencies, as this could introduce issues for users
+   # who need to update packages
    reYamlSeparator <- "^\\s*(?:---|[.][.][.])\\s*$"
    yamlSeparators <- grep(reYamlSeparator, contents)
    yamlRange <- seq.int(
--- a/src/cpp/session/modules/SessionTutorial.R
+++ b/src/cpp/session/modules/SessionTutorial.R
@@ -247,7 +247,18 @@
    
    tutorialFile <- tutorialFiles[[1]]
    contents <- readLines(tutorialFile, encoding = "UTF-8", warn = FALSE)
-   yaml <- rmarkdown:::parse_yaml_front_matter(contents)
+   
+   # find YAML header (if any). note that we avoid using rmarkdown here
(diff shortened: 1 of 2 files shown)


src/cpp/session/modules/SessionTutorial.R
```diff
@@ -247,7 +247,18 @@
    
    tutorialFile <- tutorialFiles[[1]]
    contents <- readLines(tutorialFile, encoding = "UTF-8", warn = FALSE)
-   yaml <- rmarkdown:::parse_yaml_front_matter(contents)
+   
+   # find YAML header (if any). note that we avoid using rmarkdown here
+   # just because we want to avoid
```
gtritchie, line 252: Incomplete comment?

src/cpp/session/modules/SessionTutorial.R
```diff
@@ -247,7 +247,18 @@
    
    tutorialFile <- tutorialFiles[[1]]
    contents <- readLines(tutorialFile, encoding = "UTF-8", warn = FALSE)
-   yaml <- rmarkdown:::parse_yaml_front_matter(contents)
+   
+   # find YAML header (if any). note that we avoid using rmarkdown here
+   # just because we want to avoid 
+   reYamlSeparator <- "^\\s*(?:---|[.][.][.])\\s*$"
```
gtritchie, line 253: Is this a common regex, i.e. we use it elsewhere for this purpose? Don't know that we have any way to share such things in our embedded R code, mainly curious if this is tried and tested vs. something new.

src/cpp/session/modules/SessionTutorial.R
kevinushey: Oops. Looks like I lost my train of thought!

src/cpp/session/modules/SessionTutorial.R
kevinushey: We use it for highlight rules here:

https://github.com/rstudio/rstudio/blob/3e709a0b099ee3894d4b2cd85424ae2eaf5c3e54/src/gwt/acesupport/acemode/rmarkdown_highlight_rules.js#L156-L164

And we have a similar version here:

https://github.com/rstudio/rstudio/blob/3e709a0b099ee3894d4b2cd85424ae2eaf5c3e54/src/cpp/core/YamlUtil.cpp#L32-L36

The Pandoc [spec](https://pandoc.org/MANUAL.html#extension-yaml_metadata_block) says:

> A YAML metadata block is a valid YAML object, delimited by a line of three hyphens (---) at the top and a line of three hyphens (---) or three dots (...) at the bottom.

So, strictly speaking, the second version is wrong, and so is the version in this PR -- although in practice, users almost always use `---`.

PR #7393, opened 2020-07-20
utility scripts for cleaning up RStudio state (desktop and server)

gtritchie:
Adding utility scripts I use all the time for blowing away RStudio state on Posix or Windows.

Do you think these are appropriate to live in the repo? I have been keeping them in my own repo and cloning from there, but these seem useful to have widely available.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -1,0 +1,5 @@
+# Utility Scripts
+
+This folder contains scripts intended for use by engineers working on the RStudio IDE.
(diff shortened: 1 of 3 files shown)


kevinushey:
LGTM!

I think it's worth including these sorts of scripts in the repository, as long as they're used and maintained by at least one IDE team member, and have the potential to be useful by others. I think both these scripts fit the bill.

PR #7401, opened 2020-07-21
Properly resize buffer during group enumeration

kfeinauer:
Fix bug where user group enumeration was not properly checking return code of `getgrouplist` to resize the buffer when needed. 

Fixes https://github.com/rstudio/rstudio-pro/issues/1796

### DIFF
--- a/src/cpp/core/system/PosixGroup.cpp
+++ b/src/cpp/core/system/PosixGroup.cpp
@@ -109,10 +109,6 @@ Error userGroups(const std::string& userName, std::vector<Group>* pGroups)
    if (error)
       return error;
 
-   // get the groups for the user - we start with 100 groups which should be enough for most cases
-   // if it is not, resize the buffer with the correct amount of groups and try again
-   int numGroups = 100;
-
    // define a different gid type if we are on Mac vs Linux
    // BSD expects int values, but Linux expects unsigned ints
 #ifndef __APPLE__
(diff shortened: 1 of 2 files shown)


src/cpp/core/system/PosixGroup.cpp
```diff
@@ -121,13 +117,12 @@ Error userGroups(const std::string& userName, std::vector<Group>* pGroups)
    typedef int GIDTYPE;
 #endif
 
-   boost::shared_ptr<GIDTYPE> pGids(new GIDTYPE[100]);
-   while (!getgrouplist(userName.c_str(), user.getGroupId(), pGids.get(), &numGroups))
+   // get the groups for the user - we start with 100 groups which should be enough for most cases
+   // if it is not, resize the buffer with the correct amount of groups and try again
+   int numGroups = 100;
+   boost::shared_ptr<GIDTYPE> pGids(new GIDTYPE[numGroups]);
+   while (getgrouplist(userName.c_str(), user.getGroupId(), pGids.get(), &numGroups) == -1)
```
kevinushey, line 124: FWIW I think we can remove the while loop here and just have two explicit calls to `getgrouplist()`. If I'm reading the documentation correctly, if the function returns -1, then the real required number of groups will be written to `&numGroups` and we can just make one more call with that value. (It also gives us a chance to sanity-check the second call to `getgrouplist()`, which should in theory never fail)

kfeinauer: Yes, I thought of doing that as well but figured it was ok to leave as is. I'll update with this change.

kfeinauer: I am torn though because this introduces the possibility of the following sequence:

1. Initial call to `getgrouplist`
2. User is added to another group via some other mechanism
3. Second call to `getgrouplist` still fails, since the count parameter is wrong

This is not likely but it is one of the reasons I decided to keep the `while` loop - what do you think?

kevinushey: Ah, that's a very good point. I think that's motivation enough to keep things as-is.

kfeinauer:
> How would you feel about moving this functionality into User.h/cpp? Perhaps in a method like `Error getSupplementalGroups(std::vector<Group>& out_groups)`? In that case you could leave this function and have it call `User::getUserFromIdentiifier` followed by `User::getSupplementalGroups` to avoid having to make significant changes within the IDE.

That would also mean I would need to move the `Group` struct to `shared_core` as well, and lots of places in the code depend on that. I think that move is reasonable, but in a different PR (I want to scope this just to a necessary bug fix).

Issue #7201, opened 2020-06-23
Variable explorer for Python when using in Rstudio


ntluong95:
I really apprciated that recently Rstudio had a big update that now I can see the Python object right into Environment pane. Howerver, I would like to request that we need to preview the structure of dataframe like in R (please see the picture). If we can do this, I think more Python developer will try to using R as an IDE for Python. Thank you!
![image](https://user-images.githubusercontent.com/63569637/85437509-e0e3fd00-b5b4-11ea-9f90-faacd5f932a0.png)

R-Hannibal:
@ntluong95, thank you so much for finding and reporting this way to make the RStudio + Python experience better. 

We have a fix in place in our latest Daily build of the upcoming 1.4 release of RStudio (available [HERE](https://dailies.rstudio.com/)). If you get the chance will you check it out and confirm that the issue has been addressed?

melissa-barca:
Verified in 1.4.1013.

PR #7404, opened 2020-07-21
allow expansion of some Python objects in Environment pane

kevinushey:
This PR implements support for the view of expanded contents for pandas DataFrames in the Environment Pane. This PR also bundles some miscellaneous clean-up / restructuring of some of the R code used to build components of the Environment Pane.

Closes https://github.com/rstudio/rstudio/issues/7201.

Screen Shot 2020-07-21 at 11 21 44 AM


.gitignore
```diff
@@ -37,3 +37,5 @@ GPATH
 GRTAGS
 GSYMS
 GTAGS
+TODO-INTERNAL
```
kevinushey, line 40: I've adopted this convention for tracking TODO items on separate feature branches. Let me know if this feels appropriate.

src/gwt/src/org/rstudio/studio/client/workbench/views/environment/view/EnvironmentObjectList.css
```diff
@@ -77,7 +77,7 @@ td.resizeCol {
 td.valueCol
 {
    border: 1px solid #f0f0f0;
-   white-space: nowrap;
+   white-space: pre;
```
kevinushey, line 80: This is necessary so that output from `pandas.to_string()` is formatted correctly (as it uses whitespace for alignment)

PR #7410, opened 2020-07-21
Add shortcut to create new source column

melissa-barca:
Adds `Ctrl + F7` as a shortcut to create a new source column.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml
@@ -270,6 +270,7 @@ well as menu structures (for main menu and popup menus).
          
          <menu label="P_anes">
             <cmd refid="layoutEndZoom"/>
+            <cmd refid="newSourceColumn"/>
             <separator/>
             <cmd refid="layoutConsoleOnLeft"/>
             <cmd refid="layoutConsoleOnRight"/>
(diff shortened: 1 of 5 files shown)


src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml
```diff
@@ -1497,7 +1499,12 @@ well as menu structures (for main menu and popup menus).
         label="Show All Panes"
         menuLabel="_Show All Panes"
         windowMode="main"/>
-        
+
+   <cmd id="newSourceColumn"
```
gtritchie, line 1503: For easier discovery, recommend adding the command to the (increasingly crowded) View menu, probably under the Pane submenu.

src/gwt/src/org/rstudio/studio/client/workbench/ui/PaneManager.java
```diff
@@ -579,6 +579,14 @@ public void onFocusCenterSeparator()
       panel_.focusSplitter();
    }
 
+   @Handler
```
gtritchie, line 582: Should we show an alert if command is invoked and there are already MAX_COLUMN_COUNT columns?

Or, could instead disable the command when there are that many columns, and it will thus be disabled in the menu and the command-palette. Might be simpler just to keep enabled and show an alert.

src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml
melissa-barca: It is added to the View menu unless I'm misunderstanding:

image

gtritchie: Oops, my bad, sorry. Missed that.

src/gwt/src/org/rstudio/studio/client/workbench/ui/PaneManager.java
melissa-barca: Ok, I updated to show an alert when MAX_COLUMN_COUNTS is reached and when the `Allow Source Columns` pref is set to false.

PR #7413, opened 2020-07-21
Fix log file rotation failure crash

kfeinauer:
Fixes a bug where failing to rotate a log file could cause a crash. We had plumbed up the `rotateLogFile` to return a `bool` indicating whether or not it was safe to actually log after the call, but were never checking the return code.

Fixes https://github.com/rstudio/rstudio-pro/issues/1779

### DIFF
--- a/src/cpp/shared_core/FileLogDestination.cpp
+++ b/src/cpp/shared_core/FileLogDestination.cpp
@@ -155,11 +155,20 @@ struct FileLogDestination::Impl
                return false;
 
             // Now re-open the log file.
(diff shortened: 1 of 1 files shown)


kfeinauer:
Logging to stderr when logs cannot be written properly can actually be dangerous - if stderr is pointing to a buffered stream and no one is consuming while we are producing, we can effectively hang the process. 

In the future we could consider using native APIs to signal an error condition (writing to the Windows event log, writing to sylog) but that is definitely outside of the scope of this fix.

PR #7415, opened 2020-07-21
Fix move console left/right command

### LINKED ISSUE #7405: View / Panes / Console on Left or Right not working properly (1.4)
### System details

    RStudio Version : 1.4.618
    OS Version      : any
    R Version       : any

The same issue was logged and fixed in 1.3 (tracked by #7246). That fix was not sufficient for 1.4 due (presumably) to the work in progress on source columns, so opening this to track for 1.4.

- startup IDE in a clean default state, with no files open in source pane (for best chance of repro, delete all previous settings including Qt-related settings if using Desktop)
- in View / Panes menu, note checkmark on "Console on Left" menu item
- select View / Panes / Console on Right

- an editor document is opened (which is consistent with earlier versions) but the Console remains in lower-left corner
- View / Panes / Console on Right is now checked in the menu despite Console not being on right
- Problem persists even after quit and restart (i.e. Console in lower-left, menu says it's on right)

- Untitled editor document created
- Console panel switches places with Environment panel
- Menu checkbox matches actual visual state including after a restart

gtritchie:
Without this, the View / Panes / Console on Left (or Right) commands didn't change anything until you reloaded the UI.

Fixes #7405

(Was also dependent on fix made in v1.3, #7399)


src/gwt/src/org/rstudio/studio/client/workbench/ui/PaneManager.java
```diff
@@ -403,8 +403,8 @@ else if (numDocs > 0 && window.getState() == WindowState.HIDE)
 
       eventBus.addHandler(UpdateTabPanelsEvent.TYPE, event ->
       {
-         left_.replaceWindows(panes_.get(0), panes_.get(1));
-         right_.replaceWindows(panes_.get(2), panes_.get(3));
+         left_.replaceWindows(panes_.get(0), panes_.get(1), DualWindowLayoutPanel.ReplaceMode.IF_DIFFERENT);
+         right_.replaceWindows(panes_.get(2), panes_.get(3), DualWindowLayoutPanel.ReplaceMode.IF_DIFFERENT);
```
melissa-barca, line 407: Looking into it, we don't seem to ever fire `UpdateTabPanelsEvent` so I think we should get rid of this handler and the event.

src/gwt/src/org/rstudio/core/client/layout/DualWindowLayoutPanel.java
```diff
@@ -510,10 +510,16 @@ private void unhookEvents()
...
+      IF_DIFFERENT,
+      ALWAYS
+   }
    public void replaceWindows(LogicalWindow windowA,
-                              LogicalWindow windowB)
+                              LogicalWindow windowB,
+                              ReplaceMode replaceMode)
    {
-      if (windowA_ != windowA && windowB_ != windowB)
+      if (replaceMode == ReplaceMode.ALWAYS || (windowA_ != windowA && windowB_ != windowB))
```
melissa-barca, line 522: I know you didn't write the original code, but do you know why the comparisons between windows A & B are compared with `&&` as opposed to `||`? It seems like as long as one is different we want to do the replace.

gtritchie: You added that comparison in this commit:

11367597c893c239b6c0acde6a4e9d56e49e9905

melissa-barca: Wow ok, I think this is a bug and should be changed.

src/gwt/src/org/rstudio/studio/client/workbench/ui/PaneManager.java
gtritchie: Done!

src/gwt/src/org/rstudio/core/client/layout/DualWindowLayoutPanel.java
gtritchie: Per discussion I have removed that comparison completely. Thanks!

PR #7416, opened 2020-07-22
Persist last active tab on start up

melissa-barca:
This PR fixes an issue where we were attempting to select the last active source tab before we restored the source documents. It also persists the last active column's name so we can select the active tab on the correct column.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/source/SourceColumnManager.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/SourceColumnManager.java
@@ -115,10 +115,7 @@ public class SourceColumnManager implements CommandPaletteEntrySource,
       }
 
       public final native String getActiveColumn() /*-{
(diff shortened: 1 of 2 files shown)


src/gwt/src/org/rstudio/studio/client/workbench/views/source/SourceColumnManager.java
```diff
@@ -113,6 +114,13 @@ protected State()
          return JsUtil.toStringArray(getNamesNative());
       }
 
+      public final native String getActiveColumn() /*-{
+         if (this.activeColumn)
```
kevinushey, line 118: IMHO, a more idiomatic way of specifying this is JavaScript would be `this.activeColumn || ""`.

melissa-barca: Much nicer, thank you!!

PR #7434, opened 2020-07-23
provide hook for simulating RPC delays

kevinushey:
This PR makes it possible to force a delay when executing RPCs, via:

```
.Call("rs_setRpcDelay", <delay>)
```

The primary intention is to help debugging bugs that appear timing related -- for example, I've seen issues where the autocompletion results are not properly inserted in high-latency environments, and having this makes it easier to reproduce and understand the issue locally.

### DIFF
--- a/src/cpp/session/SessionModuleContext.cpp
+++ b/src/cpp/session/SessionModuleContext.cpp
@@ -69,6 +69,7 @@
 
 #include <session/http/SessionRequest.hpp>
 
+#include "SessionRpc.hpp"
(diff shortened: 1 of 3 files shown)


jmcphers:
FWIW you can also introduce latency with Chrome's devtools:

image

kevinushey:
Thanks -- I wasn't aware of that. This PR could still be useful in cases where we don't have Chrome Devtools (or for desktop, where this didn't seem to make a difference?)

PR #7456, opened 2020-07-27
Script to fetch focus-visible polyfill from npm and copy into source tree

gtritchie:
- When I added option to disable this polyfill for those who want to see less focus rectangles (#7351) I moved the files into a client bundle, forgetting that they are also used by auth page
- To fix, created a script for fetching and copying as we do for several other such dependencies; clunky, but at least documents where it comes from and where it goes.
- This also takes the most recent version, containing a couple of very minor fixes


ricardofandrade:
A few things are still not clear to me:
- Do we still use `focus-visible` anywhere other than the sign in page?
- If not, do we still need to use it during sign-in?

My expectation was that we would end up removing this dependency.

gtritchie:
> * Do we still use `focus-visible` anywhere other than the sign in page?

Yes, we use it on sign-in page.

It is also widely used in the main IDE UI, but is now optional based on a user preference and loaded on-demand from a GWT ClientBundle. This mechanism is not available on the pre-GWT sign-in page, so this change just puts the files back (as flat files) where they were before, while also continuing to bundle them for GWT.

Issue #7136, opened 2020-06-15
Installing package from Enterprise RStudio Package Manager


ddsjoberg:
Thank you for your great work on renv!

My hospital uses RStudio Package Manager (RSPM) as a repo for our in-house packages.  I've added its location to RStudio so I can use `install.packages()` to install our packages from RSPM.

![image](https://user-images.githubusercontent.com/26774684/84675068-bda4b800-aef9-11ea-9b22-6cc179f77012.png)

When I am in a renv project, I can no longer use `install.packages()` to install the package.  `renv::install()` also does not work.  Is this intended, or am I missing something?  Is there a setting to be able to install from RSPM when in an renv project?  There is a note on the packages options page, but I don't understand its meaning
![image](https://user-images.githubusercontent.com/26774684/84675281-065c7100-aefa-11ea-9457-9a447c56f517.png)

Thank you!

kevinushey:
Note to self: repository information needs to be passed along when `renv::init()` is called here.

https://github.com/rstudio/rstudio/blob/1889a1bce5322cb541b88adfe56ecdefa1b43a20/src/cpp/session/modules/SessionRenv.R#L18-L39

wangyuchen:
Same issue here on RStudio Server pro 1.3.1056-1.  The `/etc/rstudio/rsession.conf` file has `r-cran-repos` option pointing to an internal RSPM, but when creating a new project with the option "use renv with this project", the lock file created is this:

```
{
  "R": {
    "Version": "3.6.3",
    "Repositories": [
      {
        "Name": "CRAN",
        "URL": "https://cloud.r-project.org"
      }
    ]
  },
  "Packages": {
    "renv": {
      "Package": "renv",
      "Version": "0.11.0",
      "Source": "Repository",
      "Repository": "RSPM",
      "Hash": "1c3ef87cbb81c23ac96797781ec7aecc"
    }
  }
}
```

It works fine if creating a project without the "use renv with this project" option checked, and then call `renv::init()` in the project.

R-Hannibal:
@ddsjoberg, thanks again for finding and reporting this issue as well as the kind words 😊. 

We have a fix in place in our latest Daily build of the upcoming 1.4 release of RStudio (available [HERE](https://dailies.rstudio.com/)). If you get the chance will you let us know if the issue is resolved for you in that build?

rich-rstudio:
verified on Version 1.4.915

PR #7460, opened 2020-07-27
propagate repos when running renv::init()

kevinushey:
This PR ensures that the current R repositories are propagated to the child process used to run `renv::init()`. This is necessary so that the right repositories are encoded in the newly-created lockfile.

This PR also adds a small bit of infrastructure to run R functions in child processes.

Closes https://github.com/rstudio/rstudio/issues/7136.


kevinushey:
To test:

1. Set some custom R repositories; e.g. with:

```
options(repos = c(RSPM = "https://packagemanager.rstudio.com/all/latest"))
```

2. Create a new project with `renv` integration enabled.

Screen Shot 2020-10-09 at 10 22 20 AM

3. Check the result of `getOption("repos")` to validate that the previously-set repositories have been inherited by this project.

Screen Shot 2020-10-09 at 10 22 49 AM

Issue #7471, opened 2020-07-28
Can't open new files


melissa-barca:
### System details

    RStudio Edition : Both
    RStudio Version : 1.4+
    OS Version      : n/a
    R Version       : n/a

### Steps to reproduce the problem

Unknown

### Describe the problem in detail

It is possible to get into a state where the user can't open any new files in the IDE and clicking things (e.g. the Files panes) acts as a no-op. Nothing relevant appears in the logs (system or JS console). 

This is caused by a _phantom_ source column taking all the new tabs. Clearing the state removes the column and fixes the issue. 

### Describe the behavior you expected
You should always be able to open new files. All source columns should be visible. Source columns should be removed from the client state once they are closed. 

- [x] I have read the guide to submitting good bug reports at https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports . 
- [x] I have installed the latest version of RStudio and confirmed that the issue still persists.
- [ ] If I am reporting a RStudio crash, I have included a diagnostics report. https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report
- [ ] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

rich-rstudio:
verified on Version 1.4.750-1

Issue #7371, opened 2020-07-17
Error running plot command on RStudio Server


rich-rstudio:
### System details

    RStudio Edition : Server OS and Pro
    RStudio Version : Version 1.4.607

    OS Version      : openSUSE 15
    R Version       : 3.6

### Steps to reproduce the problem
1. run a plot command in console
ex. 
> x=1
> y=1
> plot(x,y)

### Describe the problem in detail
Error in RStudioGD() : 
  Shadow graphics device error: r error 4 (R code execution error)

logs:
17 Jul 2020 16:11:49 [rserver-monitor] ERROR 17 Jul 2020 16:11:49 [rsession-rich] ERROR r error 4 (R code execution error) [errormsg: Error in grDevices:::png("/tmp/Rtmpnc7ULN/2117a44c30f04a9b919989ec4936fef1.png",  : 
  X11 is not available
]; OCCURRED AT rstudio::core::Error rstudio::r::exec::evaluateString(const string&, SEXPREC**, rstudio::r::sexp::Protect*) src/cpp/r/RExec.cpp:327; LOGGED FROM: DevDesc* rstudio::r::session::graphics::handler::shadow::{anonymous}::shadowDevDesc(pDevDesc) src/cpp/r/session/graphics/RShadowPngGraphicsHandler.cpp:217

might be related to https://github.com/rstudio/rstudio/issues/2214?

### Describe the behavior you expected
user is able to run plot command without errors

- [ x] I have read the guide to submitting good bug reports at https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports . 
- [x ] I have installed the latest version of RStudio and confirmed that the issue still persists.
- [x ] If I am reporting a RStudio crash, I have included a diagnostics report. https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report
- [x ] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
What is the output of `capabilities()` on that system? I wonder if R wasn't compiled with the necessary support for X11 / Cairo graphics, or if other required system libraries are missing.

How was R installed on this system?

rich-rstudio:
capabilities()
       jpeg         png        tiff       tcltk         X11        aqua    http/ftp     sockets      libxml 
      FALSE       FALSE       FALSE       FALSE       FALSE       FALSE        TRUE        TRUE        TRUE 
[...]

PR #7476, opened 2020-07-29
Prevent access to invisible columns

melissa-barca:
Begins to address #7471 

This is a temporary failsafe to prevent the IDE from getting into a state where it tries to access a "phantom column" that is not visible to the user. This needs to be addressed more completely so that we don't have these types on invisible columns in the first place but this will prevent the IDE from essentially shutting down. 
`public SourceColumn getActive()` already existed within SourceColumnManager to return the active column or set the main column to active if there was no active column. I updated this to check the widget's presence as well and expanded the places where it's used.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/source/SourceColumnManager.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/SourceColumnManager.java
@@ -459,7 +459,9 @@ public class SourceColumnManager implements CommandPaletteEntrySource,
    // where it is better for the column to be the main column than null.
    public SourceColumn getActive()
    {
-      if (activeColumn_ != null)
+      if (activeColumn_ != null &&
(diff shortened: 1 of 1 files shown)


kevinushey:
Also, just to make sure I understand -- this is a bandaid, but ultimately the goal is to figure out how we might get into this state (ie: prevent the creation of phantom source columns entirely) but that's for a separate PR?

melissa-barca:
`getActive()` sets `activeColumn_` so once we've called it once we can be sure it has a valid option and access the variable directly.
Your understanding is correct - this is just a fix that will hopefully prevent the IDE from getting into the state you saw yesterday which required clearing your state, but ultimately we need to understand where that column is coming from and why it's getting created or not getting removed. I don't plan on closing #7371 until that's done.

kevinushey:
Thanks! This looks good to me then -- thanks for taking a look so quickly.

PR #7483, opened 2020-07-30
Do not reset active column before start up

melissa-barca:
Adjustment to #7476 which prevented access to phantom columns by refusing to access unattached columns. There are a few times we want to access columns _before_ they are attached, this PR adds an additional check before changing the active column to see if the main column itself has been attached. If not, it proceeds with the requested active column.


melissa-barca:
@kevinushey I went ahead and merged this because my fix from earlier was a little off. Let me know if you have any additional thoughts. Thanks!

PR #7494, opened 2020-07-30
set defaults in JSNI accessors for prefs

kevinushey:
This PR is a bandaid against unexpected `null` preference values. If the associated preference is not defined or does not exist for some reason, we instead return the default value defined for the preference (if any), or the "zero" value for the associated type.


kevinushey:
> This shouldn't be necessary because there's a default prefs layer that contains these values, but as prefs are delivered on session init this is a reasonable failsafe.

This was mainly motivated by an issue that @adamconroy bumped into where the `additional_source_columns` preference was left undefined and so caused issues when attempting to access + use that preference. That implies there could be an issue deeper in the preferences system, but ultimately this is just a bandaid against those sorts of issues.

> Does/should this guard against the objects themselves being null (vs missing properties)?

It doesn't -- I've made a change just now to verify that `this` itself is not `null` (or `undefined` or some other "falsy" thing).

PR #7497, opened 2020-07-30
allow customization of default Python interpreter used by reticulate

kevinushey:
This PR brings in the first bit of work for configuring the version of Python used by RStudio.

The Global Options pane now has a tab for Python, where Python-related settings can be configured (currently, this is just the default interpreter):

Screen Shot 2020-07-30 at 1 42 26 PM

Clicking the <kbd>Select...</kbd> button brings up a dialog, listing all the Python interpreters that RStudio was able to discover on the system.

Screen Shot 2020-07-30 at 1 43 28 PM

After the user has selected a version of Python they want to use,

1. The 'default Python interpreter' preference stored by RStudio is updated,
2. The `RETICULATE_PYTHON` environment variable is set, to match the RStudio preference. 

Note that there is still a lot of work to consider; e.g.

- A similar dialog for project-specific preferences;
- Resolving conflicts between the RStudio-defined default Python preference, and any start-up scripts that might try to set `RETICULATE_PYTHON`,
- The user should be prompted to restart R if they're changing the default version of Python but `reticulate` has already loaded Python in the session,

and so on. I plan to tackle these in separate PRs; I wanted to at least get the initial work into master now.


jjallaire:
Should we make the text box editable so they can override our auto-discovery and provide another path?

kevinushey:
Now with an editable text box so users can write / paste their own Python path in:

Screen Shot 2020-08-03 at 2 18 40 PM

We also give a little bit of UI indicating if there is no `python` binary at the provided path, or if we were otherwise unable to query information about that Python installation:

Screen Shot 2020-08-03 at 2 21 55 PM

src/cpp/r/R/Tools.R
```diff
@@ -1134,3 +1134,18 @@ environment(.rs.Env[[".rs.addFunction"]]) <- .rs.Env
...
    length(cap) && cap
 })
+
+.rs.addFunction("initTools", function()
+{
+   ostype <- .Platform$OS.type
+   info <- Sys.info()
+   envir <- .rs.toolsEnv()
+   
+   assign(".rs.platform.isUnix",    ostype == "unix",              envir = envir)
```
kevinushey, line 1144: I added these helpers because it's easier to use autocomplete to help with these rather than remember if we're comparing with "Unix" vs. "unix", or "Windows" vs "windows", and so on.

src/gwt/src/org/rstudio/core/client/widget/TextBoxWithButton.java
```diff
@@ -86,6 +87,11 @@ protected TextBoxWithButton(String label,
       textBox_ = new TextBox();
       textBox_.setWidth("100%");
       textBox_.setReadOnly(readOnly);
+      
+      textBox_.addValueChangeHandler((ValueChangeEvent<String> event) ->
```
kevinushey, line 91: The changes here (and below) were made to ensure that value change events fire when the underlying textbox itself is edited, which is necessary for the case where the textbox is not read-only.

src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/PythonPreferencesPane.java
```diff
@@ -0,0 +1,319 @@
...
+      
+      mismatchWarningBar_ = new InfoBar(InfoBar.WARNING);
+      mismatchWarningBar_.setText(
+            "The active Python interpreter has been changed by an R startup script.");
+      mismatchWarningBar_.setVisible(false);
+      add(spaced(mismatchWarningBar_));
+      
+      tbPythonInterpreter_ = new TextBoxWithButton(
+            "Python interpreter:",
+            "(No interpreter selected)",
```
gtritchie, line 62: Should use constant PYTHON_PLACEHOLDER_TEXT to ensure they match?

src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/PythonPreferencesPane.java
```diff
@@ -0,0 +1,319 @@
...
+
+   private final InfoBar mismatchWarningBar_;
+   private final PythonDialogResources res_;
+   private final PythonServerOperations server_;
+   private final TextBoxWithButton tbPythonInterpreter_;
+   private final SimplePanel container_ = new SimplePanel();
+   
+   private PythonInterpreter activeInterpreter_;
+   
+   private static final String PYTHON_PLACEHOLDER_TEXT = "(No interpreted selected)";
```
gtritchie, line 310: Typo? "interpreted" vs "interpreter"?

src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/python/PythonInterpreterSelectionDialog.java
```diff
@@ -0,0 +1,54 @@
...
+
+public class PythonInterpreterSelectionDialog extends ModalDialog<PythonInterpreter>
+{
+   public PythonInterpreterSelectionDialog(final JsArray<PythonInterpreter> interpreters,
+                                           final OperationWithInput<PythonInterpreter> operation)
+   {
+      super("Python Interpreters", Roles.getDialogRole(), operation);
+      setOkButtonCaption("Select");
+      
+      widgets_ = new WidgetListBox<PythonInterpreterListEntryUi>();
```
gtritchie, line 35: The listbox should have a label for screen readers (for a visual user it's obvious what it's for but it's an unlabeled control to a screen reader):

`widgets_.setAriaLabel("Python Interpreters");`

src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/python/PythonInterpreterListEntryUi.ui.xml
```diff
@@ -0,0 +1,40 @@
+<!DOCTYPE ui:UiBinder SYSTEM "http://dl.google.com/gwt/DTD/xhtml.ent">
+<ui:UiBinder xmlns:ui="urn:ui:com.google.gwt.uibinder"
+	xmlns:g="urn:import:com.google.gwt.user.client.ui">
+
```
gtritchie, line 4: Noticed this file is indented with tabs instead of spaces.

src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/python/PythonInterpreterListEntryUi.ui.xml
```diff
@@ -0,0 +1,40 @@
...
+	.path {
+		padding: 4px;
+	}
+	
+	</ui:style>
+	
+	<g:HorizontalPanel>
+		
+		<g:FlowPanel styleName='{style.iconContainer}'>
+			<g:Image styleName='{style.icon}' ui:field="uiIcon_"></g:Image>
```
gtritchie, line 31: This creates an unlabeled image (no alt text) so a screen reader will read the underlying filename which is a long guid-like identifier.

One option is to use DecorativeImage instead (which sets empty alt text), but since this icon communicates information, we should have a meaningful alt-text, instead, e.g. "Conda", "VirtualEnv", "Python".

src/cpp/session/modules/SessionPythonEnvironments.R
```diff
@@ -0,0 +1,274 @@
...
+
+.rs.addFunction("python.findPythonSystemInterpreters", function()
+{
+   interpreters <- list()
+   
+   # look for interpreters on the PATH
+   paths <- strsplit(Sys.getenv("PATH"), split = .Platform$path.sep, fixed = TRUE)[[1]]
+   for (path in paths) {
+      
+      pythonPath <- file.path(path, "python")
```
jmcphers, line 200: Should this also check for `python.exe` on Windows?

src/cpp/session/modules/SessionPythonEnvironments.cpp
```diff
@@ -0,0 +1,58 @@
...
+Error initialize()
+{
+   using namespace module_context;
+   
+   std::string pythonPath = core::system::getenv("RETICULATE_PYTHON");
+   if (pythonPath.empty())
+   {
+      core::system::setenv(
+               "RETICULATE_PYTHON",
+               prefs::userPrefs().pythonDefaultInterpreter());
```
jmcphers, line 45: If this user pref changes at runtime, do we want to forward that change to `RETICULATE_PYTHON` (if the old value of the var was also derived from the pref)?

src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/PythonPreferencesPane.java
kevinushey: Yes, we should! Thanks for catching that.

src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/python/PythonInterpreterListEntryUi.ui.xml
kevinushey: *shakes fist at Eclipse*

src/cpp/session/modules/SessionPythonEnvironments.cpp
kevinushey: My exit hatch for this question (resolving whether the environment variable was set by us, versus by the user) was to just require a session restart any time the preference was changed via the UI. (This doesn't help for users who need to change the preference programmatically though, so I think ultimately this is necessary)

PR #7499, opened 2020-07-31
allow the enter key to insert a paragraph above rmd and yaml blocks 

jjallaire:
This goal of this is similar to previous changes that allowed for gap cursor's between code blocks: to make it easier to insert content around code blocks. The gap cursor provides a "virtual" selection between code blocks that can be made either via arrow keys or the mouse. 

This change overrides the Enter key for Ace and in the case that you are at the very start of a code block that already has a valid first line, we go ahead and insert a paragraph above. This will match better with user's intuition about how to create blocks above the current one.

Currently we implement this for Rmd Chunks and YAML blocks, both of which have unambiguous patterns for a valid first line (in fact they will be broken on render w/o that valid first line). This does create some inconsistency with plain fenced code and raw blocks, but those are infrequent/advanced enough that it seems worth it to make the 90% case much better.

Ace doesn't have a documented command for "Return" so I am just using `"insertstring"` with a `\n`. This might run afoul of other considerations I'm not aware of though. 

@jmcphers LMK if you agree with this change and if so whether there is anything else we need to do to make it fully robust.

### DIFF
--- a/src/gwt/panmirror/src/editor/src/api/node.ts
+++ b/src/gwt/panmirror/src/editor/src/api/node.ts
@@ -68,6 +68,7 @@ export interface CodeViewOptions {
   lineNumbers?: boolean;
   bookdownTheorems?: boolean;
   lineNumberFormatter?: (lineNumber: number, lineCount?: number, line?: string) => string;
+  isRequiredFirstLine?: (line: string) => boolean;
 }
 
 export type NodeTraversalFn = (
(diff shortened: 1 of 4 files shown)


jjallaire:
After doing a bit more testing, I have discovered states where the "exec" hook no longer seems to work properly. Since this could fortell some corruption of internal state, I'm going to close this PR.

PR #7506, opened 2020-07-31
delegate focus requests to panmirror when visual mode active

kevinushey:
This PR fixes an issue where attempting to navigate to a file that currently has a visual editor open would open that file, but not send focus to the document within.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTarget.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTarget.java
@@ -1198,6 +1198,21 @@ public class TextEditingTarget implements
       });
    }
    
+   @Override
+   public void navigateToPosition(SourcePosition position,
+                                  boolean recordCurrent,
+                                  boolean highlightLine,
+                                  Command onNavigationCompleted)
+   {
+      ensureTextEditorActive(() -> {
+         
+         docDisplay_.navigateToPosition(position, recordCurrent, highlightLine);
+         if (onNavigationCompleted != null)
+            onNavigationCompleted.execute();
+         
+      });
+   }
+   
    // These methods are called by SourceNavigationHistory and source pane management
    // features (e.g. external source window and source columns) so need to check for
    // and dispatch to visual mode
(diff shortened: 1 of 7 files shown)


src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTarget.java
```diff
@@ -2456,7 +2471,14 @@ public void execute() {
 
    public void focus()
    {
-      view_.editorContainer().focus();
+      if (isVisualModeActivated())
```
jjallaire, line 2474: `isVisualModeActivated()` is a logical state (should it be active) whereas the actual load state (and presence of the `panmirror_` member variable that you access in `focus()`) is different (i.e. it could be logically "activated" but still in the process of loading, so `panmirror_` is `null`). I think you probably want to wrap this call in `ensureVisualModeActive()`.

jjallaire: Actually, I don't even know if that would be correct. Here is another effort I made to get focus working when a visual mode document is opened as part of navigation: https://github.com/rstudio/rstudio/blob/66cec22c35f0eee23e53d908ae93d1975076d007/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/visualmode/VisualMode.java#L628-L640

jjallaire: I think we should talk realtime about all of this on Monday. I'll send a schedule item.

kevinushey: Would it make sense to first check `isActivated()` and then next `ensureVisualModeActivated()` so the command is properly registered if we are indeed still loading?

kevinushey:
@jjallaire should be good to go now -- I created a helper function `withActivatedVisualEditor()` that basically does the same thing as `activate()` (lets you execute a command once the visual editor has been "physically" activated). Let me know if this looks like the right change.

kevinushey:
But looking again this is essentially what `ensureVisualModeActive()` does:

https://github.com/rstudio/rstudio/blob/95d2aa8ec62b793db66b4a831dcc8041c1c9d8dc/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTarget.java#L1135-L1138

So I think we can just re-use the existing helpers here?

PR #7531, opened 2020-08-05
open source companion commit for Pro #1410

ricardofandrade:


gtritchie:
When you mention a Pro issue/pr number from the open-source repo, have to give a qualified reference otherwise it ends up referring to open-source. So `#1410` causes it to be mentioned in that open-source item instead of pro, and the link goes there, too.

You can reference via `rstudio/rstudio-pro#1410` and that will do the right thing. It's irritating to be sure.

PR #7596, opened 2020-08-17
Scroll to a specified position in an unknown file

melissa-barca:
### Intent

The goal of this PR is to enhance `rstudioapi` with a call to scroll to a line in the current editor without stealing the cursor's position. It closes pro issue 1850 (https://github.com/rstudio/rstudio-pro/issues/1850). The associated API PR is found here: https://github.com/rstudio/rstudioapi/pull/187. 

### Approach

`rstudioapi::navigateToFile` is enhanced to make the `file` argument optional so that when a file is not provided, the IDE scrolls to the requested position in the file of the most recently used source window. If no arguments are provided to the call, then it does nothing. 

Rather than modifying the existing logic in the `JumpToFunction EventHandler` I added a new event `ScrollToPosition` to fire from the handler when a file name isn't provided. The previous logic executes a fair amount of code to see if the file is opened and open it if not. We could have this logic eventually call `ScrollToPosition` after the file logic but the same effect is gained from a one line call to another method so it's not needed. 

### QA Notes

You'll need to install the most recent version of `rstudioapi` for this to work. I've been doing this by typing the following in the console:
```
install.packages(githubinstall)
gh_install_packages("rstudio/rstudioapi")
```

Then you can try calling `rstudioapi::navigateToFile` with and without a `file` and with/without a `moveCursor`.

### DIFF
--- a/src/cpp/r/R/Api.R
+++ b/src/cpp/r/R/Api.R
@@ -218,10 +218,11 @@
                                               col = 1L,
                                               moveCursor = TRUE) {
    # validate file argument
-   if (!is.character(filePath)) {
+   hasFile = !is.null(filePath) && length(filePath) > 0
+   if (hasFile && !is.character(filePath)) {
       stop("filePath must be a character")
    }
-   if (!identical(filePath, character(0)) && !file.exists(filePath)) {
+   if (hasFile && !file.exists(filePath)) {
       stop(filePath, " does not exist.")
    }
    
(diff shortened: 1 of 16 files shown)


src/cpp/r/R/Api.R
```diff
@@ -213,12 +213,16 @@
...
 })
 
-.rs.addApiFunction("navigateToFile", function(filePath, line = 1L, col = 1L) {
+.rs.addApiFunction("navigateToFile", function(filePath = character(0),
+                                              line = 1L,
+                                              col = 1L,
+                                              moveCursor = TRUE) {
    # validate file argument
-   if (!is.character(filePath)) {
+   hasFile = !is.null(filePath) && length(filePath) > 0
```
kevinushey, line 221: nit: use `<-` for assignment

melissa-barca: Fixed 👍

Issue #5412, opened 2019-09-16
In RMarkdown code chunk, make code complete for folder start from project folder instead of document folder


dracodoc:
In code chunk of RMarkdown, inside a quote, tab can bring code complete for folder and files, which is very useful in input file/folder/path.

When using relative path, this seemed always start from document folder of the Rmd, not the project folder. This is understandable as knitting a Rmd usually take document folder as working folder. 

However there are also other usage of Rmd which require different patterns. Myself use Rmd to organize code/notes/old code/various attempts/design ideas, basically everything relate to a topic are in a Rmd. This kind of Rmd never get knitted as the code chunks are holding places of various code sections, not a part of a report. If I organize these Rmd into different folder for sub projects, the document folder will change and I cannot use the auto complete for folder anymore as it doesn't start from project folder -- and to run the code chunk individually, the working directory is always project folder.

Both options are valid use cases. I'm wondering what's the best approach. One option is to have a switch for two selections in UI. There is a similar option in Knit -> Knit directory, but I'm not sure if this is the right place to apply to this aspect.

Or there could be a per Rmd configuration entry, like an option inserted in header, respond to some UI options.

ronblum:
@dracodoc Thank you for bringing this up. Could you provide an example of what you do, with some sample code and steps? It would be helpful to get an idea of what the issue is that you're running into and the improvement suggestion. Thanks! Also, I believe that the autocompletion is based on the working directory (which is initially the project directory).

dracodoc:
Sorry I didn't explain it better.

Let's say the current project directory is `Project`, and there is a Rmd in a subfolder `module`

```
Project
    ├─module
           └─xx.Rmd
```

In code chunk of Rmd, if you input a "" and start to press tab inside the "", RStudio will provide list of folder and files with auto complete as relative path. When using in console or .R script, this is from working directory, but in a Rmd that in a subfolder, it start from the Rmd document folder. This is understandable as knitting will run them from Rmd document folder.

However I'm using code chunk to hold some code and often run them interactively, never knitting the whole document in some cases. Then I'll need the path to start from working directory.

ronblum:
@dracodoc Ah! Thank you. I'll mark this option as an enhancement to review as we continue developmet of RStudio.

fernandovmacedo:
I have the same issue. Is there any chance to be fixed in version 1.3?

kevinushey:
[...]

PR #7598, opened 2020-08-17
properly resolve project when completing markdown links

kevinushey:
### Intent

Ensure Markdown link completions (when the Knit Directory pref is set to Project Directory) function as expected.

Screen Shot 2020-08-17 at 11 30 51 AM

Screen Shot 2020-08-17 at 11 30 57 AM

### Approach

Properly use the source document's properties to compute the document's project path. (Previous code erroneously concluded that `project_path` referred to the project path, when in actuality it refers to the project-relative path.)

### QA Notes

Open an R Markdown document, and then (with `|` indicating the cursor position), type:

```
[](|
```

Then press `<TAB>` to request completions. Check that the completions return what you expect as the Knit Directory preference is changed.

Closes https://github.com/rstudio/rstudio/issues/5412.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -88,4 +88,5 @@
 * Fixed disappearing commands and recent files/projects when RStudio Desktop opens new windows (#3968)
 * Fixed issue where active repositories were not propagated to newly-created `renv` projects (#7136)
 * Fixed issue where .DollarNames methods defined in global environment were not resolved (#7487)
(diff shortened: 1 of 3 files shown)


kevinushey:
Squashed + merged in https://github.com/rstudio/rstudio/commit/c30aa1d203a99e497bb5cbc347099e6cb351ea68.

Issue #7637, opened 2020-08-21
R Session Aborted after refreshing Windows desktop UI


gtritchie:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.3.1058+ (including 1.3.1073 release, 1.3.1075 preview, and latest 1.4)
    OS Version      : Windows-10
    R Version       : 4.0.2

### Steps to reproduce the problem
- run Windows desktop
- right-click on blank area of toolbar, and click Reload
- QtWebEngine region refreshes and UI is redisplayed
- Problem described next doesn't always repro; if it doesn't after a couple of tries, close RStudio and restart, and try again

### Describe the problem in detail
- a second or two later, the session crashes and see this:

![R Session Aborted screenshot](https://user-images.githubusercontent.com/10569626/90932987-37529980-e3b4-11ea-8bb2-a4c36345c45a.png)

### Describe the behavior you expected
- Session does not abort

- [x] I have read the guide to submitting good bug reports at https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports . 
- [x] I have installed the latest version of RStudio and confirmed that the issue still persists.
- [x] If I am reporting a RStudio crash, I have included a diagnostics report. https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
Given that this crash only happens:

1. When the UI only is refreshed, which is a relatively rare action for a user to take; and
2. Is evidently somewhat uncommon (except that @adamconroy bumped into it frequently during development?)

are we sure this needs to be fixed for v1.3?

R-Hannibal:
Verified Fixed on: RStudio Version - 1.4.746, R Version - 4.0.2, Windows 10 x64 - Desktop

Unable to reproduce the crash on the latest daily build after aggressive refresh attempts (I am able to repro the crash state on 1.3, in the inconsistent fashion already enumerated in the issue report). Combined with @adamconroy 's report on the PR that the [fix is working for him](https://github.com/rstudio/rstudio/pull/7641#issuecomment-679276280) (the issue was reproducing 100% for him before the fix), I feel good about closing this ticket.

Note: Changing the milestone to v1.4 based off @kevin's [comment on the PR](https://github.com/rstudio/rstudio/pull/7641#issuecomment-680188751) and the absence of the change being merged into 1.3.

PR #7641, opened 2020-08-24
be more careful when cleaning up file monitor contexts

kevinushey:
### Intent

RStudio occasionally crashes after a UI refresh (Right Click -> Reload), which we would prefer not doing.

### Approach

The crash appears to occur in the file monitor's completion routine, likely running after the associated context was deleted. The fix here is to more carefully manage the lifetime of the context object, and to allow the completion handler itself to delete the context if necessary.

### QA Notes

Follow steps in https://github.com/rstudio/rstudio/issues/7637 to reproduce, but note that the issue is quite finicky -- in my cases, I could only reproduce maybe 20% of the time.

Closes https://github.com/rstudio/rstudio/issues/7637.

### DIFF
--- a/src/cpp/core/system/file_monitor/Win32FileMonitor.cpp
+++ b/src/cpp/core/system/file_monitor/Win32FileMonitor.cpp
@@ -53,6 +53,7 @@ public:
    FileEventContext()
       : recursive(false),
         hDirectory(nullptr),
+        completionsPending(false),
         hRestartTimer(nullptr),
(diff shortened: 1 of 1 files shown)


adamconroy:
This was happening 100% for me and I just tested locally and it fixes it. Thank you! It was making dev very difficult actually.

kevinushey:
Confirmed with @jmcphers that we'll just take this for v1.4.

Issue #7610, opened 2020-08-18
allow project-specific Python configuration


kevinushey:
https://github.com/rstudio/rstudio/pull/7497 made it possible to customize the Python interpreter used by `reticulate` globally; however, we should also consider allowing for per-project versions of Python.

Implementation note: we should consider how this would play together with projects using `renv`, since `renv` has its own notion of the project-default version of Python.

melissa-barca:
Verified in 1.4.1013

PR #7656, opened 2020-08-25
allow for project-specific Python options

kevinushey:
### Intent

Make it possible to configure the version of Python used on a per-project basis, by storing the Python type, version, and path. While currently only the recorded Python path is used, in the future we may use the Python type + version to automatically provision Python environments matching those requirements if requested (e.g. for project portability).

### Approach

Implement the associated pane for the project options dialog, refactoring the common UI code into a base class that can be used by both panes.

Currently, the project options are persisted as part of the `.Rproj` file -- we may want to consider if that's the right approach.

### QA Notes

- Validate that the "global" version of Python and "project-specific" version of Python can be changed independently;
- Validate that a project-specific version of Python will override the global default.

Closes https://github.com/rstudio/rstudio/issues/7610.


src/cpp/session/resources/schema/user-prefs-schema.json
```diff
@@ -1395,11 +1395,23 @@
...
+            "default": ""
+        },
+        "python_version": {
+            "title": "Python Version",
+            "description": "The Python version.",
+            "type": "string",
+            "default": ""
+        },
+        "python_path": {
+            "title": "Python Path",
```
jmcphers, line 1411: The `title` is only used in the Command Palette, which doesn't show options w/o a Title -- so by convention options that can only be correctly manipulated by dedicated UI have no title. In this case I think we should remove the title so these options don't show up in the palette.

kevinushey:
> I didn't notice it (or maybe I missed it): do we need some UI in Global Options -> Python that indicates when the Python version is being controlled at a project level? (thinking of the case where you might try to set a Python interpreter in Global Options, but nothing happens because unbeknownst to you the project level option is overriding it)

This is a good idea; I'll make this change.

kevinushey:
Here's the UI treatment now:

Screen Shot 2020-08-25 at 4 31 05 PM

kevinushey:
I've been thinking about one issue: RStudio projects (and so, the project file) are frequently shared publicly (e.g. on GitHub); however, Python paths are normally unique to a particular user. Given that, I'm considering storing the following preferences:

```
# system interpreter
PythonVersion: 3.7.4
PythonType: system

# virtual environment
PythonVersion: 3.7.4
PythonType: virtualenv
PythonEnvironment: my-virtual-environment

# conda environment
PythonVersion: 3.7.4
PythonType: conda
PythonEnvironment: my-conda-environment
```

That is, we encode the environment name, rather than its explicit path (which then in theory could be stable + shared with multiple users).

Then, in the future, we could use this to automatically find an existing environment matching the declared specs, or provision a compatible environment if not.

kevinushey:
The downside of the above approach is that it makes it a lot more challenging for user-defined tools to set up an appropriate Python path for a project. Savvy users (or, say, `renv`) could just create their own project-local environments, and set the PythonPath to point to that associated python installation.

In addition, one can always infer the Python environment type + name from the Python path, so perhaps storing paths is better -- perhaps with some sort of special aliasing for paths that we "know" are just names for an existing environment.

Issue #7665, opened 2020-08-26
Clicking "Home" from file tab when in "Home" directory terminates R session


life-droid:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.3.1073
    OS Version      :  Windows 10
    R Version       :  4.0.2

### Steps to reproduce the problem
Launch RStudio
Navigate to the File tab of the Plots pane
If in "Home" directory, click home
If not in "Home" directory click home twice

 Note: If it doesn't occur straight away click Home again, eventually it crashes

### Describe the problem in detail
Clicking the "Home" button in the File tab of the "File/Plots/Packages/Help/Viewer" pane when already in the "Home" directory causes R to abort. A popup window then comes up saying R terminated unexpectedly.

### Describe the behavior you expected
I would expect the directory not to refresh but RStudio not to crash.

- [x] I have read the guide to submitting good bug reports at https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports . 
- [x] I have installed the latest version of RStudio and confirmed that the issue still persists.
- [x] If I am reporting a RStudio crash, I have included a diagnostics report. https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

[diagnostics-report.txt](https://github.com/rstudio/rstudio/files/5132878/diagnostics-report.txt)

kevinushey:
Thank you for the bug report! I was able to reproduce with v1.3.1075 as well. Do you see this issue in the latest v1.4 daily builds? https://dailies.rstudio.com

I wasn't able to reproduce in v1.4; I wonder if this is some version of the crash described in https://github.com/rstudio/rstudio/pull/7641.

If so, that might make that PR a candidate for backporting to v1.3.

life-droid:
Hi Kevin,

I am not sure if it is related to #7641 but when I reloaded the UI via right-click after two reloads it crashed. See attached diagnostic report. I will try with the Daily Build shortly.

[diagnostics-report.txt](https://github.com/rstudio/rstudio/files/5133166/diagnostics-report.txt)

life-droid:
With the daily build the issue does not seem to be present.

System details
RStudio Edition : Desktop
RStudio Version : 1.4.747
OS Version      :  Windows 10
R Version       :  4.0.2

mikebessuille:
Closing as duplicate of #7641

[...]

Issue #7652, opened 2020-08-25
IDE OneDrive Conflict crashes IDE file explorer


yonicd:
### System details

<details> <summary>Diagnostics Report </summary>

RStudio Diagnostics Report
==========================

WARNING: This report may contain sensitive security information and / or
personally identifiable information. Please audit the below and redact any
sensitive information before submitting your diagnostics report.

Generated: Tue Aug 25 08:24:58 2020

RStudio Version
--------------------------------------------------
1.3.1073

Session Information
--------------------------------------------------
R version 4.0.2 (2020-06-22)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 19041)

Matrix products: default

locale:
[1] LC_COLLATE=English_United States.1252  LC_CTYPE=English_United States.1252    LC_MONETARY=English_United States.1252
[4] LC_NUMERIC=C                           LC_TIME=English_United States.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

loaded via a namespace (and not attached):
[1] compiler_4.0.2 tools_4.0.2    fs_1.5.0       tinytex_0.25   xfun_0.16     

System Information
--------------------------------------------------
sysname        : Windows    
release        : 10 x64     
version        : build 19041
nodename       : USCAMLJSIDI
machine        : x86-64     
login          : yoni.sidi  
user           : yoni.sidi  
effective_user : yoni.sidi  

Platform Information
--------------------------------------------------
OS.type    : windows
file.sep   : /
dynlib.ext : .dll
GUI        : RStudio
endian     : little
pkgType    : win.binary
path.sep   : ;
r_arch     : x64

Environment Variables
--------------------------------------------------
ALLUSERSPROFILE            : C:\ProgramData
APPDATA                    : C:\Users\jonathan.sidi\AppData\Roaming
CLICOLOR_FORCE             : 1
CommonProgramFiles         : C:\Program Files\Common Files
CommonProgramFiles(x86)    : C:\Program Files (x86)\Common Files
CommonProgramW6432         : C:\Program Files\Common Files
COMPUTERNAME               : USCAMLJSIDI
ComSpec                    : C:\windows\system32\cmd.exe
DISPLAY                    : :0
DriverData                 : C:\Windows\System32\Drivers\DriverData
GFORTRAN_STDERR_UNIT       : -1
GFORTRAN_STDOUT_UNIT       : -1
[...]

PR #7671, opened 2020-08-27
cherry-pick file-monitor crash fix from master

gtritchie:
We thought this crash only manifested in very obscure scenarios such as forcing the Windows Desktop IDE to reload the embedded browser but users are hitting it via routine use of the Files pane so are now backporting it from master to next 1.3 patch.

See #7641 for more details.

### Intent

Fix crashes in Windows Desktop sessions when reloading the embedded browser or changing directories in the Files pane.

### Approach

Backported fix from master. #7641

### QA Notes

This is a Windows-desktop specific issue. Test by trying scenarios in #7637, #7652, #7665.


kevinushey:
Thanks -- LGTM!

life-droid:
Thanks guys

Issue #3564, opened 2018-10-01
Feature - close all terminals


hsselman:
### Feature description
First of all, I love the terminal integration in RStudio IDE. Thanks for that! I use it quite often lately. When I started scripting commando's to send to terminal windows, I came up with this feature. 
`Close Terminal` is already an option, but I would really appreciate a `Close all terminals` option. Now I have to close them by hand, or restart.

hsselman:
Love the workaround to close the Terminal tab itself. That would work! 

I am using `rstudioapi` to launch commando's to the terminal. Love to see what you mean by creating an add-in.

gtritchie:
Ok, here's an RStudio addin that provides a new command `Close All Terminals`. This command kills any processes running in the terminals without prompting, so use with caution. A fancier version could be created to only kill terminals that aren't busy, etc., but for now just did this.

Install from github via the devtools package.

```
devtools::install_github("rstudio/terminaltools")
```
There will be a new section on the Addins dropdown on the toolbar:

![2018-10-01_12-23-56](https://user-images.githubusercontent.com/10569626/46310698-f14c7c80-c574-11e8-8ea8-a906477df0a7.png)

You can add a keyboard shortcut via Tools / Modify Keyboard Shortcuts. Type "Close All Terminals" in the search box, then click on blank area under Shortcut to enter keysequence.

![2018-10-01_12-26-38](https://user-images.githubusercontent.com/10569626/46310823-42f50700-c575-11e8-9ebf-6c7c49fc17d1.png)

For more on creating addins, see https://rstudio.github.io/rstudioaddins/.

PR #7679, opened 2020-08-28
Add commands for showing each Global Options pane directly

### LINKED ISSUE #7678: Provide command-palette access to each Global Options pane
In the command palette, you can open some of the Global Options panes directly via a command:

- "Terminal Options"
- "Pane Layout"
- "Accessibility Options"

But not the others: be nice to have commands to go directly to each of these.

(Not including the Project Options dialog in this issue, just the global options)

gtritchie:
### Intent

Provide ability to go directly to each pane in the Global Options dialog from the command palette.

Fixes #7678

### Approach

Added new commands for panes that didn't already have one (terminal, accessibility, and pane layout already had such commands). The command are:

- Global Options (goes to General pane)
- Code Options...
- Console Options...
- Appearance Options...
- Pane Layout Options...
- Packages Options...
- R Markdown Options...
- Sweave Options...
- Spelling Options...
- Git/SVN Version Control Options...
- Publishing Options...
- Terminal Options...
- Accessibility Options... (note this is a special case as it opens JUST the accessibility pane, no other panes available, this is on purpose to reduce potential confusion for screen reader users)
- Python Options...

### QA Notes

Verify that each commands opens the corresponding pane in Global Options.

Note that three of these panes can be hidden via feature limits in RSP (`allow_shell` for terminal, `allow_publish` for publishing, `allow-vcs` for version control); ensure that when such limits are set, the corresponding command is not shown in the command palette.

Also fixed a tiny issue that was introduced via #3564; if Terminal is disallowed via the "allow_shell" feature limit (RSP) the "Close all Terminals" was not being removed, causing the Tools / Terminal menu to still appear, so now hiding it

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -43,6 +43,7 @@
 * The Data Viewer now uses the `format()` methods defined for columns entries when available (#7239)
 * Add support for navigating source history with mouse forward/back buttons (#7272)
 * Improved error logging of mistyped usernames when using PAM authentication (#7501)
+* Add ability to go directly to various Global Option panes via Command Palette (#7678)
 
 ### RStudio Server Pro
 
--- a/src/gwt/src/org/rstudio/studio/client/application/Application.java
+++ b/src/gwt/src/org/rstudio/studio/client/application/Application.java
@@ -876,6 +876,12 @@ public class Application implements ApplicationEventHandlers
          removeTerminalCommands();
       }
 
+      if (!sessionInfo.getAllowVcs())
+         commands_.showVcsOptions().remove();
+
+      if (!sessionInfo.getAllowPublish())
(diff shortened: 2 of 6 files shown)


kevinushey:
LGTM!

As an aside: do you think it's worth documenting some of the required auxiliary work (like implementing these commands) when adding new panes to the Options / Project Options dialogs? As a note in the wiki, or comment on one of the Preferences base classes, or something similar.

gtritchie:
> As an aside: do you think it's worth documenting some of the required auxiliary work (like implementing these commands) when adding new panes to the Options / Project Options dialogs? As a note in the wiki, or comment on one of the Preferences base classes, or something similar.

Good idea, I'll stick a comment somewhere.

PR #7689, opened 2020-08-31
don't log monitor interrupt errors

kevinushey:
### Intent

Fix an issue where noisy file monitor errors could be logged on session restart.

### Approach

Mark file monitor errors as expected; suppress logging of expected errors.

### QA Notes

This is an issue that reproduces purely in development builds -- nothing to verify for QA.


MariaSemple:
I agree with the general principle of this, but I think it would be cleaner to add a member to `Error::Impl` like `bool ShouldBeLogged` and set and read that, rather than store it as a string value in the properties array. Especially since the properties are conceptually part of the error message and the status of whether it should be logged isn't.

kevinushey:
Thanks! That's a good point; I'll make that change.

PR #7690, opened 2020-08-31
fix range analysis warnings in newest clang

kevinushey:
### Intent

Resolve compiler warnings from `-Wrange-loop-analysis`, introduced in the newest version of Clang.

### Approach

Look at and respond to compiler warnings. Mostly fixed via adding reference qualifier `&` or, in some cases, using `auto&&` (e.g. for `std::vector<T>` where binding to potential proxies for e.g. `std::vector<bool>` could be necessary).

### QA Notes

No QA notes / testing required.

### DIFF
--- a/src/cpp/core/ConfigProfile.cpp
+++ b/src/cpp/core/ConfigProfile.cpp
@@ -88,7 +88,7 @@ Error ConfigProfile::parseString(const std::string& profileStr)
    for (const ptree::value_type& child : profileTree)
    {
       boost::optional<Level> matchingLevel;
-      for (const Level& level : sections_)
+      for (Level level : sections_)
       {
          // if the section level starts with the defined section name
          // from a prior call to addSections, we have a matching section level
@@ -109,7 +109,7 @@ Error ConfigProfile::parseString(const std::string& profileStr)
       }
 
       std::map<std::string, std::string> values;
-      for (const ptree::value_type val : child.second)
+      for (const ptree::value_type& val : child.second)
       {
          // check to see if the parameter within the section has been registered
          const std::string& paramName = val.first;
--- a/src/cpp/session/SessionMain.cpp
+++ b/src/cpp/session/SessionMain.cpp
@@ -650,7 +650,8 @@ Error rInit(const rstudio::r::session::RInitInfo& rInitInfo)
    error = rstudio::r::json::getRpcMethods(&rMethods);
    if (error)
       return error;
-   for (const json::JsonRpcMethod& method : rMethods)
+   
+   for (json::JsonRpcMethod method : rMethods)
    {
       registerRpcMethod(json::adaptMethodToAsync(method));
    }
--- a/src/cpp/shared_core/include/shared_core/json/Json.hpp
+++ b/src/cpp/shared_core/include/shared_core/json/Json.hpp
@@ -1790,7 +1790,7 @@ template <typename T>
 inline Value toJsonValue(const std::vector<T>& in_vector)
 {
    Array results;
-   for (const T& val : in_vector)
+   for (auto&& val : in_vector)
    {
       results.push_back(val);
    }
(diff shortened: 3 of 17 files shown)


src/cpp/shared_core/json/Json.cpp
```diff
@@ -586,12 +586,6 @@ Object Value::getValue<Object>() const
    return copy;
 }
 
-template<>
```
kevinushey, line 589: This method returns a pointer to stack-allocated memory (moral equivalent of `return std::string().c_str()`); given that it's unsafe I've decided to remove it. This didn't affect anything in open source; if something is effected in pro then I think we need to change those usages to be safer.

MariaSemple: I agree - I don't see a reason why the caller couldn't get the string value and call `c_str()` themselves.

src/cpp/shared_core/include/shared_core/json/Json.hpp
```diff
@@ -1790,7 +1790,7 @@ template <typename T>
 inline Value toJsonValue(const std::vector<T>& in_vector)
 {
    Array results;
-   for (const T& val : in_vector)
+   for (auto&& val : in_vector)
```
MariaSemple, line 1793: This compiles? The parameter is a `const &`, so I think we should use `const &` here to ensure that we won't accidentally gut something we shouldn't.

src/cpp/session/SessionMain.cpp
```diff
@@ -650,7 +650,8 @@ Error rInit(const rstudio::r::session::RInitInfo& rInitInfo)
    error = rstudio::r::json::getRpcMethods(&rMethods);
    if (error)
       return error;
-   for (const json::JsonRpcMethod& method : rMethods)
+   
+   for (json::JsonRpcMethod method : rMethods)
```
MariaSemple, line 654: Is `const &` incorrect here?

src/cpp/core/ConfigProfile.cpp
```diff
@@ -88,7 +88,7 @@ Error ConfigProfile::parseString(const std::string& profileStr)
    for (const ptree::value_type& child : profileTree)
    {
       boost::optional<Level> matchingLevel;
-      for (const Level& level : sections_)
+      for (Level level : sections_)
```
MariaSemple, line 91: This is a copy instead of a reference because it's an enum type (and therefore just as fast/memory efficient to copy as to copy the reference of it)?

src/cpp/shared_core/include/shared_core/json/Json.hpp
kevinushey: It does -- this actually does deduce as `const T&` for a regular `const std::vector<T>&`. This is because it basically uses the result of `decltype(*val.begin())`, and for const vectors this will return a const interator which will yield a `const T&` when dereferenced. See e.g.

https://godbolt.org/z/Ej973W

In other words, `auto&&` preserves the const-ness here because the container itself handles const-ness in its iterators.

src/cpp/session/SessionMain.cpp
kevinushey: It seems to be; at least, `-Wrange-loop-analysis` indicates that doing this avoids copies. I don't quite understand why but this makes the compiler happy.

src/cpp/core/ConfigProfile.cpp
kevinushey: Right, exactly.

src/cpp/shared_core/include/shared_core/json/Json.hpp
MariaSemple: In that case, why change this away from `const T& val` to `auto&&` when we use `const auto&` in other places? I also think both `const auto&` and `const T&` are more obvious in their meaning to the reader.

MariaSemple: Oh -  I see in the approach section it's because of the potential weirdness with `std::vector<bool>`

PR #7738, opened 2020-09-04
implement simple selection-based transformations for visual mode

kevinushey:
### Intent

Currently, most APIs in `rstudioapi` expect a traditional text-based document, and so provide information suitable for that format. Unfortunately, such APIs don't map well to the abstractions used with the visual editor, and so many of these APIs are no longer appropriate.

This PR implements a set of simple selection-based APIs that allow R packages (via addins) to receive and transform the current selection, and work both in visual mode and source mode.

### Approach

The PR is a bit long, but straightforward: we implement two API functions `.rs.api.selectionGet()` and `.rs.api.selectionSet()`, and wire those up to "do the right thing" regardless of whether the editor is in visual mode or source mode.

### QA Notes

I'll need to put together some example addins to make it easier (possible?) for QA to properly test this.

---

visual-mode-selection-api-small

### Notes

Right now, `.rs.api.selectionGet()` just returns an R list with `value` set to the current selection; e.g.

```
> print(.rs.api.selectionGet())
$value
[1] "  styled <- styler::style_text(selection$value)\n"
```

Is there anything else we should include on this object?


kevinushey:
Note also that, in visual mode, we don't transform to Markdown before reading the selection, so some extra information is lost:

Screen Shot 2020-09-04 at 3 33 05 PM

@jjallaire, just to confirm -- do you think we should just err out in this case, or is there something sensible we could do?

src/gwt/panmirror/src/editor/src/editor/editor.ts
```diff
@@ -627,6 +627,33 @@ export class Editor {
...
+    );
+
+  }
+
+  public replaceSelection(value: string): void {
+
+    // retrieve properties we need from selection
+    let {from, empty} = this.view.state.selection;
+
+    // insert text
```
jjallaire, line 644: In order to preserve the selection's marks here you want to do this:

```typescript
const tr = this.view.state.tr;
const selectionMarks = tr.selection.$from.marks();
tr.replaceSelectionWith(state.schema.text(value, selectionMarks), false);
this.view.dispatch(tr);
```

PR #7742, opened 2020-09-05
Improve User Experience with XDG directory permissions issues

### LINKED ISSUE #4505: better handle case where user lacks read/write permissions in ~/.rstudio
### System details

    RStudio Edition : Desktop
    RStudio Version : v1.2.1330
    OS Version      : macOS 10.14.3
    R Version       : 3.5.3

Run the following in your shell:

```sh
sudo chown -R root ~/.rstudio-desktop
```

Then, try launching RStudio.

Depending on what files / folders within are inaccessible, different errors can occur. In this particular case, RStudio can (if no project is opened) fail to launch, or (if a project is open) fail to close that project.

Some possibilities:

1. RStudio could attempt to repair permissions on these folders during launch if possible;
2. RStudio could notify the user if it lacks read / write permissions into the state folder.

### Notes

Prompted by https://github.com/rstudio/rstudio/issues/4501, but other incantations of this issue have been reported in the past. I think this occurs most commonly when e.g. the user's home directory is mounted as a networked filesystem (especially on Windows) but mapping of Windows <-> Unix permissions is incomplete.

MariaSemple:
### Intent

Fixes #4505 

This PR intends to improve the user experience when `~/.local/share/rstudio` and `~/.config/rstudio` are not owned by the correct user. 

### Approach

* On pref save, likely the most common problem for an end user, modify the error message that will be returned to be more descriptive:
image

* In SessionMain, test dirs for write access. If the files exist and the user doesn't have write access, attempt to repair permissions. Log messages on failures:
image

RStudio on session start up failure (happens when `~/.local/share/rstudio` is not accessible by the current user):
image

Downside: Repairing permissions generally fails for the same reason the user cannot write to these files.

### QA Notes

To reproduce this issue, do one of the following:

1. Start RStudio Desktop as root on a system on which RStudio Desktop has never been started.
2. On a system where RStudio Desktop has already been run, execute `chown -R root:root ~/.local/share/rstudio` and/or `chown -R root:root ~/.config/rstudio` to emulate the issue caused by (1).

Then, start RStudio Desktop as your regular user and attempt to modify and save something in your global preferences. Prior to this change you should see an error that says "No such file or directory".

More considerations:
* The session should still start if `~/.config/rstudio` is inaccessible, but it won't start if `~/.local/share/rstudio` is inaccessible.
* This should still work (and the logs should be correct) if the XDG paths are changed. 
* Repairing the permissions (e.g. `chown -R myuser:myuser ~/.config/rstudio`) should resolve the issue completely (i.e. no more logs about it either).


src/cpp/core/system/Xdg.cpp
```diff
@@ -108,17 +109,53 @@ FilePath resolveXdgDir(const std::string& envVar,
...
          user,
          homeDir
    );
+
+   Error permError;
+   Error error = confDir.ensureDirectory();
+   if (!error)
+   {
+      // Test permissions here.
+      error = confDir.testWritePermissions();
```
jmcphers, line 129: We call userConfigDir all the time and it's expected to be a fast, read-only operation -- probably not the right place to add a write test. This test should be done only once early at startup.

src/cpp/session/prefs/UserPrefsLayer.cpp
```diff
@@ -109,6 +109,19 @@ Error UserPrefsLayer::writePrefs(const core::json::Object &prefs)
...
 
+   // Modify the error to be more descriptive
+   if (isFileNotFoundError(error) && prefsFile_.getParent().exists())
+   {
+      error = Error(
+         error.getName(),
+         error.getCode(),
+         "Unable to save preferences. Please verify that " + 
+            prefsFile_.getParent().getAbsolutePath() +
+            " exists and it and all its children are owned by the user '" +
```
jmcphers, line 120: For a non-technical audience, how about "the folder X and all the files inside it"?

jmcphers (changes requested):

src/cpp/core/system/Xdg.cpp
MariaSemple: Should we test this at session startup, server startup or both? We do read from here in the server/desktop code for crashpad settings, I think.

src/cpp/session/prefs/UserPrefsLayer.cpp
MariaSemple: Oh, actually I realized I already changed this to `"Unable to save preferences. Please verify that " + prefsFile_.getAbsolutePath() + " exists and is owned by the user '" + system::username() + "'."`. I must not have pushed the change when I thought I did.

src/cpp/core/system/Xdg.cpp
```diff
@@ -108,17 +109,53 @@ FilePath resolveXdgDir(const std::string& envVar,
...
+         error = system::User::getUserFromIdentifier(system::username(), currUser);
+         if (!error)
+            error = confDir.changeOwnership(currUser, true);
+      }
+   }
+
+   if (error)
+   {
+      std::string message;
+      message.append("Encounterd an issue while testing config dir \"")
```
kevinushey, line 148:
```suggestion
      message.append("Encountered an issue while testing config dir \"")
```

MariaSemple: I changed the phrasing of this locally to `"Unable to access the directory \""`.

src/cpp/core/include/core/system/Xdg.hpp
```diff
@@ -62,6 +62,12 @@ FilePath userConfigDir(const boost::optional<std::string>& user = boost::none,
 // On Windows, this is 'FOLDERID_LocalAppData' (typically 'AppData/Local').
 FilePath userDataDir(const boost::optional<std::string>& user = boost::none,
                      const boost::optional<FilePath>& homeDir = boost::none);
+                     
+// This function verifies that the userConfigDir() and userDataDir() exist and are owned by the running user.
+// 
+// It should be invoked once. Any issues with these directories will be emmitted to the session log.
```
kevinushey, line 68:
```suggestion
// It should be invoked once. Any issues with these directories will be emitted to the session log.
```

src/cpp/core/system/Xdg.cpp
```diff
@@ -135,6 +136,52 @@ FilePath userDataDir(
...
+               error = dir.changeOwnership(currUser, true);
+         }
+      }
+
+      if (error)
+      {
+         std::string message;
+         message.append("Unable to access \"")
+            .append(dir.getAbsolutePath())
+            .append(":\n  ")
```
kevinushey, line 172:
```suggestion
            .append("\":\n  ")
```

src/cpp/session/prefs/UserPrefsLayer.cpp
```diff
@@ -109,6 +109,19 @@ Error UserPrefsLayer::writePrefs(const core::json::Object &prefs)
       lastSync_ = prefsFile_.getLastWriteTime();
    }
 
+   // Modify the error to be more descriptive
+   if (isFileNotFoundError(error) && prefsFile_.exists())
+   {
+      error = Error(
```
kevinushey, line 115: I wonder if we should instead amend the existing error's properties here, rather than creating a whole new error object? (Since in theory some properties of interest on the original error could be lost)

kevinushey (approved):
LGTM!

src/cpp/session/prefs/UserPrefsLayer.cpp
kevinushey: We could even add a helper method to make this a sort of "official" way of tagging this kind of information on an error (since there's a lot of places in the code base where users and support would benefit from extra tagged information). Or do we have an existing one that is suitable here?

gtritchie (approved):
LGTM! Kevin beat me to pointing out "emmitted" vs "emitted". I'm too slow today.

src/cpp/session/prefs/UserPrefsLayer.cpp
MariaSemple: The dialog that the user is presented with only displays the message of the error. All the properties are left out. The only way to set the message of  an error is at construction time. However, I do think it would be reasonable to set the original error as the cause, or at least copy over all the properties.

kevinushey: What do you think about setting or amending the message on the original error object?

Sorry for being nitpicky; I'm just trying to think if there's a sort of "best practice" that we could encode here and then reuse in other locations where we want to surface extra information for the user.

MariaSemple: I'm not sure. When I modified the `Error` class I continued to follow the existing pattern - i.e. that the message can't be updated after the error has been created. We use to take the message solely from the boost error code object. I don't know the reasoning behind the original desicion so I can't say whether it should be changed or not. @jmcphers Do you have any opinion on this?

src/cpp/core/system/Xdg.cpp
```diff
@@ -135,6 +136,52 @@ FilePath userDataDir(
...
 }
 
+void verifyUserDirs(
+   const boost::optional<std::string>& user,
+   const boost::optional<FilePath>& homeDir)
+{
+   auto testDir = [](const FilePath& dir, const ErrorLocation& errorLoc)
+   {
+      Error permError;
+      Error error = dir.ensureDirectory();
```
jmcphers, line 146: I think this might cause trouble later because there is some first-run logic that checks for the existence of these directories to figure out if we need to do certain kinds of initialization (or migration from old folders). Since the pathological case is generally that these directories exist but are unwritable, how would you feel about having this test function do nothing if the directory isn't in place yet?

jmcphers (changes requested):

src/cpp/core/system/Xdg.cpp
```diff
@@ -135,6 +136,49 @@ FilePath userDataDir(
...
+            permError = Error(
+               error.getName(),
+               error.getCode(),
+               "Missing write permissions to " + dir.getAbsolutePath() + ". Some features may not work correctly.",
+               ERROR_LOCATION);
+
+            system::User currUser;
+            error = system::User::getUserFromIdentifier(system::username(), currUser);
+            if (!error)
+               error = dir.changeOwnership(currUser, true);
```
jmcphers, line 160: On Linux you need to be root to change ownership, so I think that this is very unlikely to succeed. In the pathological case (your settings/data dir is owned by someone else), the effect here will be that we log "Unable to access [dir]: Operation not permitted" which isn't super helpful. Maybe we should just figure out who the owner is and log it if we can? Or have a different error message for the attempted ownership change?

jmcphers (changes requested):

src/cpp/core/system/Xdg.cpp
MariaSemple: In the screenshots in the description you can see that you get both errors in that case. Is that sufficient?

jmcphers: Yes, I think that's fine. Is there a circumstance where you _would_ expect `changeOwnership` to work? Does it work on Windows, for example?

MariaSemple: I'm not aware of a circumstance where this would work. I included it based on @kevinushey's suggestion in #4505 because even though I can't think of a situation where it would work I don't think it's harmful and it's possible there's some filesystem permission scenario that I'm missing where it could resolve the user's issue.

jmcphers: I really can't think of one. If you want to leave it in, could you add a comment explaining the reasoning above the change ownership call to save some future head-scratching?

This PR LGTM modulo said comment!

jmcphers (approved):

src/cpp/core/system/Xdg.cpp
MariaSemple: Done.

kfeinauer:
@MariaSemple I am seeing the more generic error page when testing out invalid permissions on the local directory:

Screen Shot 2020-10-21 at 10 23 26 AM

As a result, I'm unable to verify the ticket. Are you seeing the same behavior in the latest desktop?

PR #7756, opened 2020-09-09
Remove source code for unused Tests tab

gtritchie:
### Intent

There is source code for a "Tests" pane that would live in the Console panel. This was abandoned (test results are shown in the Build pane) but the source code was never removed.

### Approach

Remove the unused source code.

### QA Notes

Nothing to test here, source-code cleanup.


javierluraschi:
LGTM!

PR #7760, opened 2020-09-09
fix unit test hang analysis on macOS

gtritchie:
### Intent

When unit tests timeout (5 minutes), getting a stack trace of the hang on macOS wasn't working.

### Approach

- brew install pidof and gdb, and write necessary goo to .gdbinit needed on macOS
- pidof doesn't support fully qualified paths on macOS, so pass the basename only

### QA Notes

Purely a dev/build environment thing. No product or customer impact.


kfeinauer (approved):
Thanks for this!

PR #7765, opened 2020-09-10
auto-complete virtual environment names

kevinushey:
### Intent

Allow the autocompletion system to provide completion results for Python virtual environments, normally referenced via `use_virtualenv()`.

### Approach

Relatively straight-forward addition to the autocompletion system.

### QA Notes

Check that the completion results in `use_virtualenv(|` reflect the available Python virtual environments located at `~/.virtualenv`.


gtritchie (approved):
LGTM!

PR #7770, opened 2020-09-10
show correct symbol for Return key on Mac menus

### LINKED ISSUE #6524: incorrect symbol shown for Enter key in menus
### System details

    RStudio Edition : Desktop
    RStudio Version : v1.3.921
    OS Version      : macOS 10.15.3
    R Version       : 3.6.1

---

Screen Shot 2020-03-25 at 9 18 56 AM

Looks like this is true in RStudio v1.2 as well; however, the expected symbol is shown in RStudio v1.1.

Screen Shot 2020-03-25 at 9 19 42 AM

gtritchie:
### Intent

- Fixes #6524

Since RStudio 1.2 we've been showing the wrong symbol for the <kbd>Return</kbd> key in the Mac desktop IDE's menus. We were showing the symbol for <kbd>Enter</kbd>, instead.

### Approach

On Mac desktop only, notify Qt that we _really_ meant the <kbd>Return</kbd> key so it shows the expected symbol.

### QA Notes

Check that any command on Mac Desktop shows the expected symbol for commands using "Enter" in their shortcut, and that these keyboard shortcuts still work. In my testing with a full Mac keyboard, both the Enter and Return keys work as expected, but mainly we care about the Return key (the one on Mac laptops).

The code change is in Mac-only desktop code, and only impacts the main menu.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -47,6 +47,7 @@
 * R6Class method defintions are now indexed and accessible by the fuzzy finder (Ctrl + .)
 * The 'Preview' command for R documentation files now passes along RdMacros declared from the package DESCRIPTION file. (#6871)
 * Some panes didn't have commands for making them visible, now they do (#5775)
+* Show correct symbol for Return key in Mac menus (#6524)
 
 ### RStudio Server Pro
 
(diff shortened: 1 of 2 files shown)


kevinushey (approved):
LGTM!

PR #7782, opened 2020-09-14
look for versioned python interpreters (e.g. pythonx.y) on PATH

kevinushey:
### Intent

Currently, RStudio only discovers Python interpreters called `python` (or `python.exe`) on the PATH. However, multiple Python interpreters with versioned names (e.g. `python3.7`) may be available on the PATH as well. This PR makes sure such interpreters are discovered by RStudio.

### Approach

Rather than hardcoding the Python interpreter name, we list files in each directory on the PATH that match an appropriate regular expression.

Note that, in theory, one downside of this approach is that it could be slow when attempting to list files within a very large directory.

### QA Notes

Test that versioned Python executables are discovered in the Python dialog here:

Screen Shot 2020-09-14 at 10 58 33 AM

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/python/PythonInterpreterSelectionDialog.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/python/PythonInterpreterSelectionDialog.java
@@ -33,6 +33,7 @@ public class PythonInterpreterSelectionDialog extends ModalDialog<PythonInterpre
       setOkButtonCaption("Select");
       
       widgets_ = new WidgetListBox<PythonInterpreterListEntryUi>();
+      widgets_.setHeight("420px");
       widgets_.setAriaLabel("Python Interpreters");
       
       for (PythonInterpreter interpreter : JsUtil.asIterable(interpreters))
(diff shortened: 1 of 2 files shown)


src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/python/PythonInterpreterSelectionDialog.java
```diff
@@ -33,6 +33,7 @@ public PythonInterpreterSelectionDialog(final JsArray<PythonInterpreter> interpr
       setOkButtonCaption("Select");
       
       widgets_ = new WidgetListBox<PythonInterpreterListEntryUi>();
+      widgets_.setHeight("420px");
```
gtritchie, line 36: Is this related to this PR, or just a bit of UI fit and finish?

gtritchie (approved):
Minor curiosity question, but LGTM!

src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/python/PythonInterpreterSelectionDialog.java
kevinushey: Related to the PR, in that I saw that after letting more rows enter the UI it became overly tall, and so constraining the height helped it feel more natural here.

PR #7812, opened 2020-09-16
Provide meaningful name for source column landmarks and min/max buttons

### LINKED ISSUE #7805: Source columns have unreadable landmark names
### System details

    RStudio Edition : Server
    RStudio Version : 1.4.822
    OS Version      : any OS running RStudio Server
    R Version       : any

- start screen reader (I'm using VoiceOver on Safari)
- load RStudio Server
- add a couple source columns via "Add Source Column" command
- in screen reader, bring up list of Landmark regions (on VoiceOver, use VO+U and arrow keys)

- the source columns have names like "Source3ir06540t415 region" and "Sourcebaz9otp6zz75 region"

screenshot of VoiceOver landmarks rotor showing bogus source column names

- regions should have distinct but meaningful names

### Other symptoms
- The min/max buttons in the panel header also use this label, so if you tab to them you get "Minimize Source3ir06540t415", etc.

gtritchie:
### Intent

- Fixes #7805

The source columns and their min/max buttons had accessibility labels containing randomly generated letters and numbers. These need to contain meaningful (and unique) text for each source column.

### Approach

Generate an accessible name of the form "Source Column #" where # starts at 1 and increments each time a new source column is created during this instance of the IDE. Plumb that through creation and recreation of the source columns.

### QA Notes

With a screen reader running, create some source columns then list the landmarks. The source columns should now said "Source Column 1", "Source Column 2", and so on. Same with the min/max buttons on each source column: "Minimize Source Column 1" and so forth, when you get keyboard focus on them. The other landmarks are unchanged (including the standard Source pane, which remains "Source").

It is only necessary to test this with one screen reader (or it would be sufficient to view the page source and look for the phrase "Source Column" in various aria-labels).

- On VoiceOver (Mac) use VO+U then arrow keys to view the landmarks.
- On JAWS (Windows) turn on the virtual PC cursor then Ctrl+JAWS+R to list regions.
- NVDA (Windows) switch off focus mode (NVDA+spacebar), then NVDA+F7 and switch to "Landmarks" type

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/theme/PrimaryWindowFrame.java
+++ b/src/gwt/src/org/rstudio/core/client/theme/PrimaryWindowFrame.java
@@ -45,7 +45,7 @@ public class PrimaryWindowFrame extends WindowFrame
    public PrimaryWindowFrame(String title,
                              Widget mainWidget)
    {
-      super(title);
+      super(title, title);
       ThemeStyles styles = ThemeResources.INSTANCE.themeStyles();
 
       panel_ = new ClickFlowPanel();
@@ -88,12 +88,12 @@ public class PrimaryWindowFrame extends WindowFrame
    {
       subtitle_.setText(subtitle);
    }
-   
+
    public void addLeftWidget(Widget widget)
    {
       panel_.add(widget);
    }
-   
+
    private final DoubleClickState doubleClickState_ = new DoubleClickState();
    private final Label subtitle_;
    private final ClickFlowPanel panel_;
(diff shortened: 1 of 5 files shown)


melissa-barca:
My only concern here is that the numbers could get out of sync when a user removes columns. It doesn't look like we ever decrement `sourceColumnCounter_` so the numbers wouldn't be ordered when a user removes and adds columns. Of course if we did decrement it, the columns still would lose numeric order. Is this something we should be concerned about?

gtritchie:
> My only concern here is that the numbers could get out of sync when a user removes columns.

No, I don't think it's a major concern. In reality it is unlikely a non-sighted screen-reader user is going to use source columns as it would not provide any usability benefit I can think of; this is really just a dot-the-i's fix to maintain WCAG compliance.

melissa-barca (approved):
Ok, LGTM!

PR #7873, opened 2020-09-23
Keep VCS context menu visible

### LINKED ISSUE #7859: Contextual menu in git pane does not render correctly when put at the bottom
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.4.854
    OS Version      :  WINDOWS 10
    R Version       :  4.0.2

Layout should change and the git pane should be put at the bottom and not in the default at upper right. 

have a rather long list of file modified or untracked by git. 

Right click on a file (to revert for example)

When right clicking on a file, a contextual menu opens. 

If the file is on top, it prints ok. 
image

If in the middle, the box is reduced with a scroll
image

If the file is at the bottom, we can't even scroll and it impossible to use the contextual menu
image

The problem does not appear when git pane is upper right because it can overlay on the below pane it seems.

The contextual menu could be show **above** the pointer and not always **under** when we are near the edge of the pane.

gtritchie:
### Intent

- Fixes #7859 for both Git and SVN panes

### Approach

Created a helper method to calculate position of context menu so it displays above the clicked location if there isn't enough room below it. Same with left/right, if you click close to the right edge of window, it will now display to the left instead of being pushed out of the visible window.

### QA Notes

Follow steps in issue, confirm the menu now fully appears even when Git or SVN pane is in bottom region, and you are clicking close to the bottom of it.

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/widget/ThemedPopupPanel.java
+++ b/src/gwt/src/org/rstudio/core/client/widget/ThemedPopupPanel.java
@@ -99,6 +99,11 @@ public class ThemedPopupPanel extends DecoratedPopupPanel
          sizeToWindow(top, Style.Overflow.AUTO);
    }
 
+   public void setAutoConstrain(boolean autoConstrain)
+   {
+      autoConstrain_ = autoConstrain;
+   }
+
    // Size the table to the window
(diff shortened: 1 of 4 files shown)


adamconroy (approved):
Yes! Very necessary for desktop especially.

PR #7895, opened 2020-09-28
Prevent exceptions when attempting to focus a pane without tabs

### LINKED ISSUE #7880: Focus Next/Prev Pane command won't focus console-only panel
### System details

    RStudio Version : 1.4.869
    OS Version      : any
    R Version       : any

- Close all the optional tabs in the Console panel (e.g. Terminal, Jobs)
- Use Focus Next Pane (F6) or Focus Previous Pane (Shift+F6) to move focus between panes
- Watch what happens when you encounter the console

- Focus will not enter the Console from either direction

- Should be able to navigate to Console even when it's the only pane in the console panel

melissa-barca:
### Intent

Fixes #7880 

`Focus Next/Prev Pane` should be able to navigate to the Console pane even when the console is the only tab open in that pane. It should also skip over a tab panel if there are no tabs open in that panel.

### Approach

When Console is the only panel within its pane it doesn't follow its typical convention of being a WorkbenchTab. This panel cannot be closed but makes the panel appear empty. Instead of an exception, we need to default to displaying the Console.

In fixing this I realized that attempting to run these commands to navigate to an empty tabset threw a similar exception. I updated the code so it will now skip over empty tabsets and focus on the next/prev pane in reference to that the empty panel. 

### QA Notes

Text the Focus Next/Prev Pane command with a variety of different tabs open and closed. The commands are triggered by `F6` and `Shift + F6`. Workspace configuration can be updated from Global Options >> Pane Layout.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/ui/PaneManager.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/ui/PaneManager.java
@@ -667,7 +667,7 @@ public class PaneManager
       if (currentFocus == null)
          return;
 
-      focusWindow(getAdjacentWindow(currentFocus, false /* before */));
+      focusAdjacentWindow(currentFocus, false /* before */);
    }
 
    @Handler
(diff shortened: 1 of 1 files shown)


gtritchie (approved):
LGTM!

PR #7899, opened 2020-09-28
Update Zoom Source to include columns

### LINKED ISSUE #7801: Update zoom commands to account for multiple columns
Our `Zoom Left Column` command doesn't make sense when a user has multiple source columns open. We need to come up with a way where any column can be zoomed in on and rename `Zoom Left Column` to something more appropriate. There are a few different ways we can do this:

1. Create a new command for `Zoom Current Source Column` that is only active when a user is in a source column and will zoom the currently active column. Leave the existing commands as there are - possibly rename `Zoom Left Column` to `Zoom Left / Center Column`. This follows the design we used for focusing splitter positions in #7481.

2. Remove `Zoom Right Column` and `Zoom Left Column` and replace them with a single  `Zoom Current Column` which detects which column has focus and zooms it. This would put us in a better place if we decide to allow more panes/columns on the right, but requires the user to be more aware of what they have focused.

melissa-barca:
### Intent

Fixes #7801 

Enhances the `Zoom Source` command so it zooms in on the most recently accessed Source pane - this may be a Source Column.

### Approach

I updated `onLayoutZoomSource` to specify the active source column rather than the defaulting to the main column. While researching this PR I realized that our zooming animation code was unnecessary as `MainSplitPanel` has an animate function we can use to get the same effect so I removed a fair amount of the existing code as well.

### QA Notes

Open several source columns and move focus between them, then select 'Zoom Source'. The most recently selected column should be zoomed in on.


src/gwt/src/org/rstudio/studio/client/workbench/ui/PaneManager.java
```diff
@@ -925,128 +928,37 @@ else if (window.equals(getConsoleLogicalWindow()))
...
-         {
-            isAnimating_ = false;
             panel_.onSplitterResized(new SplitterResizedEvent());
-            super.onComplete();
             if (afterComplete != null)
                afterComplete.execute();
          }
-      };
+         
+         public void onLayout(Layer layer, double progress)
```
gtritchie, line 958: Is this empty method needed?

gtritchie (approved):
LGTM (one very minor question)

src/gwt/src/org/rstudio/studio/client/workbench/ui/PaneManager.java
melissa-barca: Yes because it's part of the `AnimationCallback` interface. It gets called at each step of animation but we don't need to do anything special for that here.

Issue #5979, opened 2020-01-08
Sometimes autocomplete results in duplication of the first few letters of the function name


ronblum:
### System details

    RStudio Edition : Desktop (and Server , I think)
    RStudio Version : 1.x
    OS Version      : Multiple
    R Version       : Multiple

### Steps to reproduce the problem

The issue is intermittent—I can't figure out a consistently reproducible example. But when it does happen:

1. In the console, type the beginning of a command. For example, `lengt`.
2. Hit <kbd>tab</kbd> for autocomplete.
3. Choose `length`.

### Describe the problem in detail

Sometimes, intermittently, the result is something like `lelength()`.

### Describe the behavior you expected

The result should be just the name of the function—in this example, `length()`.

- [x ] I have read the guide to submitting good bug reports at https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports . 
- [x ] I have installed the latest version of RStudio and confirmed that the issue still persists.
- [ ] If I am reporting a RStudio crash, I have included a diagnostics report. https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report
- [x ] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
Here's a way to more reliably reproduce the issue. (I think it's easier to reproduce when there is a bit of latency; e.g. when using RStudio Server.)

1. Type `rn` in the source editor.
2. Press <kbd>Tab</kbd> to request completions.
3. Quickly type `om`, before the completion list has a chance to display.
4. Accept the completion that is presented on the screen before it is dismissed.

In this case, I end up with the completion `rnrnbinom()` in the editor, rather than just `rbinom()`.

calpan:
This happens frequently for me with filename autocompletion when there is some noticeable lag between hitting Tab and seeing the popup list (possibly due to large number of files in target folders).

kevinushey:
Here's a way to reproduce this more consistently:

```
# simulate a 'laggy' autocompletion request
trace(.rs.rpc.get_completions, quote(Sys.sleep(1)), print = FALSE)
```

Then, type the following:

```
stats::ac<TAB>
```

Wait for the autocompletion list to pop up, and then accept the first completion result. You should see something like the following:

![acf](https://user-images.githubusercontent.com/1976582/94471716-78da1f80-017e-11eb-93c9-1882d649be04.gif)

kevinushey:
[...]

PR #7900, opened 2020-09-28
fix issues with completion insertion in high-latency environments

kevinushey:
### Intent

In some cases (especially in high-latency environments), attempts to insert completions would produce a "wrong" completion result. For example:

- The text would be duplicated; producing `stats::ac` -> `stats::acacf()` rather than `stats::acf()`;
- The wrong portion of text would be replaced; e.g. `stats::rn` -> `stats:rnorm()` rather than `stats::rnorm()`.

In particular, the issue normally occurs when:

1. The user requests completions (either automatically, or explicitly via TAB);
2. The user immediately begins typing new characters,
3. The completion response is received, and the popup is shown.

In particular, the completion engine was not handling the new characters added (or removed) in (2) cleanly.

### Approach

The issue was caused by some incongruences in the autocompletion context being used. New completion requests are triggered somewhat aggressively as the user types, and the completion manager tracks a single "active" completion context. Unfortunately, the "active" completion context was being updated too soon, so in the case where:

1. The user is being presented with completion results (in a popup) as part of the current set of completions,
2. A new completion request has been sent off,
3. The response from completion request (2) has not yet been received.

the "active" completion context would represent the state from (2) rather than (1), and so the completion insertion would be incorrect.

### QA Notes

Test via cases in https://github.com/rstudio/rstudio/issues/5979. Note that both R and Python scripts should be tested independently, as they use different completion manager implementations. (Hopefully that will change in the future.)

Closes https://github.com/rstudio/rstudio/issues/5979.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/console/shell/assist/CompletionManagerBase.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/console/shell/assist/CompletionManagerBase.java
@@ -129,6 +129,9 @@ public abstract class CompletionManagerBase
       if (positionChanged)
          return;
       
+      // cache context data (to be used by popup during active completion session)
+      contextData_ = data;
+      
       String line = data.getLine();
       if (completions.isCacheable())
          completionCache_.store(line, completions);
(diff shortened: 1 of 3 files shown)


kevinushey:
Merging this now so we can get testing on master. Please review post-hoc if you have a chance!

Issue #7060, opened 2020-06-08
Code completion bug: Function argument suggestion for R functions does not work anymore in v1.3


tungmilan:
Hello,

I noticed that in v1.3, function argument suggestion no longer works after the first argument of a function. Could you check if it's the case for you guys too? Thanks!

### System details

    RStudio Edition : Desktop
    RStudio Version : Version 1.3.1008
    OS Version      : Windows 10 x64 (build 16299)
    R Version       : R version 4.0.1 (2020-06-06)

### Steps to reproduce the problem

```r
library(tidyr)

iris_long <- iris %>% 
  pivot_longer(names_to = 'variable',)
```

### Describe the problem in detail

After the first comma `,`, function argument suggestion stopped working 

![](https://i.imgur.com/u9X3iV7.png)

### Describe the behavior you expected

argument suggestion should work after each comma `,`  (like in v1.2.5042)

![](https://i.imgur.com/2LCkpZN.png)

- [x] I have read the guide to submitting good bug reports at https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports . 
- [x] I have installed the latest version of RStudio and confirmed that the issue still persists.
- [ ] If I am reporting a RStudio crash, I have included a diagnostics report. https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
I see the same behavior with RStudio 1.2.5042, so this doesn't appear to be a regression.

tungmilan:
Thanks for checking @kevinushey ! So it was because of some packages got updated? Here is my session info that still has argument suggestion working

````
- Packages -------------------------------------------------------------------------------------------------------------
 package     * version     date       lib source                          
 assertthat    0.2.1       2019-03-21 [1] CRAN (R 3.6.1)                  
 backports     1.1.6       2020-04-05 [1] CRAN (R 3.6.2)                  
 callr         3.4.3       2020-03-28 [1] CRAN (R 3.6.3)                  
 cli           2.0.2       2020-02-28 [1] CRAN (R 3.6.3)                  
 crayon        1.3.4       2017-09-16 [1] CRAN (R 3.6.1)                  
 desc          1.2.0       2018-05-01 [1] CRAN (R 3.6.1)                  
 devtools      2.3.0       2020-04-10 [1] CRAN (R 3.6.2)                  
[...]

PR #7906, opened 2020-09-29
fix autocompletion of formals for piped-to S3 generics

kevinushey:
### Intent

Autocompletion of function argument names could fail when piping in an object to an S3 generic function.

### Approach

We normally try to retrieve function arguments from two locations:

1. The S3 method that would normally be dispatched to, if we can determine which method that is;
2. The S3 generic itself.

However, in the case where our attempts to find a method in (1). fail, we would end up with no completion results at all. This PR alleviates that by falling back to (2) when (1) fails.

### QA Notes

Test via example described in https://github.com/rstudio/rstudio/issues/7060. Be sure that the `tidyr` package is installed and loaded when testing.


kevinushey:
Merging now to get testing on master -- please review post-hoc.

PR #7909, opened 2020-09-29
Prune overflow data in Virtual Scroller

### LINKED ISSUE #7712: Virtual Scroller enabled console shows truncated content
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.4
    OS Version      :  Any
    R Version       :  Any

1. Ensure that the Virtual Scroller feature is turned on in Global Options ("Limit console output to a subset...")
2. In the console perform the command as shown here: https://github.com/rstudio/rstudio/pull/7568#issuecomment-675010330
3. Note the oddly truncated content

The data that is coming up from the server is coming in odd, non-contiguous chunks (this can be seen in the network pane). Even going to the `ConsoleProcess::onStdout` function shows that the truncation is happening a bit deeper down the line.

Ideally we get ALL of the data but there are incredibly degenerate cases where it'd lock up RStudio (for i in 1:99999999 etc) which is likely why we started doing this chunking in the first place. In which case if we could get it just to send the final chunk instead of disparate chunks that are intended to be trimmed anyway that would mimic the current functionality.

adamconroy:
### Intent

Fixes #7712 

Currently, only in Desktop mode as far as I can tell, the server doesn't always accurately prune the data as per the 1000 line single element hard cap that it thinks it's doing. This difference can be observed by watching the `set_client_state` updates. Even when the server is being overzealous in Desktop mode the VirtualConsole will prune this information so it matches the server.

### Approach

The VirtualScroller will now prune elements as they're added as well as after a full submit loop in the VirtualConsole is run.

### QA Notes

Ensure that we're only seeing roughly 1000 lines within a single element as per the example in the issue above. This should match the functionality of having the VirtualConsole disabled as well ("Limit console output to a subset of content" setting).

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/VirtualConsole.java
+++ b/src/gwt/src/org/rstudio/core/client/VirtualConsole.java
@@ -648,8 +648,11 @@ public class VirtualConsole
       }
 
       Entry<Integer, ClassRange> last = class_.lastEntry();
-      ClassRange range = last.getValue();
-      if (isVirtualized()) VirtualScrollerManager.prune(parent_.getParentElement(), range.element);
+      if (last != null)
+      {
+         ClassRange range = last.getValue();
+         if (isVirtualized()) VirtualScrollerManager.prune(parent_.getParentElement(), range.element);
+      }
 
       // If there was any plain text after the last control character, add it
       if (tail < data.length())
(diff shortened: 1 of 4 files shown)


gtritchie (approved):
LGTM! Are the GWT unit tests still passing (always paranoid about anything touching the VirtualConsole).

adamconroy:
Good call, the tests caught an issue where invalid data would throw an NPE. :) All passing now.

PR #7915, opened 2020-09-29
Prevent excessive set_client_state calls

### LINKED ISSUE #7785: excessive 'set_client_state' RPCs sent periodically
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.4.813
    OS Version      : macOS 10.15.6
    R Version       : 4.0.2

Start the IDE, and leave it idle for a little while. Then, open the request log.

A large number of `set_client_state` RPCs are made, evidently to update the window splitter position. However, the splitter position is not actually changed for each call.

Screen Shot 2020-09-14 at 1 32 53 PM

We should avoid calling `set_client_state` when the client state has not changed.

Note that there are no ill effects from this issue; it's ultimately just to reduce overall chattiness between the client and server.

melissa-barca:
### Intent

Fixes #7785 

We should only send `set_client_state` when a state has actually changed. This PR fixes two bugs where we were updating the workbench splitter position's state and the active source column's state when they hadn't actually changed. 

### Approach

One of the logic statements for checking the splitter position was reporting a change after it confirmed no change had occurred - it was missing a `!` in front of the check. 

`SourceColumnManager` was creating a new state every time it checked for a change in the active column, rather than when the active column changed. 

### QA Notes

See repro notes on original issue.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/ui/MainSplitPanel.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/ui/MainSplitPanel.java
@@ -93,7 +93,7 @@ public class MainSplitPanel extends NotifyingSplitLayoutPanel
          if (a.hasSplitterPos() ^ b.hasSplitterPos())
             return false;
          if (a.hasSplitterPos() &&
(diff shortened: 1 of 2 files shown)


kevinushey (approved):
LGTM!

PR #7916, opened 2020-09-29
annotate inotify errors with better message

kevinushey:
### Intent

When an inotify error occurs, the generic "ENOSPC" error code is returned, which is associated with the "No space left on device" error message. That message is misleading for users; it hints that the system might be low on disk space when the reality is that the system has been exhausted of inotify watches.

### Approach

This PR adds a new error constructor `syscallError()`, which allows the caller to both hint at the failing function, and provide their own message to provide rather than the default message associated with the particular error code.

### QA Notes

Not exactly sure how to test beyond finding a way to reproduce https://github.com/rstudio/rstudio-pro/issues/1988.

### DIFF
--- a/src/cpp/shared_core/include/shared_core/Error.hpp
+++ b/src/cpp/shared_core/include/shared_core/Error.hpp
@@ -710,6 +710,36 @@ Error systemError(
    const ErrorLocation& in_location);
 
 /**
+ * @brief Function which creates a system error associated with a particular system call.
+ *
+ * @param in_function        The error message to associate with this error.
+ * @param in_code            The error code (typically errno).
+ * @param in_location        The location of the error.
+ *
+ * @return A system error.
+ */
+Error syscallError(
+   const std::string& in_function,
+   int in_code,
+   const ErrorLocation& in_location);
+
+/**
+ * @brief Function which creates a system error associated with a particular system call.
+ *
+ * @param in_function        The error message to associate with this error.
+ * @param in_code            The error code (typically errno).
+ * @param in_message         The error message. Overrides the default error message associated with the error code.
+ * @param in_location        The location of the error.
+ *
+ * @return A system error.
+ */
+Error syscallError(
+   const std::string& in_function,
+   int in_code,
+   const std::string& in_message,
+   const ErrorLocation& in_location);
+
+/**
  * @brief Function which creates an unknown error. This should be used only when a specific error code cannot be
  *        determined.
  *
@@ -735,6 +765,8 @@ Error unknownError(const std::string& in_message, const ErrorLocation& in_locati
  */
 Error unknownError(const std::string& in_message, const Error& in_cause, const ErrorLocation& in_location);
 
+// returns the error message associated with a particular system error code
+std::string systemErrorMessage(int code);
 
 // return a printable error message from an error (depending on the error this
 // might require consulting the message, category, or name)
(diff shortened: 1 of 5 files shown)


src/cpp/shared_core/include/shared_core/Error.hpp
```diff
@@ -709,6 +709,36 @@ Error systemError(
...
+/**
+ * @brief Function which creates a system error associated with a particular system call.
+ *
+ * @param in_function        The error message to associate with this error.
+ * @param in_code            The error code (typically errno).
+ * @param in_location        The location of the error.
+ *
+ * @return A system error.
+ */
+Error syscallError(
```
MariaSemple, line 721: Could we call this `systemCallError` instead?

src/cpp/shared_core/include/shared_core/Error.hpp
```diff
@@ -709,6 +709,36 @@ Error systemError(
    const Error& in_cause,
    const ErrorLocation& in_location);
 
+/**
+ * @brief Function which creates a system error associated with a particular system call.
+ *
+ * @param in_function        The error message to associate with this error.
```
MariaSemple, line 715: I think the description here would be more accurate if it were something like `The system function for which this error occurred.`

src/cpp/shared_core/include/shared_core/Error.hpp
```diff
@@ -735,6 +765,8 @@ Error unknownError(const std::string& in_message, const ErrorLocation& in_locati
  */
 Error unknownError(const std::string& in_message, const Error& in_cause, const ErrorLocation& in_location);
 
+// returns the error message associated with a particular system error code
+std::string systemErrorMessage(int code);
```
MariaSemple, line 769: Can this be entirely defined in the cpp file? Once I pull these changes to the SDK (and release it) it will become very difficult to make signature changes to any functions in public headers (`Error.hpp` is a public header in the SDK).

MariaSemple (changes requested):
Looks good to me! Just a couple of stylistic nitpicks in the shared_core code

src/cpp/shared_core/include/shared_core/Error.hpp
kevinushey: You're right -- I had forgotten to update the text here; thanks for catching that.

src/cpp/shared_core/include/shared_core/Error.hpp
kevinushey: I've made a compromise here:

- `Error.cpp` (in shared_core) gets its own inline version, just so that we avoid changing the public interface for the SDK;
- `System.hpp` (in core) gets a public version, which can be used by other 'core' utilities (and so `LinuxFileMonitor`).

Does that sound reasonable?

MariaSemple: Ah - I missed that usage in `LinuxFileMonitor`, my mistake. I think it's fine to leave it the other way. I should be able move the declaration to the cpp file with my `pull-shared-core.sh` script. I believe I'm doing that for another function elsewhere already.

kevinushey: Done -- thanks!

kevinushey:
@MariaSemple should be ready for final review -- thanks!

MariaSemple (approved):
LGTM!

PR #7942, opened 2020-10-01
Show debug traceback while VirtualScroller is turned on

### LINKED ISSUE #7928: Console traceback not always appearing
### System details

    RStudio Version : 1.4.887
    OS Version      : any
    R Version       : any

- In 1.4, turn on the virtual scroller (Global Options / Console / Limit console display to a subset of total content)
- Create an R script with below code and source it. Then execute `f2()` again in the console.

```
f1 <- function() {
  stop("Inspect me.")
}

f2 <- function() {
  f1()
}

f2()
```
- Plain text results for the error, no traceback UI
- With the virtual scroller mode OFF, sometimes the traceback UI appears, sometimes it doesn't (e.g. try running `f2()` over and over via up-arrow + Enter); haven't figured out the pattern here seems very inconsistent (timing?)

screenshot of console showing incorrect traceback UI

Console should show the traceback UI for the initial source and for the separate `f2()` invocation as seen here in 1.3.1093:

screenshot of console showing correct traceback UI instances

adamconroy:
### Intent

Fix traceback showing up while the VirtualScroller is turned on. Also fix an out of bounds error that existed before the VirtualScroller was added.

Fixes #7928 

### Approach

The ConsoleError element is its own VirtualConsole that cannot and should not be virtualized. Luckily the flag that I added for disabling that functionality came through in a pinch here. 

I also found an out of bounds error that would cause the traceback to be added to a floating element outside of the error message itself or, in the case where there was not later element to be found, throw a null pointer error and not display at all. This was unrelated to VirtualScroller.

### QA Notes

Ensure that the traceback shows properly when the VirtualScroller is both on and off, as shown in the issue linked above.

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/VirtualConsole.java
+++ b/src/gwt/src/org/rstudio/core/client/VirtualConsole.java
@@ -488,7 +488,7 @@ public class VirtualConsole
    public void submit(String data, String clazz, boolean forceNewRange, boolean ariaLiveAnnounce)
    {
       // Only capture new elements when dealing with error output, which
-      // is only place that sets forceNewRange to true. This is just an
+      // is the only place that sets forceNewRange to true. This is just an
       // optimization to avoid unnecessary overhead for large (non-error)
       // output.
       captureNewElements_ = forceNewRange;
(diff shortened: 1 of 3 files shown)


gtritchie (approved):
Whew, was worried this would be a bigger nightmare! LGTM.

Issue #2007, opened 2018-01-18
RStudio keep hanging when "Quit Session" or saving file 


abbatidanilo:
RStudio (RStudio-1.1.414) keeps on hanging on most principal actions, for example simply quitting or opening existing files. In particular I am dealing with RMarkdown projects. I tried with different R version shifting from the latest version 3.4.3 down to 3.3.3 but with no appreciable improvement.

jmcphers:
Sorry you're having trouble! We have not been able to reproduce this, unfortunately; could you include some additional details about your system? See our contribution guide for what to include:

https://github.com/rstudio/rstudio/blob/master/CONTRIBUTING.md

abbatidanilo:
No problem Jonathan!
My fault instead: I just realized what a poor subject I put for to describe the issue...
In any case, I was in a hurry so I preferred to rollback to a previous  edition of RStudio (Version 1.0.153) because I realized than the issue persists even with older version of the R compiler.
Pity I removed the current edition with all the relevant diagnostic. I promise: I won't do it again b4 saving the sessionInfo()!
Now I am using RStudio ver 1.0.153 with R3.3.3 and it works much better  (the ones I used some weeks ago).
I do not know if others are facing the same problem (I discovered the same kind of issues described in the past <https://stackoverflow.com/questions/34532328/couldnt-quit-r-session-or-couldnt-quit-rstudio> - but I thought it was too old...)

cheers

jmcphers:
Okay, I don't think we have enough to go on here -- but if you do start seeing it again and have time to collect more diagnostics, feel free to reopen.

ronblum:
[Combing with #2036]

> Hello,
RStudio keeps on hanging when you try to quit or to save a file. In particular I am writing a RMarkdown note with nothing especial on it.
The issue happens some time using RStudio-1.0.153 but much often when using the most recent RStudio-1.1.414. I tried with the most recent R compiler 3.4.2 but I also downgrade to 3.4.0 and 3.3.3 (which is the one I currently using). I cannot say if this depends on the R compiler or the RStudio version itself, but it seems the older, the better...
I opened the same issue last week but in that case I did not send any detail. Here at least I managed to have the diagnostics-report.
I cannot provide the sessioninfo() because when it hangs, the only way is to kill the process.
OS: Windows 10 pro on an Intel i5 2540M CPU
[diagnostics-report.txt](https://github.com/rstudio/rstudio/files/1661561/diagnostics-report.txt)

[...]

PR #7951, opened 2020-10-02
Fix height for html chunk callbacks 

melissa-barca:
### Intent

Fixes pro-issue #2007 https://github.com/rstudio/rstudio-pro/issues/2007

The height used for the iframe displaying html from a chunk callback should be based on the entire document.

### Approach

Previously this height was based on the html body which was leading to the callback displaying with too short of a height.


kevinushey (approved):
LGTM!

Issue #3448, opened 2018-09-09
Cannot Exit readLines(file("stdin")) in Windows


billdenney:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.1.423 
    OS Version      : Windows 10
    R Version       : 3.5.1

### Steps to reproduce the problem

Run the following in the console:

``` r
readLines(file("stdin"))
```

### Describe the problem in detail

I cannot exit this.  `Ctrl+Z` appears to be bound to undo, so I cannot send an end file that way, and I found no other way to exit the readLines prompt.

### Describe the behavior you expected

Either return `character(0)` as occurs in RGui or have some way to exit.

kevinushey:
Reading from stdin is not currently supported.

What do you want access to `stdin` for? If you need to read user input, you should be able to use `readline()`.

billdenney:
@kevinushey, I was running an example that I found online.  I don't actually need to read `stdin` in general.

The issue is that when I ran `readLines(file("stdin"))` I could not do anything in the Rstudio console.  I had to restart the R session.  The issue that I'd like to have fixed is that the command above should not block the console.  Could `stdin` handling be updated so that `readLines(file("stdin"))` just returns `character(0)` like Rgui does?

ronblum:
Marking as a bug—even though it's unsupported, there is no way to get out of this.

chasemc:
I ran into this today. 

Accidentally typed something like `lapply(zz, readLines()) ` instead of `lapply(zz, readLines) `. Only option was to restart R. Could be annoying to some as there is no error or warning; R just hangs.  This behavior is not consistent with RGui.exe or Rscript.exe, but is consistent with R.exe.

zerweck:
Is this on anyone's radar? I just killed a huge session with many objects in the RAM just because I accidentally used the wrong autocomplete option (just like @chasemc) . Very annoying and absolutely unclear from the user perspective when it happens.

krzyklo:
Run into this trap as well = Restart R

billdenney:
Thanks!

krzyklo:
superb! Thank you

PR #7982, opened 2020-10-06
make it possible to interrupt readLines(stdin())

kevinushey:
### Intent

Currently, when R requests input from the user via `readline()` or `readLines(stdin())`, there is no way to interrupt the read -- the R session is effectively stuck and the user is forced to restart the session.

This PR makes it possible to exit that state.

Closes https://github.com/rstudio/rstudio/issues/3448.

### Approach

Previously, we tried to "guess" whether the appropriate action was to send a "true" process interrupt, or to try and send a "console" interrupt, based on the state of the prompt. This fails when R attempts to read input from the user without setting a prompt, as in `readLines(stdin())`.

To properly handle this, we now:

1. Always send a so-called "interrupt" RPC;
2. If the R session appears to be busy, then we interrupt it; if not, then we send a console-level interrupt.

### QA Notes

Test that the following states can all be interrupted (by pressing <kbd>Escape</kbd> while the Console has focus) on different platforms:

```
Sys.sleep(10)
system("sleep 10")
readLines(stdin())
readline()
readline("Enter input: ")
```

### DIFF
--- a/src/cpp/r/session/RStdCallbacks.cpp
+++ b/src/cpp/r/session/RStdCallbacks.cpp
@@ -76,9 +76,6 @@ InternalCallbacks s_internalCallbacks;
 // temporarily suppress output
 bool s_suppressOutput = false;
 
-// temporarily suppress interrupt output
-bool s_suppressNextInterruptOutput = false;
-
 class JumpToTopException
 {
 };
@@ -392,7 +389,6 @@ int RReadConsole (const char *pmt,
       // handle the interrupt as we noticed calling Rf_onintr() in that
       // environment could cause a crash
 #ifndef _WIN32
-      s_suppressNextInterruptOutput = true;
       Rf_onintr();
 #endif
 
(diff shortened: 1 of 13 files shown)


src/cpp/r/session/RStdCallbacks.cpp
```diff
@@ -426,8 +426,20 @@ void RWriteConsoleEx (const char *buf, int buflen, int otype)
    {
       if (!s_suppressOutput)
       {
+         bool isInterruptOutput =
```
kevinushey, line 429: This is sort of a hack based on how we draw the prompt. In a regular R session, when an interrupt is issued, R writes a single newline:

https://github.com/wch/r-source/blob/819477cd6dbefb8fae61a07fc13064faa29a2714/src/main/errors.c#L188

with the intention of having that ensure the input prompt is then drawn on its own line. The net result is that interrupting a command gives you a "new" prompt to work on.

RStudio instead tries a little harder to ensure the prompt is always shown on its own line, which means an extraneous newline is written to the console history in these cases. Hence, this is ultimately a hack to ensure that RStudio's output more closely mimics that of R.

jmcphers (approved):
LGTM!

Issue #7920, opened 2020-09-30
FR: "open in a new column" for files


ColinFay:
It would be awesome to be able to do "Open in a new column" when selecting one or more documents in the file pane. 

For example here, a "Open in a new column" would be very handy: 

![Screenshot 2020-09-30 at 16 05 58](https://user-images.githubusercontent.com/17936236/94696042-ec269180-0336-11eb-86b0-fd6d0488832f.png)

ronblum:
@ColinFay Thank you for the suggestion. We'll review this as we continue to work on improving RStudio. I hope this means that you're getting some good use of the multiple source columns (new in 1.4 Wax Begonia).

valerie-rstudio:
### Verify Failed

    RStudio Edition : Server
    RStudio Version : 2021.12.0-daily+80
    OS Version      : ubuntu20
    R Version       : 4.0.3

### Summary
A blank column is opened if an already open file is selected by "Open File in New Column".

### Repro Steps
1. Create a new R Script. Save and name it. Keep the file open in the editor.
2. Click File > Open File in New Column.
3. Select the newly created R Script that is already open in the original editor column.
4. **Expected**: No new column is created.
5. **Actual**: New blank column is created. File remains open in original column.

### Notes
The "Open Each File in New Columns" option from the Files pane follows the expected results.

<hr>

### Verified Fixed
Verified that...
- The "Open File in New Column" option can open any file type in a new column
- The "Open File in New Column" option no longer appears in the File menu once the maximum number of columns is reached
- There is no default keyboard shortcut for the "Open File in New Column"
- The "Open Each File in New Columns" option from the Files pane can open any file type in a new column
- The "Open Each File in New Columns" option from the Files pane opens each selected file in a new column. Once the maximum number of columns is reached, additional files are opened in the last opened column
- The "Open Each File in New Columns" option from the Files pane opens files in the order they are listed in the Files pane.
- The "Open Each File in New Columns" option from the Files pane does not open a blank column if an already opened file is selected

GCRev:
@valerie-rstudio That's the main menu File button that you're referencing right? 

If that's the case, then I think we need a new issue opened for that because its behavior was not modified for this feature, which only targeted the Files pane.

valerie-rstudio:
Opened a new ticket for the issue mentioned above https://github.com/rstudio/rstudio/issues/9891. The remainder of the functionality has been verified.

PR #7994, opened 2020-10-07
Open source doc in new column

melissa-barca:
### Intent

Allow user to open a document in a new column, rather than columns always opening with a blank doc. This is based on #7920. 

### Approach

I added a new command `openSourceDocNewColumn` that mimics the existing `openSourceDoc` command. When opening the file chooser the command creates a new column and sets it to the active column, when the user selects a file the process is the same as with `openSourceDoc` where the file is opened in the active column. If the user cancels out of the file chooser, the new column is removed. 

I looked into what it would take to open files in a new columns from the Files Pane and it ends up being more complicated as it revolves around open file events which differ between file types. I implemented being able to `Shift+Click` a file to open it in a new column on a separate branch but ultimately decided that feature is too risky for this to make it into 1.4.

I also included some general code clean up (removed unused functions and parameters, fixed spacing)
### QA Notes

You should be able to open any file type in a new column, but there should only ever be three columns open at a time. An error will be displayed if the user attempts to open a fourth column. 

image


src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml
```diff
@@ -777,6 +778,10 @@ well as menu structures (for main menu and popup menus).
          <shortcut refid="openSourceDoc" value="Ctrl+O" disableModes="emacs"/>
          <shortcut refid="openSourceDoc" value="Meta+O" if="org.rstudio.core.client.BrowseCap.isMacintosh()"/>
 
+         <shortcut refid="openSourceDocNewColumn" value="Ctrl+X Ctrl+C" disableModes="default,vim,sublime"/>
+         <shortcut refid="openSourceDocNewColumn" value="Ctrl+Shift+O" disableModes="emacs"/>
+         <shortcut refid="openSourceDocNewColumn" value="Meta+Shift+O" if="org.rstudio.core.client.BrowseCap.isMacintosh()"/>
```
gtritchie, line 783: On Mac (maybe others, didn't check) conflicts with toggleDocumentOutline shortcut.

gtritchie (changes requested):

src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml
melissa-barca: Thanks! I've changed it.

gtritchie (approved):
LGTM!

src/gwt/src/org/rstudio/studio/client/workbench/ui/PaneManager.java
```diff
@@ -674,18 +675,51 @@ public void onFocusPreviousPane()
...
+         source_.openSourceDoc(onCancelled, onCompleted);
+      }
+   }
 
+   private boolean validateNewColumnRequest()
+   {
+      if (!userPrefs_.allowSourceColumns().getValue())
+      {
+         pGlobalDisplay_.get().showMessage(GlobalDisplay.MSG_INFO, "Cannot Add Column",
+            "Allow Source Columns preference is disabled.");
```
jmcphers, line 711: This command needs to be disabled, and probably hidden, when the `allowSourceColumns` preference prevents it from working. I'd suggest handling that in the source column manager (i.e. bind a preference listener that sets the command to be enabled/visible based on the value of the pref).

src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml
```diff
@@ -777,6 +778,10 @@ well as menu structures (for main menu and popup menus).
          <shortcut refid="openSourceDoc" value="Ctrl+O" disableModes="emacs"/>
          <shortcut refid="openSourceDoc" value="Meta+O" if="org.rstudio.core.client.BrowseCap.isMacintosh()"/>
 
+         <shortcut refid="openSourceDocNewColumn" value="Ctrl+X Ctrl+C" disableModes="default,vim,sublime"/>
```
jmcphers, line 781: If this is meant to be an Emacs-like keybinding for the command, maybe we should use the closest emacs keybinding? I think C-x C-c means something else in Emacs, and we should use something like `C-x 5 f` for this command (but we should check with @kevinushey).

jmcphers (changes requested):

src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml
melissa-barca: it looks like `C-x 5 f` is taken in emacs as well. @kevinushey  do you have a recommendation?

src/gwt/src/org/rstudio/studio/client/workbench/ui/PaneManager.java
melissa-barca: Good point, I've added a listener to `SourceColumnManager` to handle this as well as updated PaneManager to hide the commands when the max column count is reached.

melissa-barca:
I added the image from `openSourceDoc` to this command as well. Here is the updated screenshot: 
image

src/gwt/src/org/rstudio/studio/client/workbench/ui/PaneManager.java
```diff
@@ -382,6 +382,14 @@ else if (numDocs > 0 && window.getState() == WindowState.HIDE)
          hiddenTabs_ = tabNamesToTabs(evt.getValue().getHiddenTabSet());
          populateTabPanel(hiddenTabs_, hiddenTabSetTabPanel_, hiddenTabSetMinPanel_);
 
+         // manage source column commands
+         boolean visible = userPrefs.allowSourceColumns().getValue() &&
+            (userPrefs.panes().getValue().getAdditionalSourceColumns() < MAX_COLUMN_COUNT);
+         
+         commands_.newSourceColumn().setVisible(visible);
+         commands_.openSourceDocNewColumn().setVisible(visible);
+         commands_.focusSourceColumnSeparator().setVisible(visible);
```
jmcphers, line 391: Here and below: it makes sense to disable the first two commands if we're already maxed out for columns (since they create columns), but it seems to me like the third shouldn't be affected. What do you think?

src/gwt/src/org/rstudio/studio/client/workbench/views/source/SourceColumnManager.java
```diff
@@ -242,6 +243,23 @@ public void onSwitchToDoc(SwitchToDocEvent event)
          }
       });
 
+      userPrefs.allowSourceColumns().addValueChangeHandler(event ->
+      {
+         // PaneManager helps manage these commands so when modifying this code, update
```
jmcphers, line 248: It's not totally clear to me why we need this code in both places ... doesn't this handler run at startup? (if not, maybe make a function that's called both at startup and in the value change handler?)

jmcphers (changes requested):

src/gwt/src/org/rstudio/studio/client/workbench/views/source/SourceColumnManager.java
melissa-barca: PaneManager just needs to help with commands that need to be adjusted based on the max column count. It tracks this because it's actually responsible for adding and removing the windows that contain the columns. I'll update the comment to be more specific. I was grouping all the commands into that category when I shouldn't have been.

src/gwt/src/org/rstudio/studio/client/workbench/ui/PaneManager.java
melissa-barca: You're right, fixed it.

jmcphers (approved):
This LGTM, but please hold off on merging until you fix the Emacs shortcut. C-x C-c is the shortcut for exiting Emacs so opening a file in a tab would definitely not be what Emacs users expect!

jmcphers:
I also think it'd also be okay to leave this command fully unbound (i.e. no default shortcut in any mode) if we can't find a suitable binding, as keyboard shortcuts are in very short supply and this is not likely to be a high frequency command, plus users who do use this command a lot can always add their own binding.

melissa-barca:
Ok, I've removed the keyboard shortcut. Thanks@

PR #8000, opened 2020-10-07
fix svn commit message encoding issue

### LINKED ISSUE #7959: RStudio wrongly encodes SVN commit messages on Windows (UTF-8 issue)
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.3.959
    OS Version      : Windows 10
    R Version       : 4.0.2

1. Create a new SVN repository. From the command-line (`cmd`), use the SVN command-line client: `svnadmin create testrepo`.
2. Checkout the SVN repository using RStudio ‘File | New Project | Version Control | Subversion’. The repository URL will look something like this if the repo was placed in `c:\folder\`: `file:///c:/folder/testrepo`.
3. Create and save a new file in the project, and then use ‘Tools | Version Control | Commit’ to commit it. Use the commit messages `RStudio: æøå`.
4. Use ‘Tools | Version Control | History’ to view the commit log.

When committing something to SVN, RStudio on Windows uses the wrong character encoding in the commit message.

Actually, the commit message is *double-encoded*. For example, when the commit message (as written in the commit message window) is:

    RStudio: æøå

It ends up as the following in the SVN repo:

    RStudio: Ã¦Ã¸Ã¥

This is the original string with the following operations performed:

  1. The original text is converted to UTF-8 (where the letters æ, ø and å ends up taking up *two* bytes each).
  2. The byte sequence from step 1 is (wrongly) *interpreted* as being in the Windows-1252 encoding (default character encoding on (Western) Windows systems). So, for example, the byte sequence for ‘æ’ becomes the two *characters* ‘Ã¦’.
  3. The result is stored as a commit message (I guess as an UTF-8 byte sequence, so the ‘Ã¦’ is stored as four bytes, two for each character).

Note that when *reading* the commit message, e.g., using the ‘History’ feature, RStudio does the right thing. So the original (wrongly-encoded) message ends up being (correctly) shown as:

    RStudio: Ã¦Ã¸Ã¥

To confirm this, I have tested the following (in the same repository):

1. Add a commit using the command-line client on Linux (UTF-8 encoding).
2. Add a commit using the command-line client on Windows (Windows-1252 encoding, though I think the client internally converts it to UTF-8).
3. Add a commit using RStudio on Windows.

In all cases, ‘svn log’ on Windows and the ‘History’ feature in RStudio on Windows shows the same commit message for each commit, which is correct for 1 and 2, and wrong for 3. Since 1 *is* correctly stored, 2 must be so too, and 3 must be incorrectly stored. That is, RStudio wrongly *encodes* commit messages for SVN on Windows. (This is, of course, only visible for non-ASCII commit messages, since ASCII is a subset of UTF-8).

The commit message should have been correctly encoded instead of double-encoded. So when viewing the commit log (either using RStudio or the SVN command-line client), it should look like `RStudio: æøå` instead of `RStudio: Ã¦Ã¸Ã¥`.

- [ ] I have installed the latest version of RStudio, and confirmed that the issue still persists. [I don’t have admin access on my (work) Windows PC, so I can’t install later versions of RStudio. But I have checked that there have been no SVN-related changes in RStudio versions later than the one I tested.]

gtritchie:
### Intent

- Fixes #7959
- On Windows, an SVN commit message containing non-ASCII characters was being mangled by a Windows encoding conversion.
- NOTE: this appears to have always been broken, so if we have any concerns about risk (or testing cost) we could consider punting this to 1.5

### Approach

- svn works internally in utf-8, and we don't do any conversion of the (utf-8) string we get from the user, so make sure it doesn't get converted to another locale on Windows by specifying that it is already utf-8

### QA Notes

Test scenario described in issue. I used https://tortoisesvn.net/ to install svn on Windows (be sure to include the command-line tools, which aren't installed by default).

Note that this change doesn't fix previously mangled commit messages, only new ones. Try a variety of characters in commit messages and make sure they show up correctly when viewing svn history from the UI and from command-line.

### DIFF
--- a/src/cpp/session/modules/SessionSVN.cpp
+++ b/src/cpp/session/modules/SessionSVN.cpp
@@ -1029,6 +1029,7 @@ Error svnCommit(const json::JsonRpcRequest& request,
 
    ShellArgs args;
    args << "commit" << globalArgs();
+   args << "--encoding" << "UTF-8";
    args << "-F" << tempFile;
 
    args << "--";


kevinushey (approved):
LGTM! I'm okay with taking this for v1.4.

jmcphers (approved):

PR #8004, opened 2020-10-08
infer appropriate encoding when reading HTML help documentation

kevinushey:
### Intent

Currently, UTF-8 encoded help is not displayed correctly in autocompletion tooltips. It would be neat if it did.

### Approach

Ensure that HTML help retrieved for a package is marked with the encoding declared in that package's DESCRIPTION file.

### QA Notes

Test with the package at https://github.com/kevinushey/nihao -- install it with:

```
remotes::install_github("kevinushey/nihao")
library(nihao)
nihao::
```

and confirm that Help documentation is displayed as expected. Note that this needs to be tested on Windows.

Screen Shot 2020-10-07 at 10 04 08 PM

### DIFF
--- a/src/cpp/session/modules/SessionHelp.R
+++ b/src/cpp/session/modules/SessionHelp.R
@@ -16,6 +16,9 @@
 # use html help
 options(help_type = "html")
 
+# cached encoding of help files for installed packages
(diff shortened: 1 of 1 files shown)


gtritchie (approved):
LGTM!

Issue #7999, opened 2020-10-07
Python Global Environment Doesn't Register First Import When Automatically Switching from R to Python


chriscardillo:
### System details

    RStudio Edition : Desktop 
    RStudio Version : 1.4.909
    OS Version      : Mac OS 10.15.2
    R Version       : 3.6.2

### Steps to reproduce the problem

Open RStudio and open a new .py file in the source editor. Make your .py file contain the following

```python
import pandas as pd
import seaborn as sns
import matplotlib.pyplot as plt

my_df = pd.DataFrame({"col1": [1, 2, 3], "col2": [4, 5, 6]})

sns.pointplot(x='col1', y='col2', data=my_df)
plt.show()
```

While your development environment is still set up for R (e.g. R REPL, R environment pane, etc.) run the the first like of the .py, `import pandas as pd`. The console will switch to Python REPL, but `pd` will not show up in the variable explorer. You will still be able to access `pd` functions, classes, etc., but it doesn't show up in the explorer. Now, if you continue to run lines, e.g. `import seaborn as sns`, you will see these subsequent variables, as RStudio is now set up as a Python environment.

### Describe the problem in detail

See above.

### Describe the behavior you expected

I expected, when I run the first line of the .py, `import pandas as pd`, that `pd` would be a variable I can see in the environment pane.

- [x ] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting a RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [ x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

chriscardillo:
@kevinushey - here you go! :)

kfeinauer:
@kevinushey Attempting to verify this on 1.4.990. I just set up Python for the first time, selected my interpreter (`/usr/bin/python`) and am trying to use Ctrl + Enter on the first line in my test.py file (which matches Chris's above). 

The IDE is not automatically switching into Python mode, and is attempting to run `import pandas as pd` as R code, so I get an R error in the console:

> Error: unexpected symbol in "import pandas"

[...]

PR #8009, opened 2020-10-08
update detect changes cache when reading environment state

kevinushey:
### Intent

Closes https://github.com/rstudio/rstudio/issues/7999.

### Approach

The 'detect changes' code maintains a cache of known objects in a Python module, and updates when that changes; however, we need to initialize that cache when the environment pane is first populated.

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/7999.


gtritchie (approved):
LGTM!

Issue #8013, opened 2020-10-08
Syntax highlighting and autocompletion when using inline R code in R Markdown


mattwarkentin:
Hi,

I was wondering if it would be possible to add a feature to the R Markdown source editor so that when you are writing code inside inline `R` code chunks that the editor could support the standard R syntax highlighting and tab-autocompletion you get when inside a fenced `R` code chunk.
````
---
title: My sweet document
output: html_document
---

This is some inline code `r head(mtcars)` that appears as gray in colour and does not allow 
for autocompletion in RStudio.

Below is a code chunk that has nice syntax highlighting and tab-autocompletion + fuzzy narrowing.
```{r}
head(mtcars)
```

````

kevinushey:
This actually did work in RStudio 1.1, so it looks like this might be a regression:

<img width="623" alt="Screen Shot 2020-10-08 at 1 15 30 PM" src="https://user-images.githubusercontent.com/1976582/95508799-632bdd80-0968-11eb-9e20-0b9d14f6d423.png">

mattwarkentin:
Hmm, interesting. I'm using __RStudio 1.4.825__, for reference.

kfeinauer:
This looks broken to me in the latest daily. I don't get any autocompletion for inline chunks.

kevinushey:
@kfeinauer: were you testing in source or visual mode? This fix should be specifically for source mode; we could consider implementing this for visual mode as well but that was outside of this PR's intent.

kfeinauer:
@kevinushey I was indeed in source mode.

kfeinauer:
@kevinushey Scratch that - I just needed to hit `Tab` to see the autocompletions. What causes this to open automatically? It doesn't do so in the inline, but does for a regular chunk.

kfeinauer:
Spoke with @kevinushey - it is expected behavior to need to use Tab or Ctrl+Space to get the autocompletions to show up in an inline chunk. Though we could consider automatically showing them as well, but that is not currently expected.

Verified 1.4.1035.

PR #8014, opened 2020-10-08
allow R completion manager in inline R code chunks

kevinushey:
### Intent

Older versions of RStudio allowed the R completer to provide completions within inline R code chunks. Losing that functionality was not intentional; this PR brings it back.

### Approach

Properly delegate to the R completion manager if we detect the cursor is within an inline R code chunk.

### QA Notes

Create an R Markdown document, and then test that R code completions are popped up for text of the form:

```
`r st|`
```

### Notes

Also comes with R highlighting of inline R code chunks, because ... why not.

Screen Shot 2020-10-08 at 2 13 24 PM

Closes https://github.com/rstudio/rstudio/issues/8013.


gtritchie (approved):
LGTM!

PR #8025, opened 2020-10-09
add Dockerfile editor mode

kevinushey:
### Intent

Bring in basic editor support for Dockerfiles.

### Approach

Ace already provides an editor mode + highlight rules for Dockerfiles, so just define an editor mode and use that.

### QA Notes

Open a sample Dockerfile, and confirm that basic syntax highlighting is present.

Screen Shot 2020-10-09 at 2 35 22 PM

Note that we definitely don't intend to provide any higher-level features at this point; this is purely to allow users to quickly edit Dockerfiles as necessary from RStudio.

---

Closes https://github.com/rstudio/rstudio/issues/5141.


kevinushey:
Merged in https://github.com/rstudio/rstudio/commit/6a8f5784eb4e66c4bdc238529c523a0e419d3866.

jasonmoy28:
Is it available on RStudio now?

PR #8027, opened 2020-10-09
Add Support for In Text Citations

dragonstyle:
### Intent

Pandoc-citeproc supports citations of both note and in text style citations (as does CSL). Previously, we always used the note form like `[@cite]`. We now support the in-text form as well, which can be either `@cite` or `@cite [suffix]`. Users may use any of the forms and expect completions, formatting, and so on to be respected and work properly. In addition, when users are inserting citations using the insert citation dialog, they may choose to use the in-text style if they are only selecting a single citation.

Citations are reversible by using the backspace key (e.g. if you mean to type a GitHub name or twitter handle you can just backspace to 'break' the citation.

### Approach

We did some plumbing to break any assumptions that citations were always enclosed by brackets. We ensured that the in-text form don't conflict with crossrefs.

### QA Notes
- Citation entries will now be blue in color, as links, cross refs.
- Citation entries are valid of the following types of forms:
```
[@cite]
[@cite; and some text @cite2]
This is some text @cite and some more text.
This is some text @cite [page 23] and more text.
```


src/gwt/panmirror/src/editor/src/api/widgets/navigation-tree.tsx
```diff
@@ -56,7 +56,6 @@ const kNavigationTreeIndent = 8;
 // like you would use to navigate a hierarchical file system.
 export const NavigationTree: React.FC<NavigationTreeProps> = props => {
   const style: CSSProperties = {
-    overflowY: 'scroll',
```
jjallaire, line 59: Is the idea that this will never overflow? Or is the overflow handled elsewhere?

dragonstyle: It is handled elsewhere (the tree itself uses overflow auto, so there is literally no reason for this). I can't explain why it made its way in other than perhaps I left it in after doing some testing of scroll behavior...

Issue #7696, opened 2020-09-01
Update to Pandoc 2.10.2 once it is released


jjallaire:

kfeinauer:
Verified binary ships with Pandoc v 2.11.1.1 in RStudio 1.4.1027

PR #8042, opened 2020-10-13
Update to Pandoc 2.11

jjallaire:
### Intent

Update to Pandoc 2.11. Resolves https://github.com/rstudio/rstudio/issues/7696

### Approach

Updates our embedded version of Pandoc to v2.11 (https://pandoc.org/releases.html#pandoc-2.11-yyyy-mm-dd). Also embeds the master branch of the **rmarkdown** package.

Note that the rmarkdown package automatically uses the *latest version* of Pandoc that it discovers on the system. This means that backwards compatibility issues created by Pandoc release are encountered immediately by R Markdown users when Pandoc is upgraded. This creates a hard requirement for the rmarkdown package to gracefully handle breaking Pandoc changes (it needs to adapt to the new behavior but keep the old behavior supported for older versions of Pandoc).

Pandoc v2.11 introduces a number of breaking changes:

1) `pandoc-citeproc` has been removed (it's functionally it now embedded in `pandoc`). This is a breaking change b/c things formerly accomplished by calling `pandoc-citeproc` now requiring calling `pandoc` w/ the requisite arguments.

2) The Pandoc AST representation for tables has become richer, which breaks code that relied on the previous AST (this will affect the `ioslides_presentation()` R Markdown format and potentially other 3rd party code which consumes table ASTs.

3) The `gfm` output format has some non-trivial output differences (headers, links, possibly other things?) from previous versions.

The R Markdown package will implement fixes/workarounds for all of these issues, which the IDE can inherit by embedding the latest version of R Markdown from GH (eventually we will simply point to the CRAN version once it's published there).

Note that you might think that a Pandoc release with breaking changes creates a reason *not* to update within the IDE. In fact it's the reverse: because the IDE doesn't have the ability to pin pandoc versions, we eventually get exposed to all of these bugs as users update pandoc in various ways. Embedding the new version sooner allows us to identify the bugs more quickly and get them fixed in pandoc and rmarkdown ASAP. Further, breaking Pandoc releases tend to force the ecosystem to depend on the newer version (as its sometimes prohibitively expensive for user packages/scripts to support multiple versions). As this occurs, if the IDE hasn't updated its version more users experience a "broken" ecosystem as the embedded version becomes more and more out of sync w/ contemporary usage. As a result it's much better for us to force the issue and take our medicine up front.

### QA Notes

You can track rmarkdown updates to address Pandoc v2.11 compatibility here: https://github.com/rstudio/rmarkdown/issues?q=is%3Aissue+is%3Aopen+label%3A%22pandoc+support%22. Once these (and any other issues we discover) are all addressed there will be a CRAN release, which we will then update the IDE to point to.

As a result of the release of Pandoc 2.11 there will immediately be very wide community testing of compatibility w/ existing documents. We can expect these issues to be reported and addressed on the Pandoc and R Markdown repos (there isn't really anything we can do in the IDE besides just benefit from fixes in Pandoc and R Markdown). If we do see issues compatibility issues reported to the IDE repo we should redirect them to the R Markdown repo.


aaronpuchert:
There seems to be a left-over declaration in [src/cpp/session/include/session/SessionModuleContext.hpp:881](https://github.com/rstudio/rstudio/blob/99a689fd9a0f7901582ff13e020baca56214c1a2/src/cpp/session/include/session/SessionModuleContext.hpp#L881):

```cpp
std::string pandocCiteprocPath();
```

PR #8047, opened 2020-10-13
use new-style remotes format for embedded packages

kevinushey:
### Intent

When embedding R packages, we use the old-style `Github*` fields for indicating the package source. Those are not supported by `renv`, and `devtools` and friends have supported the `Remotes*` syntax for some time now, so use that instead.

### Approach

Straightforward use of new `Remotes*` syntax.

### QA Notes

No test required.


kevinushey:
It looks like the CRAN release of Packrat still doesn't understand the Remotes syntax so we need to preserve the older Github style prefixes as well.

Issue #8011, opened 2020-10-08
better handle R <-> Python transitions after session restart


kevinushey:
1. Run some Python code so that the `reticulate` REPL is made active,
2. Restart R, e.g. via Session -> Restart R.

On restart, we see some issues:

1. The old Python prompt is presented, even though R is actually active right now:

<img width="461" alt="Screen Shot 2020-10-08 at 10 52 48 AM" src="https://user-images.githubusercontent.com/1976582/95495527-6ae18700-0954-11eb-9d7e-8deebc94f725.png">

2. We are still showing Python in the Environment pane:

<img width="701" alt="Screen Shot 2020-10-08 at 10 53 21 AM" src="https://user-images.githubusercontent.com/1976582/95495636-8b114600-0954-11eb-8ab2-f92c074c1e44.png">

I think we either want to ensure the prompt + Environment pane are properly reverted to reflect the "current" state in the IDE, or think about whether we could restore the Python REPL as well.

(Note: if we wanted to go the route of restoring the Python REPL, we'd probably want to consider also trying to serialize and de-serialize the Python session state, which seems somewhat perilous.)

ronblum:
Shoot. I noticed this a little back, but I didn't realize it wasn't expected—I thought it was by design. 👎

melissa-barca:
Verified in 1.4.1013

PR #8053, opened 2020-10-13
reset prompt on session restart

kevinushey:
### Intent

When working in Python mode, the prompt is set to `>>> `. Upon restart, we return to "R" mode, but we don't reset the prompt, giving the indication that the user is still in Python mode.

### Approach

Allow the Shell widget to listen for restart events, and reset the prompt after a restart has occurred.

### QA Notes

Test by:

1. Activating Python mode, e.g. by running code in a Python script,
2. Restarting the session.

Closes https://github.com/rstudio/rstudio/issues/8011.


gtritchie (approved):
LGTM!

PR #8064, opened 2020-10-14
allow digits after rcpp / cpp engine name

kevinushey:
### Intent

The [cpp11] package provides a new way to use C++ code within R packages. This PR allows us to support highlighting of `cpp11` chunks.

### Approach

Tweak the engine regex to allow for arbitrary digits after the engine name. (future-proof ourselves for the release of `cpp2050`)

### QA Notes

Test that R Markdown chunks of the form:

`````
```{cpp11}
#include <cpp11.hpp>
```
`````

are highlighted as though they are C++ code.


gtritchie (approved):
LGTM!

Issue #8065, opened 2020-10-14
cannot View objects after calling source_python


kevinushey:
At the beggining of an R script, I call:

.
library(reticulate)
use_python("C:\\Users\\jpuig\\Anaconda3\\envs\\tf_gpu\\python.exe", required = TRUE)
source_python("expICH/00_core/expICH_functions.py")
.

expICH_functions.py looks like:

import pandas as pd
import numpy as np

def function1():
.
def functionN():

As soon as the source_python function gets called, any variable I create in the R environment can't be visualized, I got the following errors:

![image](https://user-images.githubusercontent.com/9660118/96027810-b9ad8600-0e58-11eb-84fd-8ea97eeb67de.png)

_Originally posted by @jpuig005 in https://github.com/rstudio/rstudio/pull/6862#issuecomment-708573296_

melissa-barca:
Verified in 1.4.944 on macOs.

PR #8066, opened 2020-10-14
avoid defining View function in Python main module

kevinushey:
### Intent

Previously, RStudio defined a function called `View()` in the Python main module, that could be invoked to view objects using RStudio's object explorer / data viewer. This unfortunately caused the issue described in https://github.com/rstudio/rstudio/issues/8065, where calling `reticulate::source_python()` would effectively overwrite R objects in the session with those from the Python module -- including the View hook.

### Approach

Rather than implement `View()` via a function in the main module, instead use a REPL hook to detect calls to `View()` and route those to the right place. (Note that this implies that "clever" usages of View will not work, but since this is purely an RStudio concept and cannot work in a regular R session I think that is okay.)

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/8065; in particular, confirm that this works:

```
library(reticulate)
pyfile <- tempfile(fileext = ".py")
writeLines("def foo(): pass", con = pyfile)
source_python(pyfile)
View(mtcars)
```

Closes https://github.com/rstudio/rstudio/issues/8065.

### DIFF
--- a/src/cpp/session/modules/SessionReticulate.R
+++ b/src/cpp/session/modules/SessionReticulate.R
@@ -119,10 +119,6 @@
       help(...)
    }
    
-   # define View hook (for data viewer, object explorer)
-   main <- reticulate::import_main(convert = FALSE)
-   reticulate::py_set_attr(main, "View", .rs.reticulate.viewHook)
-   
    # ensure matplotlib hooks are injected on load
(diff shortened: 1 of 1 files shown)


gtritchie (approved):
LGTM!

PR #7861, opened 2020-09-22
Reset active column after a document pop out 

melissa-barca:
### Intent

Fixes #6475 

Fixes issue where `Save / Save As...` was disabled after a viewer object was popped out from main source. Also fixes an issue that was found while testing that prevented data viewers from being returned to the source window.  

### Approach

When a document is popped out from the main column, reset the active editor to the next viewable one if one exists. 

### QA Notes

Issue #6475 has step-by-step repro instructions. Also make sure that the popped out data viewer can be returned to the main window.

PR #8095, opened 2020-10-16
Fix 'Close All Except Current' command

melissa-barca:
### Intent

Fixes #7861 

The `Close Other Documents` command was broken by the multiple columns code -- if the user was prompted to save any other document prior to closing, the wrong document was left open.

### Approach

The code existed to save the editor prior to executing commands that could reset the active editor, but it wasn't actually being used.

### QA Notes

See original ticket.


kevinushey (approved):
LGTM!

PR #8112, opened 2020-10-19
hardcode location for POSTGRES librarys on macOS

gtritchie:
# Intent

Mac package builds should include the postgreql dependencies in the generated bundle, otherwise the application may not run on a non-developer machine.

On some configurations this was not happening, and cmake was deciding these libraries are part of the system and not including them. Specifically, find_library was resolving to:

`PQ_LIB:FILEPATH=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk/usr/lib/libpq.tbd`

### Approach

Hardcode path to library on macOS only. Confirmed with a local build that this results in all dependencies going into the resulting package build.

### QA Notes

Dev / build thing, nothing to test

### DIFF
--- a/src/cpp/CMakeLists.txt
+++ b/src/cpp/CMakeLists.txt
@@ -41,8 +41,8 @@ add_definitions(-DBOOST_BIND_GLOBAL_PLACEHOLDERS)
 # they currently do not work with our source
 add_definitions(-D_WEBSOCKETPP_NO_CPP11_MEMORY_=1)
 
-# the non-strict masking option was identified as a potentially crashing 
-# issue due to reliance on undefined C++ behavior, and was removed entirely in 
+# the non-strict masking option was identified as a potentially crashing
+# issue due to reliance on undefined C++ behavior, and was removed entirely in
 # websocketpp 0.7 (RStudio is using 0.5.1 at this time)
 add_definitions(-DWEBSOCKETPP_STRICT_MASKING)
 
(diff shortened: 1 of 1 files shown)


kfeinauer (approved):
LGTM - Could you sanity check that it's loading the libs from the framework directory by running `otool -L .../MacOS/RStudio | grep libpq` and ensuring it's pointing to the correct file in the Frameworks directory after installing the resulting build package?

src/cpp/CMakeLists.txt
```diff
@@ -415,7 +415,11 @@ if(UNIX)
    find_library(DL_LIB "dl")
    find_library(SQLITE_LIB "sqlite3")
    get_filename_component(SQLITE_LIB "${SQLITE_LIB}" REALPATH)
-   find_library(PQ_LIB "pq")
+   if (APPLE)
+      set(PQ_LIB "/usr/local/lib/libpq.dylib")
```
kfeinauer, line 419: @gtritchie Also if you don't mind let's put a comment here about why we're specifically setting the path on OSX (qt bundle script will only bundle dylibs, so if this is pointing to an xcode library spec file it won't get copied into the bundle)

gtritchie: Good idea, done.

gtritchie:
MacOS otool -L RStudio | grep libpq
        @executable_path/../Frameworks/libpq.5.dylib (compatibility version 5.0.0, current version 5.13.0)

PR #8117, opened 2020-10-20
Ensure log file gets reopened if needed

### LINKED ISSUE #7898: rserver file log not updated after forking
### System details

    RStudio Edition : Server
    RStudio Version : 1.4 (latest)
    OS Version      : 
    R Version       : 

1. Configure a setup that will cause rserver to log errors after startup. This is easiest if you enable launcher sessions but never start the launcher, as errors should periodically be emitted.

2. Configure rserver to write to a file via /etc/rstudio/logging.conf

```
[@rserver]
log-level=debug
logger-type=file
log-dir=/var/log/rstudio
```

3. Restart rstudio-server - notice that rserver does not write the errors to the log file.

4. Run rserver directly via `/usr/lib/rstudio-server/bin/rserver --server-daemonize 0`. Notice that the log file is written properly. 

The log file is not written properly. This is because the process of daemonizing rserver (after initializing the file log), closes the opened file descriptor for the log file. Then, for the entire rest of the run of the process, we think the file descriptor is open without issue, even though it is not.

This is a result of the changes in https://github.com/rstudio/rstudio-pro/commit/39ae1d4b70a3a56f889f5cd286f0bd3ef6375446#diff-2c1270e62fb7a3c1f00670caee81f7b0.

Logs to show up in the rserver.log file after daemonizing.

MariaSemple:
### Intent

Fixes #7898.

Ensure that we are able to log to a file from the server even after we daemonize.

### Approach

It's not enough to test the goodbit/state of the output stream because even it doesn't know it's in a bad state until a real write has been attempted. Writing an empty string is not sufficient.

To resolve this, each time we write a log to the stream, we check the goodbit on the stream. If the goodbit is not set, we re-open the stream and re-attempt the write operation. Most of the time, the goodbit is good and the stream will not be re-opened.

### QA Notes

This issue only affects the rserver process. If the logger for the rserver process is configured as described in #7898, the log should contain more messages than just `20 Oct 2020 00:16:28 [rserver] INFO Reading rserver configuration from /etc/rstudio/rserver.conf`, which is the only entry I get prior to the fix. The output you see in the file should be the same (excepting times and barring confguration changes) as the logs you get in syslog when the file logger is not enabled.

### DIFF
--- a/src/cpp/shared_core/FileLogDestination.cpp
+++ b/src/cpp/shared_core/FileLogDestination.cpp
@@ -208,18 +208,36 @@ void FileLogDestination::writeLog(LogLevel in_logLevel, const std::string& in_me
       return;
 
    // Lock the mutex before attempting to write.
-   boost::unique_lock<boost::mutex> lock(m_impl->Mutex);
+   try
+   {
+      boost::lock_guard<boost::mutex> lock(m_impl->Mutex);
 
-   // Open the log file if it's not open. If it fails to open, log nothing.
-   if (!m_impl->LogOutputStream && !m_impl->openLogFile())
-      return;
+      // Open the log file if it's not open. If it fails to open, log nothing.
+      if (!m_impl->openLogFile())
+         return;
 
-   // Rotate the log file if necessary. If it fails to rotate, log nothing.
-   if (!m_impl->rotateLogFile())
-      return;
+      // Rotate the log file if necessary. If it fails to rotate, log nothing.
+      if (!m_impl->rotateLogFile())
+         return;
 
-   (*m_impl->LogOutputStream) << in_message;
-   m_impl->LogOutputStream->flush();
+      (*m_impl->LogOutputStream) << in_message;
+      m_impl->LogOutputStream->flush();
+
+      // If the output stream has bad state after writing, it might have been closed. Try re-opening it and writing the
+      // message again. Often it is not possible to tell that a stream has failed until a write is attempted.
+      if (!m_impl->LogOutputStream->good())
+      {
+         if (!m_impl->openLogFile())
+            return;
+
+         (*m_impl->LogOutputStream) << in_message;
+         m_impl->LogOutputStream->flush();
+      }
+   }
+   catch (...)
+   {
+      // Swallow exceptions because we'd trigger recursive logging otherwise.
+   }
 }
 
 } // namespace log


src/cpp/shared_core/FileLogDestination.cpp
```diff
@@ -208,18 +208,36 @@ void FileLogDestination::writeLog(LogLevel in_logLevel, const std::string& in_me
...
+      m_impl->LogOutputStream->flush();
+
+      // If the output stream has bad state after writing, it might have been closed. Try re-opening it and writing the
+      // message again. Often it is not possible to tell that a stream has failed until a write is attempted.
+      if (!m_impl->LogOutputStream->good())
+      {
+         if (!m_impl->openLogFile())
+            return;
+
+         (*m_impl->LogOutputStream) << in_message;
```
kevinushey, line 233: Is there ever a case where we could end up duplicating a message in the log output here? (Is it worth worrying about? Seems like maybe not since this should be a rare error case in theory at least)

src/cpp/shared_core/FileLogDestination.cpp
```diff
@@ -208,18 +208,36 @@ void FileLogDestination::writeLog(LogLevel in_logLevel, const std::string& in_me
...
+         m_impl->LogOutputStream->flush();
```
kevinushey, line 234: Would it be worth checking if we get another log failure here? Is there anything we can or should do if both attempts to write to the logfile fail?

kevinushey (approved):
LGTM! Questions are mainly for my own understanding.

src/cpp/shared_core/FileLogDestination.cpp
MariaSemple: I considered this - I think it's technically possible that some or all of a message is written before the stream fails and the `badbit` is set, but in my opinion it would be better to have the message logged twice than to be in the situation where the message (and subsequent messages) isn't logged at all.

I also tried testing the streams goodness by writing `""` to the stream, but that is a no-op it appears. I'd also rather have one log appear twice than have some non-log data appear in the stream every time we are going to log something.

src/cpp/shared_core/FileLogDestination.cpp
MariaSemple: If both attempts fail, I don't think a third attempt to open it will work, and we can't attempt to log the failure because that will result in recursive logging attempts, even if there is another log destination available. 

These changes are primarily to address what happens when we daemonize. We fork the original process and continue running in the child (for process group reasons, IIUC), which means that the output stream that was opened in the original process looks open in the child (the shared_ptr is not null, the goodbit is set on the stream), but really it's not open. In that case the behaviour we get is that the first attemp to write to the log file fails completely (no message is written) and we now detect the failure and re-open the log. To the user or admin, it appears as though the output stream never closed.

Issue #8145, opened 2020-10-21
"Create project as subdirectory of" field blank on state reset; prevents creation of project


kevinushey:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.4.dev
    OS Version      : macOS 10.15.7
    R Version       : 4.0.3

### Steps to reproduce the problem

Reset RStudio's state, using `scripts/clean-rstudio`. Then, launch a new copy of RStudio and try to create a new package project.

### Describe the problem in detail

The "Create project as subdirectory of" box is blank, and attempts to create a package silently fail.

<img width="550" alt="Screen Shot 2020-10-21 at 2 05 19 PM" src="https://user-images.githubusercontent.com/1976582/96787460-a47bae80-13a6-11eb-8740-50dd577fe37b.png">

### Describe the behavior you expected

We should provide a default project location (home directory?), rather than leaving that entry blank.

Note that while the resolution here is simple (ensure a proper sub-directory is selected) the fact that there is no UI response when the attempt to create a project fails makes me think this is worth fixing in 1.4.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting a RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
Also true for the Global Options dialog:

<img width="603" alt="Screen Shot 2020-10-21 at 2 08 35 PM" src="https://user-images.githubusercontent.com/1976582/96787651-f7556600-13a6-11eb-87a1-8f95a4b5b5c3.png">

kfeinauer:
Verified fixed 1.4.990

PR #8146, opened 2020-10-21
avoid setting empty value for default directory locations

kevinushey:
### Intent

Normally, directory-chooser widgets require the user to supply a non-empty value. However, the default value for their associated preferences was the empty string, which could lead to this state for factory-fresh RStudio installations.

### Approach

Two changes, just to be safe:

1. Disallow empty strings, and force those to be interpreted as the user's home directory;
2. Update the default value of the preference.

#### Tools -> Global Options...

Check that `Default working directory (when not in a project)` is set to `~`:

Screen Shot 2020-10-21 at 4 10 35 PM

#### File -> New Project -> New Directory -> New Project

Check that `Create project as subdirectory of:` is set to `~`:

Screen Shot 2020-10-21 at 4 11 19 PM

---

Closes https://github.com/rstudio/rstudio/issues/8145.

### DIFF
--- a/src/cpp/session/resources/schema/user-prefs-schema.json
+++ b/src/cpp/session/resources/schema/user-prefs-schema.json
@@ -25,7 +25,7 @@
         },
         "initial_working_directory": {
             "type": "string",
-            "default": "",
+            "default": "~",
             "title": "Initial working directory",
             "description": "The initial working directory for new R sessions."
         },
@@ -625,7 +625,7 @@
         },
         "default_project_location": {
             "type": "string",
-            "default": "",
+            "default": "~",
             "title": "Default new project location",
             "description": "The directory path under which to place new projects by default."
         },
--- a/src/cpp/session/resources/schema/user-prefs-schema.json
+++ b/src/cpp/session/resources/schema/user-prefs-schema.json
@@ -25,7 +25,7 @@
         },
         "initial_working_directory": {
             "type": "string",
-            "default": "~",
+            "default": "",
             "title": "Initial working directory",
             "description": "The initial working directory for new R sessions."
         },
@@ -625,7 +625,7 @@
         },
         "default_project_location": {
             "type": "string",
-            "default": "~",
+            "default": "",
             "title": "Default new project location",
             "description": "The directory path under which to place new projects by default."
         },
--- a/src/gwt/src/org/rstudio/studio/client/projects/ui/newproject/NewDirectoryPage.java
+++ b/src/gwt/src/org/rstudio/studio/client/projects/ui/newproject/NewDirectoryPage.java
@@ -16,6 +16,7 @@ package org.rstudio.studio.client.projects.ui.newproject;
 
 import com.google.gwt.aria.client.Roles;
 import org.rstudio.core.client.ElementIds;
+import org.rstudio.core.client.StringUtil;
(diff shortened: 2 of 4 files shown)


gtritchie (commented):
Generic question, are there any Pro-code only instances of this pattern that need to be handled as well?

Also, is this a regression (curious to know if it was done on purpose for some other reason or has always been this way; it sounds vaguely familiar).

If it's not a regression in 1.4, then probably wouldn't meet the bar for 1.4 at this stage?

kevinushey:
It is indeed a regression -- those fields are never empty in 1.2 or older; it looks like 1.3 may have been affected by the same issue though.

I note that we have the notion of an "initial_working_directory" preference, which is used here:

https://github.com/rstudio/rstudio/blob/6da40861669dafb40c6c019bbafe4446dcb254dd/src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/GeneralPreferencesPane.java#L414

But we also have the older version of the default working directory, based on the session info here:

https://github.com/rstudio/rstudio/blob/6da40861669dafb40c6c019bbafe4446dcb254dd/src/gwt/src/org/rstudio/studio/client/workbench/model/SessionInfo.java#L253-L260

And IIUC, in 1.2 and older we would've used the version tagged in SessionInfo. So perhaps we should use that if the preference is empty / unset?

jmcphers:
>  So perhaps we should use that if the preference is empty / unset?

We should. The idiom used here is usually a "computed pref" -- that allows us to have prefs that default to system or admin provided values but can be overridden by users.

That said, this was an issue in 1.3 and empirically was not a major issue so I don't think warrants a fix in 1.4 at this point. How would you feel about making this change in Juliet Rose instead?

kevinushey:
I would prefer to fix this in 1.4 -- this behavior is pretty sub-par.

new-package-project

gtritchie:
In flows I tried other than New Package (e.g. New Project, New Shiny Application) there is a warning displayed when you click the create button with an empty directory field; it seems specific to New Package (maybe others, I certainly didn't try all possibilities):

jmcphers:
There is a "default project directory" setting (`session-default-new-project-dir`) that we should follow here if it's set.

gtritchie:
Confirmed this was introduced by #5069 and wasn't part of some intentional change to start with empty path (which would have been weird, I agree).

kevinushey:
We actually already treat as `initial_working_directory` as a computed preference:

https://github.com/rstudio/rstudio/blob/2abff95c9c3a6f386a738b1f651190d7afe1714a/src/cpp/session/prefs/UserPrefsComputedLayer.cpp#L56-L57

Does setting an empty default cause that preference from the computed layer to be ignored? Or is something else going on here?

kevinushey:
Now updated to use the proper computed defaults, as retrieved from SessionInfo.

I suspect there might be a separate issue here with GWT's user preference accessor not reading from the computed layer, though. @jmcphers, does that sound plausible?

Either way, I believe this is the right fix for 1.4.

jmcphers:
>  GWT's user preference accessor not reading from the computed layer, though. @jmcphers, does that sound plausible?

Some further investigation may be warranted but if this were generally broken I'd expect a lot more things to not work (for example the computed layer is also where we detect the git executable location).

jmcphers (approved):

PR #8158, opened 2020-10-22
Fix cross-device global replace

melissa-barca:
### Intent

Global replace was silently failing when the file was on a different machine than RStudio. This was reported in #7736 which I'm leaving open until we can get confirmation from the user that it is fixed as there has been lots of back and forth. 

### Approach

Pass the `overwrite` flag to the move command when performing the replace; otherwise cross-machine moves that overwrite a file are prevented. 

### QA Notes

Perform find and replace where the file containing the replace is not on the same machine as the IDE.

### DIFF
--- a/src/cpp/session/modules/SessionFind.cpp
+++ b/src/cpp/session/modules/SessionFind.cpp
@@ -576,7 +576,8 @@ private:
             error = setPermissions(tempReplaceFile_.getAbsolutePath(), filePermissions_);
 #endif
             if (!error)
-               error = tempReplaceFile_.move(FilePath(currentFile_));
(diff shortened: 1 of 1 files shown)


kevinushey (approved):
LGTM!

It feels worthwhile making this the default behavior for `move()`, since otherwise it'll be very easy for us to forget whether we want to allow cross-device "moves" (and I assume we almost always will want that). Would you mind filing that as a separate issue?

melissa-barca:
@kevinushey I had the same thought as I made this change. I've opened #8161.

kevinushey:
Thanks!

PR #8184, opened 2020-10-25
Revert "Fix chunk highlighting issue"

melissa-barca:
Reverts rstudio/rstudio#8093

See comments on #8088. This fix seems to have done more harm than good as now there are issues happening with the standard editor in a single column. I'm reverting this  for now and will work on a new fix.


melissa-barca:
FYI @jmcphers & @kevinushey

PR #8194, opened 2020-10-26
Fix issues with chunk highlighting with multiple columns

melissa-barca:
### Intent

To properly highlight chunks before and after adding or removing a column, the Ace Editor needs to retain its `AceBackgoundHighlighter` handlers while the object detaches. 

### Approach

With a single column the editor widget does not detach until it is closed so we used the `onAttachOrDetach` function to clear out any handlers. With multiple columns we will attach, detach, and reattach the widget but never want to detach the handlers. I removed the handler clean up and the attach handler as it was only responsible for clean up.

### QA Notes

We want to try different orderings of editing markdown files, adding columns, and switching between visual and standard source mode and confirm that chunks are properly highlighted.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/ace/AceBackgroundHighlighter.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/ace/AceBackgroundHighlighter.java
@@ -19,7 +19,6 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
-import org.rstudio.core.client.HandlerRegistrations;
 import org.rstudio.core.client.JsVector;
 import org.rstudio.core.client.JsVectorInteger;
 import org.rstudio.core.client.ListUtil;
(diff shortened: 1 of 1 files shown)


kevinushey:
My concern here is that if we're not cleaning up the handlers, we could "leak" them so that the associated editor instance is never properly cleaned up even after it's closed / removed.

Is there any other event or signal that the editor instance is about to be removed / destroyed / closed that we could use here?

kevinushey:
One possibility:

Ace has a function 'destroy()' that can be called on the Ace instance to clean up the editor. Maybe we should call that when a document is closed?

https://github.com/ajaxorg/ace/blob/a505922a01395ab77a878c08da06eb09b7822d23/lib/ace/editor.js#L2756-L2776

Then, we could also listen for the "destroy" Ace signal here, similar to "changeselection" (and others):

https://github.com/rstudio/rstudio/blob/98bbcd3cac5f4f722f46b105d8a80dbcef3c8e9c/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/AceEditorWidget.java#L300-L310

And the highlighter could listen for that "destroy" signal, and use that as a signal that it should remove its handlers.

That said this is a little much for 1.4 so maybe we should just take the PR as-is (and acknowledge that we're leaking some event handlers, but they're associated with an Ace instance that is no longer used and so shouldn't affect the session too adversely)

melissa-barca:
@kevinushey I created #8196 and assigned it to myself to handle this after 1.4. My thoughts were that we could add a close function to `DocDisplay` that `EditingTarget` called while closing and then `AceEditor` could tell `AceBackgroundHighlighter` to clear the handlers but your solution is more straight forward. It's good to know about the native ace signals. Thanks!

Issue #8124, opened 2020-10-20
Change default Preview PDF setting on MacOS to RStudio Viewer or System Viewer


ronblum:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.3.1093 and 1.4.926
    OS Version      : MacOS
    R Version       : 4.0.3

### Steps to reproduce the problem

On MacOS:
1. Knit an R markdown file to PDF
    - Correct: The Rmd knits and the PDF file is created
    - Unexpected: The PDF file does not open up automatically
2. Tools -> Global Options -> Sweave -> Preview PDF... -> RStudio Viewer or System Viewer
3. Knit the R markdown file to PDF
    - Correct: The Rmd knits and the PDF file is created
    - Correct: The PDF file does opens up automatically

### Describe the problem in detail

The default value of the Preview PDF... option is "(No Preview)".

Note that on these are set by default on Linux (Evince) and Windows (Sumatra).

### Describe the behavior you expected

Change the default value to "System Viewer" or "RStudio Viewer" (which used to be how we did it in 1.2).

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting a RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
Marking as 1.4 as this should be a simple + safe fix, and the issue effects a lot of users.

kfeinauer:
Verified fixed on 1.4.990

PR #8197, opened 2020-10-26
default to RStudio PDF viewer on macOS

kevinushey:
### Intent

The default PDF previewer had unintentionally been set to (No preview) with RStudio 1.4 on macOS. This was actually caused by us attempting to request a desktop synctex viewer; however, we do not bundle any such viewer on macOS.

### Approach

When computing the PDF previewer preference, choose the RStudio viewer on macOS, and the default synctex viewer otherwise.

### QA Notes

Test that, for a brand new RStudio 1.4 on macOS, that the default previewer is not set to (No preview).

Closes https://github.com/rstudio/rstudio/issues/8124.


gtritchie (approved):
LGTM!

Issue #8187, opened 2020-10-25
Building ai-blog project opens web browser with Marketo cross-domain request


gtritchie:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.4.999 (c3bd072b, 2020-10-23) - Qt 5.15.1
    OS Version      : macOS 11.0 Beta (20A5395g)
    R Version       : 3.6.1

### Steps to reproduce the problem
- install "distill" package
- clone repo https://github.com/rstudio/ai-blog
- open project in IDE
- build website in Build tab

### Describe the problem in detail
- generated website is displayed in RStudio viewer, but a tab is also opened in the system web browser saying "This page is used by Marketo Forms 2 to proxy cross domain AJAX requests."

<img width="815" alt="screenshot showing unexpected page loaded in web browser" src="https://user-images.githubusercontent.com/10569626/97116884-84593180-16bd-11eb-8db4-9c049554d4a0.png">

### Describe the behavior you expected
- no "Marketo Forms" proxy page opens in web browser

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting a RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

gtritchie:
> Aside, the code at play here contains workarounds for a Qt issue that is supposed to have already been fixed as of Qt 5.12.3 (#2491) and in theory could be rolled back.

In Qt 5.12.8 the load attempt ("...proxy cross domain AJAX requests...") first gets caught in `DesktopWebpage.cpp: WebPage::onUrlIntercepted()` which causes it to be added to the `s_subframes` array; then it is seen in `DesktopWebPage.cpp: WebPage::acceptNavigationRequest()` where it ends up in `isSafeHost()` here (which returns false and blocks the load attempt):

```
   // if this appears to be an attempt to load an external page within
   // an iframe, check it's from a safe host
   if (!isLocal && s_subframes.count(url))
   {
      s_subframes.erase(url);
      return isSafeHost(host);
   }
```

[...]

PR #8199, opened 2020-10-27
allow Qt 5.15.x redirects

kevinushey:
### Intent

Qt 5.14 added a new type of navigation, [NavigationTypeRedirect](https://doc.qt.io/qt-5/qwebenginepage.html#NavigationType-enum), which we were disallowing in requests to `acceptNavigationRequest()`. This PR allows us to accept redirects.

In addition, prior releases of RStudio needed to work around a Qt bug in `acceptNavigationRequest()`, which basically required us to infer the navigation type in a roundabout way. That workaround is no longer required.

**NOTE**: We should consider whether we want to always allow redirects. If a host is considered "safe", do we also want to consider their redirects (even to a separate host / origin) as safe? Or is the right fix here to just disallow the `marketo` redirect altogether? Or do we whitelist `marketo` but disallow other redirects?

### Approach

Straightforward removal of old workaround + handling of new navigation type.

### QA Notes

Closes https://github.com/rstudio/rstudio/issues/8187. Test via notes in https://github.com/rstudio/rstudio/issues/8187.

### DIFF
--- a/src/cpp/desktop/DesktopWebPage.cpp
+++ b/src/cpp/desktop/DesktopWebPage.cpp
@@ -470,6 +470,14 @@ bool WebPage::acceptNavigationRequest(const QUrl &url,
       {
          return true;
       }
+#if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)
+      else if (navType == NavigationTypeRedirect)
+      {
+         return
+               this->url().authority() == url.authority() ||
+               isSafeHost(host);
+      }
+#endif
       else
       {
          // when not allowing external navigation, open an external browser


src/cpp/desktop/DesktopWebPage.cpp
```diff
@@ -46,13 +46,6 @@ namespace {
 WindowTracker s_windowTracker;
 std::mutex s_mutex;
 
-// NOTE: This variable is static and hence shared by all of the web pages
-// created by RStudio. This is intentional as it's possible that the
-// request interceptor from one page will handle a request, and a _different_
-// page will handle the new navigation attempt -- so both pages will need
-// access to the same data.
-std::map<QUrl, int> s_subframes;
```
gtritchie, line 54: Given where we're at in 1.4, I wouldn't make such a big change now; use of this map predates the workarounds made for the specific Qt issue, so worry there are other side effects/reasons we might be missing here.

Plus we've never tested reliability of NavigationType in the context of Qt 5.12.8, which this code will still be using in everything but Mac desktop, so hate to open up that can of worms.

Maybe open a separate tracking issue to revisit cleaning this up in juliet-rose when we have a bit more time to validate?

src/cpp/desktop/DesktopWebPage.cpp
```diff
@@ -382,15 +370,12 @@ bool isSafeHost(const std::string& host)
...
-//
-//    https://bugreports.qt.io/browse/QTBUG-56805
-//
-// so we avoid querying it here.
 bool WebPage::acceptNavigationRequest(const QUrl &url,
-                                      NavigationType /* navType*/,
+                                      NavigationType navType,
                                       bool isMainFrame)
 {
+   qDebug() << url << "[" << navType << "]";
```
gtritchie, line 377: This should be removed.

src/cpp/desktop/DesktopWebPage.cpp
```diff
@@ -466,7 +443,13 @@ bool WebPage::acceptNavigationRequest(const QUrl &url,
       {
          return true;
       }
-      else
+#if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)
+      else if (navType == NavigationTypeRedirect)
+      {
+         return true;
```
gtritchie, line 449: Should this be doing `return isSafeHost(host);` instead of just `return true`?

src/cpp/desktop/DesktopWebPage.cpp
```diff
@@ -466,7 +443,13 @@ bool WebPage::acceptNavigationRequest(const QUrl &url,
...
+      }
+#endif
+      else if (navType == NavigationTypeLinkClicked)
```
gtritchie, line 452: Nervous about relying on navType in Qt 5.12.8, just because it introduces the need to retest this on all platforms. How about this just stays as "else", thus wrapping all use of navType in the QT_VERSION >= 5.15 and making this change a no-op for anything but Mac w/5.15?

src/cpp/desktop/DesktopWebPage.cpp
```diff
@@ -358,7 +346,8 @@ void initSafeHosts()
    safeHosts_ = {
       ".youtube.com",
       ".vimeo.com",
-      ".c9.ms"
+      ".c9.ms",
+      ".marketo.com"
```
gtritchie, line 350: Don't think we need to add this to safe hosts; the issue wasn't that this wasn't being allowed (which is also the case in RStudio 1.2 and 1.3 and didn't seem to be a concern) but that we are now opening it in system browser.

src/cpp/desktop/DesktopWebPage.cpp
```diff
@@ -330,11 +323,6 @@ void WebPage::closeRequested()
 
 void WebPage::onUrlIntercepted(const QUrl& url, int type)
 {
-   if (type == QWebEngineUrlRequestInfo::ResourceTypeSubFrame &&
```
gtritchie, line 333: I'd leave this in (per earlier comment about minimizing the change).

src/cpp/desktop/DesktopWebPage.cpp
```diff
@@ -418,14 +403,6 @@ bool WebPage::acceptNavigationRequest(const QUrl &url,
       return true;
 #endif
    
-   // if this appears to be an attempt to load an external page within
-   // an iframe, check it's from a safe host
-   if (!isLocal && s_subframes.count(url))
```
gtritchie, line 423: Would leave this as-is.

gtritchie (changes requested):
Summary of comments: reduce change so it doesn't change anything with Qt < 5.15, including not adding marketo as a safe host.

gtritchie:
> do you think that's still okay for 1.4?

Makes sense to me. Main concern I have (in general) with this overall issue is are there requests that we were formally allowing with Qt 5.12.8 that are now somehow blocked?


> Thanks @gtritchie -- I've made the much more narrow change you suggested.
> 
> I also allow redirects when the authority of the requested URL matches that of the current page (basically allowing internal redirects) -- do you think that's still okay for 1.4?

Makes sense to me. My main concern with this issue, in general, is "are there request we were allowing through in Qt 5.12.8 that are now blocked?" It _seems_ like this change should handle anything reasonable.

gtritchie (approved):
LGTM!

gtritchie:
Be sure to merge to feature/qt-5-15-mac-only, not to master.

Issue #8156, opened 2020-10-22
Run selected line(s) bug in 1.4.953


mike-lawrence:
### System details

    RStudio Edition : Desktop
    RStudio Version :  1.4.953
    OS Version      :  Ubuntu 20.04
    R Version       : 4.0.3

### Steps to reproduce the problem
Enter the following in a text editor:
```r
library(tidyverse)
# a = 
	(
		tibble(v1=1)
	)
```
Then place the cursor anywhere after or below the hash, hit ctrl-enter, and the `library(tidyverse)` part is run, never the rest of the code.

### Describe the problem in detail
I have an admittedly idiosyncratic style whereby I prototype processing chains enclosed in parens with an assignment at the beginning that is commented-out at first. On Rstudio 1.3.1093 and earlier, the IDE had no problems with this and I could hit ctrl-enter on any line inside the parens to run the whole chunk (it also worked if I placed the cursor on the commented-out assignment line, on the opening paren line, and on the close paren line). I just downloaded 1.4.953 and observe that even when the cursor is inside the parens, the code inside the parens is not run by ctrl-enter, instead the line immediately preceding the commented-out assignment line is run. 

### Describe the behavior you expected
If the cursor is placed anywhere on or after the commented-out assignment line, I expected ctrl-enter to run the contents of the parens-encapsulated chunk. 

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting a RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
@mike-lawrence Thank you for raising this issue!

1) I can reproduce the above in RStudio Desktop 1.4.953 on Ubuntu 20.
2) Things work fine in RStudio Desktop 1.4.953 on MacOS 11 beta 10 and RStudio Server 1.4.953 on Red Hat 8.2.

mikebessuille:
This sounds related to another bug Kevin was recently looking at...?

melissa-barca:
[...]

PR #8202, opened 2020-10-27
handle state arrays in getState() method

kevinushey:
### Intent

For each line in the document, Ace tracks the current "state" for highlight rules. This is either a string for a single state, or an array giving the set of states that are active for that line. This is typically used for e.g. multi-line strings, where the state tells us that we are (for example) in a double-quoted multiline string, identified by the `qqstring` state.

When rainbow parentheses are activated, we see two main issues:

1. getState() returns an array rather than a string;
2. A dummy `#tmp` state is placed at the front of the state array.

### Approach

This PR handles the two above cases, by ensuring that `getState(row)` returns whatever the "primary" state is.

### QA Notes

Test manually with the test cases in https://github.com/rstudio/rstudio/blob/master/src/gwt/test/test-multiline-expr.R, and ensure each case works with rainbow parentheses enabled / disabled.

Closes https://github.com/rstudio/rstudio/issues/8156. (thanks @melissa-barca for catching that!)

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/ace/EditSession.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/ace/EditSession.java
@@ -29,7 +29,34 @@ public class EditSession extends JavaScriptObject
    }-*/;
    
    public native final String getState(int row) /*-{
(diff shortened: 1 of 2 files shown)


melissa-barca:
LGTM!

Issue #8204, opened 2020-10-27
Removing a connection in the Connections pane removes all connection history


ronblum:
### System details

    RStudio Edition : Desktop and Server
    RStudio Version : 1.4.962
    OS Version      : MacOS 11 and Red Hat 8.2
    R Version       : 4.0.3

### Steps to reproduce the problem

In the Connections pane:
1. Click on `New Connection`
2. Create a new connection, e.g. to a database using a DSN, or to Spark, etc., and connect.
3. Close the connection by clicking on the "disconnect" icon: 
![image](https://user-images.githubusercontent.com/31009952/97349323-ba391a00-1865-11eb-947c-8c635fe5ed29.png)
4. Return to the pane's starting screen by clicking on the left arrow: 
![image](https://user-images.githubusercontent.com/31009952/97349610-0dab6800-1866-11eb-8ad1-1d8c4f2488e3.png)
    - Note that the connection is now listed in the pane
5. Click on `New Connection`
6. Create a new connection, e.g. to a database using a DSN, or to Spark, etc., and connect.
7. Close the connection by clicking on the "disconnect" icon
8. Return to the pane's starting screen by clicking on the left arrow.
9. Click on one of the connections.
10. Remove the connection by click on the trash icon: 
![image](https://user-images.githubusercontent.com/31009952/97349765-41868d80-1866-11eb-92dd-0277b5b4b219.png)
    - We are back at the main screen of the Connections pane, but all of the connections are gone.

### Describe the problem in detail

All the connections get removed.

### Describe the behavior you expected

Only the connection that was marked for removal should be removed; the rest should still be available.

I'm not sure if this qualifies as a regression, a new issue, or a variant of an existing issue (#6336), so I'm marking it for 1.4 for the moment.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting a RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
Thanks @ronblum! I can reproduce this as well with your instructions.

melissa-barca:
Verified in 1.4.997

PR #8205, opened 2020-10-27
fix issue with removal of connection entry

kevinushey:
### Intent

Currently, attempting to remove a single connection from the Connections history actually removes _all_ connections from the Connections history. That seems bad. We should not do that.

### Approach

The issue appears to arise in the way we try to modify the JSON Array when removing a connection.

If I understand correctly, we were using the [erase-remove idiom](https://en.wikipedia.org/wiki/Erase%E2%80%93remove_idiom) here to remove a range of elements from the array, but this seems to not do what we expect here. A safe fix for 1.4 is to just create a brand new JSON array and add JSON elements to that, except for the connection to be removed.

### QA Notes

Follow @ronblum's excellent notes in https://github.com/rstudio/rstudio/issues/8204.

Closes https://github.com/rstudio/rstudio/issues/8204.

### DIFF
--- a/src/cpp/session/modules/connections/ConnectionHistory.cpp
+++ b/src/cpp/session/modules/connections/ConnectionHistory.cpp
@@ -31,7 +31,7 @@ namespace connections {
 namespace {
 
 
-bool isConnection(const ConnectionId& id, json::Value valueJson)
+bool isConnection(const ConnectionId& id, const json::Value& valueJson)
 {
    if (!json::isType<json::Object>(valueJson))
    {


kfeinauer (commented):
This is problematic indeed, but we have a test case explicitly in `JsonTests.cpp` that shows this idiom should work properly:

https://github.com/rstudio/rstudio-pro/blob/23b55d0deaf3e4c9ac857a1cfc106c346d02d629/src/cpp/shared_core/json/JsonTests.cpp#L400-L422

I think it would be worth investigating why this works for the simple test case and not for the connections array.

kfeinauer:
@kevinushey Could you test changing `isConnection(const ConnectionId& id, json::Value valueJson)` to use `const json::Value&` to see if that fixes the issue?

kevinushey:
Spent a bit more time looking, and it looks like the call to `remove_if` here is inadvertently setting each element in the array to `null`:

https://github.com/rstudio/rstudio/blob/4b02253c0e11483dcd0d991d8e8906d5931b8f41/src/cpp/session/modules/connections/ConnectionHistory.cpp#L121-L124

The `isConnection()` predicate lives here:

https://github.com/rstudio/rstudio/blob/4b02253c0e11483dcd0d991d8e8906d5931b8f41/src/cpp/session/modules/connections/ConnectionHistory.cpp#L34-L44

and strangely, changing that to accept `valueJson` by reference rather than value also fixes the issue.

kevinushey:
Oops; I saw your comment just after I finished writing that up + posting. You're exactly right that using `const&` also fixes the issue.

kfeinauer (approved):

PR #8217, opened 2020-10-28
Add null check to prevent overwriting file with 'NULL'

### LINKED ISSUE #8209: RStudio editor destroys files
### System details

    RStudio Edition :  Desktop
    RStudio Version : 1.3.1093
    OS Version      : Windows 10 Home
    R Version       : 4.0.3

(This issue is also posted on [StackOverflow](https://stackoverflow.com/questions/64561862/rstudio-editor-destroys-files?noredirect=1#comment114160729_64561862))

Create a file `temp.R` containing some code and verify you can read it
```r 
scan(file = "temp.R", what = "", sep = "\n")
## Read 2 items
## [1] "c(\"There is stuff in this file\"," "  \"extending over two lines.\")"  
```
The rest is described under details below.

Subsequently I try to edit the file `temp.R`
```r
edit(file = "temp.R")
```
At this point, I see the editor window with only the string `NULL` in it.

So I press the "Cancel" button

Now read the file again:
```r
scan(file = "temp.R", what = "", sep = "\n")
## Read 1 item
## [1] "NULL"
```

Obviously, I expected to find the original contents of `temp.R` when the editor opened. I also hoped that `temp.R` would still be intact after I pressed "Cancel". Instead, even though I canceled the edit, it has already replaced the file and my code is gone. I speculate that the overwrite of the file happened *before* the editor opened.

#### Additional notes
  * I *do* observe the same issue if I use an external editor [via `options(editor = <path to editor>)`] rather than the build-in one.
  * I first observed this issue in another context, doing something like `fix(foo)` where `foo` is a function. If I made a syntax error, I'd see a parse error message but not the expected message about trying `foo <- edit()` to fix it up (see the help page for `edit`). If I did try `foo <- edit()`, I'd get the same blank edit window and my function `foo` would be overwritten with `NULL`.

- [ x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x ] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x ] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

melissa-barca:
### Intent

Fixed #8209 

Fix the `edit` call so it doesn't overwrite files when writing to a file on Windows.

### Approach

We need to recognize the user is requesting to edit a file by the name being missing **or** null. Previously, we were only checking if name was missing and interpreting NULL as a string to write to the file.

### QA Notes

See original ticket.

### DIFF
--- a/src/cpp/session/modules/SessionEnvironment.R
+++ b/src/cpp/session/modules/SessionEnvironment.R
@@ -446,7 +446,7 @@
 {
    # if 'name' is missing, we're likely being invoked by
    # 'utils::file.edit()', so just edit the requested file
-   if (missing(name))
+   if (missing(name) || is.null(name))
       return(.Call("rs_editFile", file, PACKAGE = "(embedding)"))
    
    # otherwise, we're more likely being invoked by 'edit()', which


kevinushey (approved):
LGTM!

Issue #8213, opened 2020-10-28
Shiny Run Document dropdown empty


gtritchie:
### System details

    RStudio Edition : Desktop and Server
    RStudio Version : 1.4.980 (tried back to 1.4.100 with same results)
    OS Version      : any (tried Mac Catalina and Big Sur, Windows-10 for desktop, Centos7 for server)
    R Version       : various

### Steps to reproduce the problem
- File / New File / R Markdown...
- Configure for Shiny as follows:

<img width="532" alt="New R Markdown dialog with Shiny selected" src="https://user-images.githubusercontent.com/10569626/97470699-1606b000-1905-11eb-9321-5b0be9bf9dad.png">

- Click on the little arrow next to Run Document button on editor toolbar

### Describe the problem in detail
- Get a little grey box, no dropdown

<img width="352" alt="screenshot of little rectangle shown instead of dropdown menu" src="https://user-images.githubusercontent.com/10569626/97471065-90373480-1905-11eb-815b-9a3f2085b1a6.png">

### Describe the behavior you expected
- Dropdown as shown here in 1.3.1093 with Record Test and Run Tests options

<img width="594" alt="dropdown showing " src="https://user-images.githubusercontent.com/10569626/97470821-3afb2300-1905-11eb-8505-5bfa80e0a554.png">

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting a RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
Note: the menu appears to show up after the document has been saved.

<img width="515" alt="Screen Shot 2020-10-28 at 1 42 11 PM" src="https://user-images.githubusercontent.com/1976582/97494368-71e03180-1923-11eb-81c9-754085db87d0.png">

<img width="520" alt="Screen Shot 2020-10-28 at 1 42 31 PM" src="https://user-images.githubusercontent.com/1976582/97494375-73115e80-1923-11eb-94ea-64c1408aae2e.png">

kevinushey:
The plot thickens:

[...]

PR #8219, opened 2020-10-28
update format list after loading visual mode

kevinushey:
### Intent

Normally, when the editing toolbar is initialized, changes are sequenced roughly as:

- manage UI elements
- do Shiny-specific things

When the `visualMode_` object is initialized, it also calls performs the 'manage UI elements' step, so we effectively have:

- manage UI elements
- do Shiny-specific things
- manage UI elements

and that extraneous attempt to manage UI elements stomps on the Shiny-specific actions we had taken.

### Approach

Re-arrange initialization so that visual mode is initialized before we update the R Markdown format list.

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/8213. It may also be worth verifying that other R Markdown runtimes (e.g. from https://rmarkdown.rstudio.com/authoring_shiny_prerendered.html; `runtime: shiny_prerendered`) are handled properly.

Closes https://github.com/rstudio/rstudio/issues/8213.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTarget.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTarget.java
@@ -1920,13 +1920,6 @@ public class TextEditingTarget implements
          docDisplay_.addOrUpdateBreakpoint(breakpoint);
       }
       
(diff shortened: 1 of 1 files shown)


gtritchie (approved):
LGTM! Thanks for jumping on this fix so quickly!

PR #8224, opened 2020-10-29
pull over small script tweaks from Qt 5.15 work

gtritchie:
### Intent

Pull over some generic improvements from the aborted Qt 5.15 branch. These are minor dependency installation script tweaks.

### Approach

- update install-qt.sh with latest from Qt (https://code.qt.io/cgit/qbs/qbs.git/tree/scripts/install-qt.sh); nothing new specific to our current usage but might as well update it so it's one less thing to worry about if we end up having to update Qt for some other reason
- install-qt-macos displays warning that the scripted Qt installer is only intended for build machines, not dev machines

### QA Notes

Nothing to test; purely dev-box/build-box impacting (and very minor at that).

### DIFF
--- a/dependencies/common/install-qt.sh
+++ b/dependencies/common/install-qt.sh
@@ -59,10 +59,13 @@ Positional arguments
         target and toolchain.
 
 Options
-  --directory <directory>
+  -d, --directory <directory>
         Root directory where to install the components.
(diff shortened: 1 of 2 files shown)


kevinushey (approved):
LGTM!

PR #8231, opened 2020-10-29
fixups for C++ autoindent tests

kevinushey:
### Intent

Shake the dust off of the C++ auto-indent tests.

### Approach

- Don't try to include the now-removed `background_highlighter.js` (background highlighting is done in https://github.com/rstudio/rstudio/blob/bugfix/c++-autoindent-tests/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/ace/AceBackgroundHighlighter.java, and is not relevant for these tests)

- Skip a couple more tests that aren't relevant for the open brace indentation tests.

- Make it easier to find failed tests by appending `[FAIL]` to broken tests (so once can then just search for `[FAIL]` in the document)

### QA Notes

Nothing for QA to test; this is mainly to ensure the C++ autoindent test suite runs without errors.

Closes https://github.com/rstudio/rstudio/issues/8229.


jmcphers (approved):
LGTM, thanks for taking care of this so quickly!

PR #8232, opened 2020-10-29
always enable save commands for new source docs

kevinushey:
### Intent

Ensure that newly-opened R Markdown documents can be saved.

### Approach

The `newDoc()` function tried to enable / disable the Save command depending on whether the IDE already had an active editor. However, creation of a new document (usually) implies that a new editor will be created and become active, so we should check that state only after the document has been created.

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/8228.

Closes https://github.com/rstudio/rstudio/issues/8228.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/source/SourceColumn.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/SourceColumn.java
@@ -982,7 +982,7 @@ public class SourceColumn implements BeforeShowEvent.Handler,
 
                   // toggle save commands after the editor has been opened,
                   // in case this action has created and focused a new editor
-                  if (activeEditor_ != null)
+                  if (SourceColumn.this == manager_.getActive())
                   {
                      target.forceSaveCommandActive();
                      manageSaveCommands(true);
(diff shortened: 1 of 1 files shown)


src/gwt/src/org/rstudio/studio/client/workbench/views/source/SourceColumn.java
```diff
@@ -979,13 +978,14 @@ public void onResponseReceived(SourceDocument newDoc)
...
+                  EditingTarget target = addTab(newDoc, Source.OPEN_INTERACTIVE);
 
-                  if (contents != null)
+                  // toggle save commands after the editor has been opened,
+                  // in case this action has created and focused a new editor
+                  if (activeEditor_ != null)
                   {
                      target.forceSaveCommandActive();
-                     manageSaveCommands(active);
+                     manageSaveCommands(true);
```
melissa-barca, line 988: The active parameter actually represents whether or not this is this column is the current active column. (`activeEditor_` is only populated for the active column) You'll see in `manageSaveCommands` we use `getNextActiveEditor()` instead of directly accessing the active editor, this is a function that will return the editor that would become active if the column became active. 
When you open a new doc it gets set to the active editor so this code works fine except in the very unlikely case where this function gets called after the user has moved on and is working with a different file. Really it doesn't make a difference, but I think we should pass `this == manager.getActive()` here.

melissa-barca (changes requested):

src/gwt/src/org/rstudio/studio/client/workbench/views/source/SourceColumn.java
kevinushey: Got it -- thanks for explaining!

melissa-barca (approved):
LGTM!

Issue #6475, opened 2020-03-15
Save / Save As greyed out


HanOostdijk:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.3.904
    OS Version      : Windows 10 x64 (build 18363)
    R Version       : 3.6.0 (2019-04-26)

Problem could be the same as [5722](https://github.com/rstudio/rstudio/issues/5722).
The status of that is [repro needed]. This could mean (??) that you need another person to reproduce it. Then here it is.

### Steps to reproduce the problem
In my case it is enough to show a data.frame in a separate window. See detail description.

### Describe the problem in detail

- Start RStudio
- No edit panel is open and therefore open an edit panel via menu: `File|New File|R Script`
- Enter the line `df = tibble::tibble(x=1:5,y=1:5)`
- Execute the line with the `source` button.
- I see `source('~/.active-rstudio-document')` in the console panel and `df` is created and visible in the `data` section of the `Environment panel`
- display `df` via the button in the `data` section.
- if I set focus on the edit panel and click on the `File` menu I see the following options
![image](https://user-images.githubusercontent.com/13506389/76706431-afbb7d00-66e7-11ea-8ace-ae5c5e821495.png)
- in other words: no problems with save and saveas
- when I set focus on the panel where `df` is played and click the `show in new window` button, the data.frame is show in a separate panel as expected
- if I set focus again on the edit panel and click on the `File` menu I see the following options
![image](https://user-images.githubusercontent.com/13506389/76706542-7b948c00-66e8-11ea-95d3-1a3d0912f20a.png)
- in other words: the save options are no longer available
- closing the separate panel makes no difference: the save options are still not available
  
### Describe the behavior you expected
Also in the case of the existence of a separate window to view a data.frame I expect that I can 'saveas' code in an edit window when that window has focus.

- [x] I have read the guide to submitting good bug reports at https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports . 
- [ ] I have installed the latest version of RStudio and confirmed that the issue still persists.
** daily build version **
- [ ] If I am reporting a RStudio crash, I have included a diagnostics report. https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report
- [ ] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
Thanks for the reproducible example! I can reproduce this now on Windows and Linux with RStudio Desktop.

[...]

PR #8250, opened 2020-10-30
update save commands on window focus

kevinushey:
### Intent

When focus is transferred between different source windows, the state for certain commands is not always updated. This PR ensures that save commands are enabled / disabled as appropriate when an RStudio window gains focus.

### Approach

Use a Window focus handler to update save commands as appropriate.

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/6475.

Closes https://github.com/rstudio/rstudio/issues/6475.

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/command/AppCommand.java
+++ b/src/gwt/src/org/rstudio/core/client/command/AppCommand.java
@@ -219,7 +219,12 @@ public class AppCommand implements Command, ClickHandler, ImageResourceProvider
 
    public void setEnabled(boolean enabled)
    {
-      if (enabled != enabled_)
+      setEnabled(enabled, false);
+   }
+   
+   public void setEnabled(boolean enabled, boolean force)
+   {
+      if (force || enabled != enabled_)
       {
          enabled_ = enabled;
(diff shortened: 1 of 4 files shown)


src/gwt/src/org/rstudio/core/client/command/AppCommand.java
```diff
@@ -219,7 +219,12 @@ public boolean hasCommandHandlers()
 
    public void setEnabled(boolean enabled)
    {
-      if (enabled != enabled_)
+      setEnabled(enabled, false);
+   }
+   
+   public void setEnabled(boolean enabled, boolean force)
+   {
+      if (force || enabled != enabled_)
```
kevinushey, line 227: For context: RStudio maintains command "state" both on the GWT side, and also on the Desktop side for RStudio Desktop. (This indicates whether the Save command should be enabled in e.g. the File menu.)

I was seeing some cases where the client-side save command was marked as enabled, but the desktop-side save command was still marked as disabled. Having `force` here allows us to ensure that the client + desktop command states are sync'ed in the case where this happens.

melissa-barca (approved):
LGTM! Thanks for looking into this!

PR #8252, opened 2020-10-30
Handle multibyte characters in long global replace previews

melissa-barca:
### Intent

When determining which characters to highlight in the Find in Files pane we need to account for multibyte characters. 

### Approach

When adjusting the output of long strings we determine which characters to remove from the front based on the utf8 character values rather than bytes. 

### QA Notes

This issue only happens with search and replaces performed on lines with over 3000 characters that include a multibyte utf8 character e.g. Я


kevinushey (approved):
LGTM!

PR #8253, opened 2020-10-30
fix for undo on Windows w/Qt

kevinushey:
### Intent

When Ctrl is held down, Qt gives us malformed key events which are not interpreted properly in some cases by the visual mode editor instance.

### Approach

Detect undo specifically in this case, and do some gymnastics to re-fire an undo request.

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/7960. Closes https://github.com/rstudio/rstudio/issues/7960.

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/command/ShortcutManager.java
+++ b/src/gwt/src/org/rstudio/core/client/command/ShortcutManager.java
@@ -33,6 +33,7 @@ import org.rstudio.core.client.command.KeyMap.KeyMapType;
 import org.rstudio.core.client.dom.DomUtils;
 import org.rstudio.core.client.events.NativeKeyDownEvent;
 import org.rstudio.studio.client.RStudioGinjector;
+import org.rstudio.studio.client.application.Desktop;
 import org.rstudio.studio.client.application.events.EventBus;
 import org.rstudio.studio.client.application.events.WarningBarClosedEvent;
 import org.rstudio.studio.client.events.EditEvent;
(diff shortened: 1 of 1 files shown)


gtritchie:
Did you check other common Ctrl+whatever combinations to see if they also exhibit corruption? Curious about Ctrl+Y for redo, for example.

I did try the SimpleBrowser example with Qt 5.15.1 and both Ctrl+Z and Ctrl+Y work on French keyboard with prosemirror demo, but neither worked with Qt 5.12.8-based example.

kevinushey:
All Ctrl-prefixed keys are corrupt in this way, but this was the only command that seemed to be affected (for whatever reason).

I wanted the narrowest-possible fix for 1.4; do you think we should do something more globally here?

gtritchie:
> I wanted the narrowest-possible fix for 1.4; do you think we should do something more globally here?

No, agree with keeping it scoped as small as possible. I poked around a bit and couldn't find a Qt issue exactly matching this behavior, though did see some similar things such as this one reported on MacOS for Qt 5.15 (sigh). https://bugreports.qt.io/browse/QTBUG-85660

So seems like a problematic area for Qt and the upgrade to 5.15 would probably have fixed us on some platforms, broken us in new ways on others.

gtritchie (approved):
LGTM!

kevinushey:
I think Ctrl + Y might work just out of virtue of it being in the same location on the keyboard for both QWERTY and AZERTY layouts. 🙃

cderv:
> I think Ctrl + Y might work just out of virtue of it being in the same location on the keyboard for both QWERTY and AZERTY layouts. 🙃

@kevinushey In fact, CTRL + Y does not work for me in the IDE even in source mode (WINDOWS 10 and AZERTY) 🙄 Not using undo often so did not see that before... 

Do you want a new issue ?

kevinushey:
Oh no 😭  I'll re-open the old issue.

Issue #8260, opened 2020-11-02
File path autocompletion fails in Rmarkdown document


jmbuhr:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.4.993
    OS Version      : pop-os 20.10
    R Version       : 4.0.3

### Steps to reproduce the problem

Create an Rmarkdown document in a subdirectory of an RStudio project.
In a code chunk, type opening and closing quotation marks (`""`) and activate the file path autocompletion
with Ctrl+Space or Tab.

### Describe the problem in detail

The following error-box appears: `R code execution error`
with the console message:

```
Error in isNamespaceLoaded(pkg) : 
  attempt to use zero-length variable name
```

This does not happen if you explicitly add the dot (`.`) to signify
the current directory in the quotation marks before triggering
autocompletion. `""` Fails, `"."` works.

### Describe the behavior you expected

File path autocompletion works.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting a RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

mikebessuille:
Could we please repro this on a supported OS to see if it's only a pop-os issue?

mikebessuille:
In any case, this is not a critical issue for 1.4.  There is an easy workaround for this.

jmbuhr:
I was just able to reproduce it on the latest Windows 10, using the same steps.

kevinushey:
This needs to be fixed in 1.4 and the fix will be trivial.

kevinushey:
I've definitely seen this issue before but I'm now struggling to reproduce. @jmbuhr can you provide any extra context? Can you share a project or something similar where you can reliably reproduce this behavior?

jmbuhr:
I noticed something weird I haven't considered before! It is not about the location of the Rmd document in the project, it is the existence of a folder named `data` in the same directory

kevinushey:
Thanks! I can reproduce given that information.

melissa-barca:
Verified in 1.4.1000.

PR #8269, opened 2020-11-02
don't display help for file completion items

kevinushey:
### Intent

When scrolling through an autocompletion list, RStudio always attempts to display help for a completion item. However, some completion types (e.g. files and directories) do not have any relevant completion items. Still, we request help for these as though they were R objects, and so sometimes we ended up providing irrelevant help. For example, from RStudio 1.2:

Screen Shot 2020-11-02 at 11 49 12 AM

In older versions of RStudio, we "successfully" provided the wrong help topic; in the 1.4 dailies, we erred when reading this help.

### Approach

Fix the underlying issue causing the R error to appear when requesting help, and disallow help requests for file / directory / string completion items.

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/8260#issuecomment-720673265. In particular, with `|` denoting the cursor position:

```
dir.create("data")
"|"
```

Try to request completions. You should see the error as RStudio erroneously attempts to display help for the directory completion item.

Closes https://github.com/rstudio/rstudio/issues/8260.

### DIFF
--- a/src/cpp/session/modules/SessionHelp.R
+++ b/src/cpp/session/modules/SessionHelp.R
@@ -577,7 +577,7 @@ options(help_type = "html")
    if (length(package) && nzchar(package))
    {
       packagePath <- system.file(package = package)
-      if (file.exists(packagePath))
+      if (nzchar(packagePath))
       {
          encoding <- .rs.packageHelpEncoding(packagePath)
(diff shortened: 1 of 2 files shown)


gtritchie (approved):
Wacky! LGTM.

PR #8272, opened 2020-11-02
Fix an issue where the Sqlite DB is corrupted by writing to a stale file handle

MariaSemple:
### Intent

Rich and Karl found this issue with my initial fix for #7898 (PR #8117). This PR ensures that stale file handles are not kept.

### Approach

After the server is daemonized, explicitly reload the loggers. This has two benefits over going back to the old way we used to write to log files (open a new file handle everytime we write a log and close it immediately after):

* it's more efficient to open the file handle once and keep it open until we're done with it.
* the changes to `shared_core` will be easier to incorporate into the SDK than moving more code out of `core` into `shared_core`, which can sometimes be quite work intensive.

### QA Notes

* When using the server with a file logger and the Sqlite DB, it should not become corrupted.
* Logs should be written to the file destination correctly.


kevinushey (approved):
LGTM!

jmcphers (approved):

Issue #8256, opened 2020-10-31
Failure to load R when starting projects on Windows long path (versions 1.4xxx)


javbermejo:
### System details

    RStudio Edition :  Desktop
    RStudio Version :  Previews all 1.4xxx above (works with 1.3xxx)
    OS Version      :  Windows
    R Version       :  4.03

### Steps to reproduce the problem
Attempting to load a project in a long trajectory fails to load R and.

### Describe the problem in detail
When loading a project stored in a "long path" directory trajectory (Windows), R crashes. Works OK with RStudio versions 1.3xxxx

### Describe the behavior you expected
Load properly (as in versions 1.3xxx)

- [x ] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x ] If I am reporting a RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x ] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

![image](https://user-images.githubusercontent.com/73740700/97776930-a4309f80-1b6c-11eb-8024-8756d13dee15.png)
![image](https://user-images.githubusercontent.com/73740700/97776944-b9a5c980-1b6c-11eb-9162-55f4bde251fd.png)

kevinushey:
> When loading a project stored in a "long path" directory trajectory (Windows), R crashes. Works OK with RStudio versions 1.3xxxx

It would help if you could provide an example path wherein this issue reproduces.

javbermejo:
"C:\Users\javie\Madrid Digital\HGUGM CARDIOLOGIA - Investigación Imagen - Investigación Imagen\Indices of AS\Stats Indices of AS\Indices of AS.Rproj"
Opens OK with any version < 1.4xxxx

Yoshinobu-Ishizaki:
Same error occurs on me. 
I think this is related to the windows path which contains non-ascii character.
When I intend to create a new R project at "C:\\日\" folder ( 日 is a one of Japanese character), RStudio crashes with a same message below.

![image](https://user-images.githubusercontent.com/36043439/98119796-75118980-1ef0-11eb-8514-12f60a6106e2.png)


```
sessionInfo()
R version 3.6.3 (2020-02-29)
[...]

PR #8286, opened 2020-11-04
ensure profiler resources path is UTF-8

kevinushey:
### Intent

Attempts to open projects with non-ASCII characters in the pathname could fail in some cases. This was caused by an attempt to create a directory associated with a mis-encoded path on startup, which (in some cases) could be an error rather than a warning.

### Approach

Ensure that the profiler path is UTF-8 encoded, and also treat errors as warnings so they don't prohibit session start.

### QA Notes

On Windows: test that projects with multibyte characters can be created / opened.

Closes https://github.com/rstudio/rstudio/issues/8256.

### DIFF
--- a/src/cpp/r/RSexp.cpp
+++ b/src/cpp/r/RSexp.cpp
@@ -1329,6 +1329,18 @@ SEXP create(const std::map<std::string, std::string>& map, Protect* pProtect)
    return listSEXP;
 }
 
+SEXP createUtf8(const std::string& data, Protect* pProtect)
+{
+   SEXP strSEXP;
+   pProtect->add(strSEXP = Rf_allocVector(STRSXP, 1));
+
+   SEXP charSEXP;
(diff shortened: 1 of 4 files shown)


jmcphers (approved):
LGTM!

javbermejo:
RStudio Version 1.4.1010: 1) Session provides a warning but not crashing. 2) However, impossible to restore the workspace due to failure to translate to native encoding:

WARNING: Failed to restore workspace from 'C:/Users/javie/Desktop/HGUGM CARDIOLOGIA - InvestigaciÃ³n Imagen - InvestigaciÃ³n Imagen/Indices of AS/Stats Indices of AS/.RData'
Reason: unable to translate 'C:/Users/javie/Desktop/HGUGM CARDIOLOGIA - Investigaci<f3>n Imagen - Investigaci<f3>n Imagen/Indices of AS/Stats Indices of AS/.RData' to native encoding

kevinushey:
@javbermejo thanks for reporting back. Could you file a separate issue?

PR #8287, opened 2020-11-04
prevent Save As from incorrectly being disabled

### LINKED ISSUE #8277: Can not use "Save As"
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.4.999
    OS Version      : Windows 10 Enterprise 64-bit
    R Version       : 4.0.3 64-bit

Open any R script then try saving it as a different file

The option is not available

rstudio_201103094930

Save file as should work

melissa-barca:
### Intent

Fixes #8277 

`Save As` was disabled whenever `Save` was disabled preventing users from renaming documents.

### Approach

This bug was introduced with #8250. As mentioned in that issue, even when Save As is set properly the client side can lose sync with the desktop side causing the command to appear disabled so we need to force any updates to the command's state. Previously we were using the same logic to determine whether Save and Save As should be enabled, this PR updates the check for Save As.

### QA Notes

There are good repro steps in the original issue. Since this was introduced with a prior fix for #6475 we should test both issues.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/source/SourceColumn.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/SourceColumn.java
@@ -16,8 +16,6 @@ package org.rstudio.studio.client.workbench.views.source;
 
 import com.google.gwt.core.client.JsArrayString;
 import com.google.gwt.core.client.Scheduler;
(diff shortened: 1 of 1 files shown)


kevinushey (approved):
LGTM!

Issue #8293, opened 2020-11-04
JS exception when attempting to read console history


kevinushey:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.4.dev
    OS Version      : macOS 10.15.7
    R Version       : 4.0.3

### Steps to reproduce the problem

Run some R code that generates an error; e.g.

```
stop("Oh no!")
```

### Describe the problem in detail

On occasion, a JS error is seen in the console:

<img width="1920" alt="Screen Shot 2020-11-04 at 1 56 26 PM" src="https://user-images.githubusercontent.com/1976582/98172142-ad3faa80-1ea5-11eb-8967-350e6a7fddae.png">

This effects later attempts to access the console history.

### Describe the behavior you expected

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting a RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kfeinauer:
Verified fixed 1.4.1007

PR #8294, opened 2020-11-04
fix offset check when reading history

kevinushey:
### Intent

When attempting to access the most recent item in the command line history, we would (on occasion) read out of bounds of our history list, leading to a JS exception.

### Approach

Clamp the history offset to the range of the history list.

### QA Notes

Reproduce via notes in https://github.com/rstudio/rstudio/issues/8293.

Closes https://github.com/rstudio/rstudio/issues/8293.


kevinushey:
Might also fix https://github.com/rstudio/rstudio/issues/7928?

gtritchie (approved):
LGTM!

Issue #8302, opened 2020-11-05
Rainbow parentheses in 1.4.999-1 lose color in lines below raw string


dlependorf:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.4.999-1
    OS Version      : Windows 10 version 2004
    R Version       : 4.0.3

### Steps to reproduce the problem

1. Turn rainbow parentheses on in the options
2. Open a new script tab
3. Type a line with parentheses to test out the rainbow coloring
4. Type a line below the first line with a raw string, with the R 4.0 syntax of r"(some string)"
5. Type a line with parentheses below the line with the raw string

### Describe the problem in detail

When rainbow parentheses are turned on, any parentheses in a script file in lines below a line with a r"()" raw string are no longer colored. They appear as the same color as normal text (generally black for light themes and white for dark themes).

<img width="623" alt="rainbow parens" src="https://user-images.githubusercontent.com/10523844/98292450-9494ca80-1f61-11eb-93a1-f1d078f14225.png">

If that raw string line is either deleted or commented out, rainbow parentheses color instantly comes back for the affected lines. Alternatively, as soon as that "r" preface to the raw string is deleted, the color also comes back.

### Describe the behavior you expected

The expectation is that rainbow parentheses persist regardless of what is in the rest of the script.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting a RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kfeinauer:
Verified fixed 1.4.1013

PR #8308, opened 2020-11-06
avoid stomping on rainbow parens state with raw strings

kevinushey:
### Intent

In our tokenizer, we keep track of both the rainbow parentheses count as well as the complement for a raw string (if any). Those were using the same indices into the state array, and so could stomp on each other in some cases.

(Note: I experimented with both using an object to keep track of this state, as well as a "named" array, but both seemed to have problems; likely based on how Ace uses + manipulates that array. We could fix it with an Ace patch, but that's definitely out of scope for 1.4)

### Approach

Use a different index for raw strings.

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/8302.

Closes https://github.com/rstudio/rstudio/issues/8302.

### DIFF
--- a/src/gwt/acesupport/acemode/r_highlight_rules.js
+++ b/src/gwt/acesupport/acemode/r_highlight_rules.js
@@ -253,13 +253,12 @@ define("mode/r_highlight_rules", ["require", "exports", "module"], function(requ
           
           // initialize stack
           stack = stack || [];
-          stack.length = 2;
 
           // save current state in stack
           stack[0] = state;
 
           // save the expected suffix for exit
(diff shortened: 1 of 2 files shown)


adamconroy:
I also experimented with using named arrays when I was originally working on this, was very frustrating seeing odd errors being thrown for without a good reason. Full Typescript Ace rewrite time? 🤣

adamconroy (approved):

gtritchie:
> Full Typescript Ace rewrite time? 🤣

I think switching to Monaco editor would be a better use of our time, but it's not a small task.

PR #8311, opened 2020-11-06
Fix Close All Except Current with unsaved files

### LINKED ISSUE #8295: Close all except current does not close all
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.4.1002
    OS Version      :  Windows 10
    R Version       : 4.0.2

* Open RStudio
* Open somes files with an unsaved new Rscript in the middle
* Select one of the file (before last one for example) 
* Close All Except Current - Don't save when asked
* The file is there but is not the only kept 

close-all

Previous similar issue reported : https://github.com/rstudio/rstudio/issues/8092

Only the selected file should be kept

melissa-barca:
### Intent

Fixes #8295 

Certain combinations of opening files with "dirty" files (requiring a prompt to save) prevent `Close All Except Current` from closing all columns. 

### Approach

Previously we set the active editor to keep track of the editor to leave open and reset it when a change was triggered by the prompt to save other editors. We were only resetting the variable tracking the active editor and not which editor was actually active in `DocTabLayoutPanel`. This fix selects the active editor when it is set manually.

There was also an issue with multiple columns that left all the tabs open when the current editor was in the left most column due to columns being removed while iterating the column list. We now iterate over the list of column names rather than the columns themselves to make sure we reach each one while closing tabs. This is a marginally slower approach but shouldn't be noticeable since the column list size should always be small (currently always less than 5).

### QA Notes

There are good examples in #8295 but we also need to make sure things work as expected with multiple columns open.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/source/SourceColumn.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/SourceColumn.java
@@ -28,7 +28,6 @@ import org.rstudio.core.client.ResultCallback;
 import org.rstudio.core.client.StringUtil;
 import org.rstudio.core.client.command.AppCommand;
 import org.rstudio.core.client.command.CommandBinder;
-import org.rstudio.core.client.dom.WindowEx;
 import org.rstudio.core.client.events.*;
 import org.rstudio.core.client.files.FileSystemItem;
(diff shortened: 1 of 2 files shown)


kevinushey (approved):
LGTM!

PR #8312, opened 2020-11-06
remove duplicate copy of HasContextMenuHandlers interface

gtritchie:
### Intent

Remove unnecessary source file HasContextMenuHandlers.java; this interface is already defined exactly the same way in GWT itself.

Obviously not actually needed for 1.4, but very small cleanup. If we want to be super-strict please mark for julette-rose and I'll rebase the PR.

### Approach

Delete source file and fix the two locations that were referencing it, confirmed it compiles and check that right-click context-menu continues working in various locations. Also checked that there aren't any additional instances to worry about in pro-only code.

### QA Notes

Nothing specific to test here, just code cleanup related to custom right-click context menu.


jmcphers (approved):
LGTM! There's no possible regression here as far as I can tell so let's just take it for 1.4.

Issue #8321, opened 2020-11-09
Unable to load workspace in Windows due to encoding translation failure


javbermejo:
### System details

    RStudio Edition :  Desktop
    RStudio Version :  1.4.1010
    OS Version      :  Windows 10
    R Version       :  4.03

### Steps to reproduce the problem

When loading a project stored in a windows path which includes non UTF8 characters, workspace is not loaded. The full error in RStudio reads:

WARNING: Failed to restore workspace from 'C:/Users/javie/Desktop/HGUGM CARDIOLOGIA - InvestigaciÃ³n Imagen - InvestigaciÃ³n Imagen/Indices of AS/Stats Indices of AS/.RData'
Reason: unable to translate 'C:/Users/javie/Desktop/HGUGM CARDIOLOGIA - Investigaci<f3>n Imagen - Investigaci<f3>n Imagen/Indices of AS/Stats Indices of AS/.RData' to native encoding

### Describe the problem in detail
The Windows path is 'C:/Users/javie/Desktop/HGUGM CARDIOLOGIA - Investigación Imagen - Investigación Imagen/Indices of AS/Stats Indices of AS/.RData'

Tried different locales but error persists:
Sys.setlocale(locale = "Spanish")
[1] "LC_COLLATE=Spanish_Spain.1252;LC_CTYPE=Spanish_Spain.1252;LC_MONETARY=Spanish_Spain.1252;LC_NUMERIC=C;LC_TIME=Spanish_Spain.1252"
Warning message:
In list.files(cacheDir) :
  unable to translate 'C:/Users/javie/Desktop/HGUGM CARDIOLOGIA - Investigaci<f3>n Imagen - Investigaci<f3>n Imagen/Indices of AS/Stats Indices of AS/.Rproj.user/58CB7782/explorer-cache' to native encoding

### Describe the behavior you expected

- [ x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting a RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [ x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
Closed by https://github.com/rstudio/rstudio/pull/8322.

[...]

PR #8322, opened 2020-11-09
use UTF-8 path when restoring global environment

kevinushey:
### Intent

On Windows, users might be working within paths / projects that cannot actually be represented in the current locale. This can cause various issues.

### Approach

Use UTF-8 paths instead.

### QA Notes

Test that RStudio functions as expected (and you can save + load an R workspace) on Windows when the path contains non-ASCII characters; e.g. letters or words from a language not using the Latin alphabet.

Closes https://github.com/rstudio/rstudio/issues/8321.


jmcphers (approved):
This LGTM, but we don't have a live branch for 1.5 yet (and shouldn't as it's not the next release). Could you rebase to 1.4-juliet-rose and update NEWS-1.4-juliet-rose there?

PR #8325, opened 2020-11-09
specify formula switch in brew list invocations

gtritchie:
### Intent

Eliminate warning messages from `install-dependencies-osx`:
```
"Warning: Calling `brew list` to only list formulae is deprecated! Use `brew list --formula` instead."
```

### Approach

Supply the switch suggested by the warning message.

### QA Notes

Purely a dev and build environment thing. Nothing to test.


jmcphers (approved):
Thanks for keeping these clean!

Issue #8331, opened 2020-11-10
Python interpreter 'Select' button doesn't open on Enter


melissa-barca:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.4.1013
    OS Version      : macOS Catalina 10.15.7
    R Version       :  3.6.1

### Steps to reproduce the problem

1. Navigate to Global Options or Project Options Python tab
2. Tab to focus on `Select...` button.
3. Press `Enter` key

### Describe the problem in detail
The file chooser doesn't open. This does work for all other similar buttons in Global/Project options.

Adding to 1.4 because I assume this is a quick fix. We can remove if I'm wrong. 

### Describe the behavior you expected
The directory chooser should open.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting a RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

gtritchie:
FWIW, spacebar is the standard key to clicking a button via keyboard, but we do support Enter in most situations as well, so ideally we'd do it here, too.

melissa-barca:
Verified in 1.4.1024

PR #8333, opened 2020-11-10
register Enter handler only on textbox

kevinushey:
### Intent

The handler here was being registered on the entire widget, making it active for both the textbox and the Select button. This was preventing the Select button from reacting to Enter keypresses. 

### Approach

Scope the handler to the textbox only.

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/8331.

Closes https://github.com/rstudio/rstudio/issues/8331.


gtritchie (approved):
LGTM!

melissa-barca (approved):

Issue #8338, opened 2020-11-10
allow forward / backward navigation in Help pane with mouse buttons


kevinushey:
It would be useful if one could navigate backwards / forwards within the Help pane using the mouse side buttons, similar to how these buttons behave by default in a browser.

mikebessuille:
Mario verified this in 1.4-Juliet-Rose

PR #8339, opened 2020-11-10
allow forward / back in Help pane with mouse side buttons

### LINKED ISSUE #7462: Navigate Viewer and Help pane history's using mouse forward/back buttons
#7272 added the ability to navigate source history using a mouse with forward and back buttons (when mouse is in a source pane).

@yihui suggested supporting similar thing for Viewer history, and if we do that then probably should also do it for Help pane.

kevinushey:
### Intent

Make it possible to navigate backwards / forwards using mouse buttons.

### Approach

Fairly straightforward implementation of mouse handlers + JS callbacks to power the feature.

### QA Notes

Test that your mouse's side buttons allow you to navigate forwards / backwards within the Help pane.

Closes https://github.com/rstudio/rstudio/issues/8338.
Closes #7462.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/application/DesktopHooks.java
+++ b/src/gwt/src/org/rstudio/studio/client/application/DesktopHooks.java
@@ -34,7 +34,7 @@ import org.rstudio.core.client.widget.MessageDialog;
 import org.rstudio.core.client.widget.Operation;
 import org.rstudio.studio.client.application.events.EventBus;
 import org.rstudio.studio.client.application.events.LauncherServerEvent;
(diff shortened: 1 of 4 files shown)


gtritchie:
From Slack chat:

- existing issue for adding this to Help pane: #7462
- be sure to test Desktop; there are known issues with Qt 5.12 and these mouse buttons, see #7272

kevinushey:
Thanks @gtritchie: you were indeed right that the same gymnastics were required to ensure this works on desktop.

jmcphers (approved):
LGTM, but we definitely don't need this for 1.4. Want to merge it to Juliet Rose?

PR #8342, opened 2020-11-11
ensure F1 works for Python help when triggered from console

kevinushey:
### Intent

<kbd>F1</kbd> was not properly triggered the 'go to help' command for Python code when triggered from the R console.

### Approach

Unfortunately, it seems like the underlying bug lives in `reticulate`, and may be somewhat hairy to fix (it involves different help handler code that I'm not super familiar with). Until we have that sorted out, the simplest way forward is to just have our own patch in RStudio.

### QA Notes

Verify that help for Python objects can be requested both in Python scripts and R scripts. For example, in R:

```
library(reticulate)
np <- import("numpy", convert = TRUE)
np$random
```

And in Python:

```
import numpy as np
np.random
```

Closes https://github.com/rstudio/rstudio/issues/5982.

### DIFF
--- a/src/cpp/session/modules/SessionHelp.R
+++ b/src/cpp/session/modules/SessionHelp.R
@@ -377,10 +377,21 @@ options(help_type = "html")
 
 .rs.addJsonRpcHandler("show_custom_help_topic", function(helpHandler, topic, source) {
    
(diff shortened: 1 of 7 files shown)


gtritchie (approved):
LGTM!

PR #8356, opened 2020-11-12
Ensure files moved to Source Pane are visible

### LINKED ISSUE #8353: Removing extra source pane closes files
**Steps to reproduce:**
* Open several R files in the editor.
* Add a new column to the layout having `Source` type (`Tools` -> `Global Options` -> `Pane Layout` -> `Add column`).
* Move some opened file from the default source pane into the new source pane.
* Remove the newly added column (same settings page from where it was added).

**Expected behavior:** should move the files that were opened to the next available source pane.

**Actual behavior:** the files are no longer opened in RStudio.

**Setup:** RStudio 1.4.953.

melissa-barca:
### Intent

Closes #8353 

If a source columns with unsaved files is closed, the files are moved to the main source area. However, if the main source pane was minimized this wasn't apparent so we now pop it open. 

### Approach

When a file is added to a pane, ensure the pane is visible. Also added text to the Global Options pane so that the behavior is clear to the user. Depending on feedback, we may change this behavior in 1.5.

### QA Notes

Open several files in a new column and minimized the main source pane. Add text without saving to one of the files. Remove the column from the Global Options pane and confirm the unsaved file is now visible in the source pane.


gtritchie (approved):
LGTM!

Issue #8361, opened 2020-11-12
loading package in forked child process might cause hang


akersting:
### System details

    RStudio Edition : Server
    RStudio Version : 1.3.1093 patched with #6492
    OS Version      : Linux
    R Version       : 3.5.3 / 4.0.3

### Steps to reproduce the problem
Not easily possible, but see example below.

### Describe the problem in detail
As a result of #2597, #6492 checks for fork in package event callbacks. However, even with this patch applied a forked session might still hang when loading a package there. After facing this in the wild, I attached gdb to the stuck child:

Backtrace:
```
#0  0x00007f205912029c in __lll_lock_wait () from target:/lib/x86_64-linux-gnu/libpthread.so.0
#1  0x00007f2059119714 in pthread_mutex_lock () from target:/lib/x86_64-linux-gnu/libpthread.so.0
#2  0x0000564a29d9529b in ?? ()
#3  0x0000564a29d93643 in ?? ()
#4  0x0000564a29ea20ac in ?? ()
#5  0x00007f205935bfc5 in do_dotcall (call=0x564a2e4cfc40, op=<optimized out>, args=<optimized out>, env=<optimized out>) at dotcode.c:1252
```

RStudio was built without debug symbols, but:
```
(gdb) f 5
#5  0x00007f205935bfc5 in do_dotcall (call=0x564a2e4cfc40, op=<optimized out>, args=<optimized out>, env=<optimized out>) at dotcode.c:1252
1252	dotcode.c: No such file or directory.
(gdb) info locals
ofun = 0x564a29ea1f70
retval = <optimized out>
cargs = {0x564a2d775e88, 0x564b8e60b108, 0x564a2bb86620, 0x4008000000000000, 0x564a2ba5fc40, 0x7fff00000001, 0x7f20594aec48 <R_PPStackSize>, 0x564a2ba60308, 0x564a2e444408, 0x7f2000000000, 0x564a2ba60308, 0x564b303039d8, 0x564a2e444478, 
  0x87000000bb, 0x564a2ba60420, 0x564a2ba60420, 0x564a2ba603b0, 0x564a2c134e10, 0x564a2ba603b0, 0x564a2ba60420, 0x7f20593226d0 <tmp_cleanup>, 0x564b0c5c2660, 0x0, 0x100000000, 0x564a00000000, 0x564a2ca16658, 0x7fffd3b61100, 0x564b0c5c2970, 
[...]

PR #8364, opened 2020-11-12
don't enque client events from forked sessions

kevinushey:
### Intent

Avoid attempts to enque client callbacks from forked sessions. (Most commonly, these are package events from forked processes.)

### Approach

Simple check of `wasForked()` flag.

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/8361.

Closes https://github.com/rstudio/rstudio/issues/8361.


jmcphers (approved):
*enqueue!*

R-Hannibal:
@kevinushey - I know you said test via notes on 8361, but I am still not sure how to verify this after reading through the ticket 🤦‍♂️. Any help you can offer will be greatly appreciated.

kevinushey:
Reproducing this, I think, is challenging. In theory this should reproduce the issue, at least some of the time:

```
print(parallel::mcparallel({
  library("dplyr")
}))
```

But you may not have success since the issue is ultimately intermittent.

Issue #8357, opened 2020-11-12
Restarting R while writing Python has unintended consequences 


melissa-barca:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.4.1024
    OS Version      : macOS Catalina
    R Version       : 3.6.1

### Steps to reproduce the problem

1. Open a python script and run at least one line so that the python interpreter is started in the console. e.g. `string1 = "testing"`
2. Restart the R session without exiting the interpreter or closing RStudio. 
3. Attempt to run the line again, note the line attempts to run as R without opening the python interpreter.
4. Run another line of the python script and the interpreter will open. 

### Describe the problem in detail
Restarting the R Session while in a python interpreter prevents the first line of python in the new session from triggering `reticulate::repl_python()`. 

### Describe the behavior you expected
`reticulate::repl_python()` would be triggered before the python code is run. 

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting a RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

melissa-barca:
Verified in 1.4.1028

PR #8366, opened 2020-11-12
reset to R language after session restart

kevinushey:
### Intent

The RStudio front-end tries to keep track of the active language (Python vs. R). However, it was not properly tracking this state following a session restart.

### Approach

Since the console will revert to R after a session restart, we should also reset the tracked language to R as well.

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/8357.

Closes https://github.com/rstudio/rstudio/issues/8357.


melissa-barca (approved):
LGTM!

Issue #5984, opened 2020-01-09
A pane window won't keep minimised when working with Python. 


Fred-Wu:
I am currently using RStudio 1.2.5036 on a Linux machine. 

I setup my Pane Layout with History alone in the lower left, so that I can minimise it when working with R, and it is kept minimised just fine. 

However, when working with Python, everytime I run some codes, that window will be triggered and un-minimised automatically. 

Is this something that can be fixed?

ronblum:
@Fred-Wu Thank you for filing the issue. Is it limited to Python, or also R code? Also, what version of R and Linux are you using?

Fred-Wu:
> @Fred-Wu Thank you for filing the issue. Is it limited to Python, or also R code? Also, what version of R and Linux are you using?

I am working with R 3.6.2 on Ubuntu 18.04. 
It is limited to Python. Working with R is fine.
However, I can confirm that windows version of RStudio has no such an issue.

ronblum:
I am also able to reproduce this in
- RStudio Desktop 1.3.776 on MacOS 10.15.3
- RStudio Server 1.3.776 on Ubuntu 18.04.3 via MacOS 10.15.3 and Windows 10

but it works fine on
- RStudio Desktop 1.3.776 on Windows 10

Also, I see this in some other panes, like Environment.

This is a duplicate of #3084, but I'll close that one since this is already tagged.

--------

From @hrbrmstr in #3084:

> Using the desktop dailies (current version for this issue 1.2.780), macOS 10.14.
> 
> I'm a "console on the right" person and have history pane minimized below the source pane since I never really use it. When processing R code it stays minimized.
> 
> When i start a python code file in a source pane tab and start executing lines the history pane always unminimizes even when I re-minimize it.
> 
> Further, when going back to an R source pane tab in the same RStudio session, the first execution of a line of code causes a minimized history pane to unminimize, but it stays unminimized on subsequent R line executions.
> 
> I'll gladly clarify and/or post a screen cap when I get some time/if needed.

tjtnew:
Using a similar pane setup as that in #3084 (i.e. history below source and minimized) but not using python and I get the same unminimizing behaviour if I load a help file from console (e.g. `?ls`).  Other commands to not seem to have an effect.

hrbrmstr:
[...]

PR #8370, opened 2020-11-13
don't raise History pane on console reset

kevinushey:
### Intent

When a console history file is loaded, we automatically raise + focus the History pane. This is a bit too aggressive, especially in the context of reticulate since we need to switch + maintain separate history files for R + Python.

### Approach

Don't raise the History pane.

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/5984.

Closes https://github.com/rstudio/rstudio/issues/5984.


melissa-barca (approved):
LGTM!

PR #8379, opened 2020-11-13
don't migrate license metadata

gtritchie:
### Intent

- Fixes https://github.com/rstudio/rstudio-pro/issues/2129

### Approach

- Don't migrate verify file or license file to new XDG location; leave in same location for backwards/forwards compatibility

### QA Notes

- See issue for more details


jmcphers (approved):
It'd be nice if we could use the new XDG location (as it's configurable) but this is really the zero-friction change for existing customers so agree it's what we should do at this point.

PR #8382, opened 2020-11-14
use timer to collate multiple loadFinished calls

kevinushey:
### Intent

On rare occasions, multiple `loadFinished()` signals will be sent to the main window when attempting to load RStudio, with the first of those signal(s) setting `ok = false` with the final setting `ok = true`. This causes issues for us on load as we expect to only ever receive a single `loadFinished()` signal.

### Approach

Use a timer to collect and wait for other incoming `loadFinished()` signals, to see if we eventually receive a "success" signal.

### QA Notes

Unfortunately not sure how to reproduce consistently, but users encountering this issue will hopefully be able to report back.

Closes https://github.com/rstudio/rstudio/issues/8270.


MariaSemple (approved):
LGTM!

Issue #8391, opened 2020-11-15
Arrow in the console brings wrong statement after session restart


david-cortes:
**Steps to reproduce:**
* Start a new R session.
* Write and execute this command in the console: `print("before restart")`
* Click `Session` -> `Restart R`.
* Write and execute this command in the console: `print("after restart")`.
* Click `Session` -> `Restart R`.
* Write and execute this command in the console: `print("after restart2")`.
* Focus on the console and press the up arrow.

**Expected behavior:** should bring up `print("after restart2")`.

**Actual behavior:** brings up `print("before restart")`

**Setup:** RStudio 1.4.953.

ronblum:
Thanks @david-cortes ! I can reproduce this on MacOS 11.0.1 as well in RStudio Desktop 1.4.1032. This is not an issue in 1.3.1093, so this appears to be a regression.

mikebessuille:
Are we losing all the history after the restart?  or are we just showing the wrong command?  If the command history is lost, we should try to fix this for 1.4

kevinushey:
I think the issue here is a bit more general. It seems like, after a restart, the first-executed command is not added to the history. So, a similar example here would be:

1. Execute `print("About to restart!")`.
2. Restart the R session,
3. Execute `print("I restarted!")`,
4. Press Up to access command history.

The `print("I restarted!")` command is not entered into the history, even though the rest of the command history is preserved.

kevinushey:
Looks like it's caused by https://github.com/rstudio/rstudio/commit/f18ee378c24a991fa506fed00344a54c89705a27.

kfeinauer:
Verified fixed 1.4.1038

PR #8402, opened 2020-11-16
fix usage of addHistory flag

kevinushey:
### Intent

A recent PR (https://github.com/rstudio/rstudio/pull/8053) made some changes to ensure the console prompt was properly reset after a session restart, to avoid the case where e.g. a Python prompt was preserved even after the session restart returned to R mode.

This had the unintended side effect of causing the next-executed command to be omitted from the console history. This was due to my misunderstanding of the `addHistory` flag; I thought "surely we do not want to add the prompt to the console history" but this actually sets a flag that controls whether the next-processed command is added to the console history.

### Approach

We _do_ want the next-executed command to be added to the console history, so add it!

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/8391.

Closes https://github.com/rstudio/rstudio/issues/8391.


gtritchie (approved):
LGTM, and don't call me Shirley.

PR #8406, opened 2020-11-17
Stop using Ctrl+FKey shortcuts on macOS

### LINKED ISSUE #8401: keyboard shortcut for adding source column not working on Mac desktop
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.4.1033
    OS Version      : macOS 11.0.1, also repros on Catalina 10.15.7
    R Version       : 3.6.1

- start desktop IDE on macOS
- <kbd>Ctrl+F7</kbd> to add a source column (as seen under View / Panes / Add Source Column)

- nothing happens

- source column should be added (up the maximum of 3)

gtritchie:
### Intent

- The Ctrl+F1 through Ctrl+F8 shortcuts have existing meanings on macOS and shouldn't be used by applications
- They are either not passed through to the application at all, or come through inconsistently especially in RStudio Desktop
- Fixes #8401

### Approach

- On Mac change following shortcuts from Ctrl+FKey to Cmd+FKey:
    - Cmd+F1 - activate VCS pane
    - Cmd+F2 - activate Build pane
    - Cmd+F6 - activate Find in Files pane
    - Cmd+F7 - new source column
- On Mac, get rid of shortcut for Activate Connections altogether (was Ctrl+F5 which is reserved but Cmd+F5 is used for VoiceOver on macOS; given nobody has cared don't think eliminating it will be a big problem and users can always add their own mapping)

### QA Notes

Test that these shortcuts now work on Mac and are unchanged on Windows/Linux.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml
@@ -736,7 +736,7 @@ well as menu structures (for main menu and popup menus).
          <shortcut refid="activateBuild" value="Meta+F2" if="org.rstudio.core.client.BrowseCap.isMacintosh()"/>
          <shortcut refid="activateBuild" value="Ctrl+F2" if="!org.rstudio.core.client.BrowseCap.isMacintosh()"/>
          
-         <!-- Not using Cmd+F5 on Mac, it is the standard shortcut for toggling VoiceOver -->
+         <!-- Not using Cmd+F5 (aka Meta+F5) on Mac, it is the standard shortcut for toggling VoiceOver -->
          <shortcut refid="activateConnections" value="Ctrl+F5" if="!org.rstudio.core.client.BrowseCap.isMacintosh()"/>
          
          <shortcut refid="activateFindInFiles" value="Meta+F6" if="org.rstudio.core.client.BrowseCap.isMacintosh()"/>
--- a/src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml
@@ -731,11 +731,15 @@ well as menu structures (for main menu and popup menus).
          <shortcut refid="activatePackages" value="Ctrl+7"/>
          <shortcut refid="activateEnvironment" value="Ctrl+8"/>
          <shortcut refid="activateViewer" value="Ctrl+9"/>
-         <shortcut refid="activateVcs" value="Ctrl+F1"/>
-         <shortcut refid="activateBuild" value="Ctrl+F2"/>
-         <shortcut refid="activateConnections" value="Ctrl+F5"/>
-         <shortcut refid="activateFindInFiles" value="Ctrl+F6"/>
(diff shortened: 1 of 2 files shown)


src/gwt/www/docs/keyboard.htm
```diff
@@ -886,7 +886,7 @@ <h2 id="views">Views</h2>
     <tr>
       <td>Add source column</td>
       <td>Ctrl+F7</td>
-      <td>Ctrl+F7</td>
+      <td>Command+F7</td>
```
melissa-barca, line 889: Nit: Should we be consistent and use `Cmd+F7` instead of typing out Command?

src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml
```diff
@@ -731,11 +731,15 @@ well as menu structures (for main menu and popup menus).
...
-         <shortcut refid="activateVcs" value="Ctrl+F1"/>
-         <shortcut refid="activateBuild" value="Ctrl+F2"/>
-         <shortcut refid="activateConnections" value="Ctrl+F5"/>
-         <shortcut refid="activateFindInFiles" value="Ctrl+F6"/>
-         <shortcut refid="newSourceColumn" value="Ctrl+F7"/>
+         <shortcut refid="activateVcs" value="Meta+F1" if="org.rstudio.core.client.BrowseCap.isMacintosh()"/>
+         <shortcut refid="activateVcs" value="Ctrl+F1" if="!org.rstudio.core.client.BrowseCap.isMacintosh()"/>
+         <shortcut refid="activateBuild" value="Meta+F2" if="org.rstudio.core.client.BrowseCap.isMacintosh()"/>
+         <shortcut refid="activateBuild" value="Ctrl+F2" if="!org.rstudio.core.client.BrowseCap.isMacintosh()"/>
+         <shortcut refid="activateConnections" value="Ctrl+F5" if="!org.rstudio.core.client.BrowseCap.isMacintosh()"/>
```
melissa-barca, line 738: Should we add a comment here that `Meta+F5` is reserved on mac so someone doesn't add it while trying to be helpful?

melissa-barca (changes requested):

src/gwt/www/docs/keyboard.htm
gtritchie: This page is primarily used by screen-reader users (it replaces the inaccessible shortcut popup when screen-reader mode is enabled). "Command" is pronounced correctly with screen readers, but "Cmd" can have issues.

gtritchie: Oops, I see I have some that are "cmd", let me fix that! :-)

src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml
gtritchie: Was thinking we need a reference like that, there are so many platform-specific shortcut rules and I've yet to find a comprehensive site. Anyway, this one does seem very likely to get put back in given the pattern surrounding it so I will add a comment.

melissa-barca (approved):
LGTM!

Issue #8198, opened 2020-10-26
The Connections pane hangs indefinitely when attempting to re-establish a Spark connection via a New R Script.


R-Hannibal:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.4.972
    OS Version      : macOS 11.0 Beta (Big Sur)
    R Version       : 4.0.3

### Steps to reproduce the problem

1. Login to the IDE > Navigate to Connections pane > New Connection > Spark > Select 'OK' on the New Connection dialogue.
2. After the connection is established, disconnect from the connection ('Disconnect from a connection' button) > Yes.
3. Select the Connect dropdown in the Connections pane > select New R Script.

### Describe the problem in detail
A 'Connecting' message and progress spinner appear in the Connections pane and hangs there indefinitely (never seen it clear, stopped timing after 5 minutes).

![Connections Pane Hang](https://user-images.githubusercontent.com/62304541/97232700-a4671e80-17ab-11eb-995b-48080b0cf5ea.jpg)

### Describe the behavior you expected
I'd expect the same behavior seen in 1.3.1093: A new R Script is opened with the connection code, but the Connections pane itself remains static until the code in the R Script is run by the IDE user.

![Connections Pane Correct](https://user-images.githubusercontent.com/62304541/97232831-e6906000-17ab-11eb-9542-761a99d252aa.jpg)

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting a RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

R-Hannibal:
Adding this issue to 1.4 for consideration in the main release or patch since it is a regression from 1.3 functionality (it may not be serious enough for either, but I'd like it looked at 👍).

kevinushey:
I wasn't able to reproduce with macOS 10.15.7 (Catalina). Could this be specific to Big Sur?

R-Hannibal:
@kevinushey - Just re-tested after the Big Sur update to 11.0.1, and I'm still getting the same behavior. My ability to confirm on Catalina is limited since I updated to Big Sur without a Catalina back-up plan (😒).

[...]

PR #8430, opened 2020-11-18
fix precedence in null check

kevinushey:
### Intent

Closes https://github.com/rstudio/rstudio/issues/8198.

### Approach

Use parentheses to properly guard against `null` position when accessing its values. (🤦 )

### QA Notes

Test via @R-Hannibal's notes in https://github.com/rstudio/rstudio/issues/8198.


adamconroy (approved):
Oh Lazy Eval, why can't you read our minds!

PR #8432, opened 2020-11-18
provide an internal API for closing all source tabs without saving

### LINKED ISSUE #8425: Add internal command for closing all buffers without saving
For automation, it would be very helpful to have a command to close all open buffers (including satellite windows and source columns) without prompting to save modified buffers.

Perhaps simply: `.rs.api.closeAllSourceBuffersWithoutSaving()` or something equally ugly. Don't think it should be hooked to a menu or even the command-palette as accidentally triggering would be Bad (tm).

gtritchie:
### Intent

- When authoring Selenium automation it can be helpful to close all existing sources tabs without worrying about saving
- Fixes #8425

### Approach

- Invoke from console via `.rs.api.closeAllSourceBuffersWithoutSaving()`
- Made the name intentionally long and cumbersome to reduce likelihood of accidentally invoking it

### QA Notes

- All code added will only be hit if this internal API is invoked from Console, so very safe

### DIFF
--- a/src/cpp/r/R/Api.R
+++ b/src/cpp/r/R/Api.R
@@ -720,6 +720,10 @@
    
 })
 
+.rs.addApiFunction("closeAllSourceBuffersWithoutSaving", function() {
+   .Call("rs_documentCloseAllNoSave", PACKAGE = "(embedding)")
+})
+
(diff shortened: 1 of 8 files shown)


adamconroy (approved):
🙏

Issue #8453, opened 2020-11-22
Empty Rmd file freezes RStudio 1.4 in Bookdown project


GegznaV:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.4.1048
    OS Version      : Windows 10
    R Version       : 4.0.3

### Steps to reproduce the problem

1. Create a new `Bookdown` project.
2. Create a new empty Rmd document.
3. Save empty Rmd document in a project as, e.g., `00--empty-file.Rmd`
4. Try to close RStudio (it hangs until you try to close it again).
5. Try to open this project (sometimes some pales do not load, the expressions are not evaluated in the Console, and there are problems to close the project/exit RStudio).

I add the project in which the problem reproduces on my PC:
[bookdown-demo-proj.zip](https://github.com/rstudio/rstudio/files/5579973/bookdown-demo-proj.zip)

### Describe the problem in detail

After **step 4**, RStudio hangs in this state:
![01--rs-hangs-on-trying-to-close](https://user-images.githubusercontent.com/12725868/99911374-94842100-2cfc-11eb-89ae-33af8e3eddf9.gif)
Until this button is clicked again:
![image](https://user-images.githubusercontent.com/12725868/99911386-a9f94b00-2cfc-11eb-8702-e320e84f3d07.png)

The issues after performing **step 5** are illustrated below (I began to record the screen about 20 seconds after opening the project).
As you can see, some panes do not initialize, mathematical calculations are not evaluated in the console.
The green bar below is a time line:
![02-rs-hangs-after-restart](https://user-images.githubusercontent.com/12725868/99911415-ec228c80-2cfc-11eb-98d0-3b1eb179785f.gif)

### Describe the behavior you expected

I expect RStudio to work without freezing when empty Rmd files are present in an RStudio project.

I used only an empty Rmd file, but other situations should be tested too. E.g., no freezing should be present with:

- Empty files with other extensions.
- Rmd files that have no headings (that start with `#`) in the text.
- other non-bookdown projects.

*** 
- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting a RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
[...]

PR #8464, opened 2020-11-23
avoid running pandoc when we have empty .Rmd file

kevinushey:
### Intent

Closes https://github.com/rstudio/rstudio/issues/8453.

### Approach

Pandoc will hang if we try to run it and give an empty string as input.

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/8453.


jmcphers (approved):

mikebessuille:
The turnaround on this kind of issue is extraordinary.   :)

Issue #8465, opened 2020-11-23
no output shown when executing inline code chunk


kevinushey:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.4.dev
    OS Version      : macOS 11.0.1
    R Version       : 4.0.3

### Steps to reproduce the problem

Create a new R Markdown document, and then try to run an inline R chunk. For example, with the code:

`````
`r runif(1)`
`````

### Describe the problem in detail

A blank white output box is shown.

<img width="144" alt="Screen Shot 2020-11-23 at 3 05 35 PM" src="https://user-images.githubusercontent.com/1976582/100025655-111b0c00-2d9e-11eb-9907-883d4305506f.png">

### Describe the behavior you expected

The output of the executed command should be shown.

<img width="173" alt="Screen Shot 2020-11-23 at 3 05 27 PM" src="https://user-images.githubusercontent.com/1976582/100025671-15dfc000-2d9e-11eb-954b-df2636b70969.png">

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting a RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

melissa-barca:
Verified in 1.4.1063

PR #8466, opened 2020-11-23
don't use virtual console for inline chunk output

kevinushey:
### Intent

Closes https://github.com/rstudio/rstudio/issues/8465.

### Approach

Disable virtualized console for outputs generated from inline code chunks. Also add a safety fix in virtualscroller.js to ensure we do have a scroller element to inspect when adding output.

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/8465.

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/virtualscroller/virtualscroller.js
+++ b/src/gwt/src/org/rstudio/core/client/virtualscroller/virtualscroller.js
@@ -87,7 +87,12 @@ var VirtualScroller;
       }
 
       // jump to latest button
-      if (this.scrollerEle.parentElement.getElementsByClassName("jump-to-latest-console").length < 1) {
+      var hasJumpToConsoleBtn =
+        this.scrollerEle &&
+        this.scrollerEle.parentElement &&
+        this.scrollerEle.parentElement.getElementsByClassName("jump-to-latest-console").length < 1;
+
+      if (hasJumpToConsoleBtn) {
         this.jumpToLatestButton = document.createElement("div");
         this.jumpToLatestButton.classList.add("jump-to-latest-console");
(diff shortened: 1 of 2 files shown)


src/gwt/src/org/rstudio/core/client/virtualscroller/virtualscroller.js
```diff
@@ -87,7 +87,12 @@ var VirtualScroller;
       }
 
       // jump to latest button
-      if (this.scrollerEle.parentElement.getElementsByClassName("jump-to-latest-console").length < 1) {
+      var hasJumpToConsoleBtn =
+        this.scrollerEle &&
+        this.scrollerEle.parentElement &&
+        this.scrollerEle.parentElement.getElementsByClassName("jump-to-latest-console").length < 1;
```
adamconroy, line 93: Unfortunately these invariants will just hide bugs. Those two cases are invalid for how this is written. A scroller element with a parent is necessary for the virtualizer, at least at the moment. We could put a check higher up to throw an error early, like when it's traversing up the DOM to find the scroller element, but the assumption is that this class is written expressly to work inside an `ace_scroller` element.

adamconroy (changes requested):

src/gwt/src/org/rstudio/core/client/virtualscroller/virtualscroller.js
kevinushey: Roger that. I've made a change so that we throw a plain error if we cannot resolve the scroller element, so at least errors will be more "obvious" for us. Let me know if this looks okay.

adamconroy:
👍

adamconroy (approved):

PR #8472, opened 2020-11-24
use more robust test for writable library

kevinushey:
### Intent

In some cases, RStudio might mis-detect a non-writable R library as writable. This PR intends to be a bit more robust to those cases.

### Approach

Rather than trust the POSIX permissions, check that we can explicitly create and remove a dummy file within the library directory. (This is necessary since, in some cases, POSIX permissions might report that a directory is writable even if the user doesn't actually have write access to that directory.)

### QA Notes

Unfortunately, not sure how to reproduce, but this was reported as occurring with an installation of R living on a NFS.

Closes https://github.com/rstudio/rstudio-pro/issues/2184.

### DIFF
--- a/src/cpp/r/R/Tools.R
+++ b/src/cpp/r/R/Tools.R
@@ -681,9 +681,31 @@ environment(.rs.Env[[".rs.addFunction"]]) <- .rs.Env
 })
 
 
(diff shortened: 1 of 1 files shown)


jmcphers (approved):
LGTM, please update NEWS before merge!

Issue #8489, opened 2020-11-29
Comments in C++ are continued when they shouldn't


david-cortes:
**Steps to reproduce:**
* Create some C++ file in RStudio.
* Inside some C++ function, write a comment with old-style `/* */` marks.
* Put the focus at the end of the comment and press enter - e.g.
```c++
void some_fun(int &a)
{
  /* this is a comment */<focus here and press enter>
  a += 1;
}
```

**Expected behavior:** should switch to the beginning of a new line.

**Actual behavior:** continues the comment, even through it was already closed - i.e.:
```c++
void some_fun(int &a)
{
  /* this is a comment */
   * 
   */
  a += 1;
}
```

**Setup:** RStudipo 1.4.1056

ronblum:
Thanks @david-cortes ! I can reproduce this in RStudio Desktop 1.4.1063 on MacOS 11.1 beta, as well.

kevinushey:
Marking as 1.4 as the fix here will be small + low risk, and it effects the fundamental editing experience in C++ documents.

kfeinauer:
Verified fixed 1.4.1073. The multiline comment is now only continued if it is not yet complete (has a matching */).

PR #8496, opened 2020-12-01
fixup C++ multi-line comment behaviors

kevinushey:
### Intent

Closes https://github.com/rstudio/rstudio/issues/8489.

### Approach

Tweak the way we detect whether `/**` is "incomplete" and so we should complete the comment with `*/`. We now perform detection based on whether the next line already has a `*` comment marker on the next line.

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/8489.


gtritchie (approved):
LGTM!

PR #8522, opened 2020-12-02
check for rstudio pam profile on load

kevinushey:
### Intent

Check that `/etc/pam.d/rstudio` exists on rserver startup, and notify the developer if it is not available.

Screen Shot 2020-12-02 at 2 34 07 PM

### Approach

Simple checking + writing to stderr.

### QA Notes

Developer only; no QA required.


gtritchie (approved):
LGTM!

Issue #8524, opened 2020-12-03
avoid overflow in dialog box on big sur


kevinushey:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.4.dev
    OS Version      : macOS 11.0.1 (Big Sur)
    R Version       : Any

### Steps to reproduce the problem

Load and then attempt to update an R package. For example:

```
library(MASS)
install.packages("MASS")
```

### Describe the problem in detail

Shared from Slack:

<img width="280" alt="Screenshot 2020-12-03 at 10 37 04" src="https://user-images.githubusercontent.com/1976582/101068440-97c5ab00-354d-11eb-81b7-0c851d5dc683.png">

### Describe the behavior you expected

No overflowing on dialog boxes.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting a RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

gtritchie:
Verified on Big Sur 11.0.1 with RStudio 1.4.1077:

![2020-12-03_15-30-55](https://user-images.githubusercontent.com/10569626/101101093-a7101d00-357c-11eb-9929-8dccbb050246.png)

PR #8527, opened 2020-12-03
avoid dialog overflow via word-smithing

kevinushey:
### Intent

Closes https://github.com/rstudio/rstudio/issues/8524.

### Approach

Tweak wording to avoid overflowing dialog box used on Big Sur.

Screen Shot 2020-12-03 at 11 18 35 AM

### QA Notes

Test that the dialog box presented when running:

```
library(MASS)
install.packages("MASS")
```

does not overflow, as the one in https://github.com/rstudio/rstudio/issues/8524 does.


gtritchie (approved):
LGTM. Thank goodness we don't currently localize into German, we'd need to cut this text in half at least.

Issue #8511, opened 2020-12-02
Carriage Returns cause degenerate output with Limit Console Output turned on


adamconroy:
### System details

    RStudio Edition : Desktop or Server
    RStudio Version : 1.4
    OS Version      : 
    R Version       : 

### Steps to reproduce the problem

1. Turn on Limit Console Output
2. Follow the output posted by Kevin https://github.com/rstudio/rstudio/issues/8428#issuecomment-732379178 and notice that output gets jumbled around when using carriage returns.

### Describe the problem in detail

In certain cases, particularly when using multiple carriage returns in a single line, the output gets jumbled and not what you would expect. I have a pretty thorough set of tests that I'm using to narrow down exactly what's happening but I don't quite have my finger on it yet, the other issues that cropped up while investigating #8428 were pulling me away from the primary cause.

### Describe the behavior you expected

Carriage returns always bring the cursor to the beginning of the line.

adamconroy:
https://gist.github.com/adamconroy/862b08e80b6766c7c212bd12b40f09de

This is a gist of the ways I'm testing right now. There are different symptoms based on whether there are unicode characters, new lines, where the carriage returns are, the length of lines...and more. It's made narrowing down exactly what's going on tricky, especially with some non-regression weird functionality showing up as well.

mikebessuille:
We need to either fix this, or disable the console limiting feature until it's fixed.

kfeinauer:
Verified fixed 1.4.1078.

PR #8528, opened 2020-12-03
Ensure that VirtualConsole ranges don't write outside of VirtualScroller buckets

adamconroy:
### Intent

Fix https://github.com/rstudio/rstudio/issues/8511

### Approach

What was happening is VirtualConsole was doing in-range insertions based on the original parent instead of always relative to the original range element (that will be inside a VirtualScroller bucket). Change the `insertAfter` functions to always be relative to the original range.

### QA Notes

Make sure that the output in the test cases Kevin put in #8428 match with Limit Console Output turned on/off.


adamconroy:
*insert big sigh of relief and face palm*

kevinushey (approved):
LGTM!

PR #8530, opened 2020-12-03
Add "rename" command to document tab context menu

### LINKED ISSUE #8374: Add Rename File to document tab context menu
Once the document tabs have a context menu, it should contain the Rename File command. This will need to be aware of whether the tab contains a file-based entity (e.g. shouldn't be shown for non-document tabs that be displayed in the source region, columns, and satellite windows).

This request was extracted from #1664

gtritchie:
### Intent

Ability to rename a file by right-clicking on its tab is a common feature in IDEs.

Fixes #8374

### Approach

Added rename behavior to document tab context menu.

Can't directly use the rename command itself because it assumes the target is the currently selected editor which isn't necessarily the case when you right-click on a tab.

### QA Notes

Verify that "Rename" only appears when you right-click on tabs of actual saved documents, not on "Untitled" docs or non-document tabs. Behavior should be identical (for better or worse) with the existing File / Rename command on main menu.


adamconroy:
This is great!

adamconroy (approved):

PR #8570, opened 2020-12-08
also use tinytex when compiling .Rnw-generated .tex files

kevinushey:
### Intent

We added the ability to use [tinytex](https://yihui.org/tinytex/) a while back, when compiling standalone `.tex` files. This PR extends that to also work with the `.tex` files rendered from a Sweave (`.Rnw`) document.

### Approach

Relatively straight-forward application + use of the existing `runTinytex()` function; now used to render the `.tex` file generated after weave.

### QA Notes

Ensure that the `tinytex` R package is installed and in use:

```
options(timeout = 3600)
library(tinytex)
tinytex::install_tinytex()
```

Ensure that you've configured RStudio to use tinytex when rendering `.tex` files:

Screen Shot 2020-12-08 at 3 08 02 PM

Then try creating a `.Rnw` document, and then rendering it. For example:

`````
\documentclass{article}

\begin{document}

<<>>=
print(1 + 1)
@

\end{document}
`````

Confirm that the rendered PDF is opened after successful compilation, and that the Compile PDF tab (next to the console) reports that tinytex was used:

Screen Shot 2020-12-08 at 3 18 30 PM

Closes https://github.com/rstudio/rstudio/issues/2788.

### DIFF
--- a/src/cpp/session/modules/tex/SessionCompilePdf.cpp
+++ b/src/cpp/session/modules/tex/SessionCompilePdf.cpp
@@ -628,7 +628,7 @@ private:
       }
       else if (prefs::userPrefs().useTinytex())
       {
-         runTinytex();
+         runTinytex(targetFilePath_);
       }
       else
       {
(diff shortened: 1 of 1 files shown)


jmcphers (approved):
LGTM, please update NEWS when you merge!

Issue #8588, opened 2020-12-09
C++ compilation database needs to be invalidated when compiler updated


kevinushey:
### System details

    RStudio Edition : Both
    RStudio Version : All
    OS Version      : macOS 11.0.1
    R Version       : 4.0.3

### Steps to reproduce the problem

Distilled from discussion in https://github.com/rstudio/rstudio/issues/4685#issuecomment-741950741. Roughly, the steps would be:

1. Use RStudio in a project with C++ sources,
1. Close RStudio,
1. Upgrade your C++ compiler (e.g. on macOS, upgrade command line tools),
1. Re-open RStudio in that project.

### Describe the problem in detail

Diagnostics may fail to run in that project, as the cached compilation arguments + headers are no longer valid.

### Describe the behavior you expected

We should invalidate our compilation database if we detect that the compiler has been updated.

---

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting a RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
Closed by https://github.com/rstudio/rstudio/pull/8591.

astayleraz:
I verified that the correct diagnostics are emitted following the steps outlined. I verified on Mac desktop Version 1.4.1572 with R 4.0.3. Closing as fixed.

PR #8591, opened 2020-12-09
rebuild compilation database when compiler changes

kevinushey:
### Intent

For projects using C++ source files, RStudio maintains some internal state describing how files in that project should be compiled. This data was not being invalidated when the compiler was changed, leading to cases where a new compiler was being used with old compiler flags, which were not always compatible and lead to issues.

### Approach

Include the compiler's `--version` output as part of the compilation database, and invalidate the database if that changes. (The compiler is queried from R via `R CMD config {CC/CXX}`. This check is not perfect, but should be "good enough" for 99% of users.)

### QA Notes

This is a bit challenging to test. You will have to:

1. Create a project using C++ sources,
2. Edit some C++ files within the project, and confirm that C++ diagnostics are working as expected,
3. Change the default compilers used by R (this can be done via `~/.R/Makevars`),
4. Restart RStudio, and try editing files in the project again. Confirm that the correct diagnostics are emitted.

Closes https://github.com/rstudio/rstudio/issues/8588.

### DIFF
--- a/src/cpp/core/libclang/SourceIndex.cpp
+++ b/src/cpp/core/libclang/SourceIndex.cpp
@@ -259,7 +259,7 @@ TranslationUnit SourceIndex::getTranslationUnit(const std::string& filename,
    // report to user if requested
    if (verbose_ > 1)
    {
-      std::cerr << "COMPILATION ARGUMENTS:" << std::endl;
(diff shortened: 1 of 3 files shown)


gtritchie (approved):
LGTM!

PR #8595, opened 2020-12-10
Update openssl on Windows to v1.1.1i

### LINKED ISSUE #8574: rstudio on windows depends on an OpenSSL version with known vulnerabilities.
From https://github.com/rstudio/rstudio/blob/master/dependencies/windows/install-openssl/install-openssl.R

```
URL <- "https://www.openssl.org/source/openssl-1.1.1d.tar.gz"
```

Check https://www.openssl.org/news/vulnerabilities-1.1.1.html for an overview of security vulnerabilities that have been fixed since then.

gtritchie:
### Intent

- Update to latest bug fix release of OpenSSL 1.1.1i
- Fixes #8574

### Approach

- Built new version and uploaded archive to S3
- Updated dependency scripts and CMake file

### QA Notes

- Should be covered by routine usage and testing; openssl is historically very stable between these minor releases


kevinushey (approved):
LGTM! Always nice when existing build scripts "just work" with a new version of software...

gtritchie:
Windows desktop only; I would let it be covered by community usage and overall final regression passes, it is a very low risk change.

PR #8619, opened 2020-12-15
adapt test script to changes in #8614

kevinushey:
### Intent

Ensure that `rsession` tests can still run even with the changes in #8614.

### Approach

Make sure `R_HOME`, `R_DOC_DIR` and `R_LIB_DIR` are always defined. In addition, force `rsession` to use a version of `libR.dylib`, since we explicitly don't link directly to the R library (so users can choose different versions of R at runtime).

### QA Notes

No QA; this is just to get tests up and running again.

### DIFF
--- a/src/cpp/rstudio-tests.in
+++ b/src/cpp/rstudio-tests.in
@@ -131,6 +131,16 @@ else
     RSTUDIO_SESSION_BIN="rsession"
 fi
 
+# set R environment variables needed by rsession tests
+export R_HOME=$(R --slave --vanilla -e "cat(paste(R.home('home'),sep=':'))")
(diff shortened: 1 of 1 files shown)


kfeinauer (approved):
LGTM!

gtritchie:
@kevinushey Do you need to do something similar for the Windows test script?

https://github.com/rstudio/rstudio/blob/master/src/cpp/rstudio-tests.bat.in

kevinushey:
@gtritchie Thanks for catching that -- the Windows tests appear to be equally upset. I'll take a look.

kevinushey:
The Windows tests appear to be happy now -- they were broken before because I was building against an experimental version of R.

Issue #8552, opened 2020-12-07
Limit Console Output always scrolls to the bottom


adamconroy:
### System details

    RStudio Edition : <Desktop or Server
    RStudio Version : 1.4.X
    OS Version      : 
    R Version       : 

### Steps to reproduce the problem

1. Turn on Limit Console Output
2. Create a function that outputs constantly to the console
3. Observe that the console will always scroll to the bottom

### Describe the problem in detail

It's difficult to navigate around the console while it continues to output because it always scrolls to the bottom.

I spent a day investigating this and the cause isn't what I thought it was. I was originally doing this on purpose and thought that I could just comment this line out and we'd be good: https://github.com/rstudio/rstudio/blob/eca45fbb0a356667241d406fd82c529b598c0347/src/gwt/src/org/rstudio/core/client/virtualscroller/virtualscroller.js#L268

But it is not the case. Something else is scrolling it to the bottom. I tried commented out pretty much everything in virtualscroller.js and wasn't able to find the cause in a few hours so going to just file this issue and put it on Juliet-Rose.

### Describe the behavior you expected

If you're scrolled up it doesn't auto scroll to the bottom when new content is appended to the console.

valerie-rstudio:
### Verify Failed

    RStudio Edition : Desktop
    RStudio Version : 1.4.1527
    OS Version      : mac OS 11.1
    R Version       : 4.0.2

### Failure Points
When a long running function completes its final output, the console scrolls to the bottom.
1. Spam the console with any input until you can scroll.
2. Run this R code `for (x in 1:5)
{
  print(x)
  Sys.sleep(5)
}`
3. While the above code is running, scroll all the way up on the console.
4. **Expected Result:** The console remains in the same scroll position.
5. **Actual Result:** The console scrolls to the bottom once the code is finished.

### Points Verified
- When a long running function outputs to the console, and it is not the final output of that function, then the console maintains the users scroll position.

adamconroy:
@valerie-rstudio I believe that we are properly maintaining the functionality without the option turned on. In the testing I just did now it only scrolls to the bottom when the long running command starts or stops but it doesn't scroll as output is added during the `for` loop. This is the same functionality as without the "limit output" option turned on so I think we're in the clear here. Can you double check that?

kevinushey:
There are two components to the behavior to be verified here:

1. If the user has scrolled console history to the bottom, then it should remain at the bottom, even as new output is streamed in.

[...]

Issue #8544, opened 2020-12-05
Resizing or even clicking console border moves focus to the end


david-cortes:
**Steps to reproduce:**
* Print enough output in the console so as to make it scrollable.
* Scroll to the middle of the console output.
* Hover the mouse around the border of the console pane from which it's possible to resize it by dragging a left click.
* Click the border, but don't even move the mouse (so as not to resize anything).

**Expected behavior:** should do nothing, and resizing should ideally maintain the focus.

**Actual behavior:** focus moves to the end.

**Setup:** RStudio 1.4.1078.

ronblum:
@david-cortes I can reproduce this using RStudio 1.4.1078 on MacOS 11.1. I believe you use Debian and Windows? Which would mean the issue occurs in RStudio Desktop on multiple operating systems.

Note: This is not an issue in RStudio Server, using Safari, Chrome, and Firefox on MacOS 11.1.

mikebessuille:
Until this past weekend, the console had "Limit console display to a subset of total content" enabled by default.  As of this weekend it is disabled by default.  Can you verify which setting you had when you encountered this issue?

adamconroy:
@ronblum im moving to juliet rose on the basis that it's like due to the Limit feature that we have turned off, can you confirm?

ronblum:
Confirmed! Turning off the option makes the problem go away. The problem returns when when it's turned back on.

valerie-rstudio:
### Verified Fixed

    RStudio Edition : Desktop
    RStudio Version : 1.4.1527
    OS Version      :  mac OS 11.1
    R Version       :  4.0.2

Verified that when user attempts to resize console, the console maintains users scroll position with the Limit Console Output option either on or off.

Issue #8529, opened 2020-12-03
RStudio v1.4.1074-1 Preview Bug - Console jumps around


jzadra:
I just installed the RStudio v1.4.1074-1 Preview. The console jumps around. Hitting "latest" does something, but doesn't solve it. Hitting escape a few times will stop it temporarily, but then it will start again after scrolling.

https://imgur.com/a/og7sxtl

─ Session info ───────────────────────────────────────────────────────────────────────────────────────────────────────

version R version 4.0.2 (2020-06-22)
os macOS Catalina 10.15.7
system x86_64, darwin17.0
ui RStudio
language (EN)
collate en_US.UTF-8
ctype en_US.UTF-8

gtritchie:
May already be covered by other fixes (or at least other existing issues) but will let @adamconroy make that determination.

lrdegeest:
Also seeing this. Same version of RStudio. Windows machine.

adamconroy:
This is likely the same as the above issue. I will need a reprex.

kevinushey:
@jzadra @lrdegeest: can you please provide a reproducible example? Preferably, a snippet of code that we could copy + paste to run in RStudio, and then scroll around and see the jumpy console issue you're referring to.

jzadra:
Sorry I uninstalled it because it was unusable, so I can't test.  It happened as soon as the console in the new session had enough output that it started to scroll.  

One thought to try if it isn't occurring is that I have my tibble options set to print 50 rows instead of the default (10 i think?).  I also have the GUI setting for "Limit output line length to:" at 100.

My tibble options are set in the options block of my .Rprofile:
`  tibble.print_max = 70, tibble.print_min = 50`

adamconroy:
@jzadra thanks, I'll see if I can reproduce it with those settings. An example data frame would go a long way as well but I'll do my best.

For reference you can go to options -> Console -> Limit console output and that will restore the old console functionality so you don't have to uninstall.

valerie-rstudio:
### Verified Fixed

    RStudio Edition : Desktop
    RStudio Version : 1.4.1597
    OS Version      : mac OS 11.2.1
    R Version       : 4.0.3

Verified that the console no longer appears to be jumping around. Tested with the tibble settings `tibble.print_max = 70, tibble.print_min = 50`. Also tested various console scrolling and focus functionality with Limit Console Output turned off to ensure no other bugs were introduced.

Issue #8504, opened 2020-12-01
Jumping console with wide tibble


avsmith:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.4.1056
    OS Version      :  MacOS 10.15.7
    R Version       :  4.0.3

### Steps to reproduce the problem

- Load tibble with lots of columns. (I have an 5002 row/85 column example, and 2398 row/75 column example)
- Then try to look at snippet of tibble (by entering the tibble name)

### Describe the problem in detail

The console screen then jumps, with a blue "Latest" button appearing in lower right. Rather than 10 rows being displayed, there is a cycle of tibbles being displayed with varying numbers of rows (<= 10). The console then jumps around.

I can stop the console jumping by displaying a narrower tibble. I have to get the wider tibble to scroll off the top of the screen to stop the jumping.

### Describe the behavior you expected

A non-jumping tibble shown in the console.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting a RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [ ] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

gtritchie:
Try turning off the "Limit console display to a subset..." option. We are working on several fixes in this area but did recently turn this option on by default.

![screenshot of global options highlighting mentioned option](https://user-images.githubusercontent.com/10569626/100808220-975fcf80-33e8-11eb-8db3-e6b775598ceb.png)

gtritchie:
Adding to 1.4 so it gets on the radar.

adamconroy:
Hi @avsmith, thanks for the report. Can I get a minimal reprex as well the size of your screen (if you're on a 13" MacBook or a 40" iMac)?  I couldn't repro it on my machine so there might be a specific thing I'm missing about the configuration.

avsmith:
My examples are with sensitive data, so I wasn’t able to even make a screencast of the behavior. I’ll try to make an example in a bit.

FWIW, the suggestion to turn off "Limit console display to a subset…” did make the jumping stop.

-a

david-cortes:
Not sure if #8535 is the same issue, but it's still happening in 1.4.1078.

[...]

PR #8623, opened 2020-12-16
Fix various virtualscroller autoscrolling bugs

adamconroy:
### Intent

Fixes https://github.com/rstudio/rstudio/issues/8544
Fixes https://github.com/rstudio/rstudio/issues/8504
Fixes https://github.com/rstudio/rstudio/issues/8529
Fixes https://github.com/rstudio/rstudio/issues/8552

Wanted to get #8568 also in here but there's some weirdness related to GWT going on 😢 

### Approach

The weird Tibble related bugs were due to how I was counting the number of items in a bucket. A super wide Tibble is a ton of unique `span` elements that can all be on a single line. This caused the calculation of how large a bucket is to be wildly inaccurate so it didn't know if it was scrolled all the way up or down. The `isBucketFull` function in `virtualscroller.js` solves this by just counting newlines in the `innerText` which seems to work pretty well for now, though I did notice a performance hit which might need some attention.

The console always scrolling to the bottom is a...more complex issue. It happens in a lot of places based on a whole bunch of different events and there are a few different `scrollToBottom` insertion points that are hard to grab in a catch all. The `BottomScrollPane` makes a lot of assumptions based on events that happen entirely within GWT and the state doesn't seem to update cleanly between pure JS and the GWT event loop. I refactored out some of the built in autoscrolling and conditionally moved that to virtualscroller.js but there is still more work to be done. This [comment](https://github.com/rstudio/rstudio/issues/8568#issuecomment-745748922) in #8568 shows some of the complexity inherent in having the content and scrolling managed in multiple places.

### QA Notes

The most important thing in testing this will be to ensure that nothing with the Limit Console Output feature turned off is broken. Always good to be a little extra cautious when changing rerouting old code.

### DIFF
--- a/src/cpp/rstudio-tests.in
+++ b/src/cpp/rstudio-tests.in
@@ -131,6 +131,16 @@ else
     RSTUDIO_SESSION_BIN="rsession"
 fi
 
+# set R environment variables needed by rsession tests
+export R_HOME=$(R --slave --vanilla -e "cat(paste(R.home('home'),sep=':'))")
+export R_DOC_DIR=$(R --slave --vanilla -e "cat(paste(R.home('doc'),sep=':'))")
+export R_LIB_DIR=$(R --slave --vanilla -e "cat(paste(R.home('lib'),sep=':'))")
+
+# On macOS, we need to tell the rsession executable where to look for libR.dylib
+if [ "$(uname)" = "Darwin" ]; then
+   export DYLD_INSERT_LIBRARIES="$R_LIB_DIR/libR.dylib"
+fi
+
 # Run core tests
 if [ -z "$TEST_SCOPE" ] || [ "$TEST_SCOPE" == "core" ]; then
     echo Running 'core' tests...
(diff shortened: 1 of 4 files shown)


kevinushey:
LGTM!

R-Hannibal:
@adamconroy - Could you provide any more specificity to the QA notes on this PR?

`...ensure that nothing with the Limit Console Output feature turned off is broken...`

I absolutely agree with caution, but if the only way to truly verify this fix is a general regression of IDE functionality, we may want to check the individual tickets and then allow a combination of early adopter usage and certification checks to provide that more general coverage.

Note: Apologies if you meant specific testing around the console with the Limit Output option turned on. I'm reading that as a general regression of the IDE and it's scaring me 😆.

kevinushey:
Testing this is hard because it's hard to quantify all of the ways that users might interact with the Console pane, and the ingrained expectations they have on the way it works. It's only after one of those expectations is violated that the issue starts to stand out.

I would test things like:

- Clicking on the console pane sends focus to console input;
- The console stays scrolled to the bottom when lots of streaming output is emitted;
- The other various glitches reported in the past are ironed out.

Basically, you need to think and behave like a regular R user and see if anything has changed.

PR #8626, opened 2020-12-16
Don't switch to visual mode if there are unprocessed source capsules

jjallaire:
Fixes https://github.com/rstudio/rstudio/issues/8617

### DIFF
--- a/src/gwt/panmirror/src/editor/src/nodes/rmd_chunk/rmd_chunk-capsule.ts
+++ b/src/gwt/panmirror/src/editor/src/nodes/rmd_chunk/rmd_chunk-capsule.ts
@@ -21,6 +21,7 @@ import { pandocAttrReadAST, kCodeBlockAttr, PandocAttr, kCodeBlockText } from '.
 import { uuidv4 } from '../../api/util';
 
 export function rmdChunkBlockCapsuleFilter() {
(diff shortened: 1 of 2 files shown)


jmcphers (approved):
LGTM!

jjallaire:
@jmcphers and @mikebessuille I'd like to make the case that this fix be included in v1.4 proper. My reasoning:

1) When it occurs it results in user data/content loss. We can only speculate about how often it might occur, however....

2) A user just sent me an Rmd to repro a bug in the Distill package and lo and behold it had an unprocessed source capsule in it! So unscientifically it seems like this can happen probably more frequently that we'd like. It could even happen to hundreds or even thousands of Rmds.

3) The fix is *extremely* safe. No new code paths, just a regex that when matches disqualifies visual mode and using an existing code path we have for implementing the disqualification.

I don't think that merging this would disqualify any regression testing we have already done as I think we can reason well about it's locality of impact.

mikebessuille:
I could get behind that.  Especially if the fix is made to Master this weekend, because then we can still cut an RC on Monday. :)

jjallaire:
Okay, I'll. merge it to master now. Thank you! I will sleep much better at night knowing there is no way we can ever munge user data.

Issue #8655, opened 2020-12-22
Searches for hyphens in visual mode don't return matches


ronblum:
### System details

    RStudio Edition : Desktop and Server
    RStudio Version : 1.4.1096
    OS Version      : MacOS 11.1 and Safari on MacOS 11.1 pointing at Red Hat 8.3
    R Version       : 4.0.3

### Steps to reproduce the problem

1. Open an R markdown file in visual mode.
2. Type a string that contains a hyphen, e.g. `hi-there`
3. <kbd>Edit -> Find</kbd>
4. Search for the string containing the hyphen.
    - No search results appear.

### Describe the problem in detail

Hyphens don't show up in search results.

### Describe the behavior you expected

Searching for hyphens in visual mode should identify matches.

Note:
1) This works in source mode.
2) This works for other non-alphanumeric characters, such as:
    - En-dash: `–`
    - Em-dash: `—`
    - Brackets:`[`
    - Emojis: `😃`

[I'm marking this as 1.4 to review it for Wax Begonia. I expect we'll decide handle it later, and if we do, I suggest we fix it in Juliet Rose.]

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting a RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

jjallaire:
This is caused by the `u` flag (for unicode regexes) which invalidates our conversion from string to regex here: https://github.com/rstudio/rstudio/blob/master/src/gwt/panmirror/src/editor/src/behaviors/find.ts#L78

Discussion here: https://esdiscuss.org/topic/why-is-regexp-u-a-syntax-error

jjallaire:
Will likely need to use Unicode escapes: https://javascript.info/regexp-unicode

astayleraz:
I verified the fix in mac desktop Version 1.4.1514. Find in visual mode now shows words with hyphens.

PR #8664, opened 2020-12-29
visual editor: correctly handle finding words w/ punctuation

jjallaire:
Fixes https://github.com/rstudio/rstudio/issues/8655

The cause of the problem was using normal regex character escapes (e.g. `\-`) in a unicode regex. See discussion here: https://esdiscuss.org/topic/why-is-regexp-u-a-syntax-error

The fix is to use full unicode escape sequences (e.g. `\u002d`) rather than simple ascii character based escapes.

@jmcphers Could you review to make sure there aren't any hidden gotchas w/ this approach?


jmcphers (approved):
LGTM!

PR #8676, opened 2021-01-04
Add selectors to find in files for selenium tests

melissa-barca:
### Intent

Add selectors to the buttons and table in the `Find in Files` pane so we can access them in unit tests to be added in rstudio/rstudio-ide-automation.

### Approach

Added Class IDs for Fast Select Tables and Left Right Toggle Buttons.

### QA Notes

This shouldn't need any additional QA. We'll know it's working as expected when I submit the unit test.


src/gwt/src/org/rstudio/core/client/widget/FastSelectTable.java
```diff
@@ -107,6 +110,11 @@ public void onKeyDown(KeyDownEvent event)
          }
       });
    }
+   
+   public void setClassId(String name)
+   {
+      table_.setClassName(ClassIds.FAST_SELECT_TABLE + "_" + name);
```
gtritchie, line 116: Can this potentially result in class names with spaces in them?

src/gwt/src/org/rstudio/core/client/widget/LeftRightToggleButton.java
```diff
@@ -85,10 +86,21 @@ public LeftRightToggleButton(String leftLabel, String rightLabel,
...
       Roles.getButtonRole().setAriaPressedState(right_, leftIsOn ? PressedValue.FALSE : PressedValue.TRUE);
    }
 
+   public void setClassId()
+   {
+      ClassIds.assignClassId(getElement(),
+         ClassIds.LEFT_RIGHT_TOGGLE_BTN + " _ " +
+            left_.getInnerText() + "_" +
+            right_.getInnerText());
+      ClassIds.assignClassId(left_, ClassIds.LEFT_TOGGLE_BTN + "_"  + left_.getInnerText());
```
gtritchie, line 100: Same as other question, does this potentially result in class ids with spaces in them?

gtritchie (changes requested):

src/gwt/src/org/rstudio/core/client/widget/LeftRightToggleButton.java
melissa-barca: You're right, thanks! I've fixed the issue.

src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/FindOutputPane.java
```diff
@@ -134,6 +135,7 @@ public void onKeyUp(KeyUpEvent event)
       FormLabel replaceLabel = new FormLabel("Replace with: ", replaceTextBox_);
       replaceToolbar.addLeftWidget(replaceLabel);
       replaceToolbar.addLeftWidget(replaceTextBox_);
+      Roles.getButtonRole().setAriaLabelProperty(replaceTextBox_.getElement(), "Replace with");
```
gtritchie, line 138: Why was this added? The replaceTextBox_ already has an accessible label via the above `FormLabel replaceLabel`.

gtritchie (changes requested):

src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/FindOutputPane.java
melissa-barca: Thanks for pointing this out, I've removed it

gtritchie (approved):
LGTM!

PR #8686, opened 2021-01-06
detect newer plumber tags

### LINKED ISSUE #8118: Recognize all plumber files
When a "traditional" Plumber file is being edited:

```r
library(plumber)

#* @apiTitle Plumber Example API

#* Echo back the input
#* @param msg The message to echo
#* @get /echo
function(msg = "") {
    list(msg = paste0("The message is: '", msg, "'"))
}
```

the IDE presents the option to Run API:
image

However, in version 1.0.0, Plumber now includes a "tidy interface", so APIs can also be defined as follows:
```r
library(plumber)

pr() %>% 
  pr_get(path = "/echo",
         handler = function(msg = "") {
           list(msg = paste0("The message is: '", msg, "'"))
         })
```

It would be nice if the IDE presented the same Run API option when editing these files as well

UPDATE (2021-01-05)
Per comments later in this issue, the recommended best practice we should support is:

```r
library(plumber)

#* @plumber
function(pr) {
  pr %>%
    pr_get(path = "/echo",
           handler = function(msg = "") {
             list(msg = paste0("The message is: '", msg, "'"))
           })
}
```

gtritchie:
- Fixes #8118

### Intent

Light-up IDE plumber integration when newer tags are seen, not just the original plumber tags.

### Approach

Updated the regex based on list from plumber team.

### QA Notes

Validate scenario as described in #8118

### DIFF
--- a/NEWS-1.4-juliet-rose.md
+++ b/NEWS-1.4-juliet-rose.md
@@ -18,6 +18,7 @@
 * Change default per-user install folder to %LocalAppData%\Programs on Windows (#8598)
 * Cmd+U now toggles underlining in the visual editor on macOS (#8656)
 * Improve YAML cursor position after omni-insert in the visual editor (#8670)
+* Detect newer plumber tags when enabling plumber integration (#8118)
 
 ### RStudio Server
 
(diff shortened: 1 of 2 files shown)


kevinushey (approved):
LGTM!

Issue #8185, opened 2020-10-25
RStudio and reticulate


henry090:
Hi @kevinushey. RStudio preview version now throws an error if I run the following command:

```
> rstudioapi::restartSession()

Restarting R session...

Error in py_call_impl(callable, dots$args, dots$keywords) : 
  Evaluation error: Unable to access object (object is from previous session and is now invalid).
Error in py_call_impl(callable, dots$args, dots$keywords) : 
  Evaluation error: Unable to access object (object is from previous session and is now invalid).
Error in py_call_impl(callable, dots$args, dots$keywords) : 
  Evaluation error: Unable to access object (object is from previous session and is now invalid).
```

Now, is it an issue for RStudio desktop or reticulate?

kevinushey:
Reprex:

```
library(reticulate)
x <- py_eval("1 + 1", convert = FALSE)
rstudioapi::restartSession()
```

kevinushey:
Fixed in 1.4-JR via https://github.com/rstudio/rstudio/pull/8736.

astayleraz:
I verified this in Version 1.4.1507-1 and R 3.6.0 and was able to run the code provided with non errors. I also tested this in the mac desktop Version 1.4.1514.

braverock:
I think this issue or a closely related one has arisen a recent RStudio Preview Version 1.4.1722

```
Error in py_call_impl(callable, dots$args, dots$keywords) : 
  Evaluation error: Unable to access object (object is from previous session and is now invalid).
```
Behavior seems somewhat inconsistent, I can make it go away with a completely clean environment and restart of RStudio server.  I don't have a small reproducible example yet...

kevinushey:
@braverock please let me know if you've able to find a reproducible example; if you are please feel free to file a new issue.

JosiahParry:
I am now encountering this issue. 

``` r
library(reticulate)
use_condaenv("geo")
GDP <- import("geopandas")
LIBPYSAL <- import("libpysal")

fp <- file.path(LIBPYSAL$examples$root, "guerry", "Guerry.shp")

guerry <- GDP$read_file(fp)

py_get_attr(guerry, "geometry")
#> Error in py_get_attr_impl(x, name, silent): Unable to access object (object is from previous session and is now invalid)
```

<sup>Created on 2022-11-02 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


Conda-env: 

```yaml
name: geo
channels:
[...]

PR #8736, opened 2021-01-13
properly handle Python null pointers when building Environment pane

kevinushey:
### Intent

When a session is restarted, it's possible that some R objects acting as proxies for Python objects are saved and then later restored as null pointers. We need to handle these specially to avoid errors on restart, as reported in https://github.com/rstudio/rstudio/issues/8185.

### Approach

Check for null pointers up-front and handle them specifically.

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/8185.

Closes https://github.com/rstudio/rstudio/issues/8185.


gtritchie (approved):
LGTM, be sure to add to NEWS

Issue #8698, opened 2021-01-07
RStudio loads xfun and yaml packages for no reason


arencambre:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.4.1102
    OS Version      : Windows 10 20H2
    R Version       : 4.0.3

### Steps to reproduce the problem
Open RStudio and do nothing else. In my case, a project is not loaded, but an unsaved R script and a view of an object named `x` come up since RStudio restores my prior state.

### Describe the problem in detail
When I update packages, I am prompted to restart R:
![image](https://user-images.githubusercontent.com/411495/103903289-5468da00-50c1-11eb-9842-8d53e6a2429d.png)

That happens because one of the packages ready for an update is `xfun`, and rsession.exe has a lock on xfun.dll (and yaml.dll):
![image](https://user-images.githubusercontent.com/411495/103904183-8cbce800-50c2-11eb-86a4-98bd50ed21e7.png)

This is despite that RStudio says neither package is loaded:
![image](https://user-images.githubusercontent.com/411495/103904905-87ac6880-50c3-11eb-8909-6a7209f6fbb2.png)

Reviewing [xfun's documentation](https://cran.r-project.org/web/packages/xfun/index.html), the **Reverse imports** section suggests it's used mainly with libraries associated with Rmds and also possibly with other libraries.

The problem is I have no Rmds open:
![image](https://user-images.githubusercontent.com/411495/103904362-caba0c00-50c2-11eb-8c0a-5ca305cffca5.png)

Pressing **Yes** on that dialog caused a successful update of `xfun`. However, after closing and reopening RStudio, it still puts a lock on `xfun.dll` and `yaml.dll`.

### Describe the behavior you expected
RStudio does not cause rsession.exe to load libraries that are needed for functionality I am not using.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting a RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

arencambre:
Something similar is happening on a different PC:
[...]

PR #8739, opened 2021-01-14
avoid auto-loading yaml, tinytex packages

kevinushey:
### Intent

On startup, we were auto-loading the `yaml` and `tinytex` packages. This is undesirable, especially on Windows, as it makes it very challenging to update these packages within an RStudio session.

### Approach

- `tinytex`: use our own shim for finding the root of a `tinytex` installation, modeled after the function itself.
- `yaml`: use [yaml-cpp](https://github.com/jbeder/yaml-cpp) rather than the R `yaml` package for parsing YAML when enumerating tutorials.

### QA Notes

First, ensure the `tinytex` and `yaml` packages are installed:

```
install.packages(c("tinytex", "yaml"))
```

Then, open the Tutorial pane, and restart the R session. After restarting the R session, check the output of:

```
loadedNamespaces()
```

Confirm that neither `yaml` nor `tinytex` are in the list of installed packages.

---

Closes https://github.com/rstudio/rstudio/issues/8698.


jmcphers (approved):
Nice, there are a couple of other places where we can now also stop depending on the yaml R package (we can get those later).

Issue #8549, opened 2020-12-07
Python cp949 encoding error


issactoast:
### System details

    RStudio Edition : 
    RStudio Version : 1.4.1056
    OS Version      : Windows10
    R Version       : 4.0.3 (2020-10-10)

### Steps to reproduce the problem
print korean letters in Rstudio python py file and run
``
print("안녕")
``
### Describe the problem in detail
If I run the print Korean word using the print function, it shows the following error.
```
reticulate::repl_python()
Python 3.8.3 (C:/Users/issac/miniconda3/python.exe)
Reticulate 1.18 REPL -- A Python interpreter in R.
>>> print("안녕")
UnicodeEncodeError: 'cp949' codec can't encode character '\xc8' in position 1: illegal multibyte sequence
```
I check out the same code in miniconda python, it shows Korean 안녕 correctly.

### Describe the behavior you expected
print Korean word in console "안녕"

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting a RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

melissa-barca:
Thank you for reporting this issue. I can reproduce in 1.4.1081 on Windows. 
@kevinushey do you think this is something we can get into 1.4?

kevinushey:
I don't think this quite meets the bar for 1.4 since we're so close to release, but let's definitely fix for 1.4-JR.

astayleraz:
I tested this in desktop Version 1.4.1514 and get the following error still `> reticulate::repl_python()
Python 3.6.12 (C:/Users/IEUser/AppData/Local/r-miniconda/envs/r-reticulate/python.exe)
Reticulate 1.18.9005 REPL -- A Python interpreter in R.
>>> print("안녕")
UnicodeEncodeError: 'charmap' codec can't encode characters in position 0-1: character maps to <undefined>
>>> print("你好")
UnicodeEncodeError: 'charmap' codec can't encode characters in position 0-1: character maps to <undefined>`

kevinushey:
[...]

PR #8748, opened 2021-01-15
use Python-compatible unicode escapes

kevinushey:
### Intent

Ensure that unicode inputs are handled correctly when the Python REPL is active.

### Approach

When escaping unicode text as console input, use an escaping format that is compatible with both the R and Python REPLs.

### QA Notes

Test that the following can be run from a Python script on Windows:

```
print("안녕")
print("你好")
```

Screen Shot 2021-01-15 at 2 52 32 PM

---

https://github.com/rstudio/rstudio/issues/8549.


gtritchie (approved):
LGTM!

Issue #8567, opened 2020-12-08
Long character variables get truncated in console


david-cortes:
**Steps to reproduce:**
* Create a very large character variable in R that would span hundreds of lines.
* Print said variable through `cat(long_char)`.

**Expected behavior:** should print all the lines if possible, tell whether truncation occurred if not, or make it deducible from what's showing at the beginning of the console.

**Actual behavior:** prints only the tail of the variable. Before the print starts, the `cat` statement and whatever was there before show correctly and are still visible in the console, despite the output from `cat` having been truncated.

**Setup:** RStudio 1.4.1078.

Attached is an example serialized variable to reproduce.
[long_char.zip](https://github.com/rstudio/rstudio/files/5661426/long_char.zip)

(The file needs to be renamed to `.rds` and loaded with `readRDS`)

kevinushey:
Simpler reproducible example:

```
long <- as.character(1:1E4)
writeLines(long)
```

Note that this will be fixed in the latest daily builds (>= 1.4.1087) as we've disabled virtualized console output by default for now. For posterity, I see:

<img width="233" alt="Screen Shot 2020-12-08 at 1 55 08 PM" src="https://user-images.githubusercontent.com/1976582/101546023-ff6c6380-395c-11eb-9ed2-6f5614fe0792.png">

And also "holes" in the output:

<img width="145" alt="Screen Shot 2020-12-08 at 1 55 37 PM" src="https://user-images.githubusercontent.com/1976582/101546054-0f844300-395d-11eb-89b6-7162d9770475.png">

adamconroy:
This is the same issue as https://github.com/rstudio/rstudio/issues/7712 which supposedly was fixed in https://github.com/rstudio/rstudio/pull/7909/files

The code is untouched so I don't know why it's resurfacing.

It is a Desktop only bug which explains my issues reproducing it, forgot about that. Will have to see if that communication can be equalized between versions too, not sure why there is any difference.

valerie-rstudio:
### Verified Fixed

    RStudio Edition : Desktop
    RStudio Version : 1.4.1540
    OS Version      : mac OS 11.1
    R Version       : 4.0.3

Verified that the console output of a variable that is 10,000 lines no longer has "holes" in the output, and also clearly appears truncated (see image).
[...]

PR #8779, opened 2021-01-21
VirtualScroller: prune fix, MacOS click focus fix

adamconroy:
### Intent

Address issues:
#8740  
#8567

### Approach

The focus bug was confusing me quite a bit. Turns out MacOS treats focus differently than Windows. Windows is very aggressive about making sure that the most obvious TextArea/Input will get focus even if you don't set it in the JS, MacOS is....much less forgiving. The fix was trivial, just allow the old focus code to run and only skip the "scroll to bottom" portion.

For the pruning issue I turned back on the code that I had removed when fixing the Clear Console issues. I think I was just a little too exuberant about removing code that didn't seem to be necessary.

### QA Notes

The only potential problem area I see is possibly Clear Console getting into some funky states with the pruning put back in. I tested it in a handful of cases but I could have missed something.

For the focus issue I would just do a quick smoke across Mac/Win/Linux to make sure that it feels good on all platforms.

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/VirtualConsole.java
+++ b/src/gwt/src/org/rstudio/core/client/VirtualConsole.java
@@ -649,6 +649,13 @@ public class VirtualConsole
          match = match.nextMatch();
       }
 
+      Entry<Integer, ClassRange> last = class_.lastEntry();
+      if (last != null)
+      {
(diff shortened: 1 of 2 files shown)


kevinushey (approved):
LGTM!

Issue #8551, opened 2020-12-07
Encoding errors with Windows and non-ASCII chars in account name


melissa-barca:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.4.1081 (but also 1.3 and 1.2)
    OS Version      : Windows 10
    R Version       : 4.0.3

#### Python Repro
1. Create user with special characters in their name (e.g. Crème Brûlée)
2. Make sure Python is installed somewhere within the user's home directory. Navigate to Global Options >> Python and choose Select to find an interpreter. You'll see invalid characters
![image](https://user-images.githubusercontent.com/5323711/101382427-f2307580-3885-11eb-85a6-25776a9672c9.png)
3. Attempt to select an interpreter. You'll see an error message.
![image](https://user-images.githubusercontent.com/5323711/101382520-0f654400-3886-11eb-9d55-d9d8e08c1c56.png)

#### Find in Files Repro
1. Create user with special characters in their name (e.g. Crème Brûlée)
2. Create a file in the user's home directory containing some text.
3. Press `Ctrl + Shift + F` to bring up Find in Files. Attempt to search for text within the file you created.
4. There will be 0 results.

### Describe the problem in detail
We are misencoding the user's path and therefore can't locate it. This is likely happening in other places in the IDE as well. 

### Describe the behavior you expected
We should be able to read files and executables within the user's directory. 

- [X] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [X] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting a RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
In addition, I think those copies of Python need to be ignored as well -- they're actually just stubs, and IIRC executing them in an interactive prompt actually prompts you to install Python from the Windows Store. 🙃

melissa-barca:
Interesting, then it's not identifying the interpreter that is installed here: C:\Users\Crème Brûlée\AppData\Roaming\Microsoft\Windows\Start Menu\Programs\Python 3.9
But maybe that's just because I haven't added it to my path?

mikebessuille:
[...]

PR #8802, opened 2021-01-25
fixups for Python environments on Windows

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/8551.

### Approach

- Ensure Python path is UTF-8 encoded, as this is the expected encoding for our front-end UI.
- Don't display "invalid" interpreters in the dropdown list.

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/8551.


gtritchie (approved):
LGTM!

Issue #8805, opened 2021-01-25
ignore anaconda environments found in revdep folders


kevinushey:
<img width="994" alt="Screen Shot 2021-01-25 at 1 32 56 PM" src="https://user-images.githubusercontent.com/1976582/105768942-f2f29900-5f11-11eb-87bf-fece8dfd5874.png">

ronblum:
Verified in
- RStudio Desktop 1.4.1540 on MacOS 11.1
- RStudio Server 1.4.1533 on Red Hat 8.3

I cheated a bit; instead of using `revdepcheck::revdep_check()`, I used `reticulate::conda_create()` and changed the name of the environment to `revdep`.

Issue #8784, opened 2021-01-22
support Python installations installed via pyenv


kevinushey:
Users can install Python via [pyenv](https://github.com/pyenv/pyenv), and those normally get installed into a known location (`~/.pyenv/versions`). We should support the lookup and use of Python installations installed in this way.

ronblum:
Verified for both global and project options in
- RStudio Desktop on MacOS 11.1
- RStudio Server on Red Hat 8.3
- RStudio Server on Ubuntu 20.04.1

Note: Ran into this issue along the way. https://github.com/rstudio/reticulate/issues/944

PR #8806, opened 2021-01-25
support lookup of pyenv python; filter some conda envs

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/8805, https://github.com/rstudio/rstudio/issues/8784.

### Approach

Support the lookup of Python installations within the `~/.pyenv` folder, as installed by [pyenv](https://github.com/pyenv/pyenv) -- or similarly, `reticulate::install_python()`.

### QA Notes

First, run the following code to install Python 3.7.7 with pyenv:

```
remotes::install_github("rstudio/reticulate")
reticulate::install_python(version = "3.7.7")
```

Then, check that this installation is found when looked up. For example, I see it here on my system (among other discovered Python installations):

Screen Shot 2021-01-25 at 1 40 35 PM

### DIFF
--- a/src/cpp/session/modules/SessionPythonEnvironments.R
+++ b/src/cpp/session/modules/SessionPythonEnvironments.R
@@ -176,6 +176,7 @@
 {
    interpreters <- c(
       .rs.python.findPythonSystemInterpreters(),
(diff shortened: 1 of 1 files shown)


gtritchie (approved):
LGTM!

Issue #8785, opened 2021-01-22
show Stop button when running long-running Python computation


kevinushey:
For example, try running the following code in Python:

```
import time
time.sleep(10)
```

No Stop button is displayed, and so the computation cannot be interrupted via UI.

mikebessuille:
yes for Juliet-Rose, assuming the functionality for interrupting Python is already done.

astayleraz:
I verified in Version 1.4.1527 that running python the user can now stop what is running. Closing as fixed.

Issue #8763, opened 2021-01-19
allow reticulate Python REPL to be interrupted


kevinushey:
Note: needs to be implemented in `reticulate`, but filing here to scope as part of 1.4-JR work.

kevinushey:
Fixed via https://github.com/rstudio/reticulate/pull/928.

For testing, first install the development version of `reticulate` with:

```
remotes::install_github("rstudio/reticulate")
```


Then, in a Python script, try to run the following Python code:

```
import time
time.sleep(10)
```

Then, try to interrupt execution in a couple ways; e.g. by pressing Ctrl + C, or Escape, or by pressing the Stop button.

astayleraz:
I verified on mac Version 1.4.1527 and r 4.0.3 that I was able to interrupt execution pressing Ctrl + C, or Escape, or by pressing the Stop button. Closing as fixed.

PR #8807, opened 2021-01-25
allow user to interrupt Python REPL

kevinushey:
### Intent

This PR ensures that the Stop button is displayed in the Console Pane when the Python REPL is busy, and clicking that will interrupt the associated Python computation.

Addresses https://github.com/rstudio/rstudio/issues/8763.
Addresses https://github.com/rstudio/rstudio/issues/8785.

### Approach

Listen for reticulate "busy" events, and use those to control the display + enabled-ness of the R interrupt command (which is handled appropriately when `reticulate` is running).

### QA Notes

Note: requires the development version of `reticulate` for interrupt support. Install it with:

```
remotes::install_github("rstudio/reticulate")
```

Then, try running a Python script with the contents:

```
import time
time.sleep(30)
```

and check that the associated computation can be interrupted.

https://user-images.githubusercontent.com/1976582/105775097-19690200-5f1b-11eb-990d-3bd36b169f73.mov

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/events/ReticulateEvent.java
+++ b/src/gwt/src/org/rstudio/studio/client/events/ReticulateEvent.java
@@ -27,7 +27,8 @@ public class ReticulateEvent extends GwtEvent<ReticulateEvent.Handler>
       }
 
       // Event data accessors ----
-      public final native String getType() /*-{ return this["type"]; }-*/;
+      public final native String getType()           /*-{ return this["type"]; }-*/;
+      public final native JavaScriptObject getData() /*-{ return this["data"]; }-*/;
 
    }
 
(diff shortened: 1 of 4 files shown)


src/gwt/src/org/rstudio/studio/client/events/ReticulateEvent.java
```diff
@@ -70,8 +76,9 @@ protected void dispatch(Handler handler)
    public static final Type<Handler> TYPE = new Type<>();
 
    // synchronize with SessionReticulate.R
-   public static final String TYPE_PYTHON_INITIALIZED = "python_initialized";
+   public static final String TYPE_PYTHON_INITIALIZED = "pythAon_initialized";
```
gtritchie, line 79: Is this change intended?

kevinushey: It definitely is not -- thanks for catching that.

Issue #5158, opened 2019-07-30
Rstudio hangs when debugging code or calling "stop", rsession at 100% CPU


slhck:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.2.1564
    OS Version      : macOS 10.14
    R Version       : 3.6.1 (Homebrew)

### Steps to reproduce the problem

I am loading a few large data files and running some computations on them. I would like to debug one of the computation steps, so I naively set a `stop("foo")` within the function that is called.

### Describe the problem in detail

Once the `stop` point is hit, I see in the console:

```
Error in (function (.df, .kpi_col, .user_segments, .lower_is_better = TRUE,  : 
    foo
```

And the R session is still doing… something. I see the 🛑button, which indicates that something is computing, I see the CPU for `rsession` at 100%, and I can't terminate or restart R from Rstudio.

Hitting _Session » Interrupt R_ or _Session » Terminate R…_ is doing nothing. Hitting _Session » Restart R_ is showing the *Restarting R…* info window, but it's not actually restarting R. I've waited for minutes, and nothing is happening.

So somehow it seems like Rstudio has lost the connection with the R session. The worst part is when the UI is unresponsive; I cannot save changes to the file I have open, and the only thing that works is completely killing Rstudio.

The same happens when I set a breakpoint at the same location. I see the "Browse" message in the console, but the debug buttons never appear, and the code editor never switches to the debug view (the one where the affected code is highlighted). The CPU goes to 100%, nothing happens for minutes.

sessionInfo():

```
R version 3.6.1 (2019-07-05)
Platform: x86_64-apple-darwin18.6.0 (64-bit)
Running under: macOS Mojave 10.14.5

Matrix products: default
BLAS:   /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib
LAPACK: /usr/local/Cellar/openblas/0.3.6_1/lib/libopenblasp-r0.3.6.dylib

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] parallel  stats     graphics  grDevices utils     datasets  methods  
[8] base     

other attached packages:
 [1] nortest_1.0-4      optparse_1.6.2     tidylog_0.1.0     
 [4] digest_0.6.19      modelr_0.1.4       broom_0.5.2       
[...]

PR #8817, opened 2021-01-26
fix issue with debugger hang when using do.call()

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/5158.

Normally, deparsing a call is cheap, as the only things embedded in a call object are symbols, calls, and literals. However, some calls (notably, those generated when a function is invoked via `do.call()`), will be very expensive to deparse, as deparsing those implies also deparsing the objects embedded in those calls.

Similarly, the "expression" associated with the promise of such a call might not actually be an expression, as `do.call()` can force early evaluation of its arguments -- and so R generated a promise mapping to an already-evaluated expression; that is, the expression is just the requested value itself.

### Approach

Detect and handle both of these cases when opening the debugger.

### Automated Tests

Unit tests included.

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/5158. Fix has some risk as it touches the code used for formatting things for display in the debugger + environment pane.

### DIFF
--- a/src/cpp/session/modules/SessionEnvironment.R
+++ b/src/cpp/session/modules/SessionEnvironment.R
@@ -377,7 +377,10 @@
 
 .rs.addFunction("functionNameFromCall", function(val)
 {
-   call <- attr(val, "_rs_call", exact = TRUE)
+   call <- .rs.nullCoalesce(
+      attr(val, "_rs_call", exact = TRUE),
+      val
+   )
    
    if (is.function(call[[1]]))
       return("[Anonymous function]")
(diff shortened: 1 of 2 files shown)


src/cpp/session/modules/SessionEnvironment.R
```diff
@@ -364,16 +377,83 @@
...
+         object
+      }
+      else
+      {
+         type <- .rs.explorer.objectType(object)
+         as.name(sprintf("<%s>", type))
+      }
+   }
    else
-      as.character(substitute(call))
```
kevinushey, line 371: Note: I believe this worked somewhat by accident, since calling `as.character()` on a call returns a character vector of the same length as the call. For example:

```
> call <- quote(c(a, b, c))
> as.character(call)
[1] "c" "a" "b" "c"
```

But since we only ever looked at the first element of the vector, it just so happened to produce the right result.

https://github.com/rstudio/rstudio/blob/58a3c0f1d2a51977f9d5e24b956ab02f948a0fcf/src/cpp/r/RSexp.cpp#L884-L894

(note that we just extract the first element of the associated vector, and silently ignore the rest)

jmcphers (approved):
It makes me just a little nervous to change this but I think we have enough runway left in Juliet Rose for it. Nice work. Needs a NEWS entry!

Issue #8460, opened 2020-11-23
Panes without tabs should always be minimized


melissa-barca:
### System details

    RStudio Edition : Both
    RStudio Version : 1.3+
    OS Version      : All
    R Version       : All

### Steps to reproduce the problem

1. Navigate to the Pane Layout section of Global Options and hide all tabs from one or both tab sets.
2. Select Apply.

### Describe the problem in detail
In 1.3 we added the ability for a user to hide all tabs in a pane, however a gray square is still displayed in its place (see picture). This pane will pop open in other circumstances as well. 

### Describe the behavior you expected
Empty panes should default to always being minimized. 

![image](https://user-images.githubusercontent.com/5323711/99985863-7d851380-2d7c-11eb-9404-486221640564.png)

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting a RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

astayleraz:
I tested this on Version 1.4.1534 and I saw the bottom pane minimize but the top one was not minimizing when I cleared all items set to show. Setting back to in progress.

astayleraz:
Verified in mac desktop Version 1.4.1541 that panes are minimizing and displaying based on if there are items/tabs in those panes or not. Closing as fixed.

PR #8828, opened 2021-01-27
Minimize panes without tabs and show panes with newly added tabs

melissa-barca:
### Intent

Fixes issue addressed in #8460 

1. When a user updates their pane layout through Global Options or `rstudio-prefs.json`, hide newly empty tab panes or show panes that were previously empty but now have tabs. 
2. Updated code that restores panes to "default" in several cases. Previously we always restored a pane's `WindowState` to `NORMAL`, this update sets the state to `MINIMIZE` if the pane is an empty tab pane. 

### Approach

1. Replace code that iterated over all panes and set their state to normal with a function that determines if the state should be `NORMAL` or `MINIMIZED`. 
2. Update `panes` user preference `ValueChangeHandler` to check if a tab set pane has newly become emptied or possible and if so, update the state accordingly. 

### Automated Tests

https://github.com/rstudio/rstudio-ide-automation/issues/150

### QA Notes

1. See #8460 
2. Configure your layout so one tab pane is completely empty, then zoom into any non-empty pane. Zoom the pane a second time to return to your original configuration and the hidden pane should remain hidden.


gtritchie (changes requested):
The issue is marked v1.5 so either need to wait for 1.5 to do this or change issue to be 1.4-juliet-rose if you think it meets the bar.

Also needs a NEWS entry.

melissa-barca:
@gtritchie Thanks, I've updated the original issue and NEWS.md

gtritchie (approved):
LGTM!

Issue #8594, opened 2020-12-10
Add "whole word" filter to "Find in files" 


warnes:
I often make use of the `Find in Files` tool. Currently, matching a whole word requires enabling `Regular expression` and adding `\b` before and after the word.

It would be helpful to have a `Whole word` checkbox, as is already present in the code editor's `Find` tool.

mikebessuille:
leaving in 1.4-JR assuming it is low-risk

valerie-rstudio:
### Verified Fixed

    RStudio Edition : Desktop
    RStudio Version : 1.4.1533
    OS Version      : mac OS 11.1
    R Version       : 4.0.2

Verified that Find in Files now has a "whole word" checkbox that causes find to only give the search results where the whole search word or entire search phrase is found. Also verified that the "whole word" checkbox and "regex" checkbox cannot be checked at the same time (if one is checked, and the other is clicked, the first is deselected and the other is selected).

PR #8836, opened 2021-01-28
Add word option to Find in Files

melissa-barca:
### Intent

Adds a "Whole word" option to Find in Files to be consistent with our in-editor Find / Replace. This allows the user to search for a complete word or phrase.

### Approach

Added a "Whole word" checkbox and field to the Find in Files state. When checked, the frontend passes the query to the backend as a regular expression padded with `\b` tags e.g. `\bsearchphrase\b`. The backend doesn't know the difference between a regular express and whole word query. 

For simplicity's sake the user cannot enter a regular expression when the Whole word box is checked to prevent unexpected results. The thought is that if the user has a more complicated query than they'd likely want to include the `\b` tags themselves. 

### Automated Tests

rstudio/rstudio-ide-automation#155 includes the infrastructure for testing the Find in Files Dialog and Pane with simple tests that they can be loaded with the expected elements. I've included the new checkbox in this test as well as added aria labels to the other checkboxes so they can be found by automated tests. The PR is on hold until this one gets merged.
Testing the functionality of the new option will be handled as part of rstudio/rstudio-ide-automation#99 which is currently in progress. 

### QA Notes

See original request in #8594

### Checklist

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/FindInFilesDialog.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/FindInFilesDialog.java
@@ -70,7 +70,7 @@ public class FindInFilesDialog extends ModalDialog<FindInFilesDialog.State>
             path: path,
             regex: regex,
             caseSensitive: caseSensitive,
-            wholeWord: wholeWord, 
+            wholeWord: wholeWord,
             filePatterns: filePatterns,
             excludeFilePatterns: excludeFilePatterns,
             resultsCount: 0,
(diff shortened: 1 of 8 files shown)


src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/FindInFilesDialog.java
```diff
@@ -181,6 +187,26 @@ public FindInFilesDialog(OperationWithInput<State> operation)
       setExampleIdAndAriaProperties(spanPatternExample_, txtFilePattern_);
       setExampleIdAndAriaProperties(spanExcludePatternExample_, txtExcludeFilePattern_);
 
+      Roles.getCheckboxRole().setAriaLabelProperty(checkboxCaseSensitive_.getElement(), 
+         checkboxCaseSensitive_.getText());
+      Roles.getCheckboxRole().setAriaLabelProperty(checkboxWholeWord_.getElement(),
+         checkboxWholeWord_.getText());
+      Roles.getCheckboxRole().setAriaLabelProperty(checkboxRegex_.getElement(),
+         checkboxRegex_.getText());
```
gtritchie, line 195: These three `setAriaLabelProperty` calls are unneeded and should be removed. The CheckBox control already produces proper HTML, in terms of accessibility. ARIA should always be a last resort if semantically correct HTML isn't possible.

More details, the underlying HTML produced by this control is essentially:

```html
<input type="checkbox" id="some_unique_id"/>
<label for="some_unique_id">Visible Label</label>
```

Another way to tell is that if you click on the label text, the checkbox itself will toggle. This is entirely due to the association established by `for` in the label element, and is the preferred native HTML way to do this.

gtritchie (changes requested):

src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/FindInFilesDialog.java
melissa-barca: Thanks for the explanation about ARIA, that wasn't clear to me. I've updated the code to set ElementId for the elements.

gtritchie (approved):
LGTM!

Issue #7295, opened 2020-07-07
Tabbing between "find" and "replace" boxes


burchill:
In RStudio, pressing Ctrl+F when editing a file will open up a pane for finding and replacing, and put the text cursor in the "find" textbox.  In previous versions of RStudio, you used to be able to press the Tab key to move directly from the find textbox to the "replace" textbox, which was incredibly helpful for a quick find-and-replace.  Now, pressing the Tab button moves the selection to the "Next" button, then the "Prev" button, then the "All" button, and _then_ the replace textbox.

It would be great if this could be brought back, or you could set up a key binding to select the "replace" textbox.

gtritchie:
Adding a screenshot for context:

<img width="528" alt="screenshot of find replace bar" src="https://user-images.githubusercontent.com/10569626/86848449-e46e8e00-c062-11ea-9631-2e31af79c579.png">

Another fix would be to rearrange these controls so replace box is right after find box; it is an accessibility requirement that focus goes in the same order the controls are visually arranged.

burchill:
Yes, I thought that might be the case, changing the order would be good as well

config-i1:
I agree this is very annoying. Pressing "Tab" four times instead of one breaks the work flow for me. Previously, this worked perfectly, now it's a pain.

config-i1:
Any update on this issue? v1.4.1103 still has this.
Just a though, allowing users to change the order would resolve this... :)

gtritchie:
Quick and safe change, will include in 1.4-juliet-rose.

valerie-rstudio:
### Verified Fixed

    RStudio Edition : Desktop
    RStudio Version : 1.4.1540
    OS Version      : mac OS 11.1
    R Version       : 4.0.3

Verified that by default the Find Panel controls are able to be tabbed through in the order they appear on screen. When the Command Palette option "Tab key moves focus directly from find to replace text in find panel" is checked and the focus is on the Find textbox, focus moves over to Replace box when tab is pressed. Once Replace is in focus, the controls listed after Replace on the Find Panel can be tabbed through in the order they appear on the screen.

PR #8869, opened 2021-02-01
option to enable legacy tab key behavior in editor's Find panel

gtritchie:
### Intent

Requested in #7295

screenshot of the find panel

In RStudio 1.3 (and newer) I made most UI controls accessible via keyboard tab sequence, and accessibility rules are that the tab key should move through controls in the same order they are represented on screen.

Some users prefer the old behavior in the Find panel, where hitting tab when focus is on the "Find" text moves directly to the "Replace" text, skipping over the "Next/Prev/All" buttons. Currently you must tab through those controls (so 4 presses of tab key instead of one).

### Approach

Added a user preference to reenable the old behavior and conditionally set the exact same behavior found in RStudio 1.2 and earlier. Preference is shown as "Tab key moves focus directly from find text to replace text in find panel" in the command palette. I did not add it to the Global Options dialog. If the find panel is displayed and the option is changed, must close and reopen the find panel for the option to take effect.

screenshot of command palette showing new option

Note the code that sets custom tabindexes is taken verbatim from RStudio 1.2 sources.

### Automated Tests

Not necessary; this is a fringe case that must be enabled by the user and is known to break accessibility validation.

### QA Notes

Mainly, verify that tab key behavior is no different by default in the Find panel between 1.4_wax_begonia and 1.4_juliet-rose.

If you turn on the option, verify that the tab key behavior is the same as RStudio 1.2 and earlier, mainly that when focus is in "find", hitting tab once moves focus to "replace".

### Checklist

### DIFF
--- a/src/cpp/session/resources/schema/user-prefs-schema.json
+++ b/src/cpp/session/resources/schema/user-prefs-schema.json
@@ -1213,6 +1213,12 @@
             "title": "Tab key always moves focus",
             "description": "Tab key moves focus out of text editing controls instead of inserting tabs."
         },
+        "find_panel_legacy_tab_sequence": {
+            "type": "boolean",
+            "default": false,
+            "title": "Tab key behavior in find panel matches RStudio 1.2 and earlier",
+            "description": "In source editor tab panel, tab key moves focus directly from find text to replace text."
+        },
         "show_focus_rectangles": {
             "type": "boolean",
             "default": true,
--- a/src/cpp/session/resources/schema/user-prefs-schema.json
+++ b/src/cpp/session/resources/schema/user-prefs-schema.json
@@ -1216,8 +1216,8 @@
         "find_panel_legacy_tab_sequence": {
             "type": "boolean",
             "default": false,
-            "title": "Tab key behavior in find panel matches RStudio 1.2 and earlier",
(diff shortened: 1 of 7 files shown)


src/cpp/session/resources/schema/user-prefs-schema.json
```diff
@@ -1213,6 +1213,12 @@
             "title": "Tab key always moves focus",
             "description": "Tab key moves focus out of text editing controls instead of inserting tabs."
         },
+        "find_panel_legacy_tab_sequence": {
+            "type": "boolean",
+            "default": false,
+            "title": "Tab key behavior in find panel matches RStudio 1.2 and earlier",
```
kevinushey, line 1219: I feel like we should describe the behavior in terms of what it does, as opposed to the version of RStudio it's associated with. What do you think about having the title match the description in this regard? "Tab key moves focus directly from find to replace" or similar?

kevinushey: (mainly because this is the text that shows up in the command palette and so it's the text most users will see)

gtritchie: True, it's super-bizarre the way I've got it. I'll try something more like what you're suggesting.

kevinushey (approved):

PR #8875, opened 2021-02-02
fix bugs when changes applied to Pane Layout

melissa-barca:
### Intent

Address testing feedback from #8460.

### Approach

The (sometimes) invisible Presentation tab caused the Pane Manager to act as if a tab was being displayed when one was not. I updated the code to consider a TabSet with only the Presentation Tab present while Presentation is not active to be empty. With additional testing, I realized minimizing panes to hide them wasn't sufficient when both sets were empty and in the same column so I updated the code to adjust the splitter in this case. 

### Automated Tests

rstudio/rstudio-ide-automation#150

### Checklist


gtritchie (approved):
LGTM!

PR #8880, opened 2021-02-03
Better autoscrolling with Virtualized Console

adamconroy:
### Intent

There are two different kinds of ways to add text to the console: adding new elements and appending to existing elements. Before this PR we only scrolled to the bottom in the first situation. The intent is to always scroll to the bottom now

Addresses: https://github.com/rstudio/rstudio/issues/8504

### Approach

Always check the scroll status in the VirtualConsole.submit function and use a larger window for detecting whether we are near the bottom.

### Automated Tests

No automated tests added for this at the moment.

### QA Notes

Verify that we always properly keep the scrolling at the bottom with all permutations of "Limit Console Visible" on and off with these two lines:

`while (TRUE) { message("polluting the console"); Sys.sleep(0.5) }`

`while (TRUE) { print("polluting the console"); Sys.sleep(0.5) }`

Also that it stops scrolling if you scroll up a bit while the function is operating.

### Checklist

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/virtualscroller/virtualscroller.js
+++ b/src/gwt/src/org/rstudio/core/client/virtualscroller/virtualscroller.js
@@ -53,12 +53,13 @@ var VirtualScroller;
       self._isAtTopBucket = self._isAtTopBucket.bind(self);
       self._isBucketHidden = self._isBucketHidden.bind(self);
       self._isBucketFull = self._isBucketFull.bind(self);
-      self._jumpToBottom = self._jumpToBottom.bind(self);
       self._moveWindow = self._moveWindow.bind(self);
       self._onParentScroll = self._onParentScroll.bind(self);
       self._showBucket = self._showBucket.bind(self);
       self.append = self.append.bind(self);
       self.clear = self.clear.bind(self);
+      self.scrollToBottom = self.scrollToBottom.bind(self);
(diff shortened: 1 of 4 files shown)


src/gwt/src/org/rstudio/core/client/virtualscroller/virtualscroller.js
```diff
@@ -192,8 +193,14 @@ var VirtualScroller;
       return this.scrollerEle.scrollTop < 4;
     },
 
-    _scrolledToBottom: function() {
-      return Math.abs(this.scrollerEle.scrollHeight - this.scrollerEle.offsetHeight - this.scrollerEle.scrollTop) < 4;
+    scrolledToBottom: function() {
+      return Math.abs(this.scrollerEle.scrollHeight - this.scrollerEle.offsetHeight - this.scrollerEle.scrollTop) < 50;
```
kevinushey, line 197: Could `this.scrollerEle` ever be `null` here? (Mainly asking because `scrollToBottom` has a check for this)

kevinushey (approved):

src/gwt/src/org/rstudio/core/client/virtualscroller/virtualscroller.js
adamconroy: possible! added a null check, thanks!

PR #8897, opened 2021-02-04
fix issue with "close all others" context menu sometimes closing everything

gtritchie:
### Intent

- Addresses #8855
- Bug was triggered when last document (i.e. document that last had focus) was in different source column than the document whose tab context menu was used to invoke "close all others"

### Approach

- As code looped through the columns and editors, the check to see if a given document was the one that wanted to stay open only worked if that document was in the same source column as the last active document
- We already passed in the column containing the given tab, so use it for the test in addition to the "active column" (note, I'm not 100% sure we need to do both, but taking the lowest risk approach here)

### Automated Tests

- https://github.com/rstudio/rstudio-ide-automation/issues/178

### QA Notes

- Try various combinations of opening multiple documents in multiple source columns, then right-clicking on tab and using "Close All Others" command
- Try clicking on the active document, on an inactive document in the same column, and on an inactive document in a different source column (and whatever other combos come to mind)
- Be sure everything closes except the one you clicked on, and you are prompted if any of the files have unsaved modifications

### Checklist

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/source/SourceColumnManager.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/SourceColumnManager.java
@@ -1480,8 +1480,7 @@ public class SourceColumnManager implements CommandPaletteEntrySource,
       {
          final CPSEditingTargetCommand command = (EditingTarget target, Command continuation) ->
          {
-            if (!StringUtil.isNullOrEmpty(excludeDocId) &&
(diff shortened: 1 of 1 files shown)


melissa-barca (commented):
I don't think we should still check the active column here. We are combining active and a defined `docId` because of the changes made in rstudio/rstudio#8357 to replace relying on the `activeColumn` with relying on a passed `docId`. With that change we should've included this change from `activeColumn_` to `column`. In the interest of making the code easier to read I think we should remove the active column.

melissa-barca (approved):
LGTM!

Issue #8940, opened 2021-02-09
Show Traceback doesn't function when Limit visible console output on


kevinushey:
### System details

    RStudio Edition : Desktop [Open Source]
    RStudio Version : 1.4.1560
    OS Version      : macOS Big Sur 10.16
    R Version       : R version 4.0.3 (2020-10-10)

### Steps to reproduce the problem

Run a command that produces an error with a traceback; for example:

```
f <- function() stop()
g <- function() f()
h <- function() g()
k <- function() h()

k()
```

Then, try to click Show Traceback in the console output.

<img width="679" alt="Screen Shot 2021-02-09 at 1 23 04 PM" src="https://user-images.githubusercontent.com/1976582/107430398-f3765c80-6ad9-11eb-939c-d7b9bcfb10a4.png">

### Describe the problem in detail

No traceback is shown.

### Describe the behavior you expected

The traceback should be shown.

If I turn 'Limit visible console output' back to off, restart the IDE, and then re-run these steps, it appears to work.

adamconroy:
Very interesting. I fixed this back in October, at least it appeared that I had in my dev environment. I wonder what is happening. https://github.com/rstudio/rstudio/pull/7942/files

astayleraz:
Verified in Version 1.4.1621-2 that the traceback is working in the console with the limit enabled and disabled and added an automated test to check this. Closing issue.

Issue #7918, opened 2020-09-30
Virtual scroller enabled console truncates output but not associated prompt


gtritchie:
### System details

    RStudio Edition : Server (probably desktop, too, but I only tried server)
    RStudio Version : 1.4.878
    OS Version      : CentOS 7
    R Version       : 3.6.2

### Steps to reproduce the problem

This was noted during a PR https://github.com/rstudio/rstudio/pull/7568#discussion_r470294749 . Still happens so opening an issue.

- Turn on Virtual Scroller: Global Options / Console: "Limit console display to a subset of total content"
- In the console, do:

```
for (i in 1:2000) cat(c(i, "\n"))
```

### Describe the problem in detail
As expected, the output is truncated to about 1000 lines. However, the prompt for the command that generated the truncated output is still shown, making it difficult to realize content was dropped.

### Describe the behavior you expected
Same behavior as when Virtual Scroller is off; the prompt gets truncated with the rest of the output.

Note, I don't know how important this is in real-world use. It seems like it could be troublesome, though.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting a RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
I'm still seeing code cut off in unexpected ways with this feature on:

<img width="521" alt="Screen Shot 2020-10-15 at 3 38 24 PM" src="https://user-images.githubusercontent.com/1976582/96193191-8290b080-0efc-11eb-88cd-fdbce6a2304d.png">

It seems like, after a particular piece of console input, we're only retaining the last 1000 lines of output.

After reloading the UI (Right click -> Reload), I see after trying to scroll back to the top:

<img width="576" alt="Screen Shot 2020-10-15 at 3 42 14 PM" src="https://user-images.githubusercontent.com/1976582/96193474-16627c80-0efd-11eb-8444-b07b0057cd6c.png">

So it seems like the console history is then truncated in the "regular" way.

kevinushey:
The original issue still reproduces for me.

astayleraz:
[...]

PR #9027, opened 2021-02-25
VirtualScroller bugfix: truncation indication, ConsoleError interactivity

adamconroy:
### Intent

To address: https://github.com/rstudio/rstudio/issues/8940

I've had to rewrite the element manipulation to be compatible with raw Javascript. Unfortunately, when an element is added to the DOM through an external source, GWT loses the bindings are we aren't able to manipulate the elements through GWT variables anymore. I rewrote the functionality with plain DOM manipulation functions.

To address: https://github.com/rstudio/rstudio/issues/7918

I've added `[...]` to the top of the output to indicate that it has been truncated. I'm open to ideas for alternatives to put there but this is very similar to what Visual Studio does when outputting large collections. I don't think clearing the console cutting off everything (including the command prompt) like we do without the VirtualScroller is the right move as it kind of defeats the purpose of the feature entirely.

RexuPhv

### Automated Tests

There are no automated tests for this feature as there is already some coverage.

### QA Notes

Verify that ConsoleErrors with traceback function properly with VirtualScroller both on and off.

### Checklist

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/common/debugging/ui/ConsoleError.ui.xml
+++ b/src/gwt/src/org/rstudio/studio/client/common/debugging/ui/ConsoleError.ui.xml
@@ -74,8 +74,8 @@
                   ui:field="showTracebackImage"
                   width="9"
                   height="9"
-                  styleName="{style.errorCommandImage}"/>
-           <g:Anchor styleName="{style.errorCommandText}" ui:field="showTracebackText">
+                  styleName="{style.errorCommandImage} show_traceback_image"/>
+           <g:Anchor styleName="{style.errorCommandText} show_traceback_text" ui:field="showTracebackText">
               Show Traceback
            </g:Anchor>
         </g:HTMLPanel>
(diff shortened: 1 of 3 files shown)


src/gwt/src/org/rstudio/studio/client/common/debugging/ui/ConsoleError.ui.xml
```diff
@@ -74,8 +74,8 @@
                   ui:field="showTracebackImage"
                   width="9"
                   height="9"
-                  styleName="{style.errorCommandImage}"/>
-           <g:Anchor styleName="{style.errorCommandText}" ui:field="showTracebackText">
+                  styleName="{style.errorCommandImage} show_traceback_image"/>
```
kevinushey, line 77: Can you clarify why we need these extra class names here? Are we unable to query for the class name generated by GWT in the associated `style` object?

kevinushey (approved):
LGTM; thank you for slogging through this!

My main comment is whether the separate hard-coded class names are necessary, or if we could still query / use the style names generated by GWT for each UI element.

src/gwt/src/org/rstudio/studio/client/common/debugging/ui/ConsoleError.ui.xml
adamconroy: I just double checked and unfortunately GWT is assigning the same class to the bindings, both tex telements "Show Traceback" and "Rerun" have the class `org-rstudio-studio-client-common-debugging-ui-ConsoleError_ConsoleErrorUiBinderImpl_GenCss_style-errorCommandText`.

gtritchie (approved):
LGTM! Sorry for delay, I looked at it yesterday and forgot to actually approve it.

Issue #4109, opened 2019-01-06
Commenting not working for .c files


david-cortes:
**Steps to reproduce:**
* Open some `.c` file in RStudio
* Select some piece of text
* Press Ctrl + Shift + c

**Expected behavior:**
* Should comment it out with `//` just like in a `.cpp` file

**Actual behavior:**
* Nothing happens

ronblum:
@david-cortes Thanks for reporting this! We'll take a look at it as we continue development.

Confirmed with RStudio Desktop 1.2.1225-1. Marking as a bug, though it might really be an enhancement.

MichaelChirico:
I would also be fine with <kbd> Ctrl </kbd> <kbd> / </kbd> working since that's the command in Jupyter/Atom & makes more visual/intuitive sense in the case of actually adding `/` to comment

david-cortes:
Still reproducible in the latest daily build.

ronblum:
This is working in
- RStudio Server 1.4.1620 on Red Hat 8.3 via Firefox on MacOS 11.3
- RStudio Server 1.4.1620 on SLES 15 SP2 via Safari on MacOS 11.3
- RStudio Desktop 1.4.1620 on MacOS 11.3

PR #9051, opened 2021-03-03
Allow comment toggling and comment reflow for C files

cm421:
### Intent

Allow the comment toggling shortcut (Ctrl + Shift + c) on `.c` files, just like it exists for `.h`, `.hpp`, and `.cpp` files. See issue #4109

### Approach

C and CPP files are handled by the same java CppFiletype class. The option enabling comment toggling was only enabled for CPP files. Simply moving it out of a check for CPP-type (specifically `.h`, `.hpp`, `.cc`, and `.cpp` extensions) files allows comment toggling for both C and CPP files.

 I also noticed an option for enabling comment reflow, and included that for C files as well, since functionally it would be the same for C and CPP files

### Automated Tests

N/A

### QA Notes

N/A

### Checklist


MariaSemple (approved):
This looks great, thanks!

Issue #9082, opened 2021-03-10
Allow GWT to detect desktop mode for both Qt and Electron


gtritchie:
During development we want GWT to detect both the existing QtWebEngine-based desktop app and the future Electron-based app.

gtritchie:
Implemented via https://github.com/rstudio/rstudio/pull/9085

astayleraz:
Verified in Version 1.5.104 I am still seeing the desktop variation of the UI. Closing issue.

PR #9085, opened 2021-03-10
make GWT detect both Electron and QtWebEngine desktop user agents

gtritchie:
### Intent

First non-prototype Electron commit!

During development of the Electron-based desktop, want the GWT code (i.e. the user interface) to detect and function with both the QtWebEngine-based desktop, and the Electron desktop.

Addresses #9082

### Approach

Detect desktop mode both via the QtWebEngine user agent and the Electron user agent.

### Automated Tests

N/A

### QA Notes

Test that the existing desktop app (Qt) still loads the desktop variation of the UI. Most obvious symptom is whether the main menu is displayed as part of the HTML (server) or the native menus are populated (desktop).

### Checklist

### DIFF
--- a/src/gwt/src/org/rstudio/studio/RStudio.gwt.xml
+++ b/src/gwt/src/org/rstudio/studio/RStudio.gwt.xml
@@ -4,9 +4,9 @@
 
    <!-- GWT module dependencies                                    -->
    <inherits name='com.google.gwt.user.User'/>
-   
+
    <inherits name='com.google.gwt.widgetideas.SliderBar'/>
-    
+
    <inherits name='com.sksamuel.gwt.GwtWebsockets'/>
 
    <!-- Gin module dependencies                                    -->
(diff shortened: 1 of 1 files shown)


src/gwt/src/org/rstudio/studio/RStudio.gwt.xml
```diff
@@ -26,8 +26,9 @@
       if (window.desktop)
          return "true";
 
-      // also true when run under QtWebEngine, which defers hook injection
-      return window.navigator.userAgent.indexOf("QtWebEngine") > 0 ? "true" : "false";
+      // also true when run under Electron or QtWebEngine, which defer hook injection
+      return ((window.navigator.userAgent.indexOf("Electron") > 0) ||
+              (window.navigator.userAgent.indexOf("QtWebEngine") > 0)) ? "true" : "false";
```
adamconroy, line 31: I believe we should change this to >= 0 just to be consistent with the return values of indexOf, 0 implies that the string starts with that value.

adamconroy (changes requested):

src/gwt/src/org/rstudio/studio/RStudio.gwt.xml
adamconroy: (if this is actually Javascript)

gtritchie: Per our discussion we'll just go with the existing pattern here.

Issue #4059, opened 2018-12-18
Input/output error when using textbox in Open file menu


dfalty:
### System details

    RStudio Edition : Server
    RStudio Version : 1.1.456 & 1.2.1194-1
    OS Version      : RHEL 7.3

### Steps to reproduce the problem
1. Select File -> Open File
2. Try typing out the file to open (same results using full path as relative)
3. Select Open
4. Observe empty source pane open and error popup

### Describe the problem in detail
![image](https://user-images.githubusercontent.com/15237845/50188663-f30e1c80-02e7-11e9-9183-bbf30f51971e.png)

GWarsow:
Is there some estimate, when this bug will be fixed?

fthommen:
This problems is still there in the newest RStudio version and it has been confirmed by RStudio as a bug in Decemebr 2018.  Can we expect it to be solved soon?

R-Hannibal:
Verified Fixed on: RStudio Version - 1.4.1652-1, R Version - 4.0.4, RHEL8 - Server 

Files are now successfully opening via File > Open File... > typing in partial or full file name and hitting Enter to open the file.

Comparison tested with `1.4.1647-1` where the issue was still reproducible. Tested on RDP as well and was not able to reproduce the issue there.

fthommen:
When will version `1.4.1652-1` be available for download?  We currently only find `1.4.1106-5` in the download area.

R-Hannibal:
@fthommen - You can currently access the 1.4.1652-1 build (that's the pro build, but open source builds are also available) on our [daily builds](https://dailies.rstudio.com/) page. 

Warning, these are early builds of our next release so they may be slightly less stable than `1.4.1106-5`.

fthommen:
@R-Hannibal We have a pro license.  I'll check the daily builds page.  thanks a lot for the speedy answer.

PR #9153, opened 2021-03-25
fix issue with opening file with explicitly typed name

kevinushey:
### Intent

When a user attempts to open a file by explicitly typing the name of that file (rather than selecting an entry from the file-list widget), the attempt to open that file fails. This PR intends to fix that.

Addresses https://github.com/rstudio/rstudio/issues/4059.

### Approach

The underlying issue here is that there are two notions of the "active" file:

- The current selection, if any, in the Files list widget;
- What the user has typed in the filename at the top of the widget.

Because selecting an item always updates the filename at the top of the widget, that should be taken as the correct filename to open.

### Automated Tests

https://github.com/rstudio/rstudio-ide-automation/issues/279

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/4059. The change should only effect RStudio Server, but it's worth verifying in RStudio Desktop as well.

### Checklist


jmcphers (approved):
LGTM!

Issue #9116, opened 2021-03-16
forward PATH environment variable when running rmarkdown render


kevinushey:
From JJ:

> When I set the Python Environment for a project, I see that reticulate uses it by default and that it shows up in the Terminal. However, when I execute a Quarto render command it's not on the PATH (Quarto render happens in the normal course of R Markdown render, it's essentially a custom knit handler that we automatically use when we see the "format" key and Quarto is installed).
> 
> Right now my workaround is to add this to .Rprofile:
> 
> Sys.setenv(PATH=paste(path.expand("~/opt/miniconda3/envs/quarto-cli/bin"), Sys.getenv("PATH"), sep=":"))
> 
> However I don't think most users will think/know to do this. What would you think about adding the configured Python environment to the PATH whenever we execute an Async R Process? 

I think this change would happen around here:

https://github.com/rstudio/rstudio/blob/b0f09584d8f231604c5ecf3ba2d114e64b3e3f0e/src/cpp/session/modules/rmarkdown/SessionRMarkdown.cpp#L553-L599

ronblum:
@kevinushey Does this apply with R Markdown in general or is it specific to a custom renderer/using Quarto?

kevinushey:
Quarto in this context is just a special kind of R Markdown output type; it motivates this change but I think in general we do want to ensure that the requested version of Python (or the associated environment) is on the PATH for renders.

ronblum:
Tested in
- RStudio Desktop 1.4.1660 on MacOS 11.3
- RStudio Server 1.4.1660 on SLES 15 SP2 using Chrome from MacOS 11.3

by setting both global and project options. With no project, the path of the global Python interpreter is added to `PATH`, and in a project, the project-level path is added. 

However, rendering the R Markdown file is changing the `PATH` for the R session, not just scoped in the rendering. So in the Console:

```
> # Before Knitting
> Sys.getenv("PATH")
[1] "/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin:/Applications/VMware Fusion.app/Contents/Public:/Library/TeX/texbin:/Applications/RStudio-pro-1.4.1660-1.app/Contents/MacOS/postback"
> 
> # After Knitting
> Sys.getenv("PATH")
[...]

PR #9169, opened 2021-03-29
ensure configured copy of Python is placed on PATH for render

kevinushey:
### Intent

Users might configure RStudio to use a particular version of Python, via Global Options or Project Options. While the selected version of Python is visible to reticulate via `RETICULATE_PYTHON`, it's also useful for the requested copy of Python to be available on the PATH as well.

Addresses https://github.com/rstudio/rstudio/issues/9116.

### Approach

Relatively straightforward munging of PATH in our render code.

### Automated Tests

None included.

### QA Notes

Configure RStudio to use a particular copy of Python, then create an R Markdown document and check the output of `Sys.getenv("PATH")` from an R chunk. You should see something like:

Screen Shot 2021-03-29 at 4 35 52 PM

where the first PATH entry corresponds to the directory name of the configured version of Python.

### Checklist


gtritchie (approved):
LGTM!

PR #9172, opened 2021-03-30
A simple mocha chai setup against our electron-forge skeleton

adamconroy:
Usage:

1. Write tests in in src/node/desktop/test
2. Run `npm run test`

Currently bootstrapping the Electron framework through electron-forge isn't seamless with mocha chai. However, this basic setup seems to be the standard in other Electron projects (they just have have more tooling built up to set up the environment through bash scripts etc).


gtritchie (approved):
LGTM, will need some tweaking when I get to removing electron-forge, and we'll be using yarn instead of npm, but good to get going!

Issue #9190, opened 2021-04-05
Unit tests failing on official Mac build machine


gtritchie:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.4 (master as of 2021/04/05)
    OS Version      : macOS build machine
    R Version       : any

### Steps to reproduce the problem
Look at Mac build results from official builder. Marked as unstable.

### Describe the problem in detail
It looks like the tests are failing due to being unable to fine libpq.5.dylib

### Describe the behavior you expected
No failures for unit tests.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
Unit tests are now running _and_ passing on CI. Joy!

https://build.rstudioservices.com/blue/organizations/jenkins/IDE%2Fmacos-v1.4/detail/macos-v1.4/1215/pipeline/44

PR #9203, opened 2021-04-07
use install_name_tool to fix paths at install time

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/9190.
Addresses https://github.com/rstudio/rstudio/issues/6890.

Previously, we were running `install_name_tool` for package builds at build time, rather than install time. This causes problems when we try to run unit tests with built-but-not-yet-installed RStudio components, as the processes will end up looking for libraries in the wrong locations.

This PR removes those build-time `install_name_tool` usages, and instead just fixes the library paths at install time for the requisite components.

### Approach

Relatively straight-forward switch from using post-build CMake commands to install-time CMake commands (with existing tools).

### Automated Tests

These are the automated tests!

### QA Notes

Nothing for QA.

### Checklist

### DIFF
--- a/package/osx/cmake/fix-library-paths.cmake
+++ b/package/osx/cmake/fix-library-paths.cmake
@@ -42,7 +42,15 @@ if(EXISTS "@RSESSION_ARM64_PATH@")
 		COMMAND
 		"@CMAKE_CURRENT_SOURCE_DIR@/fix-library-paths.sh"
 		"${CMAKE_INSTALL_PREFIX}/RStudio.app/Contents/Frameworks/arm64"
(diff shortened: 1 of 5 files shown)


gtritchie (approved):
This seems OK, insomuch as I understand it. Did you run a full Mac package build locally and confirm the resulting dmg contained a working RStudio.app?

kevinushey:
> This seems OK, insomuch as I understand it. Did you run a full Mac package build locally and confirm the resulting dmg contained a working RStudio.app?

I did (and fortunately everything was happy). Also confirmed that unit tests can run in the package build folder (which is what Jenkins does).

Issue #9188, opened 2021-04-05
terminal should place configured version of Python on the PATH


kevinushey:
Chatting with @jjallaire: If the user has configured a global / project version of Python for RStudio, that version of Python should be placed on the PATH.

kevinushey:
It looks like getting this right is a bit more challenging than it looks like at first glance. When launched as a login shell, bash (and presumedly, other shells) take the inherited path, but push other relevant path entries in front as appropriate. For example:

```
$ PATH=__INHERITED_PATH__ /bin/bash --login --norc -c 'echo $PATH | tr ":" "\n"'
/Users/kevinushey/.yarn/bin
/Users/kevinushey/.config/yarn/global/node_modules/.bin
/usr/local/bin
/Library/TeX/texbin
/usr/local/bin
/usr/local/bin
/usr/bin
/bin
/usr/sbin
/sbin
/Library/TeX/texbin
/usr/local/share/dotnet
/opt/X11/bin
~/.dotnet/tools
/Library/Apple/usr/bin
__INHERITED_PATH__
```

Note that `__INHERITED_PATH__` is stuck at the bottom. Ultimately, this implies that we need to find a way to run our own custom code (modifying the PATH) after the shell has processed any login-related startup scripts.

ronblum:
This works as expected in RStudio Server 1.4.1692. However, in RStudio Desktop on MacOS 11.3, when opening a new terminal, the following also happens:

1. Set the Python interpreter (global or project—I don't think it matters).
2. `Terminal`
3. `echo $PATH`
    - Python's parent directory is at the start of the path.
4. Change Python interpreter.
5. In the same terminal as before, `echo $PATH`
    - The new interpreter's parent directory is at the start of the path.
    - A different interpreter's parent directory is second to last in the path, right before RStudio's .../postback directory.
6. Open a new terminal
7. `echo $PATH`
    - The path is now correct.

mikebessuille:
Moving this ticket back to "verifying" - I think it's now been verified?

elbamos:
Hope this is helpful and not counterproductive -- if the user's chosen python is a conda or virtualenv environment, shouldn't the correct behavior be to call the appropriate `activate` script rather than adjusting the `PATH` directly?  My concern is that the path-editing is going to interfere with whatever path editing is taking place in the user's startup script; or some greater confusion will occur if the user tries to change their active python environment from the terminal.

kevinushey:
[...]

Issue #6855, opened 2020-05-12
Use wsl.exe to launch WSL shell instead of bash.exe


gtritchie:
On Windows-10 builds starting with Spring 2018 update, the RStudio Terminal should detect and launch WSL via `wsl.exe` instead of `bash.exe`. Use of `bash.exe`  (which isn't actually bash but the WSL launcher) is considered deprecated in these newer versions of WIndows.

Deprecation of bash mentioned here: https://docs.microsoft.com/en-us/windows/wsl/interop

gtritchie:
Bump

Danny-dK:
Is this still being worked on? I get:

[bsh_rstudio.webm](https://github.com/rstudio/rstudio/assets/45395070/b7fb6a18-c97b-477c-bb6b-c11f238bfc4f)

Opening a terminal that is set to 'Bash (Windows Subsystem for Linux)' seems to not do anything. For less then a second I can see it is trying to load, but then I can't do anything in this screen. If I set the terminal in global options to custom shell binary and point to C:/Windows/System32/wsl.exe everything works fine.

Danny-dK:
@kevinushey The thing is that if I go to `System32` and start bash.exe, that still works. A shell window is opening that starts my Ubuntu 20.04LTS distro under WSL2. Just not starting in Rstudio terminal.

Danny-dK:
Hmm, if I set the custom terminal location to wsl.exe everything works, if I set it to bash.exe it does not work. Perhaps point rstudio to wsl instead of bash.exe when `win-wsl-bash` is used as the option? (I don't know if it is these lines https://github.com/rstudio/rstudio/blob/8244ebc49cc3188f04aa26bb2c2377aeda315817/src/cpp/session/modules/SessionTerminalShell.cpp#L111C3-L123C55)

gtritchie:
> ...Perhaps point rstudio to wsl instead of bash.exe when `win-wsl-bash` is used as the option?

Yes, that is what this issue was opened for. I'll look at finally making that change (better late than never?).

Danny-dK:
> Yes, that is what this issue was opened for.

Ah sorry, then I misunderstood. Hope that the issue since 2020 will be resolved soon.

gtritchie:
> Ah sorry, then I misunderstood. Hope that the issue since 2020 will be resolved soon.

No worries, sorry this issue hasn't been fixed already! I'll be working on it soon.

gtritchie:
[...]

PR #9205, opened 2021-04-07
place python on the PATH for new terminal sessions

kevinushey:
### Intent

Users can configure the version of Python to be used by RStudio and reticulate in a couple ways:

- By explicitly setting the `RETICULATE_PYTHON` environment variable;
- Via the UI, with global / project-specific options.

However, none of these configured versions of Python would propagate as the "default" to the terminal. This PR tries to make that happen.

Addresses https://github.com/rstudio/rstudio/issues/9188.

### Approach

Making this work was a bit tricky, because ultimately we want to change the terminal process's `PATH` _after_ it has finished initializing (e.g. running profile startup scripts). The solution I landed on was to run `bash` as an interactive non-login shell, but then do all the things that a login shell would normally do -- that is, sourcing of `/etc/profile` and `~/.bash_profile`.

Note that the PR, in its current state, only works for the default Bash shell. Will need to confer with @gtritchie if this approach would be workable for other shells as well.

### Automated Tests

None yet.

### QA Notes

To test, try the following:

1. Using the RStudio Global Options dialog, tell RStudio to use a particular copy of Python.
2. Launch a new terminal in RStudio.
3. Run `which python`. Check that it matches the RStudio-configured version of Python.
4. Run `echo $RETICULATE_PYTHON`. Check that it matches the RStudio-configured version of Python.

We should also test that this behavior can be opted-out / disable via the associated option in Tools -> Global Options... -> Terminal.

@gtritchie also note:

> This feature must be tested on both Mac and Windows desktops, and with all available shell types (on Windows they’ll need to install a WSL flavor from Store so it’s available to test). Testing on Linux would be nice, too, but suspect Mac provides sufficient coverage

So this does require a bit of smoke-testing on all of the different flavors of bash we use on different operating systems.

### Checklist


src/cpp/session/SessionConsoleProcess.cpp
```diff
@@ -251,6 +253,67 @@ void ConsoleProcess::commonInit()
...
+                  &*options_.environment,
+                  "RETICULATE_PYTHON",
+                  reticulatePython);
+      }
+   }
+   
+   // if we're using bash, then we'll use a custom rcfile to ensure our
+   // hooks get run when the shell is launcehd
+   if (getShellType() == TerminalShell::ShellType::PosixBash ||
+       getShellType() == TerminalShell::ShellType::GitBash)
```
gtritchie, line 280: Should this also be made to work in WSLBash, too? Or is that a problem because WSL is really a separate Linux environment so the paths (and Python) it has access to can be quite different. Either way we might need some sort of warning about that (maybe a new section in the terminal support article documenting this feature's behavior and any limitations)?

src/cpp/session/SessionConsoleProcess.cpp
```diff
@@ -251,6 +253,67 @@ void ConsoleProcess::commonInit()
                            module_context::activeSession().id());
 #endif
    }
+   
+   // if RStudio has been configured to use Python,
+   // place that copy of Python on the PATH
```
gtritchie, line 258: Also, given the potential complexity of people's terminal shell configs, some might not want this happening at all, so I'd strongly recommend adding a terminal option to disable it. Probably in the Terminal / General / Shell section, and could be keyed to the selected shell (if they select a shell that doesn't support this feature maybe replace the checkbox with a message about Python settings not being automatically set).

src/cpp/session/SessionConsoleProcess.cpp
```diff
@@ -251,6 +253,67 @@ void ConsoleProcess::commonInit()
...
+   {
+      FilePath zDotDir =
+            session::options().rResourcesPath().completeChildPath("terminal/zsh");
+      
+      core::system::setenv(
+               &*options_.environment,
+               "ZDOTDIR",
+               zDotDir.getAbsolutePath());
+   }
+
```
gtritchie, line 308: Also, can we document a way to make this work with a custom shell config? People sometimes use Custom so they can tweak how we launch bash, zsh, etc.

gtritchie (changes requested):
A few initial comments / questions. Overall this might be fine, but I am worried about it going into a release at this late stage and users finding a bunch of corner cases we aren't aware of. With a master switch to enable/disable I'm less worried.

.gitignore
```diff
@@ -46,3 +46,4 @@ GSYMS
 GTAGS
 TODO-INTERNAL
 *.pid
+.zsh_history
```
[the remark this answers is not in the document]
kevinushey: I think this is caused by setting `ZDOTDIR` -- it effects where `.zsh_history` is written.

Which I think also makes it clear that if we're doing this dance with `ZDOTDIR` then we need to set `HISTFILE` as well to the expected default location...

src/cpp/session/SessionConsoleProcess.cpp
```diff
@@ -251,6 +253,67 @@ void ConsoleProcess::commonInit()
                            module_context::activeSession().id());
 #endif
    }
+   
+   // if RStudio has been configured to use Python,
```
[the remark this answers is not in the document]
kevinushey: My impression was we'd want this for all terminal processes -- e.g. if you do something like:

```
.rs.api.terminalExecute("python --version")
```

that should still set the PATH so the "right" version of Python is found. Does that make sense?

src/cpp/session/SessionConsoleProcess.cpp
kevinushey: I think we could make this work for WSL as well -- let me try to test that.

kevinushey:
@jmcphers @gtritchie: I think this should be ready for re-review.

src/cpp/session/SessionConsoleProcess.cpp
kevinushey: I've added an option to the Terminal preferences pane:

Screen Shot 2021-04-21 at 12 02 06 PM

Does that look reasonable?

kevinushey:
As a side note, just to explain the gymnastics for `zsh`...

When `zsh` is launched as a login shell, it sources a number of scripts in order:

```
$ zsh -l
/Users/kevinushey/.zshenv
/Users/kevinushey/.zprofile
/Users/kevinushey/.zshrc
/Users/kevinushey/.zlogin
kevinushey@cdrv-2 rstudio-v14 %
```

In each case, the user can customize `ZDOTDIR` to point to a different location if desired. This is done, for example, to allow users to place other dotfiles (other than `.zshenv`) in a separate location; e.g. `~/.local/config/zsh` or something similar, for people who feel strongly about the dotfiles that are placed in `$HOME`.

Because of that, in order to ensure that our hooks are properly sourced, we need to keep track of:

1. The directory containing our own `zsh` dotfiles, and
2. The "current" `zsh` ZDOTDIR, in case the user has decided to customize it.

The `_ZDOTDIR` environment variable is used to store the "real" (user-defined) `ZDOTDIR` environment variable, if it has been set.

src/cpp/session/SessionConsoleProcess.cpp
```diff
@@ -251,6 +253,142 @@ void ConsoleProcess::commonInit()
...
+               return true;
+            }
+               
+            case TerminalShell::ShellType::CustomShell:
+            {
+               std::string shellPath = options_.shellPath.getAbsolutePath();
+               
+#ifdef _WIN32
+               if (boost::algorithm::ends_with(shellPath, "bash.exe"))
+                  return true;
```
gtritchie, line 300: Using `bash.exe` to launch WSL is actually considered deprecated, so you should also detect `wsl.exe`.

gtritchie: See https://github.com/rstudio/rstudio/issues/6855 for details.

src/cpp/session/SessionConsoleProcess.cpp
```diff
@@ -251,6 +253,142 @@ void ConsoleProcess::commonInit()
...
+      if (!reticulatePython.empty())
+      {
+         auto isBashShell = [&]()
+         {
+            switch (getShellType())
+            {
+            
+            case TerminalShell::ShellType::PosixBash:
+            case TerminalShell::ShellType::GitBash:
+            case TerminalShell::ShellType::WSLBash:
```
gtritchie, line 289: I don't see how WSL can work. It's essentially a remote shell into a Linux environment, and the paths are different. You could get into the game of mapping c:\python to /mnt/c/python, I suppose.

The WSL environment may contains its own Linux-native python installation(s).

Perhaps trickiest, the "bash.exe" for launching wsl isn't actually bash (it's a generic launcher; see other comment below about wsl.exe). So it can very well be zsh or something else inside, you don't know (though you might be able to discover such things using various wsl.exe switches; but then you have to worry about which version of Windows-10 it is; some features of wsl aren't in some still-supported Windows-10 versions).

src/cpp/session/SessionConsoleProcess.cpp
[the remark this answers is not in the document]
gtritchie: Isn't this path also hit for things like running Git commands, and some other ConsoleProcess usages (which are declining over time as we move things into background jobs)? Might be fine (or even correct) to ensure Python matches in some of those cases, so probably the right thing. Just looking for opportunities to reduce the test matrix.

src/cpp/session/SessionConsoleProcess.cpp
gtritchie: Pending other discussion on WSL support (more details in later comments), might need to change the wording (a WSL user can have bash or wsl or some other shell inside the Linux environment, but I don't think this feature is going to work for any of those and the wording suggests it will).

Maybe have the checkbox text more closely match the shell options, so on Windows should say "(Git Bash only)?

Could perhaps hide the checkbox entirely when a shell is selected (Custom, PowerShell, Command Prompt, possibly WSL) that doesn't support this feature. In my mind it makes more sense for this checkbox to live under the Shell section rather than adding Miscellaneous, since it is only applicable to certain Shell settings (e.g. like the "Custom shell command-line options" control).

gtritchie (changes requested):

src/cpp/session/SessionConsoleProcess.cpp
gtritchie: I've added a TODO for myself to update the terminal support article in the Juliet-rose timeframe, and will include mention of this feature. I haven't updated the article for 1.4 at all yet: https://support.rstudio.com/hc/en-us/articles/115010737148-Using-the-RStudio-Terminal

src/cpp/session/SessionConsoleProcess.cpp
kevinushey: Hmm -- the challenge with this, as I understand it, is that `wsl.exe` will actually run whatever shell happens to be configured as the default; e.g. it's the gateway to (probably, but not necessarily) bash. Do you think it's safe for us to still assume it's bash here?

kevinushey: I see now you said exactly that below!

src/cpp/session/SessionConsoleProcess.cpp
kevinushey: If this is the case, then I'd feel more comfortable just leaving WSL out of this PR.

src/cpp/session/SessionConsoleProcess.cpp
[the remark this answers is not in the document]
kevinushey: You are right -- for this PR, I intended this to only run for terminal processes. I'll tweak that to make this a reality.

kevinushey:
Now with mildly tweaked UI:

Screen Shot 2021-04-26 at 3 48 22 PM

And the [ ] Enable Python Integration preference is only shown when using bash / zsh; e.g. we don't show it when using `fish`:

Screen Shot 2021-04-26 at 3 50 15 PM

gtritchie (approved):
LGTM! I kinda want the "Enable Python integration" checkbox to have a ? link with more details on what that means, but I'm comfortable putting that in the Terminal Support Article (which I'll update for juliet-rose).

Issue #9219, opened 2021-04-12
Issue with annotation database connection *only* in RStudio


compbiolevings:
### System details

    RStudio Edition :  Desktop
    RStudio Version :  1.4.1106
    OS Version      :  Kubuntu 20.04
    R Version       :  4.0.5

### Steps to reproduce the problem

Open RStudio Desktop, enter 'library(org.Hs.eg.db)' or any other 'org...db' (this happens for 'org.Hs.eg.db', 'org.Mm.eg.db' and 'org.Dm.eg.db' for me).  It's not an issue with R because when I run R in terminal or VSCode this issue never happens and everything works perfectly.

### Describe the problem in detail

So when I try to load an 'org.' annotation library/db, it loads all required packages, but when it gets to the 'org.' package, it gives an error message.  Here is example code with the error: 

```
> sessionInfo()
R version 4.0.5 (2021-03-31)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 20.04.2 LTS

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/blas/libblas.so.3.9.0
LAPACK: /usr/lib/x86_64-linux-gnu/lapack/liblapack.so.3.9.0

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8    LC_PAPER=en_US.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C             LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

loaded via a namespace (and not attached):
[1] compiler_4.0.5 tools_4.0.5   
> library(org.Hs.eg.db)
Loading required package: AnnotationDbi
Loading required package: stats4
Loading required package: BiocGenerics
Loading required package: parallel

Attaching package: ‘BiocGenerics’

The following objects are masked from ‘package:parallel’:

    clusterApply, clusterApplyLB, clusterCall, clusterEvalQ, clusterExport, clusterMap, parApply,
    parCapply, parLapply, parLapplyLB, parRapply, parSapply, parSapplyLB

The following objects are masked from ‘package:stats’:

[...]

PR #9229, opened 2021-04-13
catch errors when invoking connectionsObserver methods

kevinushey:
### Intent

Packages can update the Connections pane via the [Connections Contract](https://rstudio.github.io/rstudio-extensions/connections-contract.html). However, it's possible that R errors can occur if the associated R package violates the Connections contract in some way. These errors should not be fatal, since they only imply that the Connections pane will fail to update -- the underlying connection will likely still work without issue.

Addresses https://github.com/rstudio/rstudio/issues/9219.

### Approach

Wrap any connectionObserver methods with `tryCatch()`, and report errors to the user in a minimally-actionable way. At least, they can learn the offending package, and suppress these errors later.

Screen Shot 2021-04-13 at 12 28 33 PM

### Automated Tests

None included.

### QA Notes

Install the latest release of RSQLite (2.2.6), and then try running the following in RStudio:

```
BiocManager::install("org.Hs.eg.db")
DBI::dbConnect(
   RSQLite::SQLite(),
   dbname = system.file("extdata/org.Hs.eg.sqlite", package = "org.Hs.eg.db"),
   cache_size = 64000L, 
   synchronous = "off",
   flags = RSQLite:::SQLITE_RO,
   vfs = "unix-none"
)
```

You should see an error like the following:

Screen Shot 2021-04-13 at 12 29 46 PM

### Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -127,3 +127,4 @@
 * Fixed issue causing Project Sharing to fail to set Access Control Lists when using NFS v4 and `username@domain` security principals (Pro #2415)
 * Fixed issue where dialog boxes [e.g. Git commit, Installing Pro Database drivers] could fail to show output with Limit Console Output turned on (Pro #2284)
 * Fixed issue preventing Kubernetes sessions from starting due to incorrect SSL certificate checking on websocket connections; make websocket connections support the `verify-ssl-certs` option (Pro #2463)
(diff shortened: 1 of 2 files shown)


jmcphers (approved):
This LGTM and I do think we should take it for Juliet Rose due to the impact.

PR #9231, opened 2021-04-13
Add option to upload an unzipped archive if unzip is not on PATH

cm421:
### Intent

When uploading a zip archive, if unzip is not located on a user's PATH, offer to upload the zip archive without extracting instead of crashing or causing errors.

Addresses #9151 

### Approach

This adds a couple features:

1. If unzip is not on PATH, try `/usr/bin/unzip` as well. Adding other paths to check is easy
2. If unzip cannot be found, display a message asking if uploading the archive, without extracting, is okay
Upload without unzipping

3. Play nicely with the overwrite message, which will display normally if uploading the archive will overwrite an already existing archive in the same directory
Upload zip confirming override

### Automated Tests

I think some automated tests that do something like this would be more than enough:

- Sanity Check: 
  - Uploading a zip archive when `unzip` is on PATH should upload and extract the zip normally
- Backup Path Check: 
  - Uploading a zip archive when `unzip` is not on PATH, but is in `/usr/bin/unzip`, should upload and extract the zip normally
- Upload w/o Extracting Check: 
  - Uploading a zip archive when `unzip` is not on PATH or in `/usr/bin/unzip` should display a prompt asking the user if uploading without extracting is okay
    - Yes -> archive gets uploaded, but not extracted
    - No -> archive does not get uploaded
- Overwrite Compatibility Check: 
  - If uploading a zip archive will overwrite an already existing zip archive, the user should be warned, as usual

### QA Notes

### Checklist

### DIFF
--- a/src/cpp/session/modules/SessionFiles.cpp
+++ b/src/cpp/session/modules/SessionFiles.cpp
@@ -544,6 +544,8 @@ void handleFilesRequest(const http::Request& request,
 const char * const kUploadFilename = "filename";
 const char * const kUploadedTempFile = "uploadedTempFile";
 const char * const kUploadTargetDirectory = "targetDirectory";
+const char * const kIsZip = "isZip";
+const char * const kUnzipFound = "unzipFound";
 
 Error writeTmpData(const FilePath& tmpFile,
                    const char* buffer,
@@ -583,10 +585,12 @@ Error completeUpload(const core::json::JsonRpcRequest& request,
    
    // parse fields out of token object
    std::string filename, uploadedTempFile, targetDirectory;
+   bool unzipFound;
    error = json::readObject(token, 
                             kUploadFilename, filename,
                             kUploadedTempFile, uploadedTempFile,
-                            kUploadTargetDirectory, targetDirectory);
+                            kUploadTargetDirectory, targetDirectory,
+                            kUnzipFound, unzipFound);
    if (error)
       return error;
    
@@ -598,7 +602,7 @@ Error completeUpload(const core::json::JsonRpcRequest& request,
    {
       FilePath targetDirectoryPath(targetDirectory);
 
-      if (boost::ends_with(filename, "zip"))
+      if (boost::ends_with(filename, "zip") && unzipFound)
       {
          // expand the archive
          r::exec::RFunction unzip("unzip");
@@ -660,41 +664,68 @@ Error completeUpload(const core::json::JsonRpcRequest& request,
    
 Error detectZipFileOverwrites(const FilePath& uploadedZipFile,
                               const FilePath& destDir,
-                              json::Array* pOverwritesJson)
+                              const std::string& originalFilename,
+                              json::Array* pOverwritesJson,
+                              bool& rUnzipfound)
 {
-   // query for all of the paths in the zip file
-   core::system::ProcessResult result;
-   Error error = core::system::runCommand("unzip -Z1 " + uploadedZipFile.getAbsolutePath(),
-                                          core::system::ProcessOptions(),
-                                          &result);
-   if (error)
-      return error;
+   // unable to use R's unzip here in worker thread, using system's unzip instead
+   // try a couple locations for unzip, in case it's not on user's PATH
+   std::vector<std::string> pathsToTry({
+                                       "unzip",
+                                       "/usr/bin/unzip"});
+   for (auto &path : pathsToTry) {
+      core::system::ProcessResult result;
+      Error error = core::system::runCommand(path + " -Z1 " + uploadedZipFile.getAbsolutePath(),
+                                             core::system::ProcessOptions(),
+                                             &result);
 
-   if (result.exitStatus == 0)
-   {
-      std::vector<std::string> zipFileListing;
-      boost::split(zipFileListing, result.stdOut, boost::is_any_of("\n"));
+      if (error)
+         return error;
 
-      // check for overwrites
-      for (std::vector<std::string>::const_iterator
-           it = zipFileListing.begin();
-           it != zipFileListing.end();
-           ++it)
+      // If any of the paths were valid, check for overwrites and return
+      if (result.exitStatus == 0)
       {
-         // don't count empty lines
-         if ((*it).empty()) continue;
+         rUnzipfound = true;
 
-         FilePath filePath = destDir.completePath(*it);
-         if (filePath.exists())
-            pOverwritesJson->push_back(module_context::createFileSystemItem(filePath));
+         std::vector<std::string> zipFileListing;
+         boost::split(zipFileListing, result.stdOut, boost::is_any_of("\n"));
+
+         // check for overwrites
+         for (std::vector<std::string>::const_iterator
+              it = zipFileListing.begin();
+              it != zipFileListing.end();
+              ++it)
+         {
+            // don't count empty lines
+            if ((*it).empty()) continue;
+
+            FilePath filePath = destDir.completePath(*it);
+            if (filePath.exists())
+               pOverwritesJson->push_back(module_context::createFileSystemItem(filePath));
+         }
+
+         return Success();
+      }
+      else if (result.exitStatus == 127)
+      {
+         // bash return code 127 - "command not found"
+         // try another unzip location
+         continue;
+      }
+      else
+      {
+         return systemError(result.exitStatus,
+                            "Unexpected result for unzip command",
+                            ERROR_LOCATION);
       }
    }
-   else
-   {
-      return systemError(result.exitStatus,
-                         "Unexpected result for unzip command",
-                         ERROR_LOCATION);
-   }
+
+   // If we get here, there were no serious errors, but unzip was not found
+   // Try uploading just the .zip file, without unzipping
+   rUnzipfound = false;
+   FilePath zipPath = destDir.completePath(originalFilename);
+   if (zipPath.exists())
+      pOverwritesJson->push_back(module_context::createFileSystemItem(zipPath));
    
    return Success();
 }
@@ -1050,9 +1081,10 @@ bool handleFileUploadRequestAsync(const http::Request& request,
    FilePath destPath = destDir.completeChildPath(pUploadState->fileName);
 
    json::Array overwritesJson;
+   bool unzipFound = false;
    if (isZip)
    {
-      Error error = detectZipFileOverwrites(pUploadState->tmpFile, destDir, &overwritesJson);
+      Error error = detectZipFileOverwrites(pUploadState->tmpFile, destDir, pUploadState->fileName, &overwritesJson, unzipFound);
       if (error)
       {
          writeError(error);
@@ -1070,10 +1102,13 @@ bool handleFileUploadRequestAsync(const http::Request& request,
    uploadTokenJson[kUploadFilename] = pUploadState->fileName;
    uploadTokenJson[kUploadedTempFile] = pUploadState->tmpFile.getAbsolutePath();
    uploadTokenJson[kUploadTargetDirectory] = destDir.getAbsolutePath();
+   uploadTokenJson[kUnzipFound] = unzipFound; // Include here to be used by completeUpload()
 
    json::Object uploadJson;
    uploadJson["token"] = uploadTokenJson;
    uploadJson["overwrites"] = overwritesJson;
+   uploadJson[kUnzipFound] = unzipFound; // Include here to be used by GWT
+   uploadJson[kIsZip] = isZip;
    json::setJsonRpcResult(uploadJson, &response);
 
    // response content type must always be text/html to be handled
--- a/src/cpp/session/modules/SessionFiles.cpp
+++ b/src/cpp/session/modules/SessionFiles.cpp
@@ -585,7 +585,7 @@ Error completeUpload(const core::json::JsonRpcRequest& request,
    
    // parse fields out of token object
    std::string filename, uploadedTempFile, targetDirectory;
-   bool unzipFound;
+   bool unzipFound = false;
    error = json::readObject(token, 
                             kUploadFilename, filename,
                             kUploadedTempFile, uploadedTempFile,
@@ -666,14 +666,15 @@ Error detectZipFileOverwrites(const FilePath& uploadedZipFile,
                               const FilePath& destDir,
                               const std::string& originalFilename,
                               json::Array* pOverwritesJson,
-                              bool& rUnzipfound)
+                              bool* pUnzipfound)
 {
    // unable to use R's unzip here in worker thread, using system's unzip instead
    // try a couple locations for unzip, in case it's not on user's PATH
    std::vector<std::string> pathsToTry({
-                                       "unzip",
-                                       "/usr/bin/unzip"});
-   for (auto &path : pathsToTry) {
+      "unzip",
+      "/usr/bin/unzip"});
+   for (auto& path : pathsToTry)
+   {
       core::system::ProcessResult result;
       Error error = core::system::runCommand(path + " -Z1 " + uploadedZipFile.getAbsolutePath(),
                                              core::system::ProcessOptions(),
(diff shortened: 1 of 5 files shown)


src/cpp/session/modules/SessionFiles.cpp
```diff
@@ -1070,10 +1102,13 @@ bool handleFileUploadRequestAsync(const http::Request& request,
    uploadTokenJson[kUploadFilename] = pUploadState->fileName;
    uploadTokenJson[kUploadedTempFile] = pUploadState->tmpFile.getAbsolutePath();
    uploadTokenJson[kUploadTargetDirectory] = destDir.getAbsolutePath();
+   uploadTokenJson[kUnzipFound] = unzipFound; // Include here to be used by completeUpload()
```
cm421, line 1105: I have unzip found in two places because both GWT and completeUpload() need access to the flag, and I couldn't see an obvious way to access the flag in the token in GWT.

src/gwt/src/org/rstudio/studio/client/workbench/views/files/FilesUpload.java
```diff
@@ -67,22 +67,27 @@ public void execute()
          new OperationWithInput<PendingFileUpload>() {
             public void execute(PendingFileUpload pendingUpload)
             {
-               // confirm overwrites if necessary
+               Boolean unzipFound = pendingUpload.getUnzipFound();
+               Boolean isZip = pendingUpload.getIsZip();
                FileUploadToken token = pendingUpload.getToken();
```
cm421, line 72: I could see how to get values out of pendingUpload, but not from token. So unzipFound is in both. Maybe not a big deal, but if there's a cleaner way to do it I'm all ears.

src/gwt/src/org/rstudio/studio/client/workbench/views/files/FilesUpload.java
```diff
@@ -67,22 +67,27 @@ public void execute()
...
+                          true
+                          );
                }
-               else
-               {
-                  completeFileUploadOperation(token, true).execute();
+               else {
+                  // Upload and warn of overwrites, if any
+                  checkForFileUploadOverwrite(pendingUpload, token).execute();
                }
```
cm421, line 91: Is there any cleanup I need to do here at the end, particularly if the user cancels before `checkForFileUploadOverwrite()` is called?

src/gwt/src/org/rstudio/studio/client/workbench/views/files/FilesUpload.java
```diff
@@ -67,22 +67,27 @@ public void execute()
...
+                          false,
+                          checkForFileUploadOverwrite(pendingUpload, token),
+                          () -> {},
+                          true
+                          );
                }
-               else
-               {
-                  completeFileUploadOperation(token, true).execute();
+               else {
```
kevinushey, line 88:
```suggestion
               else
               {
```

src/gwt/src/org/rstudio/studio/client/workbench/views/files/FilesUpload.java
```diff
@@ -67,22 +67,27 @@ public void execute()
...
                   globalDisplay_.showYesNoMessage(
-                       MessageDialog.WARNING, 
-                       "Confirm Overwrite",
-                       confirmFileUploadOverwriteMessage(pendingUpload), 
-                       false, 
-                       completeFileUploadOperation(token, true), 
-                       completeFileUploadOperation(token, false), 
-                       false);
+                          MessageDialog.WARNING,
+                          "Unzip not found on PATH",
```
kevinushey, line 80: Some users might not not be familiar with the concept of the PATH, so I would suggest slightly more plain language:

```suggestion
                          "unzip not found,
```

src/gwt/src/org/rstudio/studio/client/workbench/views/files/FilesUpload.java
```diff
@@ -67,22 +67,27 @@ public void execute()
...
+                          "Unzip was not found on users PATH. Would you like to upload the zip archive without unzipping?",
```
kevinushey, line 81: For error / warning messages, I think a good format to adopt is:

1. What happened,
2. Why this matters,
3. What do you want to do

```suggestion
                          "The unzip system utility could not be found. unzip is required for decompressing .zip archives after upload. Would you like to upload the zip archive without unzipping?",
```

src/gwt/src/org/rstudio/studio/client/workbench/views/files/FilesUpload.java
kevinushey: I think the main requirement here is that `completeFileUploadOperation(token, false)` is called if the upload is cancelled.

src/cpp/session/modules/SessionFiles.cpp
kevinushey: I'm probably missing something, but couldn't you add a JSNI accessor to the FileUploadToken class (similar to https://github.com/rstudio/rstudio/pull/9231/files#diff-484e33a9fc038459f5594561f5a619c7ce46dc66e8cedc788d32e0c5b89c9b39R35-R37)?

src/gwt/src/org/rstudio/studio/client/workbench/views/files/FilesUpload.java
kevinushey: This is fine to me as well, but I think you could just access it from the FileUploadToken in both cases.

kevinushey (commented):
Overall LGTM; comments are mainly around style + tweaking of messages.

src/gwt/src/org/rstudio/studio/client/workbench/views/files/FilesUpload.java
MariaSemple: I would recommend something really plain language like "Unable to open the archive because the archive opening tool (unzip) could not be found." Maybe something about checking with the sys admin if they want the archive to be extracted?

MariaSemple: Just saw Kevin's next message - really like that suggestion.

src/cpp/session/modules/SessionFiles.cpp
```diff
@@ -660,41 +664,68 @@ Error completeUpload(const core::json::JsonRpcRequest& request,
...
+      {
+         // bash return code 127 - "command not found"
+         // try another unzip location
+         continue;
+      }
+      else
+      {
+         return systemError(result.exitStatus,
+                            "Unexpected result for unzip command",
+                            ERROR_LOCATION);
```
MariaSemple, line 719: I don't think we want to supply `result.exitStatus` to `systemError` here. I know we do it in general, but I think it leads to some confusing error messages because it ends up being mapped to an `errno` value. Exit codes from processes have a different meaning than `errno`s of the same value. For example, exit code 127 (as you used in your code above) indicates that the executable could not be found. `errno` 127 means "Key has expired" (`EKEYEXPIRED`). 

I'm not sure what we should pass in place, but I think we should include the exit status in the message instead of as the error code. @kevinushey Do you have any ideas of what would be a better error type? Maybe we need to make a new class of errors for process exits.

kevinushey: In this case, the exit code of 127 is handled explicitly above, so I think this is okay. (The other alternative would be to test for existence of the program; e.g. using `core::system::findProgramOnPath()` and then consider reporting a FileNotFoundError or something similar if it, or any hard-coded hint locations, cannot be resolved)

src/cpp/session/modules/SessionFiles.cpp
cm421: That was the first thing I tried, but for whatever reason copying the exact same JSNI accessors into FileUploadToken causes a reference error:
Screen Shot 2021-04-14 at 11 54 34 AM

Using the same code in the PendingFileUpload object does not cause that error, and I couldn't figure out why. I'm likely making some newbie mistake here. 

Here's snippets of what I tried:
FileUploadToken.java:
```
public class FileUploadToken extends JavaScriptObject
{
   protected FileUploadToken()
   {
   }

   public final native Boolean getUnzipFound() /*-{
      return this.unzipFound;
   }-*/;

   public final native Boolean getIsZip() /*-{
      return this.isZip;
   }-*/;
}
```

FileUpload.java:
```
public void execute(PendingFileUpload pendingUpload)
            {
               FileUploadToken token = pendingUpload.getToken();
               Boolean unzipFound = token.getUnzipFound();             <=== Reference Error
               ...
```

kevinushey: Weird! That looks correct to me. In some cases, rebuilding GWT from scratch with `ant clean; ant devmode` resolves the issue. Is that worth a shot? If it doesn't help, then I think leaving things as-is is fine -- we occasionally see inexplicable GWT errors like this that we usually, unfortunately, lack the time to fully diagnose or understand...

src/cpp/session/modules/SessionFiles.cpp
MariaSemple: I was just using 127 as an example as it has a nice specific meaning as an exit code. An exit code of 1 just means the process exited abnormally, but `errno` 1 is `EPERM` (operation not permitted). Exit odes between 1 and 127 (or 2 and 126, depending on your opinion) have whatever meaning the application developer chooses. Exit codes `128 + n` indicate the process exited with the signal `n`. There are no exit codes that line up with `errno` (at least not on purpose in the general case).

kevinushey: Ah, sorry, you're exactly right. I guess we just need a more generic "unzip returned error code %s" error / message here.

cm421: Would it be more clear if we used some "dummy/close enough" system error code, and then including the actual return code/error messages in the message?

``` suggestion
         // Not really a good error code to indicate that our bash command had an error
         // *nix:  1 = EPERM - Operation not permitted
         // WIN32: 1 = ERROR_INVALID_FUNCTION - Invalid function
         return systemError(1,
                            "Unexpected result for unzip command: " + 
                            std::to_string(result.exitStatus) + 
                            " - " + 
                            result.stdOut +
                            ": " +
                            result.stdErr,
                            ERROR_LOCATION);
```

MariaSemple: This might be too big for your change, but I think we'd be better off creating a different category of errors like "ProcessExitCode" errors or something. In the meantime, I think you should change `systemError` to `unknownError` so that at least we don't provide a message that looks like it has meaning (e.g. `"system error 1 (Operation not Permitted): <message here>"`) when really it doesn't. At least `unknownError` won't give false information.

src/gwt/src/org/rstudio/studio/client/workbench/views/files/FilesUpload.java
cm421: Fixed. All control paths will eventually call `completeFileUploadOperation()`

src/cpp/session/modules/SessionFiles.cpp
```diff
@@ -583,10 +585,12 @@ Error completeUpload(const core::json::JsonRpcRequest& request,
    
    // parse fields out of token object
    std::string filename, uploadedTempFile, targetDirectory;
+   bool unzipFound;
```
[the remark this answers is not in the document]
cm421: Absolutely agreed. Fixed!

src/cpp/session/modules/SessionFiles.cpp
```diff
@@ -660,41 +664,68 @@ Error completeUpload(const core::json::JsonRpcRequest& request,
    
 Error detectZipFileOverwrites(const FilePath& uploadedZipFile,
                               const FilePath& destDir,
-                              json::Array* pOverwritesJson)
+                              const std::string& originalFilename,
+                              json::Array* pOverwritesJson,
+                              bool& rUnzipfound)
```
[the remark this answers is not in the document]
cm421: I went back and forth on which way to do it. Thanks for the clarification. Fixed

src/cpp/session/modules/SessionFiles.cpp
```diff
@@ -660,41 +664,68 @@ Error completeUpload(const core::json::JsonRpcRequest& request,
...
-   Error error = core::system::runCommand("unzip -Z1 " + uploadedZipFile.getAbsolutePath(),
-                                          core::system::ProcessOptions(),
-                                          &result);
-   if (error)
-      return error;
+   // unable to use R's unzip here in worker thread, using system's unzip instead
+   // try a couple locations for unzip, in case it's not on user's PATH
+   std::vector<std::string> pathsToTry({
+                                       "unzip",
+                                       "/usr/bin/unzip"});
```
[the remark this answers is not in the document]
cm421: I debated this as well, but went for the multi-line format because it made it more apparent, to me at least, that it was a list of values that could be added to, and not a function call. That's probably overthinking it, but I've fixed the indentation.

src/cpp/session/modules/SessionFiles.cpp
```diff
@@ -660,41 +664,68 @@ Error completeUpload(const core::json::JsonRpcRequest& request,
...
+   for (auto &path : pathsToTry) {
```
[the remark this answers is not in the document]
cm421: Fixed!

src/cpp/session/modules/SessionFiles.cpp
cm421: Ah! Thank you! I hadn't noticed `unknownError` was an option. Using that for now.

src/cpp/session/modules/SessionFiles.cpp
cm421: No luck with cleaning/rebuilding, but I did catch a break by adding in getters for all the members of the JSON object, not just the two that I needed. Now it works great! Definitely not the behavior I expected

src/gwt/src/org/rstudio/studio/client/workbench/views/files/FilesUpload.java
cm421: Screen Shot 2021-04-14 at 9 30 14 PM

src/gwt/src/org/rstudio/studio/client/workbench/views/files/FilesUpload.java
cm421: fixed

src/gwt/src/org/rstudio/studio/client/workbench/views/files/model/FileUploadToken.java
```diff
@@ -21,4 +21,24 @@
...
+
+   public final native String getUploadTempFile() /*-{
+      return this.uploadedTempFile;
+   }-*/;
+
+   public final native String geTargetDirectory() /*-{
+      return this.targetDirectory;
+   }-*/;
+
+   public final native Boolean getUnzipFound() /*-{
```
kevinushey, line 37: A small nit (not needed to change in this PR): GWT's JSNI is usually able to handle conversion to "native" Java types; e.g. you could use a plain `boolean` here to make the code (in my subjective opinion) cleaner below.

src/gwt/src/org/rstudio/studio/client/workbench/views/files/model/FileUploadToken.java
```diff
@@ -21,4 +21,24 @@
...
+
+   public final native String getFilename() /*-{
+      return this.filename;
+   }-*/;
+
+   public final native String getUploadTempFile() /*-{
+      return this.uploadedTempFile;
+   }-*/;
+
+   public final native String geTargetDirectory() /*-{
```
kevinushey, line 33: Missing a `t`:

```suggestion
   public final native String getTargetDirectory() /*-{
```

kevinushey (approved):

MariaSemple (approved):
Besides Kevin's last two comments, LGTM!

PR #9261, opened 2021-04-23
Bugfix/busy session async

jeffvroom:
There are still some TODO comments here that I'm going to go over with Jonathan. Any feedback on how to make this better is much appreciated :) 

### Intent

Supports new configuration options to allow immediate or delayed async rpc and handling of selected operations offline - so they run even when the main thread is busy. 

### Approach

There are four new configuration options to rsession.conf:

    # settings to keep connectivity alive
    session-async-rpc-enabled=1
    session-async-rpc-timeout-ms=0

    # settings to process requests offline
    session-handle-offline-enabled=1
    session-handle-offline-timeout-ms=500

When either timeout-ms option is 0, the HttpConnection main request thread is used to convert all rpc requests to async or handle requests offline immediately. When it is non zero, the new OfflineService polling thread will do so if it wakes up and sees that the R session is currently executing, and the request was received longer than the specified timeout.

Periodically, offline service emits memory changed events in this same loop since those are only dispatched from the main thread.

An isMainThread check is added to key methods in r::exec so that the system logs and error and avoids stack corruption if an offline operation tries to use R. The app changes required include isWebsiteProject() - checking if R markdown is installed and findProgram(..) that calls R's Sys.which. There are TODOs in the code about how improving my solution to these problems.

The async-rpc requests are disabled from client_init until the first get_events request is made since the client only expects regular responses during setup.

Added a true/false user preference "Session protocol debug". When true, it logs debug messages to the file ~/.local/share/rstudio/log/rsession-(user).log for requests received and events sent from the rsession process. 

### Automated Tests

TBD: I would like to script up some tests that will kick off a system("sleep 60"), then verify that some operations can still proceed, and that we can queue up lots of console input messages, then interrupt the sleep.  It would be great to run these tests against a variety of configurations. We can also setup a mode to run all of the automated tests with various configuration options so that we can get more complete coverage of these new code paths.

Add a test in which R code updates the system PATH to include an executable checked by the rsession's findProgramOnPath.

### QA Notes

The defaults should not change any behavior right now as these options are disabled by default. 

Enabling and setting either or both of the timeout-ms values to 0 will ensure all requests are handled offline immediately if they are offlineable. If not, they are converted to async. With timeout-ms > 0, the request flow will only be altered if the R main thread is kept occupied for longer than the timeout-ms value. In other words, you will only see async rpcs used, or requests handled offline intermittently.

Also notice that when handle-offline is enabled, memory changed events should continue to be delivered even if R is busy.

There are two utilities in the code that have been changed to support "offline" requests: rmarkdownAvailable() and findProgramOnPath(). The rmarkdownAvailable() flag is now cached at startup. The findProgramOnPath() now uses the rsession's path env var which on Windows might drift out of sync if R makes changes to the PATH on the fly. 

### Checklist

- [ x] This PR passes all local unit tests


src/cpp/session/SessionRpc.cpp
```diff
@@ -332,6 +381,24 @@ Error initialize()
 
    RS_REGISTER_CALL_METHOD(rs_invokeRpc);
 
+   // TODO: asyncRpc - add option to registerRpcMethod?
+   s_offlineableUris.insert("/rpc/save_document");
```
kevinushey, line 385: Is there a way we can further validate that we aren't unintentionally introducing any issues with having these specific RPCs run on separate threads? I believe we'll have caught the "obvious" ones (interacting with R from a separate thread will cause immediate crashes) but worry about more subtle issues that might sneak by.

I wonder if a local development build with thread sanitizers enabled could help catch any other potential issues: https://clang.llvm.org/docs/ThreadSanitizer.html

Are we considering this, or any part of these changes, for 1.4-JR? If we are, I wonder if we could take the async RPC work but leave the offline-service work for 1.5, as it seems a bit more risky. Does that make sense?

src/cpp/session/SessionSourceDatabase.cpp
```diff
@@ -1033,6 +1034,11 @@ void onRemoveAll()
 
 SEXP rs_getDocumentProperties(SEXP pathSEXP, SEXP includeContentsSEXP)
 {
+   if (!r::exec::isMainThread())
```
kevinushey, line 1037: Not necessarily for this PR, but I wonder if we should instead make an appropriate change to our `RS_REGISTER_CALL_METHOD()` macro here:

https://github.com/rstudio/rstudio/blob/2171c9fa3bd775a8ddcdc5d35085df1717c88829/src/cpp/r/include/r/RRoutines.hpp#L75-L87

That way, we could have a way of detecting when any registered method is called from a separate thread (which should never happen), as opposed to just the ones we remember to sprinkle this on.

src/cpp/session/http/SessionHttpConnectionUtils.cpp
```diff
@@ -212,6 +197,10 @@ bool checkForSuspend(boost::shared_ptr<HttpConnection> ptrConnection)
       }
       else
       {
+         if (http_methods::connectionDebugEnabled())
+         {
+            LOG_DEBUG_MESSAGE("handle session suspend");
```
kevinushey, line 202: Does this (and the below stubs) need to be implemented before merging this PR?

src/cpp/r/RExec.cpp
```diff
@@ -132,7 +134,13 @@ Error evaluateExpressionsUnsafe(SEXP expr,
 {
    // detect if an error occurred (only relevant for EvalTry)
    int errorOccurred = 0;
-   
+
+   if (!isMainThread())
+   {
+      LOG_ERROR_MESSAGE("evaluateExpression called from thread other than main");
+      return rCodeExecutionError("Attempt to eval R on thread other than main", ERROR_LOCATION);
```
jmcphers, line 141: Let's include `str` as an error property to make it clearer what we're trying to do

src/cpp/session/modules/rmarkdown/SessionRMarkdown.cpp
```diff
@@ -75,6 +75,14 @@ namespace session {
 
 namespace {
 
+// TODO: asyncRpc - save_document_diff will call isWebsiteProject() that
+// needs to check the R runtime for whether Rmarkdown is installed. This is
+// not entirely accurate since if the first call to isWebsiteProject() is made
+// offline, it will return the wrong thing. Can we just cache a flag globally
+// whether RMarkdown is available - compute at startup and update when new libs
+// are added or will that state ever change?
+bool s_lastWebsiteProject;
```
kevinushey, line 84: I believe that once RStudio has finished initialization, it is not possible to change the project type within that same session without restarting RStudio (the front-end, not just the rsession itself). Given that, I think it would be reasonable to cache the project type on startup, and then read or use that as appropriate.

src/cpp/r/RExec.cpp
```diff
@@ -229,6 +237,10 @@ void SEXPTopLevelExec(void *data)
    
 Error executeSafely(boost::function<void()> function)
 {
+   if (!isMainThread())
+   {
+      LOG_ERROR_MESSAGE("executeSafely called from thread other than main");
```
jmcphers, line 242: Use `LOG_WARNING` when continuing

kevinushey (commented):
Overall LGTM -- awesome work!

My comments are mainly nits, plus some questions about whether some of the more local code could be generalized and reused in other places.

src/cpp/r/RExec.cpp
jmcphers: Or just return an error :-)

src/cpp/r/RExec.cpp
```diff
@@ -431,7 +462,12 @@ Error RFunction::call(SEXP evalNS, bool safely, SEXP* pResultSEXP,
          error.addProperty("symbol", functionName_);
       return error;
    }
-   
+   if (!isMainThread())
+   {
+      LOG_ERROR_MESSAGE("Attempt to call R function: " + functionName_ + " on thread than main");
```
jmcphers, line 467: nit: other than

src/cpp/session/SessionAsyncRpcConnection.hpp
```diff
@@ -0,0 +1,109 @@
+/*
+ * SessionRpc.hpp
```
jmcphers, line 2: Nit: header title

src/cpp/session/SessionClientEventQueue.cpp
```diff
@@ -71,7 +73,14 @@ bool ClientEventQueue::setActiveConsole(const std::string& console)
 }
 
 void ClientEventQueue::add(const ClientEvent& event)
-{ 
+{
+   if (http_methods::connectionDebugEnabled())
```
jmcphers, line 77: Consider adding a pref for this so we can all benefit!

src/cpp/session/SessionOfflineService.cpp
```diff
@@ -0,0 +1,302 @@
...
+         }
+
+         offlineThread_.detach();
+      }
+   }
+   catch(const boost::thread_interrupted&)
+   {
+      // the main thread is the one who calls stop() and it should 
+      // NEVER be interrupted for any reason
+      LOG_WARNING_MESSAGE("thread interrupted during stop");
```
jmcphers, line 117: Should indicate that this is the OfflineService thread

src/cpp/session/http/SessionHttpConnectionImpl.hpp
```diff
@@ -61,7 +61,8 @@ class HttpConnectionImpl :
    HttpConnectionImpl(boost::asio::io_service& ioService,
                       const HeadersParsedHandler& headersParsed,
                       const Handler& handler)
-      : socket_(ioService), headersParsedHandler_(headersParsed), handler_(handler)
+      : socket_(ioService), headersParsedHandler_(headersParsed), handler_(handler),
+        receivedTime_(boost::posix_time::microsec_clock::universal_time())
```
jmcphers, line 65: Use a monotonic clock

src/cpp/session/modules/rmarkdown/SessionRMarkdown.cpp
```diff
@@ -75,6 +75,14 @@ namespace session {
 
 namespace {
 
+// TODO: asyncRpc - save_document_diff will call isWebsiteProject() that
+// needs to check the R runtime for whether Rmarkdown is installed. This is
+// not entirely accurate since if the first call to isWebsiteProject() is made
+// offline, it will return the wrong thing. Can we just cache a flag globally
+// whether RMarkdown is available - compute at startup and update when new libs
```
jmcphers, line 82: Cache it in onDeferredInit

jmcphers (changes requested):

src/cpp/r/ROptions.cpp
```diff
@@ -78,6 +78,11 @@ int getBuildOptionWidth()
 
 SEXP getOption(const std::string& name)
 {
+   if (!r::exec::isMainThread())
```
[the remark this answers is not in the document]
jeffvroom: Yes, the rpc request for save_document_diff that we want to run offline called this sequence:

 isWebsiteProject() ->  rmarkdownPackageAvailable()  -> haveMarkdownToHTMLOption() -> getOption("rstudio.markdownToHTML")

Jonathan and I just spoke about caching the state for rmarkdownPackageAvailable() in an onDeferredInit() handler. 

When getOption is called from the wrong thread, it seems to cause R to longjmp that thread back into the R interpreter, so you end up with 2 threads in that main interpreter loop.

src/cpp/session/SessionOfflineService.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+// Called for each connection in mainConnectionQueue until an offlineable request is found.
+bool offlineConnectionMatcher(boost::shared_ptr<HttpConnection> ptrHttpConn,
+                              boost::posix_time::ptime now)
+{
+   if (!rpc::isOfflineableRequest(ptrHttpConn))
+      return false;
+
+   // Weird clock problem!
+   if (now < ptrHttpConn->receivedTime())
```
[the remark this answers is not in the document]
jeffvroom: Yes, good idea. Jonathan also suggested using steady_clock to avoid a system clock change from affecting this.

src/cpp/session/SessionRpc.cpp
jeffvroom: I will give it a go but suspect it might have too many false positives. It's a pretty hard problem to identify meaningful race conditions. We could have one big mutex around the logic in "handleConnection" but that might cause more problems than it's worth?  For now, I've been carefully reviewing the code in each offlineable request but you are right that it's very possible to miss something. There could be race conditions or also buried references to r::exec features.

Right now, this is scheduled only for 1.5 but I agree that the async-rpc stuff is a lot clearer cut and makes a big difference just by itself.

src/cpp/session/http/SessionHttpConnectionUtils.cpp
jeffvroom: Those are just part of a debug logging feature, so you'll see "handle session interrupt" or "handle save_document" in the logs one for each request received. I recognize it reads kind of funny in the code now that you mention it :)

src/cpp/session/SessionRpc.cpp
jeffvroom: Unfortunately, the thread sanitizer does have lots of false positives. I only looked into one in rserver where it complained about a thread race involving boost regex's use of a buffer pool from two different threads, even though from reading the code,  only one thread would be using one buffer at a time. In the docs, they say it doesn't work for C++11 mode very well so that might be a contributing factor.

kevinushey: That is unfortunate! Thanks for taking the time to give that a whirl -- if all we saw was false positives, I think we can still take that as a good sign (assuming that false negatives would be rare).

src/cpp/session/SessionModuleContext.cpp
```diff
@@ -1156,14 +1156,27 @@ FilePath findProgram(const std::string& name)
    else
    {
       std::string which;
+      FilePath resultPath;
+
+      // TODO: switch over to using findProgramOnPath all the time here - for now, doing it both ways to ensure there
+      // are no edge cases in findProgramOnPath that make a difference to app code
       Error error = r::exec::RFunction("Sys.which", name).call(&which);
+      Error dbgError = system::findProgramOnPath(name, &resultPath);
```
kevinushey, line 1164: For some extra context, we've seen things like this manifest on Windows, which (as I understand it) is caused by the fact that R links statically to the C runtime, whereas RStudio dynamically links to the C runtime. Because there are two separate C runtimes, there are also two separate `environ` blocks (among other things) which can cause issues if we need to ensure we're respecting the PATH as seen and mutated by the R process.

Some other examples where we've had to do similar gymnastics (mainly just FYI; not directly relevant for this PR):

https://github.com/rstudio/rstudio/blob/30872eaea5fce0fe66ded0f91bfe16cf24dbe0fd/src/cpp/r/include/r/RUtil.hpp#L21-L26

https://github.com/rstudio/rstudio/blob/30872eaea5fce0fe66ded0f91bfe16cf24dbe0fd/src/cpp/r/RUtil.cpp#L323-L350

src/cpp/session/SessionClientEventQueue.cpp
jeffvroom: I added a preference so we can turn this on and off without restarting.

src/cpp/session/modules/rmarkdown/SessionRMarkdown.cpp
jeffvroom: The rmarkdownAvailable() method is getting called when initializing some other module, before the onDeferredInit handler so I just am caching it the first time it's used. Even if that module goes away, there's a call to it in SessionRMarkdown's onDeferredInit() so it will definitely be called.

src/cpp/session/SessionModuleContext.cpp
jeffvroom: So maybe for windows, we'll need to get hold of R's version of the PATH variable during init. Thanks, that's a super helpful piece of info to keep in mind as well!

src/cpp/session/SessionSourceDatabase.cpp
jeffvroom: This could be useful but I'm going to hold off on this change for now because it's not obvious and will make tricky code even trickier :) I think that we mainly need these guards to prevent RStudio calls into the R runtime off the main thread. As I understand it, by the time these calls are made by the R runtime, we should have already caught the problem.  I do have checks the callbacks: polledEventHandler and rReadConsole but they turned out not to be that helpful since the stack was corrupted already by the time we got there on the wrong thread.

src/cpp/r/RExec.cpp
```diff
@@ -50,6 +50,8 @@ namespace {
 
 bool s_wasInterrupted;
 
+MainThreadFunction s_mainThreadFunction;
```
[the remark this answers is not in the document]
jeffvroom: If it's ok with you, how about we wait for the 2nd process to need "multi-threaded" R access? It looks like rsession is the only one in our team's builds but maybe it's used like shared_core elsewhere? Even though static init might work just fine, it feels safer to me that the check get turned on explicitly in the API.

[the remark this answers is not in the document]
kevinushey: That's a good point -- that seems fine to me.

src/cpp/session/SessionSourceDatabase.cpp
kevinushey: Sounds good!

src/cpp/session/SessionHttpMethods.cpp
```diff
@@ -844,6 +940,49 @@ void registerGwtHandlers()
...
+ * to have DEBUG so these messages show up. This prevents the user from needing to configure logging.conf
+ * for the session just to turn on debug logging.
+ */
+void initSessionDebugLog()
+{
+   if (s_sessionDebugLogCreated)
+      return;
+   s_sessionDebugLogCreated = true;
+
+   // TODO: should we have a preference for the path to this log file?
```
jmcphers, line 956: Generally we avoid committing TODOs since it makes it hard to use them to check for remaining work in a PR. 

Since this is a dev debug mechanism I don't think it needs to be configurable; if someone really wants to put the logs somewhere else they can always use the XDG vars.

src/cpp/session/SessionOfflineService.cpp
```diff
@@ -0,0 +1,301 @@
...
+
+            // Wait for the client to be initialized before any offline handling
+            if (!httpConnectionListener().eventsActive())
+               continue;
+
+            // If R is not occupying the main thread, we'll continue to wait.
+            // TODO: asyncRpc - are there other skippable operations, not from R - if so, we could remove this test
+            if (!console_input::executing())
+            {
+               // TODO: asyncRpc log a warning here if mainConnectionQueue has more than 5 entries
```
jmcphers, line 229: As above -- do these TODOs need to be resolved before commit? If you intend to commit them, let's change them to "consider"

jmcphers (approved):

jeffvroom:
Fixes issue:

https://github.com/rstudio/rstudio/issues/2038

Issue #9106, opened 2021-03-15
Stan diagnostics handler duplicated on save


kevinushey:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.4.1612
    OS Version      : macOS 11.2.3
    R Version       : 4.0.4

### Steps to reproduce the problem

Create a Stan document, with e.g.

```
file.edit("~/test.stan")
```

Then, execute this in the R console:

```
count <- 0
trace(.rs.stan.runDiagnostics, quote(.GlobalEnv$count <- .GlobalEnv$count + 1))
```

Then, try editing and saving the file a couple of times. Watch the `count` variable increase in the Environment pane.

### Describe the problem in detail

The save handlers are duplicated, causing the diagnostics engine to be run multiple times for each save. The `count` variable starts incrementing multiple times on each save.

### Describe the behavior you expected

Handlers should not be duplicated.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting a RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
The fix should be simple:

https://github.com/rstudio/rstudio/blob/aced3d1872861c879a2f7a523768bd20bb51707a/src/gwt/src/org/rstudio/studio/client/workbench/views/console/shell/assist/StanCompletionManager.java#L243-L257

Those handlers need to be saved as a local variable and then detached in the detach() method.

mikebessuille:
If this is going to resolve #9055, and if it's easy to fix, we should definitely fix this soon.

ronblum:
Using the example above, this is getting incremented twice each time, which is better than before, but still greater than 1.

nicokist:
@ronblum That is what I found as well.

ronblum:
Testing note: Switching back and forth between the Editor and Console panes triggers the issue, as well, so check both with that and with updating/saving the file.

ronblum:
Validated in
- RStudio Desktop 1.5.168 on MacOS 12 beta
- RStudio Server 1.5.168 on SLES 15 SP2 via Chrome on MacOS 12 beta
[...]

PR #9279, opened 2021-04-27
Bugfix/rstan duplicate save

cm421:
### Intent

Completion managers that extend the `CompletionManagerBase` class have duplicated handlers that fire twice for every one event they're registered for. This was noticed in #9106 specifically for `.stan` files when saving, but it's been occurring for all file types that use `CompletionManagerBase`, including:
- Markdown files
- Python files
- SQL files
- Stan files

### Approach

The handlers need to be registered after the completion manager objects are fully instantiated. Completion mangers now take a factory-ish approach using a "funstructor" `create()` method in place of the, now private, constructor to:
1. create the object
2. register the handlers
3. return the created object.

### Automated Tests

Probably not needed

### QA Notes

#9106 describes how to check the autosave handler for `.stan` files. Saves (and autosaves) should increment `count` by one. Clicking inside the editor of a `.stan` file for the first time should also trigger a single `OnFocus` event

### Checklist


kevinushey (approved):
LGTM!

PR #9283, opened 2021-04-28
support R graphics engine 14

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/9251.

Note that this PR is smaller than it appears to be; a substantial amount of the code here is moving our graphics device description struct definitions to their own header file, just for a bit of extra cleanliness.

### Approach

We had missing handlers for the newly-implemented R graphics engine methods (setPattern, setClipPath, setMask and their release equivalents); this PR implements those.

### Automated Tests

This test could be exercised by attempting to create some sample plots, and so feels worthwhile. I'll log an issue.

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/9251#issuecomment-825412073.

### Checklist


src/cpp/r/session/graphics/RShadowPngGraphicsHandler.cpp
```diff
@@ -192,6 +241,23 @@ Error shadowDevDesc(DeviceContext* pDC, pDevDesc* pDev)
          Error err = r::exec::executeString(code);
          if (err)
             return err;
+         
+         // stop-gap fix for crash with usages of newer graphics device APIs
+         // added in R 4.1.0 / graphics engine 14
+         // https://github.com/rstudio/rstudio/issues/9251
+         if (R_GE_getVersion() >= 14)
```
kevinushey, line 248: This is really the "core" of the PR -- we override these callbacks on the R graphics device. See https://github.com/rstudio/rstudio/issues/9251#issuecomment-825412073 for extra motivation.

src/cpp/r/session/graphics/RGraphicsDevDesc.cpp
```diff
@@ -19,530 +19,6 @@
 
 #include <R_ext/RS.h>
 
-// compatability structs for previous graphics engine versions
```
kevinushey, line 22: These were all just migrated to RGraphicsDevDescVersions.hpp.

kevinushey:
Looks like I was able to get things working after all:

Screen Shot 2021-04-28 at 11 29 56 AM

src/cpp/core/include/core/FileInfo.hpp
```diff
@@ -38,7 +38,8 @@ class FileInfo
       : absolutePath_(), 
         isDirectory_(false), 
         size_(0), 
-        lastWriteTime_(0)
+        lastWriteTime_(0),
+        isSymlink_(false)
```
jmcphers, line 42: Is this change meant to be included? (don't see it referenced anywhere)

src/cpp/core/system/file_monitor/FileMonitor.cpp
```diff
@@ -420,7 +420,8 @@ class RegistrationCommand
 
 public:
    RegistrationCommand()
-      : type_(None)
+      : type_(None),
+        recursive_(false)
```
jmcphers, line 424: Also seems not to be used?

jmcphers (approved):

src/cpp/core/include/core/FileInfo.hpp
kevinushey: This was caught by ASAN; we need to explicitly initialize the member to `false` otherwise when we reference it (as evidently we do in some cases) we'll end up reading uninitialized memory.

PR #9301, opened 2021-05-03
add "indices" to spellcheck domain wordlist

tofof:
Both _indexes_ and _indices_ are acceptable English pluralizations of index, but "[_indices_ is generally preferred in mathematical, financial, and technical contexts](https://grammarist.com/usage/indexes-indices/)," such as in, say, a statistical computing application.


jmcphers (approved):
LGTM, thank you for this improvement!

PR #9332, opened 2021-05-12
ensure R session is initialized before running tests

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/9331.

### Approach

Use a separate `runTests()` callback that can be triggered at an appropriate point after RStudio has finished initializing R.

Note that some parts of initialization (specifically, deferred init) is not run at this point -- should we consider running that for tests too?

### Automated Tests

N/A

### QA Notes

N/A

### Checklist

### DIFF
--- a/src/cpp/r/RExec.cpp
+++ b/src/cpp/r/RExec.cpp
@@ -425,7 +425,6 @@ Error RFunction::call(SEXP evalNS, bool safely, SEXP* pResultSEXP,
    // verify the function
    if (functionSEXP_ == R_UnboundValue)
    {
-      LOG_ERROR_MESSAGE("Failed to find function: '" + functionName_ + "'");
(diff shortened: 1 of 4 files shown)


src/cpp/r/RExec.cpp
```diff
@@ -425,7 +425,6 @@ Error RFunction::call(SEXP evalNS, bool safely, SEXP* pResultSEXP,
    // verify the function
    if (functionSEXP_ == R_UnboundValue)
    {
-      LOG_ERROR_MESSAGE("Failed to find function: '" + functionName_ + "'");
```
kevinushey, line 428: Removed as this message is somewhat noisy + the caller should be responsible for logging the error is appropriate anyhow.

jmcphers (approved):

Issue #7660, opened 2020-08-25
Serialize the Java and C++ portions of the official build


gtritchie:
During the "official build" we allow cmake to trigger the Java (GWT) and C++ builds at the same time, subject to the `-j` value in the Jenkinsfile.

We've started hitting issues where builds are being terminated due to OOM errors, and suspect it's related to parallelization (Java being a huge fan of using oodles of memory).

Trying a temporary fix to reduce number of cores we allow cmake to use (#7659), but we should try serializing Java and C++ so they can each take full advantage of the available cores.

gtritchie:
Purely a build-time thing, nothing to test.

PR #9336, opened 2021-05-12
Streamline VS Code + Docker Dev Experience

jonyoder:
### Intent

To prepare for the Platform team's upcoming launcher work, I've been experimenting with ways to streamline setting up a dev environment.

The intent is not to change normal dev practices, but to introduce a streamlined way for occasional RSP developers to quickly ramp up without a complex dev environment setup. This PR is a result of that work, and it includes:

* A new readme (`VSCODE.md`) for getting started with Docker + VS Code
* Modifies the `.devcontainer` Dockerfile to invoke the dependency install scripts individually. This has two benefits:
  * Allows you to simply rebuild your Docker container periodically when dependencies change.
  * Calling scripts individually results in better layering; a change in one dependency doesn't require re-installing all dependencies.
* Sets `RSTUDIO_TARGET=Server` by default so configuration can be run without changes in the devcontainer approach. This does not affect configuration in other environments.
* Makes the `Xmx` and `Xss` memory settings configurable (same defaults for normal development). For devcontainer/VSCode development, these settings are customized to 1000M/16M to make building in Docker more likely to succeed.
* Throttles parallel build jobs to `4` (results in `-j 4`) for better results when building on Docker.
* Sets the correct include paths so you can jump to Boost sources in VS Code.
* Adds support for a `GWT_BUILD_ALL` env var. When set to `NO`, this suppresses the GWT build in the `all` target. I've found that the GWT build often fails when running in parallel with C++ builds. This is automatically suppressed in the devcontainer, but unaffected for other dev approaches.
* Adds a couple of VS Code plugins for auto-install in the devcontainer.

### Approach

The above changes should not affect other development environments like developing in a VM, developing locally on Linux or Mac/Windows, CI builds, or building with the docker build script. It only affects development when cloning the repo in a container volume using Docker + VS Code.

With the changes in this PR, I've been able to open this repo in VS Code and build it successfully with no configuration changes or failures.

### Caveats

* Makes some assumptions (like `-j 4`) that may not be ideal for everyone developing in Docker. I tried to find reasonable defaults that should work in most cases.
* The Dockerfile changes for better dependency layering may require a bit of attention from time to time, in particular if the dependencies installed by `install-common` change.
* The GWT build must be run separately when in the VS Code devcontainer environment. Personally, I prefer this, and I expect it's fine, but I could be wrong.

### Questions

I'm unclear on the current Java requirements. I ran into trouble when installing both Java 8 and 11, but you can resolve it by configuring a `JAVA_HOME` for the devcontainer. VS Code is unhappy about Java 8, but likes Java 11. Since GWT seems to build fine with Java 11, I simply removed Java 8 to simply things. Is this ok?

### Test Notes

To test with this branch:
* Make sure your Git SSH Key is in your SSH agent.
* Install the `Remote - Containers` VS Code extension.
* Choose "Remote-Containers: Clone Repository in Named Container Volume" from the command palette.
* Enter `git@github.com:/rstudio/rstudio-pro/tree/jon/vscode`.

### Checklist


jonyoder:
When porting this to the open source repo, I realized that the `install-overlay` script was missing. Apart from that, the `Dockerfile` I had been using for `rstudio-pro` worked fine. Should we consider adding a dummy `install-overlay` script to this repo (that does nothing) so the Dockerfile can be ported without changes?

VSCODE.md
```diff
@@ -0,0 +1,58 @@
+# Starting Development with Visual Studio Code
+
+You can get started with development very quickly with Visual Studio Code amd Docker.
+
+## Install Prerequisites
+
+* Install Docker.
+* Install VS Code and install the "Remote - Containers" extension in VS Code.
+* Configure your Docker resources to give sufficient resources to the
```
gtritchie, line 9: This is very platform dependent (e.g. on Linux docker by default has access to all resources of the host machine), but on Docker Desktop on Mac, you can tweak via UI preferences. On Windows, using WSL2 as the docker backend (that's the default on current versions), you don't have settings for this so nothing to configure as far as I'm aware.

Windows using Hyper-V (now considered legacy) does provide knobs and dials for this, I believe.

So perhaps simplest to make this note Mac-specific to avoid Windows users hunting for something they won't find?

.devcontainer/devcontainer.json
```diff
@@ -2,26 +2,40 @@
...
   },
 
   "extensions": [
     "editorconfig.editorconfig",
     "ms-vscode.cpptools",
     "ms-vscode.cmake-tools",
     "twxs.cmake",
-    "vscjava.vscode-java-pack"
+    "vscjava.vscode-java-pack",
+    "ms-azuretools.vscode-docker",
```
gtritchie, line 32: Why include docker extension? We don't install docker inside the container, so the extension won't be usable for much.

.devcontainer/Dockerfile
```diff
@@ -68,8 +68,75 @@ ENV DEBIAN_FRONTEND=noninteractive
...
+RUN mkdir -p /opt/rstudio-tools/dependencies/common
+COPY ./dependencies/tools/rstudio-tools.sh /opt/rstudio-tools/dependencies/tools/
+
+# copy panmirror project setup (so install-common can install needed deps)
+RUN mkdir -p /opt/rstudio-tools/panmirror
+COPY ./src/gwt/panmirror/src/editor/yarn.lock /opt/rstudio-tools/panmirror/
+COPY ./src/gwt/panmirror/src/editor/package.json /opt/rstudio-tools/panmirror/
+
+# install common dependencies
+# here, we deviate from the typical "./install-common" script and instead
```
gtritchie, line 92: This makes sense from a docker workflow perspective, but when we adjust dependency scripts (e.g. add new ones, remove old ones), this may break and nothing in the local dev workflow or official builds will catch that. That's probably fine, should be infrequent, and people using the devcontainer will notice and presumably go fix it.

.devcontainer/Dockerfile
```diff
@@ -68,8 +68,75 @@ ENV DEBIAN_FRONTEND=noninteractive
...
+
+# yaml-cpp
+COPY ./dependencies/common/install-yaml-cpp /opt/rstudio-tools/dependencies/common/
+RUN cd /opt/rstudio-tools/dependencies/common && /bin/bash ./install-yaml-cpp
+
+# Qt SDK
+COPY ./dependencies/common/install-qt.sh /opt/rstudio-tools/dependencies/linux
+COPY ./dependencies/linux/install-qt-linux /opt/rstudio-tools/dependencies/linux
+RUN export QT_VERSION=5.12.8 && \
+    cd /opt/rstudio-tools/dependencies/linux && /bin/bash ./install-qt-linux
```
gtritchie, line 139: Since you build with `RSTUDIO_TARGET=Server` the Qt SDK is not used, and you can remove this from the Dockerfile. We should focus our efforts on making this workflow work for Server, and once we've moved to Electron I'll tinker with making that work for Desktop.

.devcontainer/devcontainer.json
```diff
@@ -2,26 +2,40 @@
...
+    "context": ".."
   },
 
   "settings": {
-    "terminal.integrated.shell.linux": "/bin/bash"
+    "terminal.integrated.defaultProfile.linux": "/bin/bash",
+
+    "cmake.configureArgs": ["-DRSTUDIO_TARGET=Server"],
+    "cmake.configureEnvironment": {
+      "GWT_XMX": "-Xmx1000M",
```
gtritchie, line 14: A lot of this stems from trying to build GWT at the same time as the C++ code. Known issue:  https://github.com/rstudio/rstudio/issues/7660

VSCODE.md
```diff
@@ -0,0 +1,58 @@
...
+```sh
+# Use your Github username and PAT (personal access token) as the password when prompted.
+# It's fine to cancel this process after you see the clone is starting successfully.
+git clone https://github.com/rstudio/rstudio.git
+```
+
+### Using Git with SSH
+
+For SSH, you must be running an SSH agent. Simply add your SSH key to the SSH agent
+with `ssh-add ~/.ssh/<my-key>`.
```
gtritchie, line 58: We'll want a section describing how to run RStudio Server after you've built it. Also how to debug, and how to iterate on different parts of the code. We don't know all those answers yet, but at least include the placeholder sections for us to fill in as we figure it out.

gtritchie: Also, an external contributor will need to fork the repo and make their changes there before making a PR back to us. Maybe we should mention that? Not sure yet what level of details we want.

src/gwt/CMakeLists.txt
```diff
@@ -60,7 +80,13 @@ if(GWT_BUILD)
    endif()
 
    # invoke ant to build
-   add_custom_target(gwt_build ALL)
+   if(GWT_BUILD_ALL)
```
gtritchie, line 83: In a dev workflow, we don't build GWT (typically) as part of the overall build. Instead the developer will go into `rstudio/src/gwt` and run `ant devmode`. This runs a recompile-on-demand service that is much faster than a full production build (`ant`) and supports rebuilding when you refresh the browser showing the IDE.

Pro adds some other wrinkles, but let's tackle that after dealing with open-source

gtritchie: So, if we're wanting to put the recommended dev workflow into the container, we don't want the build to include GWT. There are existing ways to tell cmake to skip it (can chat about that offline if not obvious).

Then our "how to run the server, and iterate on the code" instructions will include the step of running `ant devmode`.

This will probably eliminate the need for messing with the Java resource switches, though having that capability is worth having just in case.

gtritchie (changes requested):
Some initial comments, still a lot to figure out so we'll keep iterating! Thanks for working on this!

.devcontainer/Dockerfile
jonyoder: That's pretty much what I thought. I'll resolve this, but we can always come back to it later.

src/gwt/CMakeLists.txt
jonyoder: In this PR, we set the `GWT_BUILD_ALL` var for CMake to `NO`, so we do exclude it from the `ALL` target by default. Is this ok, or are you suggesting that we accomplish it another way?

gtritchie:
Just to capture it, I tried doing a package build on macOS on this branch (directly, not in a container or anything), and consistently hitting a GWT compilation error (unhelpful result below). Doing the same style of build against main isn't hitting this. So strong possibility this would break at least the Mac build once merged.

```
     [exec] [14:48:43] Resolved 'ide-dist' after 20s 606ms
     [java] Error: Could not find or load main class
     [java] Caused by: java.lang.ClassNotFoundException:

BUILD FAILED
/Users/gary/alt/pr/rstudio/src/gwt/build.xml:147: Java returned: 1
```

src/gwt/CMakeLists.txt
jonyoder: I removed this logic since #9885 modified the GWT build to run in isolation.

gtritchie (approved):
LGTM!

PR #9378, opened 2021-05-19
implement basic R version scanning

kevinushey:
Implements basic R version scanning, so that the version of R on the PATH is located and used by default. (This will eventually be extended with UI for selecting versions of R on Windows; this is just to help ease development for now.)


gtritchie (approved):
LGTM, but be aware I'm going to be deleting all this code in about a week.

That said, I'll certainly consult this when I get to this part!

PR #9410, opened 2021-05-25
initial bits and pieces of Electron project

gtritchie:
### Intent

Get some basic code in place so we can start collaborating on the Electron project.

### Approach

Ported some of the FilePath class and the Xdg tools, and a few other bits and pieces, started sketching out the `main` function, etc. You can build and run it (but it just exits right now).

Code is split into three folders:

1. `src/core` is for generic stuff ported from the C++ core and shared_core libraries
2. `src/main` is for the main Electron process, this is where most of the Desktop code will live
3. `src/renderer` is for the Electron renderer process; we will only be writing code here to serve as a bridge between the GWT code (loaded in the Chromium instance) and the main process

Unit tests live in matching folders: `test/core`, etc.

Run `yarn` once to install dependencies, then `yarn build` to build `src`, and `yarn test` to build and run the unit tests.

`yarn start` will start, but as I said it will quickly exit as there's no main window being created yet.

### Automated Tests

Glad you asked. Here's the code coverage, most of the uncovered lines are Windows-specific and I ran this on Mac.

```
  106 passing (138ms)

----------------|---------|----------|---------|---------|-----------------------------------------------------
File            | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s                                   
----------------|---------|----------|---------|---------|-----------------------------------------------------
All files       |   92.47 |    83.33 |      99 |   92.73 |                                                     
 environment.ts |     100 |      100 |     100 |     100 |                                                     
 err.ts         |     100 |      100 |     100 |     100 |                                                     
 file-path.ts   |   92.57 |    89.71 |   98.67 |      93 | 69,76-77,86,161,188,214,283-284,375-376,497,676,691 
 log.ts         |     100 |      100 |     100 |     100 |                                                     
 system.ts      |   94.44 |    57.14 |     100 |   94.44 | 62                                                  
 user.ts        |     100 |      100 |     100 |     100 |                                                     
 xdg.ts         |      88 |    74.47 |     100 |      88 | 54-55,60-61,124-126,265,276                         
----------------|---------|----------|---------|---------|-----------------------------------------------------
```
### QA Notes

Nothing testable here.

### Checklist


src/node/desktop/src/core/xdg.ts
```diff
@@ -0,0 +1,293 @@
...
+  FOLDERID_LocalAppData,
+  FOLDERID_ProgramData
+}
+
+/**
+ * Simplified implementation of Windows `SHGetKnownFolderPath` API using environment
+ * variables.
+ */
+export function SHGetKnownFolderPath(folderId: WinFolderID): string {
+  let envVar = "";
```
kevinushey, line 51: This is ultra bike-shedding, but should we prefer using single-quoted strings or double-quoted strings?

src/node/desktop/src/core/xdg.ts
```diff
@@ -0,0 +1,293 @@
...
+  }
+  return getenv(envVar);
+}
+
+/**
+ * Returns the hostname from the operating system
+ */
+function getHostname(): string {
+
+  if (!getHostName.hostname)
```
kevinushey, line 71: It's a little confusing here having the function name and namespace name here differ by only a single letter :-)

(Do we need to use a namespace here? Why not just have an instance variable or something in this module?)

src/node/desktop/src/main/desktop-utils.ts
```diff
@@ -0,0 +1,1026 @@
...
+ * MERCHANTABILITY OR FITNESS FOR A PARTICULAR PURPOSE. Please refer to the
+ * AGPL (http://www.gnu.org/licenses/agpl-3.0.txt) for more details.
+ *
+ */
+
+
+import { Xdg } from '../core/xdg';
+import { getenv } from '../core/environment';
+
+export namespace Desktop {
```
kevinushey, line 20: What is the motivation for using a namespace here (rather than just having these functions exported as part of the module, or defining an explicit export object or similar)?

src/node/desktop/src/main/product-info.ts
```diff
@@ -0,0 +1,19 @@
...
+ * this program is licensed to you under the terms of version 3 of the
+ * GNU Affero General Public License. This program is distributed WITHOUT
+ * ANY EXPRESS OR IMPLIED WARRANTY, INCLUDING THOSE OF NON-INFRINGEMENT,
+ * MERCHANTABILITY OR FITNESS FOR A PARTICULAR PURPOSE. Please refer to the
+ * AGPL (http://www.gnu.org/licenses/agpl-3.0.txt) for more details.
+ *
+ */
+
+export function getRStudioVersion() {
+  // TODO: need to figure out how to consume build numbers for package build, etc.
```
kevinushey, line 17: We could probably have something like `product-info.ts.in` and generate `product-info.ts` from that in CMake.

kevinushey (approved):
LGTM! Questions are mainly for my own understanding.

src/node/desktop/src/core/xdg.ts
gtritchie: There's no universal preference I've found. Some projects prefer single, some double, and of course if you want to include embedded quotes in a literal then easiest to use the other one around it.

We can pick one and configure the tools to auto-format it that way if we want, along with other conventions.

src/node/desktop/src/core/xdg.ts
gtritchie: Good catch, that's a typo. It was meant to be the same as the function.

This is a pattern used in TypeScript to work around the limitation on having static members inside functions. Typescript doesn't support it even though Javascript does.

See comment from ahejlsberg (the guy who invented TypeScript) here:

https://github.com/Microsoft/TypeScript/issues/8419

gtritchie: It's ugly, though, and obviously doesn't accomplish much type-wise, given I had it wrong.

src/node/desktop/src/main/desktop-utils.ts
gtritchie: Just playing with different approaches, honestly. Given you can just import everything from a file and give it a namespace at that point, this is probably not a worthwhile pattern. Just removed it.

src/node/desktop/src/main/product-info.ts
gtritchie: Need to learn more about versioning at the Electron API level. For example, stumbled across this, not sure what role is plays or where...

https://www.electronjs.org/docs/api/structures/product

PR #9425, opened 2021-05-27
Fix issue where unit tests running as sudo will not run correctly on CentOS 7 due to weird sudo and escaping issue

kfeinauer:
### Intent

Unit tests run with `sudo` on CentOS 7 don't actually run and instead immediately exit. See the running Launcher tests for pro builds as an example.

### Approach

Fix this. See https://github.com/rstudio/launcher/pull/100 for more details.

### Automated Tests

None necessary.

### QA Notes

Launcher tests should start running for CentOS 7. Already verified in new Launcher repo.


kevinushey (approved):
LGTM!

PR #9444, opened 2021-06-03
support build vs. release configurations for desktop

kevinushey:
### Intent

Allow for the notion of different build types within our Typescript sources, in a simple (if somewhat hacky) way.

### Approach

`src/config` is now a symlink, which can be used to point either at `../config/development` or `../config/release`. The implementations within are just simple `enum`s, which indicate the "Current" build type, which can then be tested and used as appropriate.

The commands `yarn use-dev` and `yarn use-release` are shortcuts for updating that symlink.

I haven't verified this, but I believe this should suffice since these `enum`s and their tests will be constant expressions at build time, so any unused branches should be eliminated.

I took this approach for a couple of reasons:

`tsc` supports the notion of multiple projects; e.g. as in:

```
$ tsc --help | grep -- '--project'
 -p FILE OR DIRECTORY, --project FILE OR DIRECTORY  Compile the project given the path to its configuration file, or to a folder with a 'tsconfig.json'.
```

so, in theory, we could have separate `tsconfig.json` files ... but as far as I can see, there isn't a way to pass this configuration information down to `mocha` for unit testing. (I tried, but hit a number of landmines / road blocks in the process.)

We could also consider a separate CMake configuration step (and we might need this in the future), but I was worried this would complicate development work and I didn't want to bring CMake into the development pipeline unless absolutely necessary.

### Automated Tests

Included as part of the PR.

### QA Notes

N/A

### Checklist


gtritchie (approved):
LGTM!

PR #9447, opened 2021-06-03
implement createAliasedPath

kevinushey:
Adds an implementation of `createAliasedPath()`, built on top of some of the Node primitives.


src/node/desktop/test/core/file-path.test.ts
```diff
@@ -602,11 +602,29 @@ describe('FilePath', () => {
...
+
+    it('Paths are aliased correctly, even with trailing slashes', () => {
+      const path = new FilePath('/Users/user/path/to/project');
+      const home = new FilePath('/Users/user/');
+      const aliasedPath = FilePath.createAliasedPath(path, home);
+      expect(aliasedPath).to.equal('~/path/to/project');
+    });
+
+  });
+
```
gtritchie, line 623: I'd include a few additional tests:

- One path specified with forward slashes, the other with backslashes, should still work (just because a comment in the original mentions that case)
- Test behavior with DOS paths; when run on Posix, I think the current code will not do the alias substitution on a fully qualified DOS path, but it will on Windows (so check both in test gated by process.platform?)
- Also capture resulting behavior from this comment, e.g. path isn't even a child of the home:
```
// try to compute home-relative path -- if that fails,
// or the computed path does not appear to be relative,
// then just return the original file path
```

src/node/desktop/src/core/file-path.ts
```diff
@@ -26,14 +26,19 @@ export interface FilePathWithError {
   path: FilePath;
 }
 
+const homePathAlias = '~/';
```
gtritchie, line 29: Makes sense, been through so many iterations no idea how I ended up with them as static members.

gtritchie (changes requested):

gtritchie (approved):
LGTM!

PR #9449, opened 2021-06-03
use node rather than shell for config switch

kevinushey:
Ensures compatibility with Windows.

### DIFF
--- a/src/node/desktop/package.json
+++ b/src/node/desktop/package.json
@@ -36,7 +36,6 @@
     "test": "electron-mocha -c",
     "testcover": "nyc electron-mocha -c",
     "testwip": "electron-mocha -c -g WIP",
-    "use-release": "rm -rf src/config && ln -nfs ../config/release src/config",
-    "use-dev":     "rm -rf src/config && ln -nfs ../config/development src/config"
+    "use": "node ./scripts/use.js"
   }
 }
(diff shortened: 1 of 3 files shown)


src/node/desktop/package.json
```diff
@@ -36,7 +36,6 @@
     "test": "electron-mocha -c",
     "testcover": "nyc electron-mocha -c",
     "testwip": "electron-mocha -c -g WIP",
-    "use-release": "rm -rf src/config && ln -nfs ../config/release src/config",
-    "use-dev":     "rm -rf src/config && ln -nfs ../config/development src/config"
+    "use": "node ./scripts/use.js"
```
gtritchie, line 39: How do we "use" this now (no pun intended, mostly)? `yarn use development` or something like that?

Gotta be a better way to do this sort of thing, hopefully it will come to us in a dream.

src/node/desktop/test/config/config.test.ts
```diff
@@ -23,16 +23,12 @@ import { BuildType } from '../../src/config/config';
 describe('Config', () => {
 
   it('Should reflect active configuration', async () => {
-
-    // since 'src/config' is a symlink to the 'active' build type,
-    // we resolve that for testing the 'active' build type
-    const configPath = fs.realpathSync('src/config');
-    if (configPath.includes('development')) {
-      expect(BuildType.Current).to.equal(BuildType.Development);
+    const contents = fs.readFileSync('src/config/config.ts', 'utf8');
```
gtritchie, line 26: Since the config is a .js file, why not just import it and use it as code instead of reading it and parsing it?

gtritchie (changes requested):

kevinushey:
Closing this as @gtritchie pointed out that Electron already has a flag we can check to use for "package" builds: https://www.electronjs.org/docs/api/app#appispackaged-readonly

PR #9452, opened 2021-06-04
Start converting node desktop tests to Chai's assert

cm421:
### Intent

Convert current node tests to use Chai's assert library. This is just the start, all tests will be converted to use assert.

### Approach

N/A

### Automated Tests

N/A

### QA Notes

N/A

### Checklist

### DIFF
--- a/src/node/desktop/test/core/err.test.ts
+++ b/src/node/desktop/test/core/err.test.ts
@@ -26,14 +26,14 @@ function beUnsuccessful(): Err {
   return new Error('Some error');
 }
 
-describe('Err Using Chai Asserts', () => {
+describe('WIP Err', () => {
   describe('Success helper', () => {
     it('Success return should be falsy', () => {
-      assert.isFalse(!!beSuccessful());
+      assert.isNotOk(beSuccessful());
       assert.isNull(beSuccessful());
     });
     it('Error return should be truthy', () => {
-      assert.isTrue(!!beUnsuccessful());
+      assert.isOk(beUnsuccessful());
       assert.instanceOf(beUnsuccessful(), Error);
     });
   });
--- a/src/node/desktop/test/core/err.test.ts
+++ b/src/node/desktop/test/core/err.test.ts
@@ -26,14 +26,14 @@ function beUnsuccessful(): Err {
   return new Error('Some error');
 }
 
-describe('WIP Err', () => {
+describe('Err', () => {
   describe('Success helper', () => {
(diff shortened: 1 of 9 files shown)


src/node/desktop/test/core/err.test.ts
```diff
@@ -26,16 +26,15 @@ function beUnsuccessful(): Err {
   return new Error('Some error');
 }
 
-describe('Err', () => {
+describe('Err Using Chai Asserts', () => {
```
gtritchie, line 29: Think this can just be left as-is, versus mentioning 'Chai Asserts'.

src/node/desktop/test/core/err.test.ts
```diff
@@ -26,16 +26,15 @@ function beUnsuccessful(): Err {
...
   describe('Success helper', () => {
     it('Success return should be falsy', () => {
-      expect(!!beSuccessful()).is.false;
-      expect(beSuccessful() == null);
+      assert.isFalse(!!beSuccessful());
```
gtritchie, line 32: Wondering if `isFalse` and `isTrue` force a boolean context? Be nice if we don't need the `!!` to force boolean on these. Should be a quick thing to test.

gtritchie (changes requested):
Couple minor tweaks requested, otherwise definitely prefer this to the `expect` syntax.

src/node/desktop/test/core/err.test.ts
cm421: Oops, leftover from when I was comparing chai vs node asserts. Great catch

src/node/desktop/test/core/err.test.ts
cm421: isTrue and isFalse expect its arguments to be exactly true or false. There is an .isOk and .isNotOk that accepts truthy/falsey arguments, if that's acceptable

gtritchie: For readability of tests maybe we should have our own assert specifically for `Err`, perhaps create `src/node/desktop/test/test-utils.ts` with something like:

```
// Err semantics mean "ok" when falsely, "error" when truthy
function assertSuccess(e: Err) {
  assert.isNotOk(e);
}
function assertError(e: Err) {
  assert.isOk(e);
}
```

Then can just do `assertSuccess(beSuccessfull())`, etc?

kevinushey (approved):
LGTM!

gtritchie (approved):
LGTM!

PR #9453, opened 2021-06-04
enable async and handle offline by default

jeffvroom:
changed timeouts to 200 millis for each


jmcphers (commented):
I think this change needs to include the options JSON update as well.

Also: now that these are the defaults, can/should we revert the dev-specific settings in rserver.conf?

jmcphers (approved):
LGTM!

PR #9462, opened 2021-06-07
avoid using 'realpath' (not available by default)

kevinushey:
cc: @gtritchie


gtritchie:
Happened to notice this in the vscode scripts over the weekend:

https://github.com/microsoft/vscode/blob/b91fbed9e1633063fdf2d5ca46e8756003e899e7/scripts/code.sh#L5-L14

```
if [[ "$OSTYPE" == "darwin"* ]]; then
  realpath() { [[ $1 = /* ]] &amp;&amp; echo "$1" || echo "$PWD/${1#./}"; }
  ROOT=$(dirname "$(dirname "$(realpath "$0")")"
else
  ROOT=$(dirname "$(dirname "$(readlink -f $0)")")
```

PR #9469, opened 2021-06-08
Initial skeleton for Spectron-based integration tests

gtritchie:
### Intent

Bootstrap creation of integration tests using Spectron and Mocha/Chai.

### Approach

Added Spectron, create separate mocha configs for `unit` and `integration`, and a `yarn test-int` command that runs against the dev build (not the package build).

Wrote one proof-of-concept test that launches RStudio, confirms it displays a window, and exits.

### Automated Tests

Yes. Yes it is.

### QA Notes

Nothing to test here yet; currently must run manually on a dev environment.

### Checklist

### DIFF
--- a/src/node/desktop/test/integration/startup.int.ts
+++ b/src/node/desktop/test/integration/startup.int.ts
@@ -1,5 +1,5 @@
 /*
- * main.test.ts
+ * startup.int.ts
  *
  * Copyright (C) 2021 by RStudio, PBC
  *
--- a/src/node/desktop/test/integration/startup.int.ts
+++ b/src/node/desktop/test/integration/startup.int.ts
@@ -39,7 +39,7 @@ function getMain(): string {
 }
 
 describe('Startup and Exit', function () {
-  this.timeout(10000);
+  this.timeout(20000);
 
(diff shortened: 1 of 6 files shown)


kevinushey (approved):
LGTM! Pretty darn cool that it was this easy to get integration tests set up.

src/node/desktop/test/integration/startup.int.ts
```diff
@@ -0,0 +1,62 @@
...
+
+// Find path to RStudio entrypoint
+function getMain(): string {
+  return path.join(__dirname, '../../dist/main/app.js');
+}
+
+describe('Startup and Exit', function () {
+  this.timeout(10000);
+
+  beforeEach(function () {
```
kevinushey, line 44: Does this imply that a new Electron instance is created and torn down for every test? (If so I'm guessing we'll want to tweak this in the future?)

gtritchie: Yes, we can decide how we want to chunk this up (e.g. could have all tests in a single test source file run against a single instance, etc).

Just getting something in there to build on. :-)

Issue #9473, opened 2021-06-08
unit test failure re: usage of sprintf for themes


kevinushey:
From a recent build log:

```
[2021-06-08T21:17:56.630Z] Warning (test-themes.R:1416:4): parseTmTheme handles incorrect input

[2021-06-08T21:17:56.630Z] one argument not used by format 'error parsing attribute name [68]'

[2021-06-08T21:17:56.630Z] Backtrace:

[2021-06-08T21:17:56.630Z]  1. testthat::expect_error(...) test-themes.R:1416:3

[2021-06-08T21:17:56.630Z]  3. base::sprintf("error parsing attribute name [68]", inputFileLocation)

[2021-06-08T21:17:56.630Z] 

[2021-06-08T21:17:56.630Z] Warning (test-themes.R:1422:4): parseTmTheme handles incorrect input

[2021-06-08T21:17:56.630Z] one argument not used by format 'Opening and ending tag mismatch: string line 223 and notstring [76]'

[2021-06-08T21:17:56.630Z] Backtrace:

[2021-06-08T21:17:56.630Z]  1. testthat::expect_error(...) test-themes.R:1422:3

[2021-06-08T21:17:56.630Z]  3. base::sprintf(...)

[2021-06-08T21:17:56.630Z] 

[2021-06-08T21:17:56.630Z] Warning (test-themes.R:1428:4): parseTmTheme handles incorrect input

[2021-06-08T21:17:56.630Z] one argument not used by format 'StartTag: invalid element name [68]'

[2021-06-08T21:17:56.630Z] Backtrace:

[2021-06-08T21:17:56.630Z]  1. testthat::expect_error(...) test-themes.R:1428:3

[2021-06-08T21:17:56.630Z]  3. base::sprintf("StartTag: invalid element name [68]", inputFileLocation)

[2021-06-08T21:17:56.630Z] 

[2021-06-08T21:17:56.630Z] Warning (test-themes.R:1434:4): parseTmTheme handles incorrect input

[2021-06-08T21:17:56.630Z] one argument not used by format 'StartTag: invalid element name [68]'

[2021-06-08T21:17:56.630Z] Backtrace:

[2021-06-08T21:17:56.630Z]  1. testthat::expect_error(...) test-themes.R:1434:3

[2021-06-08T21:17:56.630Z]  3. base::sprintf("StartTag: invalid element name [68]", inputFileLocation)
```

[...]

PR #9474, opened 2021-06-08
remove un-needed usages of sprintf

kevinushey:
Closes https://github.com/rstudio/rstudio/issues/9473.


kevinushey:
Tests pass locally so I think this is safe to merge.

PR #9481, opened 2021-06-09
don't allow repl_python banner to enter notebook output

jjallaire:
@kevinushey I think this is the right approach to suppressing the banner, LMK what you think.

Perhaps as annoying is that a bunch of transient objects get printed in an "R Console" output block:

Screen Shot 2021-06-09 at 10 04 07 AM

These do not get displayed when executed in JupyterLab:

Screen Shot 2021-06-09 at 10 05 52 AM

We need to figure out the right logic for suppressing that output. It seems as if JupyterLab can and does print multiple set of output, but they have some criteria that allows them to exclude that output.

### DIFF
--- a/src/cpp/session/modules/rmarkdown/NotebookQueue.cpp
+++ b/src/cpp/session/modules/rmarkdown/NotebookQueue.cpp
@@ -32,6 +32,7 @@
 
 #include <r/RCntxtUtils.hpp>
 #include <r/RInterface.hpp>
+#include <r/ROptions.hpp>
 #include <r/RExec.hpp>
 #include <r/RJson.hpp>
 #include <r/RSexp.hpp>
@@ -69,6 +70,7 @@ class NotebookQueue : boost::noncopyable
 public:
    NotebookQueue() 
    {
+      prevReticulateReplQuiet_ = false;
       // launch a thread to process console input
       pInput_ = 
          boost::make_shared<core::thread::ThreadsafeQueue<std::string> >();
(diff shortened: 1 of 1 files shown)


kevinushey:
> Perhaps as annoying is that a bunch of transient objects get printed in an "R Console" output block:

This is a consequence of `reticulate`'s auto-printing of each expression's result. The situation here is a bit tricky because, at least as an R user, with a chunk containing for example:

`````
```{r}
1 + 1
2 + 2
3 + 3
```
`````

You'd expect the output for each line to be reflected in the console output. We use that same model for Python chunks as well. However, in Jupyter, only the last produced output is shown:

Screen Shot 2021-06-09 at 11 04 33 AM

And unfortunately I don't think Python has the notion of 'visible' versus 'invisible' expression results.

Ultimately, what you see in RStudio is the same as what you'd see if you ran that code line-by-line in a regular Python console; e.g. executing that matplotlib code does show and print intermediate matplotlib objects.

E.g. using the example at:

```
>>> t = np.arange(0.0, 2.0, 0.01)
>>> s = 1 + np.sin(2 * np.pi * t)
>>> fig, ax = plt.subplots()
>>> ax.plot(t, s)
[<matplotlib.lines.Line2D object at 0x112993890>]
>>> ax.set(xlabel='time (s)', ylabel='voltage (mV)',
...        title='About as simple as it gets, folks')
[<matplotlib.text.Text object at 0x1055b9c50>, <matplotlib.text.Text object at 0x1067aa250>, <matplotlib.text.Text object at 0x10667a850>]
>>> ax.grid()
```

There's probably some way for us to improve this, but at least what we're doing is consistent.

PR #9486, opened 2021-06-09
read system root certificates from keychain for SSL connections

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio-pro/issues/2651.

On macOS, the current `getKeychainCertificates()` implementation unfortunately does not read system root certificates, which implies that SSL connections (unless we get lucky and OpenSSL finds a `certs.pem` file) will fail to connect.

### Approach

Explicitly include the root certificate keychain when searching for certificates, and disable the default OpenSSL `cert.pem` lookup on Windows + macOS (since it will point to a Homebrew-specific location that is not useful on these platforms).

### Automated Tests

Not included.

### QA Notes

Test that RDP -> RSP connections work on macOS.

### Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,5 @@
 RStudio 1.4 "Trampled Dandelion"
 
 * Removed the breaking change introduced in Juliet Rose that changed the behavior of the X-Forwarded-Proto header when RSW is behind a proxy server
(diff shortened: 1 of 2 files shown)


jmcphers (approved):
Needs a NEWS entry (@jeffvroom already added a Trampled Dandelion NEWS file)

Issue #9489, opened 2021-06-13
Python path in project file should be aliased / project-relative if possible


kevinushey:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.4.1717
    OS Version      : All
    R Version       : All

### Steps to reproduce the problem

Create an RStudio project. Then, create a local virtual environment -- for example, with something like `python -m venv .venv`. Then, via the RStudio project options Python tab, try to select that local environment.

### Describe the problem in detail

The dialog for selecting Python fails to find that local environment.

### Describe the behavior you expected

We should find local virtual environments within the project if possible.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
As part of this, we should also ensure that project-local Python references are stored as relative paths within the .Rproj file.

astayleraz:
This change is causeing failures due to the locators being changed. Example failure: https://build.rstudioservices.com/job/IDE/job/qa-opensource-automation/640/testReport/rstudio_server_pro.tests.GlobalPrefs.test_GlobalPrefGeneral/TestGlobalPrefGeneral/test_global_prefs_general_working_dir/ Its showing the locator as rstudio_tbb_clear_button_default_working_dir instead of #rstudio_tbb_button_default_working_dir 
<img width="1503" alt="Screen Shot 2021-06-17 at 3 26 48 PM" src="https://user-images.githubusercontent.com/1482677/122480495-cd7a3080-cf81-11eb-8eb8-dacc527a9b1a.png">

rich-rstudio:
verified on 2021-08.0 Build 305-1

PR #9492, opened 2021-06-14
various enhancements to python integration

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/9489.

This PR brings some enhancements to RStudio's Python integration, including:

- Projects containing a Python environment within the `.venv` directory will be automatically activated,
- Some minor UI enhancements to our Python selection dialog.

Screen Shot 2021-06-14 at 4 32 37 PM

Hover text:

Screen Shot 2021-06-14 at 4 32 42 PM

### Approach

Relatively straight-forward implementation.

### Automated Tests

TBA

### QA Notes

- In the Python selection dialog, double-clicking on a menu item suffices to select it (rather than just clicking the Select button),
- The 'Clear' button in the Python preferences pane resets the Python interpreter,
- If a project contains a Python environment in the `.venv` folder, then that version of Python is automatically placed on the PATH

(feel free to ask for more clarification if needed!)

### Checklist

### DIFF
--- a/src/cpp/session/modules/SessionPythonEnvironments.cpp
+++ b/src/cpp/session/modules/SessionPythonEnvironments.cpp
@@ -53,32 +53,31 @@ void onPrefsChanged(const std::string& /* layerName */,
    }
 }
 
-void onDeferredInit(bool)
-{
-   if (!projects::projectContext().hasProject())
-      return;
-   
-   Error error = r::exec::RFunction(".rs.python.initialize")
-         .addUtf8Param(projects::projectContext().directory())
-         .call();
-   
-   if (error)
-      LOG_ERROR(error);
-}
-
 } // end anonymous namespace
 
 Error initialize()
 {
    using namespace module_context;
    
-   events().onDeferredInit.connect(onDeferredInit);
-   
    prefs::userPrefs().onChanged.connect(onPrefsChanged);
 
    ExecBlock initBlock;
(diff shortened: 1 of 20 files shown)


jjallaire:
How about "Automatically activate project-level Python environments" ?

src/cpp/session/modules/SessionPythonEnvironments.cpp
```diff
@@ -52,12 +53,27 @@ void onPrefsChanged(const std::string& /* layerName */,
    }
 }
 
+void onDeferredInit(bool)
```
jjallaire, line 56: Will this be in time to make sure it gets onto the `PATH` of terminals that are initialized at or near to startup?

jjallaire (approved):

src/cpp/session/modules/SessionPythonEnvironments.cpp
gtritchie: I _think_ it will be OK, previous terminals don't rehydrate until displayed. Definitely worth testing.

gtritchie (approved):
LGTM! Did a quick a11y eval and no problems.

kevinushey:
One reason why we need to delay (at least a bit): user preferences aren't available until after other RStudio machinery has finished initialization. I'm going to see if there's a better place than deferred init.

PR #9493, opened 2021-06-15
remove spectron

gtritchie:
### Intent

Spectron is essentially unusable on Electron 10+ (current Electron version is 13). It uses the deprecated remote module, and requires configuring the app in a non-recommended configuration (no context isolation).

See https://github.com/electron-userland/spectron/issues/693 and https://github.com/electron-userland/spectron/issues/896

### Approach

Remove Spectron.

### Automated Tests

😭 

### QA Notes

We'll have to investigate other approaches. Right now we'll fall back to focusing on unit tests for the Electron-specific code, and our existing browser-based testing.

### Checklist


astayleraz (approved):

nowshed-imran:
Playwright has been recommended as replacement in you tagged issues.

https://github.com/microsoft/playwright

gtritchie:
Glanced at it and it seemed to be more for traditional browser-based testing (of the single-page app itself). We have that already with the server version of RStudio using Selenium; looking for something that can examine state of Electron-specific aspects such as native menus, dialogs, command-line handling, etc. (as well as seeing and manipulating the web page loaded into Electron). Haven't looked closely yet, so maybe I missed it!

vladimiry:
> Glanced at it and it seemed to be more for traditional browser-based testing (of the single-page app itself).

It's not exactly like this, see https://playwright.dev/docs/api/class-electron. So you can manipulate the main process code too, not just pure web. The point I'd like to highlight though is that when you call `evaluate` on an electron app created by `playwright` the serialization takes place. So if you want to use something inside the `evaluate` callback you have to pass it as serializable arguments.

gtritchie:
Will take a deeper look. Thanks!

PR #9505, opened 2021-06-18
automatically activate .condaenv environments; fixes for reticulate binding to Anaconda

kevinushey:
### Intent

This PR contains two main changes:

1. If the project contains a Conda environment within the `.condaenv` directory, RStudio will now automatically activate that environment on startup.

2. Add some extra glue to ensure that `reticulate` can bind to Anaconda installations on macOS. See https://github.com/ContinuumIO/anaconda-issues/issues/12478 for more details.

### Approach

The Anaconda glue is the more interesting part of the PR. Currently, Anaconda installations rely on the process rpath when discovering libraries. This unfortunately doesn't work for us, since we attempt to load Python as a library (and so don't have the rpath entries set that Anaconda is expecting).

The workaround is to set `DYLD_FALLBACK_LIBRARY_PATH` to include the Anaconda library path, but this is challenging for a couple reasons:

1. The `DYLD_FALLBACK_LIBRARY_PATH` cannot be changed after a process has already been launched. (You can change the value of the environment variable, but it will not affect future library load attempts in the current process.)
2. When RStudio launches the `rsession` process, we have no idea if the user plans to use Python, or even Anaconda Python. We also don't know where Anaconda might be installed.

The solution is to instead have RStudio place a special fallback library path in the `DYLD_FALLBACK_LIBRARY_PATH`, and have R initialize the library path when necessary. See https://github.com/rstudio/reticulate/commit/08f145b0d2f13c8f8c242ea788359406d139467d for an example.

### Automated Tests

N/A

### QA Notes

Verify, on macOS, that the following works:

1. Install Anaconda: https://www.anaconda.com/products/individual#macos
2. Install the development version of `reticulate`: `remotes::install_github("rstudio/reticulate")`
3. Try to load that environment: `reticulate::use_python("~/opt/anaconda3/bin/python", required = TRUE); reticulate::py_config()`

If all is well, you should see `reticulate::py_config()` print without issue.

### Checklist

### DIFF
--- a/src/cpp/session/SessionMain.cpp
+++ b/src/cpp/session/SessionMain.cpp
@@ -243,15 +243,7 @@ namespace {
 
 std::string s_fallbackLibraryPath;
 
-void onExit()
-{
-   if (!s_fallbackLibraryPath.empty())
-   {
-      ::remove(s_fallbackLibraryPath.c_str());
-   }
-}
-
-}
+} // end anonymous namespace
 
 bool disableExecuteRprofile()
 {
@@ -473,6 +465,7 @@ void exitEarly(int status)
 {
    stopMonitorWorkerThread();
    FileLock::cleanUp();
+   FilePath(s_fallbackLibraryPath).removeIfExists();
    ::exit(status);
 }
 
@@ -1768,9 +1761,6 @@ int main(int argc, char * const argv[])
    {
       // save fallback library path
       s_fallbackLibraryPath = core::system::getenv("RSTUDIO_FALLBACK_LIBRARY_PATH");
-      
-      // set up exit hooks
-      ::atexit(onExit);
       
       // sleep on startup if requested (mainly for debugging)
       std::string sleepOnStartup = core::system::getenv("RSTUDIO_SESSION_SLEEP_ON_STARTUP");
--- a/src/cpp/session/SessionMain.cpp
+++ b/src/cpp/session/SessionMain.cpp
@@ -1903,7 +1903,7 @@ int main (int argc, char * const argv[])
       // Mirror the R getOptions("width") value in an environment variable
       core::system::setenv("RSTUDIO_CONSOLE_WIDTH",
                safe_convert::numberToString(rstudio::r::options::kDefaultWidth));
-
+ 
       // set the rstudio user identity environment variable (can differ from
       // username in debug configurations). this is provided so that
       // rpostback knows what local stream to connect back to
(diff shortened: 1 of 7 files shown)


src/cpp/session/SessionMain.cpp
```diff
@@ -239,6 +239,20 @@ using namespace rsession::client_events;
 namespace rstudio {
 namespace session {
 
+namespace {
+
+std::string s_fallbackLibraryPath;
+
+void onExit()
```
jjallaire, line 246: Can the value we set in our executable actually persist after we exit? (i.e. do we actually need this code)

src/cpp/session/modules/SessionPythonEnvironments.R
```diff
@@ -43,11 +43,92 @@
...
+   # if this project has a local interpreter, use it
+   if (!is.null(projectDir))
+   {
+      projectPython <- .rs.python.projectInterpreterPath(projectDir)
+      if (file.exists(projectPython))
+         return(projectPython)
+   }
+   
+   # check version of Python configured by user
+   prefsPython <- .rs.readUiPref("python_path")
```
jjallaire, line 100: Quick reminder (if it isn't already resolved) to make sure that we persist python_path that is a subdirectory of the project as a relative path.

jjallaire:
Does the IDE prompt for reticulate installation when the user clicks the "Select Intepreter" button?

We should keep a note or file a separate issue that RStudio should require the CRAN version of reticulate that has the requisite changes.

jjallaire:
This probably needs some updating to reflect this subsequent PR: https://github.com/rstudio/rstudio/pull/9510

I noticed that VS code will actually run the conda activation script in the terminal when configured to bind to a project level conda environment. How hard would it be for us to do this? (and/or is it even necessary)

kevinushey:
> I noticed that VS code will actually run the conda activation script in the terminal when configured to bind to a project level conda environment. How hard would it be for us to do this? (and/or is it even necessary)

We can't really do this in the main session (or, doing it is quite challenging) since this implies running a command that modifies the process's current environment. That can only work if conda injects its own shims into the shell, so that running `conda activate` really runs a `__conda_activate` shell function that modifies the current environment. (It also implies that `conda activate` will fail if the user hasn't already called `conda init` or something similar.)

We might be able to do this in our Terminal instances, but we again run into the potential `conda init` fiasco.

Fortunately, we can basically do the same by manually setting the PATH and setting some environment variables (e.g. `CONDA_PREFIX`) to tell Conda to target a particular environment.

src/cpp/session/modules/SessionPythonEnvironments.R
kevinushey: Should be fixed up now -- thanks for spotting that.

PR #9510, opened 2021-06-20
use `env` as the project python environment directory rather than `.venv`

jjallaire:
I've been hashing out the project-specific virtual envs a bit and thinking more about directory naming. I noticed that VS Code will actually auto detect/activate any venv contained in a sub-directory with a pyvenv.cfg file. That's great flexibility, but at the same time I think we want to enforce some standards so that the IDE can depend on a predictable path (e.g. we need to screen files in the venv out of find in files results -- we  do that now for "renv" but need the analogous behavior for python envs; we also need to augment .gitignore with the venv directory).

I noted that the official python venv documentation (https://packaging.python.org/guides/installing-using-pip-and-virtual-environments/#creating-a-virtual-environment) use the convention of 'env' for project-specific environments.

I also noted that this article from the Software Carpentries (https://carpentries-incubator.github.io/introduction-to-conda-for-data-scientists/02-working-with-environments/index.html#conda-environment-sub-directory-naming-convention) recommends env for compatibility with common venv practice as well as to get the automatic .gitignore from the standard GitHub python .gitignore file. This article on using conda environments (by the same author) is highly ranked on google and recommends the same thing: https://towardsdatascience.com/managing-project-specific-environments-with-conda-406365a539ab

We could adopt VS Code's approach and be more aggressive about auto-detection (as well as make our find in files and .gitignore code similarly flexible). I think though that the community is better served by the adoption of a common, sensible convention that is based on official documentaiton/recommendations (it's nice that we can point users to the above mentioned articles and know that if they follow them precisely the right thing will happen in the IDE).

We can certainly make this more flexible if we get feedback that this is desirable, but I vote that we start w/ a simpler approach.

### DIFF
--- a/src/cpp/session/SessionModuleContext.cpp
+++ b/src/cpp/session/SessionModuleContext.cpp
@@ -2650,32 +2650,16 @@ bool isUserFile(const FilePath& filePath)
              return false;
       }
 
-      // if we are in a website project then screen the output dir
-      if (!module_context::websiteOutputDir().empty())
-      {
-         FilePath sitePath = projects::projectContext().buildTargetPath();
-         std::string siteRelative = filePath.getRelativePath(sitePath);
-         if (boost::algorithm::starts_with(
-                siteRelative, module_context::websiteOutputDir() + "/"))
-            return false;
-      }
-
-      // if we are in a quarto project then screen the output dir
-      auto quartoOutputDir = module_context::quartoConfig().project_output_dir;
-      if (!quartoOutputDir.empty())
-      {
-         FilePath projDir = projects::projectContext().directory();
-         std::string projRelative = filePath.getRelativePath(projDir);
-         if (boost::algorithm::starts_with(
-                projRelative, quartoOutputDir + "/"))
-            return false;
-      }
-
-      // screen our various virtual environment directories
-      std::vector<std::string> dirs({"packrat/", "renv/", "env/"});
+      // screen our various virtual environment directories + standard content ignores
       FilePath projPath = projects::projectContext().directory();
+      std::vector<std::string> dirs({"packrat/", "renv/"});
+      const std::vector<FilePath>& ignoreDirs = module_context::ignoreContentDirs();
+      std::transform(ignoreDirs.begin(), ignoreDirs.end(), std::back_inserter(dirs),
+                     [&projPath](const FilePath& ignorePath) {
+         return ignorePath.getRelativePath(projPath) + "/";
+      });
       std::string pkgRelative = filePath.getRelativePath(projPath);
-      for (auto dir : dirs) {
+      for (const auto& dir : dirs) {
          if (boost::algorithm::starts_with(pkgRelative, dir))
             return false;
       }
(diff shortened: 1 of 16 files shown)


jjallaire:
Note that this PR also adds `/renv/python/`  and `/.quarto/` to the standard set of directories skipped for find in files results.

src/cpp/session/modules/SessionFind.cpp
```diff
@@ -909,11 +909,15 @@ class GrepOperation : public boost::enable_shared_from_this<GrepOperation>
    bool shouldSkipFile(std::string file)
    {
       return (file.find("/.Rproj.user/") != std::string::npos ||
+              file.find("/.quarto/") != std::string::npos ||
               file.find("/.git/") != std::string::npos ||
               file.find("/.svn/") != std::string::npos ||
               file.find("/packrat/lib/") != std::string::npos ||
               file.find("/packrat/src/") != std::string::npos ||
+              file.find("/env/") != std::string::npos ||
```
kevinushey, line 917: My main worry here is that someone might reasonably have things in an `/env/` folder that isn't a virtual environment (and so would be surprised to see RStudio failing to search that folder).

kevinushey:
LGTM! I agree just using `env` for now (and becoming more flexible later if we know we need it) is fine.

src/cpp/session/SessionModuleContext.cpp
```diff
@@ -2660,11 +2660,25 @@ bool isUserFile(const FilePath& filePath)
...
+            return false;
+      }
+
+      // screen our various virtual environment directories
+      std::vector<std::string> dirs({"packrat/", "renv/", "env/"});
       FilePath projPath = projects::projectContext().directory();
       std::string pkgRelative = filePath.getRelativePath(projPath);
-      if (boost::algorithm::starts_with(pkgRelative, "packrat/"))
-         return false;
+      for (auto dir : dirs) {
```
kevinushey, line 2678:
```suggestion
      for (const auto& dir : dirs) {
```

src/cpp/session/modules/SessionCodeSearch.cpp
```diff
@@ -110,12 +111,32 @@ bool isWithinIgnoredDirectory(const FilePath& filePath)
...
+      // quarto projects
+      if (parentName == quartoProjectOutputDir)
+         return true;
       
       // packrat
       if (parentName == "packrat" &&
          parentPath.completeChildPath("packrat.lock").exists())
          return true;
       
+      // packrat
```
kevinushey, line 124: I think this is duplicated with the above?

src/cpp/session/modules/SessionCodeSearch.cpp
```diff
@@ -110,12 +111,32 @@ bool isWithinIgnoredDirectory(const FilePath& filePath)
...
+      if (parentName == "packrat" &&
+         parentPath.completeChildPath("packrat.lock").exists())
+         return true;
+
+      // renv
+      if (parentName == "renv" &&
+         parentPath.completeChildPath("settings.dcf").exists())
```
kevinushey, line 131: It's possible for a "bare" `renv` environment to not have any project settings defined, so I think this is a better check:

```suggestion
         parentPath.completeChildPath("activate.R").exists())
```

jjallaire:
@kevinushey I just pushed a change that makes the venv discovery work for any directory. We also now ignores content (for search, find in files, etc.) in any project root level directory that appears to be a venv.

kevinushey (approved):
LGTM!

PR #9514, opened 2021-06-21
implement session startup (Posix-only)

gtritchie:
### Intent

Be able to load the rsession and display the real UI, including menu population and some callbacks (essentially equivalent to where the prototype was but with somewhat less hardcoding).

Currently Posix (Mac/Linux) only, and only for dev builds; some additional Win32-specific work needed to get that standing up again, as well as work specific to package builds. 

### Approach

Location of `src/cpp` build output (on a dev machine) must be specified in `RSTUDIO_CPP_BUILD_OUTPUT` environment variable. For example, `RSTUDIO_CPP_BUILD_OUTPUT=/Users/gary/rstudio/src/cpp/build`.

R detection (Posix only) is partially implemented (Kevin did some work on this in the prototype and I have pulled it over and tweaked it a bit. It should handle common cases).

First use of node.js events (replacing the Qt signal/slots used in the existing desktop code).

The `--run-diagnostics` command-line switch is now supported, and some of the output is hooked up; you can use `yarn start-diag` to see it.

Note that quite a few changes related to this work went in separately via https://github.com/rstudio/rstudio/pull/9511.

### Automated Tests

This is getting into integration testing territory (which we don't currently have a mechanism for), but additional unit testing will also be added; wanted to get this in ASAP to unblock others.

### QA Notes

Still only building in a dev environment.

### Checklist

### DIFF
--- a/src/node/desktop/package.json
+++ b/src/node/desktop/package.json
@@ -28,7 +28,9 @@
     "ts-node": "^10.0.0",
     "typescript": "^4.3.2"
   },
-  "dependencies": {},
+  "dependencies": {
+    "uuid": "^8.3.2"
+  },
   "scripts": {
     "build": "tsc",
     "watch": "tsc -w",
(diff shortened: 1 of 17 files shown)


cm421 (approved):
LGTM!

PR #9520, opened 2021-06-22
resolve TODOs: generate launcher token and port for Electron desktop

gtritchie:
### Intent

Resolve TODOs for generating launcher token and port (currently hardcoded in the Electron app).

### Approach

- Pull in `uuid` and `crc` packages, add to NOTICE
- Hook up in code, write unit tests
- Other code changes not directly related:
  - grouped imports
  -  added `--trace-warnings` to yarn `start` and `start-diag` commands
  - did `yarn upgrade` to pull in Electron v13.1.3

### Automated Tests

Added unit tests for this new code.

### QA Notes

Internal detail covered by unit tests.

### Checklist

### DIFF
--- a/src/node/desktop/src/main/session-launcher.ts
+++ b/src/node/desktop/src/main/session-launcher.ts
@@ -13,23 +13,25 @@
  *
  */
 
+import { app, dialog } from 'electron';
 import { spawn, ChildProcess } from 'child_process';
 
 import { FilePath } from '../core/file-path';
-import { ApplicationLaunch } from './application-launch';
+import { generateShortenedUuid } from '../core/system';
 import { Err } from '../core/err';
+import { getenv, logEnvVar, setenv } from '../core/environment';
+
+import { ApplicationLaunch } from './application-launch';
 import { appState } from './app-state';
 import { ActivationEvents } from './activation-overlay';
-import { app, dialog } from 'electron';
 import { EXIT_FAILURE } from './program-status';
-import { getenv, logEnvVar, setenv } from '../core/environment';
 import { MainWindow } from './main-window';
 import { PendingQuit } from './desktop-callback';
 
 export class SessionLauncher {
-  host = '';
-  port = '40810'; // TODO: see DesktopOptions.cpp portNumber() for how to randomly generate
-  launcherToken = '7F83A8BD'; // TODO: generate short uuid
+  host = '127.0.0.1';
+  private port?: string;
+  private launcherToken?: string;
   sessionProcess?: ChildProcess;
   mainWindow?: MainWindow;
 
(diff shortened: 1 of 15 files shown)


src/node/desktop/src/main/session-launcher.ts
```diff
@@ -191,4 +193,34 @@ export class SessionLauncher {
...
+      // rsession and it's children can use it
+      if (process.platform === 'win32') {
+        const localPeer = '\\\\.\\pipe\\' + this.port + '-rsession';
+        setenv('RS_LOCAL_PEER', localPeer);
+      }
+    }
+    return this.port;
+  }
+
+  newPortNumber(): string {
```
kevinushey, line 222: nit: the function is called `newPortNumber()`, but returns a string?

src/node/desktop/src/main/session-launcher.ts
```diff
@@ -191,4 +193,34 @@ export class SessionLauncher {
   closeAllSatellites(): void {
     console.log('CloseAllSatellites not implemented');
   }
+
+  getLauncherToken(): string {
+    if (!this.launcherToken) {
```
kevinushey, line 198: Do you think there's an appropriate place where we could generate this on initialization, rather than just generating on first use (and re-checking each time thereafter)?

src/node/desktop/src/main/session-launcher.ts
```diff
@@ -191,4 +193,34 @@ export class SessionLauncher {
...
+      this.launcherToken = generateShortenedUuid();
+    }
+    return this.launcherToken;
+  }
+
+  getPort(): string {
```
kevinushey, line 204: It feels a little odd that a function that looks like a plain 'getter' also sets `this.port` as a side-effect (and the associated environment variable).

kevinushey (commented):
LGTM, but I would advocate having a separate and explicit "initialize" stage where things like `launcherToken` and the port are set (if possible).

cm421 (approved):
LGTM!

src/node/desktop/src/main/session-launcher.ts
gtritchie: Yes, that's a quirk of me keeping the original function name (but earlier on switched to storing as a string, since that's how it gets used). I can rename.

src/node/desktop/src/main/session-launcher.ts
gtritchie: I was just following the existing code. Agree it's a bit odd.

src/node/desktop/src/main/session-launcher.ts
gtritchie: Agree, just doing a literal port here.

kevinushey: Gotcha. IMHO for cases like these, we can re-examine some of the old decisions and try to clean up the implementation where appropriate; we don't necessarily need a direct port. Does that make sense, or would you prefer tracking the old implementation as closely as possible?

gtritchie: Well, not a strictly literal port, have moved this out of the Options() class and put ownership into the session-launcher. I'll do some tidying up on this before I merge. Agree too many side-effects here!

gtritchie: @kevinushey No, I agree we should cleanup as we go along, within reason.

kevinushey (approved):

PR #9532, opened 2021-06-24
disable null external pointer checks for environment pane

kevinushey:
### Intent

RStudio normally screens for objects containing null external pointers when building the Environment pane, primarily to fix crashes that occurred when introspecting certain objects in older versions of R.

This has caused a fair amount of user confusion, and in addition this workaround no longer appears to be necessary with the newer versions of R / igraph.

Addresses https://github.com/rstudio/rstudio/issues/5546.

### Approach

We now no longer search for null external pointers by default, but allow users to opt-in via a user preference if necessary.

### Automated Tests

N/A

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/5546#issuecomment-567646350. Also confirm that you can set the "Check Null External Pointers" preference to toggle the behavior.

Screen Shot 2021-06-24 at 4 48 27 PM

### Checklist


jmcphers (approved):
This LGTM but I think we should rename the user-facing title of the preference since there's not enough context to tell what it does (examine them why? for what?). Would it make it too long to mention that this refers specifically to values in the Environment pane? Maybe something like this:

> Check values in the Environment pane for null external pointers"

PR #9567, opened 2021-07-06
Adds needsDebugLog() method for conditionally formatting LOG_DEBUG

jeffvroom:
For some of my debug logging changes, I'd like a way to avoid running some code when debug logging is not enabled.

I decided against adding a new trace level or the trace-feature-flags, at least for now, as I don't think the messages we need at this stage are noisy enough. 

Let me know if there's a better way to do this or if you can think of a better name. Thanks!

### DIFF
--- a/src/cpp/shared_core/Logger.cpp
+++ b/src/cpp/shared_core/Logger.cpp
@@ -580,6 +580,11 @@ bool hasStderrLogDestination()
    return hasLogDestination<StderrLogDestination>();
 }
 
+bool needsDebugLog()
+{
+   return logger().MaxLogLevel >= LogLevel::DEBUG;
+}
+
 } // namespace log
 } // namespace core
 } // namespace rstudio
--- a/src/cpp/shared_core/Logger.cpp
+++ b/src/cpp/shared_core/Logger.cpp
@@ -412,6 +412,11 @@ void logInfoMessage(const std::string& in_message, const std::string& in_section
    logger().writeMessageToDestinations(LogLevel::INFO, in_message, in_section, in_loggedFrom);
 }
 
+bool isLogLevel(LogLevel in_logLevel)
+{
+   return logger().MaxLogLevel >= in_logLevel;
+}
+
 void refreshAllLogDestinations(const log::RefreshParams& in_refreshParams)
 {
    Logger& log = logger();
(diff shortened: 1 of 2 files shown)


src/cpp/shared_core/include/shared_core/Logger.hpp
```diff
@@ -374,6 +374,8 @@ void removeLogDestination(const std::string& in_destinationId, const std::string
  */
 void removeReloadableLogDestinations();
 
+bool needsDebugLog();
```
MariaSemple, line 377: How do you feel about something like `bool isAtLeastLogLevel(LogLevel in_level);` so it can be generalized for other use cases in the future?

Nit: This file is currently sorted alphabetically, could you move this to keep that ordering?

src/cpp/shared_core/Logger.cpp
```diff
@@ -580,6 +580,11 @@ bool hasStderrLogDestination()
    return hasLogDestination<StderrLogDestination>();
 }
 
+bool needsDebugLog()
```
MariaSemple, line 583: Nit: This file should have the functions implemented in the same order they appear in the header file. Would you mind reordering these to that effect?

MariaSemple (changes requested):

src/cpp/shared_core/include/shared_core/Logger.hpp
jeffvroom: I couldn't think of a use case for the other log levels so I put the focus on readability. I'm happy to make the change, though if you prefer.  How about:

  if (log::needsLogLevel(log::LogLevel::DEBUG))

It seems a little more readable to me.  Thanks for catching those ordering problems. I'll fix those up for sure!

MariaSemple: `needsLogLevel` doesn't feel quite right to me since the logger doesn't need a log level. We could make it shorter (e.g. `isLogLevel`). In pseudo code, I would mentally read them this way:

`if logger needs log level debug, then <blah>`
vs.
`if logger is log level debug, then <blah>`

The second one feels slightly more correct to me, but the actual name is really a nit, so up to you.

jeffvroom: I like your explanation so I changed it to isLogLevel(LogLevel::DEBUG) 

I also tried to fix the ordering and removed an unused declaration for "cleanDelims". Thanks!

MariaSemple (approved):
LGTM!

PR #9568, opened 2021-07-07
call finishInterrupt only when session has really been interrupted

jeffvroom:
### Intent

Fixes the "click interrupt three times to get the abort dialog. " part of #8865 after it was broken with async rpc. 

### Approach

With this PR, only finish the interrupt when the console busy is actually set to false

### Automated Tests

TBD - I could build an automated test to trigger the abort dialog if anyone thinks it's worth the effort

### QA Notes

Test the two ways to block the R console - both interruptible and non-interruptible and press the "Interrupt" button. 

Interruptible ops like system("sleep 30") should just return

Non-interruptible ops like the hang() example in #8865 should display the "abort R" dialog on the 3rd click in 10 seconds. 

### Checklist

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/application/ApplicationInterrupt.java
+++ b/src/gwt/src/org/rstudio/studio/client/application/ApplicationInterrupt.java
@@ -24,6 +24,7 @@ import org.rstudio.core.client.widget.MessageDialog;
 import org.rstudio.core.client.widget.ProgressIndicator;
 import org.rstudio.core.client.widget.ProgressOperation;
 import org.rstudio.studio.client.application.events.EventBus;
+import org.rstudio.studio.client.workbench.views.console.events.ConsoleBusyEvent;
 import org.rstudio.studio.client.application.events.InterruptStatusEvent;
 import org.rstudio.studio.client.application.events.ReloadEvent;
 import org.rstudio.studio.client.application.model.ApplicationServerOperations;
@@ -43,7 +44,8 @@ import com.google.inject.Provider;
 import com.google.inject.Singleton;
 
 @Singleton
-public class ApplicationInterrupt
+public class ApplicationInterrupt implements ConsoleBusyEvent.Handler
+
 {
    public interface Binder extends CommandBinder<Commands, ApplicationInterrupt> {}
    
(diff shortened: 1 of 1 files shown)


src/gwt/src/org/rstudio/studio/client/application/ApplicationInterrupt.java
```diff
@@ -186,6 +202,19 @@ public void onTerminateR()
...
+   @Override
+   public void onConsoleBusy(ConsoleBusyEvent event)
+   {
+      isBusy_ = event.isBusy();
+
+      if (!isBusy_ && pendingInterrupt_)
+      {
+         eventBus_.fireEvent(new InterruptStatusEvent(
+               InterruptStatusEvent.INTERRUPT_COMPLETED));
+         finishInterrupt(pendingInterruptHandler_);
```
kevinushey, line 215: Should we reset `pendingInterrupt_` + `pendingInterruptHandler_` here (after invoking `finishInterrupt()`)?

kevinushey (commented):
LGTM, but do we need to reset the "pending interrupt" state after we've handled a pending interrupt?

kevinushey:
Does this also work in cases where a REPL in R is running? E.g. something like:

```
for (i in 1:10) {
    response <- tryCatch(
        readline("Enter a number: "),
        interrupt = function(e) { writeLines("Caught interrupt.") }
    )
}
```

What happens if you run that code and try to repeatedly interrupt it?

jeffvroom:
Thanks, I'm super happy to have more test cases to try out. When I run this, I don't see the stop button available so I can't interrupt R when it's waiting for input. I see that the last event sent to the IDE is a busy=true so I'm guessing the IDE sees the console prompt event and turns off the stop button. Is that what we want or should you be able to interrupt it?  

Doesn't seem to be affected by these changes.

kevinushey:
You should still be able to interrupt by pressing Escape or Ctrl + C, though -- does that work as expected? (I'm wondering if you can interrupt 10 times to exit that loop, or if you get the "R is not responding" warning on the 3rd interrupt attempt. Although, I could see arguments for either behavior being correct...)

jeffvroom:
It does still let you interrupt 10 times to exit the loop. Took me a while to figure out why and a good excuse to figure out GWT's super dev mode :)  It's because interrupting it via a keypress doesn't use ApplicationInterrupt.java - it instead is called from Shell.java where there's just no logic to show the terminate dialog. It seems ok as it is to me, but I'm happy to make any changes to that other code.

kevinushey:
Ah! Got it, that makes sense. Thanks for the extra investigation -- in that case, this LGTM!

PR #9589, opened 2021-07-09
Feature/electron desktop options

cm421:
Quickly thrown together implementation of DesktopOptions so we at least have a place to put things like zoomLevel and WindowSize. Uses Electron-Store behind the scenes for now until we decide to swap it out with something else


src/node/desktop/test/unit/main/desktop-options.test.ts
```diff
@@ -0,0 +1,68 @@
...
+ */
+
+import { describe } from 'mocha';
+import { assert } from 'chai';
+
+import { TestDesktopOptions, DesktopOptionsImpl, kDesktopOptionDefaults, clearOptionsSingleton } from '../../../src/main/desktop-options';
+import { FilePath } from '../../../src/core/file-path';
+import { Err, isSuccessful } from '../../../src/core/err';
+
+const kTestingConfigFilename = process.cwd() + 'DesktopOptionsTesting';
```
gtritchie, line 22: We should write any temporary files created during tests somewhere in the temp path, and not use a fully hardcoded name (so any tests run in parallel don't fight each other; just a theoretical problem for now). I do stuff like that in the file-path tests, look for `os.tmpdir()` and `randomString()`. Maybe create a help for that in `unit-utils.ts`.

gtritchie (changes requested):
Looking good, some suggestions.

src/node/desktop/package.json
```diff
@@ -21,11 +22,13 @@
     "electron": "^13.0.1",
     "electron-mocha": "^10.0.0",
     "electron-packager": "^15.2.0",
+    "electron-store": "^8.0.0",
     "eslint": "^7.27.0",
     "eslint-config-standard": "^16.0.3",
     "eslint-plugin-import": "^2.23.3",
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-promise": "^5.1.0",
+    "ini": "^2.0.0",
```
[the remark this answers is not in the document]
cm421: Ah. That one was one I tried but decided against using. I'll clear it out

src/node/desktop/src/core/string-utils.ts
```diff
@@ -15,4 +15,12 @@
 
 export function removeDups(strings: string[]): string[] {
   return [...new Set(strings)];
-}
\ No newline at end of file
+}
+
+export function jsonLiteralEscape(str: string): string {
```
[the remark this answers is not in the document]
cm421: Well look at that. How handy! I'll probably remove this then, since it doesn't provide anything useful, and I didn't end up needing to use it

src/node/desktop/src/main/desktop-options.ts
```diff
@@ -0,0 +1,92 @@
...
+  options = null;
+}
+ 
+/**
+ * Desktop Options class for storing/restoring user desktop options.
+ * 
+ * Exported for unit testing only, use the DesktopOptions()
+ */
+export class DesktopOptionsImpl {
+   private _config = new Store({defaults: kDesktopOptionDefaults});
```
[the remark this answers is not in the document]
cm421: The C++ in me likes having hints like this about what kind of variable we're using, but if it's not a very typescript-y practice I can try and get over that.

src/node/desktop/src/main/desktop-options.ts
```diff
@@ -0,0 +1,92 @@
...
+ 
+   // Filename exposed for unit testing
+   constructor(filename = '') {
+     if (filename.length != 0) {
+       this._config = new Store({defaults: kDesktopOptionDefaults, cwd: filename});
+     }
+   }
+ 
+   public setZoomLevel(zoom: number): void {
+     this._config.set('zoomLevel', zoom);
```
kevinushey, line 73: Do we want any sectioning / notion of nested properties here? E.g.

```
this._config.set('View.zoomLevel', zoom);
```

So that related properties can be part of the same "bag".

Also, do we have any strong feelings re: what kind of casing or style we use for option names here?

src/node/desktop/src/main/desktop-options.ts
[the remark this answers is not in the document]
gtritchie: Totally get that. There's lots of code out there that does it both ways, and I don't have strong feelings other than we should be consistent. For now I started following (more or less) the Google style guide:

https://google.github.io/styleguide/tsguide.html

(also found this prettier variation of the same): https://ts.dev/style/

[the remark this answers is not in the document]
cm421: Great resource! I'll take out the `_` since that's easier than changing all of your code to match my one variable :)

src/node/desktop/src/main/desktop-options.ts
cm421: I like the idea of sections, and the module I'm using does support them. I'm planning on adding them in future work, generally mirroring the Qt sections unless someone has strong feelings otherwise.

I originally was using `ZoomLevel`, but changed it to match what the Qt code is doing (`zoomLevel`). I'm happy to go back to using a capital letter at the front though!

gtritchie: FWIW I don't have strong feelings either way. We don't have a lot of settings in the desktop.

cm421: I'll leave it for now, if that's okay, and probably change it when I add sections and constants to hold all the names

src/node/desktop/test/unit/main/desktop-options.test.ts
cm421: Added some to create either random directories or files, let me know if they're close to what you had in mind

gtritchie (approved):
LGTM!

PR #9601, opened 2021-07-13
Change DB File Scheme

MariaSemple:
### Intent

This addresses concerns that were raised on how to manage Database compatibility between versions. In particular this chart should hold true (with regards to what is supported):

                     | Dailies | Previews | Releases |
    -----------------+---------+----------+----------+
     Upgrade From    | Yes     | Yes      | Yes      |
     Upgrade to      | Yes     | Yes      | Yes      |
     Downgrade to    | No      | No       | Yes      |
     Downgrade from  | Yes     | Yes      | Yes      |

### Approach

* There is a single `CreateTables` file (one for postgres, one for sqlite) for each release. 
  * This file is used for first installation. It may be changed every release.
* There is an `AlterTables` file (one for postgres, one for sqlite) for each release.
  * Once a version is released, this file may not be modified.
  * Actions in this file should be idempotent to facilitate the `daily 1 > daily x > preview > release` upgrade path.
  * Changes to this file require an update the `DATE` portion of the file.

### Automated Tests

All server_core unit tests pass.

### QA Notes

* It's important to check both new installations and upgrades here.
* In the future, there will be other changes to the `AlterTables` file. Once that happens, it will be important to verify that this daily updates smoothly to the new daily.
* Both Postgres & sqlite dbs will need to be tested.
* We should also test that downgrading to JR or WB doesn't break anything.

### Checklist


src/cpp/core/Database.cpp
```diff
@@ -702,14 +704,134 @@ void Transaction::rollback()
...
+bool SchemaVersion::operator<(const SchemaVersion& other) const
+{
+   if (*this == other)
+      return false;
+
+   if (isEmpty())
+      return true;
+
+   if (other.isEmpty())
+      return false;
```
melissa-barca, line 759: What is the point of `isEmpty` since it always returns false?

src/cpp/core/Database.cpp
```diff
@@ -874,50 +1084,43 @@ Error SchemaUpdater::updateToVersion(const std::string& maxVersion)
...
+   for (const std::pair<SchemaVersion, FilePath>& migrationFile : files)
    {
-      // if the version has already been applied (database version is newer or same)
-      // then skip this particular migration
-      if (migrationFile.getStem() <= currentVersion)
-         continue;
+      // // if the version has already been applied (database version is newer or same)
+      // // then skip this particular migration
+      // if (migrationFile.getStem() <= currentVersion)
+      //    continue;
```
melissa-barca, line 1106: Should this commented out code and the section below be removed?
```suggestion

```

src/cpp/core/Database.cpp
```diff
@@ -874,50 +1084,43 @@ Error SchemaUpdater::updateToVersion(const std::string& maxVersion)
...
 
-      // if the migration file version is higher than the max specified version, we're done
-      if (migrationFile.getStem() > maxVersion)
-         break;
+      // // if the migration file version is higher than the max specified version, we're done
+      // if (migrationFile.getStem() > maxVersion)
+      //    break;
```
melissa-barca, line 1109:
```suggestion
```

src/cpp/server/db/make-schema.sh
```diff
@@ -3,13 +3,15 @@
 if [ -z "$1" ]
 then
    echo "Must specify at least one argument."
-   echo "Invocation: ./make-schema.sh name [sql]"
-   echo "Ex: ./make-schema.sh AddMyTable"
+   echo "Invocation: ./make-schema.sh [upgradeVersion] [sql]"
+   echo "Ex: ./make-schema.sh GO_PT sql"
```
MariaSemple, line 7:
```suggestion
   echo "Ex: ./make-schema.sh \"Ghost Orchid\" sql"
```

jeffvroom (approved):
This looks really awesome! Nice job. You did this fast and it looks really solid. I only have a couple of minor suggestions that are totally optional. 

In SchemaVersion, maybe the Flower field should be renamed to ReleaseName?  Not that we will ever give up the flower naming scheme :) 

In the update() method, I think it would be helpful to have some LOG_INFO messages stating the from and to version and an info with the SQL we are about to apply. Especially for us during development, and possibly for customers if things get messy, it will be important to know statements were run, which one caused the error. 

In a number of places, mostly in code you didn't change, we just "return error" when something goes wrong. But the database error won't include the SQL that caused the error so all we'll see to support it is something like "Invalid SQL statement" and will have to search and guess among the few possibilities. I don't want to slow you down and overly complicate this PR though so we can fix that later.

src/cpp/core/Database.cpp
```diff
@@ -702,14 +704,134 @@ void Transaction::rollback()
...
+   }
+   return *this;
+}
+
+bool SchemaVersion::operator<(const SchemaVersion& other) const
+{
+   if (*this == other)
+      return false;
+
+   if (isEmpty())
```
MariaSemple, line 755:
```suggestion
   if (isEmpty() && !other.isEmpty())
```

melissa-barca:
From @jeffvroom 
> In a number of places, mostly in code you didn't change, we just "return error" when something goes wrong. But the database error won't include the SQL that caused the error so all we'll see to support it is something like "Invalid SQL statement" and will have to search and guess among the few possibilities. I don't want to slow you down and overly complicate this PR though so we can fix that later.

For the load balancing move config to database feature I added a function to Database.hpp that will execute a query and logs an INFO message with the SQL. I wasn't planning on going back to already existing statements to use the new function instead, but maybe I can do that for this case?

src/cpp/core/Database.cpp
```diff
@@ -702,14 +704,134 @@ void Transaction::rollback()
...
+
+SchemaVersion::SchemaVersion(const SchemaVersion& other) :
+   Date(other.Date),
+   Flower(other.Flower)
+{
+}
+
+bool SchemaVersion::isEmpty() const
+{
+   return false;
```
MariaSemple, line 727:
```suggestion
   return Date.empty() && Flower.empty();
```

MariaSemple:
Here's an example of a database error log entry, where the statement failed:
```
Jul 13 11:36:01 system76-pc rserver[58037]: ERROR database error 7 (Cannot prepare statement. Fatal error. ERROR:  relation "revoked_cookie" does not exist|||LINE 1: DELETE FROM revoked_cookie WHERE expiration <= $1|||                    ^|||) [description: Could not delete expired revoked cookies from the database, description: Could not read revoked cookies from the database]; OCCURRED AT virtual rstudio::core::Error rstudio::core::database::Connection::execute(rstudio::core::database::Query&, bool*) src/cpp/core/Database.cpp:479; LOGGED FROM: int main(int, char* const*) src/cpp/server/ServerMain.cpp:770
```

jeffvroom (approved):
LGTM!

melissa-barca (approved):
LGTM!

PR #9604, opened 2021-07-14
Feature/electron desktop options

cm421:
Finished adding the rest of the desktop options. Many of these new options need more work (like fonts, but that's its own card/feature), but now they're in place with basic functionality for future work to have something that will persist between runs.

Added some code to actually save main window dimensions when the main window is closed.


gtritchie (approved):
LGTM!

src/node/desktop/src/main/desktop-options.ts
```diff
@@ -17,10 +17,49 @@
...
+const kAccessibility = 'View.Accessibility';
+
+const kLastRemoteSessionUrl = 'Session.LastRemoteSessionUrl';
+const kAuthCookies = 'Session.AuthCookies';
+const kTempAuthCookies = 'Session.TempAuthCookies';
+
+const kIgnoredUpdateVersions = 'General.IgnoredUpdateVersions';
+const kClipboardMonitoring = 'General.ClipboardMonitoring';
+
+const kRBinDir = 'WindowsOnly.RBinDir';
```
kevinushey, line 35: bike-shedding: what about `Platform.Windows.RBinDir`? And in the future we could use `Platform.MacOS` or `Platform.Linux` for other OS-specific options.

cm421: I like that significantly better, great idea!

Issue #9618, opened 2021-07-20
auto-activate anaconda python if no valid version of python (python3) on PATH


kevinushey:
The intention here is, if the user has done the work to install Anaconda Python, we should assume the user wants to use it, and so auto-activate it in IDE sessions (presuming they haven't already explicitly selected something else).

Note: other "valid" Python installations would still have precedence.

- On macOS, avoid `/usr/bin/python3`.
- On Windows, avoid Windows store Python stub.

https://github.com/rstudio/rstudio/blob/43d9a465ea13141586c94073f83f09ba1cd07fe0/src/cpp/session/modules/SessionPythonEnvironments.R#L81-L131 -- need Anaconda lookup at bottom of this code block.

Would also want to look for `python3` on the PATH for macOS + Linux here, screening out potentially "bad" versions of Python.

Also consider (time permitting): comment colors for `#| output: true` metadata comments.

Note: this issue wraps up Python-related Quarto work for 1.5 with @jjallaire.

R-Hannibal:
Moving this to Prairie Trillium for verification (code is in Ghost Orchid).

kevinushey:
We ended up backing this out -- the change here was ultimately too aggressive. See also:

https://github.com/rstudio/rstudio/issues/9952
https://github.com/rstudio/rstudio/issues/9999
https://github.com/rstudio/rstudio/pull/10000

For now, users who want to bind a version of Python to be used in their projects explicitly can do so via project preferences.

PR #9626, opened 2021-07-21
wrapping up IDE python integration items

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/9618

### Approach

- Support auto-activate for copies of Python on the PATH
- Support auto-activate for Anaconda installations
- Fix an issue that prevents `python` from being placed on the PATH when activating a base Conda installation
- Add in comment highlight rule + continue comment behaviors for `#|` chunk metadata comments.

### Automated Tests

N/A

### QA Notes

N/A

### Checklist

### DIFF
--- a/src/cpp/session/modules/SessionPythonEnvironments.R
+++ b/src/cpp/session/modules/SessionPythonEnvironments.R
@@ -78,11 +78,50 @@
   
 })
 
+.rs.addFunction("python.findWindowsPython", function()
+{
+   # NOTE: ideally, we would just parse the output of 'py --list-paths',
+   # but for whatever reason the '*' indicating the default version of
+   # Python is omitted when run from RStudio, so we try to explicitly
+   # run Python here and then ask where the executable lives.
+   #
+   # We pass '-E' here just to avoid things like PYTHONPATH potentially
+   # influencing how the python session is launched
+   pythonPath <- .rs.tryCatch(
+      system2(
+         command = "py",
+         args    = c("-E"),
+         input   = "import sys; print(sys.executable)",
+         stdout  = TRUE,
+         stderr  = TRUE
+      )
+   )
+   
+   if (inherits(pythonPath, "error"))
+      return("")
+   
+   pythonPath
+})
+
 .rs.addFunction("python.configuredInterpreterPath", function(projectDir)
 {
+   # on Windows, check if PY_PYTHON is defined; if it is, we should use 'py'
+   # to determine the version of python to be used
+   if (.rs.platform.isWindows && nzchar(Sys.which("py")))
+   {
+      pyPython <- Sys.getenv("PY_PYTHON", unset = NA)
+      if (!is.na(pyPython))
+      {
+         python <- .rs.python.findWindowsPython()
+         if (file.exists(python))
+            return(python)
+      }
+   }
+   
    # check some pre-defined environment variables
    vars <- c("RENV_PYTHON", "RETICULATE_PYTHON")
-   for (var in vars) {
+   for (var in vars)
+   {
       value <- Sys.getenv(var, unset = NA)
       if (!is.na(value))
          return(value)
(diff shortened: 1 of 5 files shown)


src/cpp/session/modules/SessionPythonEnvironments.R
```diff
@@ -101,6 +101,27 @@
...
    
+   # check for Python on the PATH -- if we have an appropriate version of
+   # python3 or python on the PATH, then use it. we ignore python installations
+   # in /usr/bin because we primarily want to target python installations that
+   # appear to have been "intentionally" installed by the user. in addition,
+   # many Linux package managers install "minimized" versions of Python onto
+   # the system path that will fail to work without extra supporting packages
+   # installed, and managing and communicating all that state to the user is
+   # challenging
+   python3 <- Sys.which("python3")
```
jjallaire, line 112: It feels to me like on Linux we may want to allow `/usr/bin/python3` to work as many linux users will just `apt-get` install the additional requirements they need (rather than get a version from PSF (which is only available as a source tarball: https://www.python.org/downloads/source/). In other words, I think if we do this then we effectively become "conda only" on Linux and I don't think that's what we want to do.

src/cpp/session/modules/SessionPythonEnvironments.R
```diff
@@ -101,6 +101,27 @@
...
+   if (nzchar(python3) && python3 != "/usr/bin/python3")
+      return(python3)
+   
+   # use a plain 'python' executable if it's not python 2
+   python <- Sys.which("python")
```
jjallaire, line 117: In quarto, I fail this check on Windows unless it's conda (I think this is to avoid taking a version of Python on the PATH that is nevertheless not the preferred version of the `py` launcher). In other words, the user might have an older version of Python 3 on the PATH, then install Python 3.9 w/ the defaults (NOT placing Python 3.9 on the PATH) but we still don't get the newer version).

jjallaire: On Windows, I also check for a definition of `PY_PYTHON`, and if it is defined use that in preference to everything.

src/gwt/acesupport/acemode/r_highlight_rules.js
```diff
@@ -227,6 +227,12 @@ define("mode/r_highlight_rules", ["require", "exports", "module"], function(requ
         regex : "#+(?!').*(?:----|====|####)\\s*$",
         next  : "start"
       },
+      {
+        // R Markdown chunk metadata comments
+        token : "comment.doc.tag",
+        regex : "#+[|].*$",
```
jjallaire, line 233: It's just a single `#` that's supported for this syntax.

jjallaire: I think the `comment.doc.tag` looks great here -- in nearly all doc themes it's identical to the comment itself, and in light themes it uses a shade of blue that is just a bit more recessive than comments which is great.

The one behavior we still need here is the termination of the metadata block: if you press `<Enter>` on a line with just `#|` it should wipe out the `#|` and leave a fully blank line in it's place (sort of like the behavior of bullets in most WYSIWYG editors)

jjallaire (changes requested):

src/cpp/session/modules/SessionPythonEnvironments.R
kevinushey: Do you think we should do any pre-flight checks (e.g. checking that `pip` and `venv` are available / functional)?

src/cpp/session/modules/SessionPythonEnvironments.R
kevinushey: If I understand correctly, I think this means:

- We should check `PY_PYTHON` at the top since if it's set it should override any other auto-discovery logic,
- We should perform the `py` check on Windows first before checking the PATH.

src/cpp/session/modules/SessionPythonEnvironments.R
kevinushey: I also wonder if we should still consider preferring Anaconda in that situation; e.g. if the first copy of Python on the path is in `/usr/bin`, but Anaconda is installed, we should use Anaconda. That is, the order would be:

- Python on the PATH from a custom location (not /usr/bin),
- Anaconda Python if found,
- Python in /usr/bin

jjallaire: Yes, I agree with this order for sure. I also agree that we should look for pip as a pre-flight check (not as sure about venv, let's discuss)

src/cpp/session/modules/SessionPythonEnvironments.R
jjallaire: Yes, for Quarto if you set the `PY_PYTHON` environment variable that does indeed trump everything.

src/cpp/session/modules/SessionPythonEnvironments.R
jjallaire: As discussed, no pre-flight check necessary.

kevinushey:
Here's what the chunk metadata auto-terminating behavior looks like:

https://user-images.githubusercontent.com/1976582/126724801-4709f1fb-bfff-409a-bcab-660fafaee08f.mov

Basically, if the user hits enter twice, then the comment prefix will be terminated.

Do you think we should _always_ terminate the chunk comment prefix when the user types enter on a line containing `#| `, or only if the last action was Enter as well?

PR #9631, opened 2021-07-22
support use of symlinks for file locks

kevinushey:
### Intent

Related to https://github.com/rstudio/rstudio-pro/issues/2460.

Some network filesystems (notably, Amazon EFS) do not support hard links, which leads to file locking failures in those environments. This PR provides an alternative that uses symlinks rather than hardlinks, for users who find the advisory file locks are still unreliable in such an environment.

### Approach

Add a new entry to `/etc/rstudio/file-locks`, which allows one to toggle between the use of symlinks vs. hard links when creating file locks. When enabled, we use `::symlink()` instead of `::link()` when creating a lock.

### Automated Tests

N/A

### QA Notes

N/A

### Checklist

### DIFF
--- a/src/cpp/session/SessionSourceDatabase.cpp
+++ b/src/cpp/session/SessionSourceDatabase.cpp
@@ -688,46 +688,50 @@ Error get(const std::string& id, bool includeContents, boost::shared_ptr<SourceD
       }
    }
    
-   if (propertiesPath.exists())
-   {
-      // read the contents of the file
-      std::string properties;
-      Error error = readStringFromFile(propertiesPath,
-                                       &properties,
-                                       options().sourceLineEnding());
-      if (error)
-         return error;
+   if (!propertiesPath.exists())
+      return core::fileNotFoundError(propertiesPath, ERROR_LOCATION);
    
-      // parse the json
-      json::Value value;
-      if (value.parse(properties))
-      {
-         return systemError(boost::system::errc::invalid_argument,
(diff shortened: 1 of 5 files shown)


src/cpp/session/SessionSourceDatabase.cpp
```diff
@@ -688,46 +688,50 @@ Error get(const std::string& id, bool includeContents, boost::shared_ptr<SourceD
...
+   if (value.parse(properties))
+   {
+      Error error = systemError(boost::system::errc::invalid_argument, ERROR_LOCATION);
+      error.addProperty("path", propertiesPath);
+      return error;
    }
-   else
+   
+   // validate we got a JSON object
+   if (!value.isObject())
```
kevinushey, line 712: This check is necessary as otherwise files in the source database containing text that is valid JSON but _not_ a JSON object could lead to a crash. (While this in theory shouldn't ever happen, better safe than sorry)

kevinushey:
@jmcphers; this PR might ultimately be superfluous if we feel advisory file locks are almost always available as an alternative where link-based locks don't work, but it might be worth pulling in some of the smaller bugfixes in the PR. Do you have thoughts?

(If we take this PR I'd also prepare a companion PR updating the Pro docs.)

jeffvroom (approved):
This looks great to me. It seems like a nice option to have in case a network FS does not support advisory locks or hardlinks. I imagine we'll need changes to the locktester to support this option?

PR #9644, opened 2021-07-23
support auto-discovery of R on Windows

kevinushey:
### Intent

- Support auto-discovery of R on Windows, using the most recently recorded version from the Windows registry.
- Also support auto-discovery of C++ build sources, for those of us too lazy to set `RSTUDIO_CPP_OUTPUT_FOLDER`.

### Approach

- Use `reg query <...>` to find what version of R is recorded in the Windows registry. This approach was taken specifically because I was having trouble installing + building https://www.npmjs.com/package/windows-registry on Windows.
- Some other heuristics for locating C++ build artefacts in non-package builds.

### Automated Tests

Not yet implemented.

### QA Notes

N/A

### Checklist

### DIFF
--- a/src/node/desktop/src/core/expected.ts
+++ b/src/node/desktop/src/core/expected.ts
@@ -13,12 +13,16 @@
  *
  */
 
-export type Expected<T> = [ T | null, Error | null ];
+export type Expected<T> = [ T, Error | null ];
 
 export function ok<T>(value: T): Expected<T> {
   return [value, null];
 }
 
-export function unexpected<T>(error: Error): Expected<T> {
-  return [null, error];
+// NOTE: we intentionally lie to the type checker here; the intention is that
+// users of Expected should only use the resulting value T if error is null or
+// unset. Effectively, this implies that "correct" usages of Expected will never
+// give the user a non-T value.
+export function err<T>(error?: Error): Expected<T> {
+  return [null as unknown as T, error ?? new Error()];
 }
--- a/src/node/desktop/src/core/expected.ts
+++ b/src/node/desktop/src/core/expected.ts
@@ -1,0 +1,24 @@
+/*
+ * expected.ts
+ *
+ * Copyright (C) 2021 by RStudio, PBC
+ *
+ * Unless you have received this program directly from RStudio pursuant
+ * to the terms of a commercial license agreement with RStudio, then
+ * this program is licensed to you under the terms of version 3 of the
+ * GNU Affero General Public License. This program is distributed WITHOUT
+ * ANY EXPRESS OR IMPLIED WARRANTY, INCLUDING THOSE OF NON-INFRINGEMENT,
+ * MERCHANTABILITY OR FITNESS FOR A PARTICULAR PURPOSE. Please refer to the
+ * AGPL (http://www.gnu.org/licenses/agpl-3.0.txt) for more details.
+ *
+ */
+
+export type Expected<T> = [ T | null, Error | null ];
+
+export function ok<T>(value: T): Expected<T> {
+  return [value, null];
+}
+
+export function unexpected<T>(error: Error): Expected<T> {
+  return [null, error];
+}
--- a/src/node/desktop/src/main/application.ts
+++ b/src/node/desktop/src/main/application.ts
@@ -26,7 +26,7 @@ import { augmentCommandLineArguments, getComponentVersions, removeStaleOptionsLo
 import { exitFailure, exitSuccess, run, ProgramStatus } from './program-status';
 import { ApplicationLaunch } from './application-launch';
 import { AppState } from './app-state';
-import { prepareEnvironment } from './detect_r';
+import { prepareEnvironment } from './detect-r';
 import { SessionLauncher } from './session-launcher';
 import { DesktopActivation } from './activation-overlay';
 import { WindowTracker } from './window-tracker';
--- a/src/node/desktop/src/main/application.ts
+++ b/src/node/desktop/src/main/application.ts
@@ -136,7 +136,8 @@ export class Application implements AppState {
       }
     }
 
-    if (!await prepareEnvironment()) {
+    const [, error] = prepareEnvironment();
+    if (error) {
       return exitFailure();
     }
 
--- a/src/node/desktop/src/main/detect-r.ts
+++ b/src/node/desktop/src/main/detect-r.ts
@@ -1,5 +1,5 @@
 /*
- * detect_r.ts
+ * detect-r.ts
  *
  * Copyright (C) 2021 by RStudio, PBC
  *
@@ -14,14 +14,14 @@
  */
 
 import { dialog } from 'electron';
+import { dirname } from 'path';
 import { execSync } from 'child_process';
 import { existsSync } from 'fs';
 
-import { logger } from '../core/logger';
 import { Environment, getenv, setVars } from '../core/environment';
-
-import { dirname } from 'path';
 import { Expected, ok, err } from '../core/expected';
+import { logger } from '../core/logger';
+
 
 interface REnvironment {
   rScriptPath: string,
@@ -53,9 +53,11 @@ function executeCommand(command: string): Expected<string> {
 }
 
 /**
- * Detect R and prepare environment for launching rsession.
+ * Detect R and prepare environment for launching the rsession process.
  * 
- * @returns true if startup should continue, false on fatal error
+ * This entails setting environment variables relevant to R on startup
+ * // (for example, R_HOME) and other platform-specific work required
+ * for R to launch.
  */
 export function prepareEnvironment(): Expected<REnvironment> {
 
--- a/src/node/desktop/src/main/detect_r.ts
+++ b/src/node/desktop/src/main/detect_r.ts
@@ -87,7 +87,7 @@ async function detectREnvironment(): Promise<REnvironment> {
   }
 
   // get R_HOME + other related R environment variables
-  const stdout = execSync(`${R} RHOME`, { encoding: 'utf-8'});
+  const stdout = execSync(`${R} RHOME`, { encoding: 'utf-8' });
   const home = stdout.trim();
   const envvars = {
     R_HOME:        `${home}`,
--- a/src/node/desktop/src/main/detect_r.ts
+++ b/src/node/desktop/src/main/detect_r.ts
@@ -15,20 +15,27 @@
 
 import { app, dialog } from 'electron';
 import { exec, execSync } from 'child_process';
-import { promisify } from 'util';
 import { existsSync } from 'fs';
-import { assert } from 'console';
 
 import { logger } from '../core/logger';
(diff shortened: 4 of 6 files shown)


gtritchie:
Neato! I won't get a chance to look until "Monday" (aka tomorrow/Saturday).

src/node/desktop/src/main/detect_r.ts
```diff
@@ -37,116 +42,98 @@ function showRNotFoundError(msg: string): void {
...
-  return true;
 }
 
-async function scanForR(rstudioWhichR: FilePath): Promise<FilePath> {
+async function scanForR(): Promise<string> {
+
+  // if the RSTUDIO_WHICH_R environment variable is set, use that
+  const rstudioWhichR = getenv('RSTUDIO_WHICH_R');
+  if (rstudioWhichR) {
+    logger().logDiagnostic(`Using RSTUDIO_WHICH_R: ${rstudioWhichR}`)
```
gtritchie, line 118: Missing semicolon (ESLint will flag as error).

src/node/desktop/src/main/detect_r.ts
```diff
@@ -37,116 +42,98 @@ function showRNotFoundError(msg: string): void {
...
-  const whichROverride = getenv('RSTUDIO_WHICH_R');
-  if (!whichROverride) {
-    dialog.showErrorBox(
-      'RSTUDIO_WHICH_R Not Set',
-      'Temporary: RSTUDIO_WHICH_R environment variable must point to R installation folder. Cannot continue.');
-    return false;
+  // scan for R
+  const R = await scanForR();
+  if (!R) {
+    throw rNotFoundErrorMessage();
```
gtritchie, line 86: Here, and all other locations where you throw rNotFoundErrorMessage: always throw an Error or subclass of an Error. So, `throw Error(rNotFoundErrorMessage())` would work.

From https://nodejs.org/dist/latest-v14.x/docs/api/errors.html:

> Class: Error
> A generic JavaScript <Error> object that does not denote any specific circumstance of why the error occurred. Error objects capture a "stack trace" detailing the point in the code at which the Error was instantiated, and may provide a text description of the error.
> 
> All errors generated by Node.js, including all system and JavaScript errors, will either be instances of, or inherit from, the Error class.
```

File: src/node/desktop/src/main/detect_r.ts [hunk 3]
```diff
@@ -13,22 +13,27 @@
...
+import { dirname } from 'path';
 
-const asyncExec = promisify(exec);
+interface REnvironment {
+  rScriptPath: string,
+  version: string,
+  envVars: Environment,
+}
+
+function rNotFoundErrorMessage(): string {
```
gtritchie, line 31: See comment later on not throwing a string directly, either use `throw Error(rNotFoundErrorMessage())` or change this function to do that for you and return an `Error`.

src/node/desktop/src/main/detect_r.ts
```diff
@@ -37,116 +42,98 @@ function showRNotFoundError(msg: string): void {
...
-async function prepareEnvironmentPosix(): Promise<boolean> {
-
-  assert((process.platform !== 'win32'));
-
-  // check for which R override
-  let rWhichRPath = new FilePath();
-  const whichROverride = getenv('RSTUDIO_WHICH_R');
-  if (whichROverride) {
-    rWhichRPath = new FilePath(whichROverride);
+  try {
```
gtritchie, line 46: I'd avoid using `Promise<boolean>`, and instead make this `Promise<void>`. See my message in Slack channel about this. There are other instances of this pattern, I'm going to go back and clean them up.

That eliminates the need for a try/catch here. An `async` function takes care of catching thrown exceptions and performing equivalent of `Promise.reject(e)` for you.

This does require callers to use promise-semantics. If using `await` then the error is made into an exception which will need to be caught, or if using promises directly, should be handled in the `.catch()` callback.

This does bring up a need for us to chat about what patterns we want to use here (exceptions, etc.) so we can agree on conventions to follow.

gtritchie: Also, we'll want to look more closely at whether we get any value from the async-ness in these routines (yes, it was my fault they were to begin with, mainly me playing around and learning, honestly).

For a desktop app like this, especially startup code, might not be worth the complexity, but haven't gone back to reason that through.

src/node/desktop/src/main/detect_r.ts
```diff
@@ -37,116 +42,98 @@ function showRNotFoundError(msg: string): void {
...
+    return prepareEnvironmentImpl();
```
gtritchie, line 47: Again, because you're calling an async function (`prepareEnvironmentImpl`) (but **not** using `await` to invoke it), it will never throw an exception! Any exceptions thrown inside it will be converted into a Promise.reject.

If you did use `await` then it could throw an exception. It all makes sense, eventually, but feel like I should make an infographic for it (including for myself).

src/node/desktop/src/main/detect_r.ts
```diff
@@ -37,116 +42,98 @@ function showRNotFoundError(msg: string): void {
...
 
   // attempt to detect R environment
-  const detectResult = await detectREnvironment(rWhichRPath, rLdScriptPath, '');
-  if (!detectResult.success) {
-    showRNotFoundError(detectResult.errMsg ?? 'Unknown error');
+  let rEnvironment : REnvironment;
+  try {
+    rEnvironment = await detectREnvironment();
+  } catch (error) {
+    showRNotFoundError(error ?? 'An unknown error occurred.');
```
gtritchie, line 62: In general shouldn't need to worry about this (`?? 'An unknown error occurred.`). I was using the pattern at one point but it seems unnecessary. We should only ever throw `Error` or subclasses of such in our code, and it is safe to assume any code we consume is doing the same (including node and Javascript itself).

src/node/desktop/src/main/detect_r.ts
```diff
@@ -37,116 +42,98 @@ function showRNotFoundError(msg: string): void {
...
-  try {
-    const { stdout } = await asyncExec('/usr/bin/which R', { encoding: 'utf-8' });
-    const R = stdout.trim();
-    if (R) {
-      return new FilePath(R);
-    }
-  } catch (error) {
-    logger().logError(error);
+  // first, look for R on the PATH
+  const stdout = execSync('/usr/bin/which R', { encoding: 'utf-8' });
```
gtritchie, line 133: This can throw an exception. Non-trivial synchronous node APIs should be assumed to throw, and the docs mention it specifically for this one:
> If the process times out or has a non-zero exit code, this method will throw. The Error object will contain the entire result from child_process.spawnSync().

So you can either handle it internally here, or let the caller deal with it.

src/node/desktop/src/main/detect_r.ts
```diff
@@ -163,62 +150,55 @@ async function scanForRPosix(rstudioWhichR: FilePath): Promise<FilePath> {
...
 
-export async function scanForRWin32(rstudioWhichR: FilePath): Promise<FilePath> {
-  assert((process.platform === 'win32'));
+function findDefaultInstallPathWin32(version: string): string {
 
-  // prefer RSTUDIO_WHICH_R
-  if (!rstudioWhichR.isEmpty()) {
-    return rstudioWhichR;
+  const keyName = `HKEY_LOCAL_MACHINE\\SOFTWARE\\R-Core\\${version}`;
+  const regOutput = execSync(`reg query ${keyName} /v InstallPath`, { encoding: 'utf-8'});
```
gtritchie, line 165: Ditto on this throwing exceptions.

src/node/desktop/src/main/utils.ts
```diff
@@ -116,6 +116,69 @@ export function rsessionExeName(): string {
...
+  }
+
+  throw rsessionNotFoundError();
+
+}
+
+function findBuildRootImpl(rootDir: string): string {
+
+  // list all files + directories in the folder
+  const files = fs.readdirSync(rootDir);
```
gtritchie, line 141: Sync node API involving the system, should assume it can throw.

gtritchie (changes requested):

src/node/desktop/src/core/expected.ts
```diff
@@ -0,0 +1,28 @@
...
+  return [value, null];
+}
+
+// NOTE: we intentionally lie to the type checker here; the intention is that
+// users of Expected should only use the resulting value T if error is null or
+// unset. Effectively, this implies that "correct" usages of Expected will never
+// give the user a non-T value.
+export function err<T>(error?: Error): Expected<T> {
+  return [null as unknown as T, error ?? new Error()];
+}
```
gtritchie, line 28: Wonder if we should combine this file with `err.ts` (don't care which survives) just so the "how to return errors from a function" techniques are in the same file for easier discovery?

gtritchie: Be good to add a unit test or two on the basic usage patterns of Expected/ok/err.

src/node/desktop/src/main/application.ts
```diff
@@ -136,7 +136,8 @@ export class Application implements AppState {
       }
     }
 
-    if (!await prepareEnvironment()) {
+    const [, error] = prepareEnvironment();
```
gtritchie, line 139: As mentioned in Slack, suggest changing `prepareEnvironment` to just return `Err` since we never use the result.

src/node/desktop/src/main/detect-r.ts
```diff
@@ -0,0 +1,241 @@
...
+}
+
+/**
+ * Detect R and prepare environment for launching the rsession process.
+ * 
+ * This entails setting environment variables relevant to R on startup
+ * // (for example, R_HOME) and other platform-specific work required
+ * for R to launch.
+ */
+export function prepareEnvironment(): Expected<REnvironment> {
```
gtritchie, line 62: Per other comment, maybe just return `Err` (can use `return Success()`).

src/node/desktop/src/main/detect-r.ts
```diff
@@ -0,0 +1,241 @@
...
+  try {
+    return prepareEnvironmentImpl();
+  } catch (error) {
+    logger().logError(error);
+    return err(error);
+  }
+
+}
+
+function prepareEnvironmentImpl(): Expected<REnvironment> {
```
gtritchie, line 73: Ditto, just return `Err`.

src/node/desktop/src/main/detect-r.ts
```diff
@@ -0,0 +1,241 @@
...
+  if (process.platform === 'win32') {
+    const binDir = dirname(scriptPath);
+    process.env.PATH = `${binDir};${process.env.PATH}`;
+  }
+
+  return ok(rEnvironment);
+
+}
+
+function detectREnvironment(): Expected<REnvironment> {
```
gtritchie, line 97: Another ditto on return `Err`.

gtritchie (changes requested):
Couple small tweaks, but could do post-merge if that's easier. Otherwise LGTM!

PR #9645, opened 2021-07-23
Feature/electron window position restore

cm421:
Add main window position restore (in addition to the already-present window size restore). Loosely mirrors the C++ desktop code. Accounts for display arrangement changing between window save/restore

### DIFF
--- a/src/node/desktop/test/unit/main/desktop-options.test.ts
+++ b/src/node/desktop/test/unit/main/desktop-options.test.ts
@@ -12,13 +12,16 @@
  * AGPL (http://www.gnu.org/licenses/agpl-3.0.txt) for more details.
  */
 
+import { BrowserWindow, Rectangle, screen } from 'electron';
 import { describe } from 'mocha';
 import { assert } from 'chai';
+import sinon from 'sinon';
 
-import { DesktopOptions, DesktopOptionsImpl, kDesktopOptionDefaults, clearOptionsSingleton } from '../../../src/main/desktop-options';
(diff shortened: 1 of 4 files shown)


src/node/desktop/test/unit/main/desktop-options.test.ts
```diff
@@ -127,4 +134,125 @@ describe('DesktopOptions', () => {
...
+    // Save bounds onto a secondary display on the right
+    DesktopOptions().saveWindowBounds(savedWinBounds);
+
+    const sandbox = sinon.createSandbox();
+    sandbox.stub(screen, 'getAllDisplays').returns(displays as Display[]);
+    const testMainWindow = sandbox.createStubInstance(BrowserWindow);
+    testMainWindow.setBounds.withArgs(savedWinBounds);
+    testMainWindow.getSize.returns([savedWinBounds.width, savedWinBounds.height]);
+    
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
cm421, line 150: Kinda annoying to have these in unit testing to keep the linter happy. I couldn't figure out how to get it to compile without adding `as any`

kevinushey (approved):
LGTM!

src/node/desktop/test/unit/main/desktop-options.test.ts
gtritchie: I'll take a peek once it is merged, should rarely need to use 'any' or do a cast like this (though sometimes it probably will be necessary, and this might be one of those times).

gtritchie (approved):
LGTM!

src/node/desktop/test/unit/main/desktop-options.test.ts
gtritchie: Fixed via #9651

PR #9651, opened 2021-07-25
eliminate "as any" and "eslint-disables" in restore window position unit tests

gtritchie:
### Intent

Unit tests for #9645 had a couple instances of "as any" assertions and corresponding es-lint disables. We should try to avoid needing these. 

### Approach

Add a variation of `createSinonStubInstance` helper named `createSinonStubInstanceForSandbox` that works with non-default sandbox. This helper uses some TypeScript magic to work around this problem (the line ` return stub as unknown as StubbedClass<T>;`).

Note, occasionally VSCode's TypeScript support loses its mind on this construct and will flag it as an error. If you close and restart VSCode it resolves. In any case TypeScript itself (and ESLint) are both happy with it.

### Automated Tests

This is some source code cleanup on unit tests, no change to product code.

### QA Notes

NA

### Checklist

### DIFF
--- a/src/node/desktop/test/unit/main/desktop-options.test.ts
+++ b/src/node/desktop/test/unit/main/desktop-options.test.ts
@@ -16,6 +16,7 @@ import { BrowserWindow, Rectangle, screen } from 'electron';
 import { describe } from 'mocha';
 import { assert } from 'chai';
 import sinon from 'sinon';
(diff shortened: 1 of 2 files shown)


cm421:
Ohhh, that's sneaky clever. Thank you for fixing that!

PR #9654, opened 2021-07-26
enable strict boolean checks

kevinushey:
Enable strict boolean checks in our Typescript code, catching some potentially unintended boolean conversions in if conditionals (among other places).

This should help in cases where we accidentally check the value of a promise in a conditional. For example:

Screen Shot 2021-07-26 at 2 31 03 PM

See https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/strict-boolean-expressions.md for more details.

### DIFF
--- a/src/node/desktop/.eslintrc.js
+++ b/src/node/desktop/.eslintrc.js
@@ -7,7 +7,7 @@ module.exports = {
     "node": true
   },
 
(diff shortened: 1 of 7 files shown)


gtritchie (approved):
LGTM!

PR #9670, opened 2021-07-28
ensure we use a pre element when computing console width

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/8696.

### Approach

We normally only apply our fonts to `<pre>` elements, as per https://github.com/rstudio/rstudio/blob/bugfix/font-size-computation-custom-fonts/src/cpp/session/resources/themes/css/fonts.css. However, when computing the Console width, we were doing so with a `<div>` element, which wouldn't have the custom font applied.

Because we were computing the width using a different font from the actual Console font, it's possible for text to overflow if the user has configured RStudio to use a font that is wider than the default font. This is most common on Windows, where the default font (Calibri) is somewhat more narrow.

### Automated Tests

N/A

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/8696.

### Checklist


jmcphers (approved):
LGTM!

PR #9682, opened 2021-07-29
implement Choose R dialog for Windows

kevinushey:
Ports the "Choose R Installation" dialog, to something resembling the native equivalent using Electron.

## Before

Screen Shot 2021-07-29 at 6 38 56 PM

## After

Screen Shot 2021-07-29 at 6 38 38 PM

There are a few remaining items:

- [ ] Consider whether we can or should refactor some of this. Should the HTML live in a template file? Should we insert the set of R installation options via Javascript after loading the window, or is the existing implementation okay?
- [ ] Detect whether the user is holding down Ctrl when RStudio is launched. We'll almost surely need a native extension module for that; e.g. https://github.com/Zysen/node-asynckeystate; https://github.com/rstudio/rstudio/blob/b75aa95170f1557c78b8abff9cb4d43865504e9f/src/cpp/desktop/DesktopWin32DetectRHome.cpp#L39-L43
- [ ] Read and store the default version of R in the desktop options.
- [ ] Do we need the GPU rendering options?

Marking this as a draft but putting the PR up for early review.


src/node/desktop/src/main/select-r.ts
```diff
@@ -0,0 +1,265 @@
...
+    // skip if we've already seen this
+    if (visitedInstallations[rInstall]) {
+      return;
+    }
+    visitedInstallations[rInstall] = true;
+
+    // check for 64 bit executable
+    const r64 = `${rInstall}/bin/x64/R.exe`;
+    if (existsSync(r64)) {
+      const html = `<option value="${r64}">[64-bit] ${rInstall}</option>`;
```
kevinushey, line 47: We should probably HTML-escape pieces here as appropriate. Need to see if Node / Electron have safe HTML templating or something similar.

gtritchie: (Repeat of a slack conversation, just for the "permanent record"): I ported our simple HTML escaper from C++ to node.js here: https://github.com/rstudio/rstudio/blob/f591ebfa21478b3c21aba9691445bc691202dc43/src/node/desktop/src/core/string-utils.ts#L23-L37

Could also look at the npm packages `escape-html` (not updated in 6 years, mind you) or `html-escaper` (more recently updated).

kevinushey: I ended up restructuring things here so that the select HTML widget is populated via IPC when the modal dialog is initialized.

src/node/desktop/package.json
```diff
@@ -41,8 +41,9 @@
     "uuid": "^8.3.2"
   },
   "scripts": {
-    "build": "tsc",
+    "build": "tsc && yarn run copy-files",
     "clean": "ts-node scripts/clean.ts",
+    "copy-files": "ts-node scripts/copy-files.ts",
```
gtritchie, line 46: Don't copy these files. Use them directly from source tree via relative location with `appState().resourcesPath().completePath()`; there are existing examples in the code for the static HTML files such as `html/error.html`. Get that working with dev builds, and it'll work with package builds (eventually, still working out some issues with Windows package builds).

src/node/desktop/scripts/copy-files.ts
```diff
@@ -0,0 +1,59 @@
+/*
```
gtritchie, line 1: You don't need this script. See other comment about using files in-place for dev builds.

FWIW, there's also a node module I pulled in as a dev dependency for WIP on the packaging for Windows that makes it easy to do recursive copying. https://github.com/timkendrick/recursive-copy.

src/node/desktop/src/ui/README.md
```diff
@@ -0,0 +1,72 @@
+
+# RStudio Electron UI
+
+This README documents the structure of custom widgets used in RStudio.
+
+For example, consider a widget called 'clock'. The following files should be created:
+
+```
+widgets/clock.ts            <required>
+widgets/clock/ui.html       <required>
```
gtritchie, line 10: Let's keep the `html` file outside the `src` tree, probably in a `resources` folder (for example, `src/node/desktop/resources/widgets`). Then refer to it using the `appState().resourcesPath().completePath()` technique mentioned elsewhere.

I'll probably end up moving the existing `html` folder into this as well (`src/node/desktop/resources/html`).

gtritchie (changes requested):
Overall approach looks great, some comments on where things should live (and don't need to copy files around for dev builds). I didn't look super-close at the actual R chooser dialog implementation nitty-gritty, but the screenshot looks great! :-)

src/node/desktop/package.json
kevinushey: How should we handle widgets which want to depend on Typescript files? (Since after compilation, the associated Javascript will be placed in the `dist/` folder, rather than compiled in the same directory.)

gtritchie: Need to look closer again, but my gut reaction is: the widgets themselves will have been compiled to javascript before they execute and thus be in the same (relative) dist location with any any files they are importing. I suspect there's an angle I'm missing with that "gut reaction" though.

For the webpacked version, everything in a given module will end up in a single JavaScript file. So all the main process code will be combined into one .js file, and all the renderer process code in another.

We do already use a tiny bit of code from the `core` library in the main preload script (something from `file-path.ts`). We can get away with that in a preload script since they can use node.js, currently. But renderer code outside the preload can't use it. There's also a small risk mentioned in #9737 that we might need to turn off most node.js even for the preload scripts.

PR #9694, opened 2021-08-03
Lower log priority of unrecognized files in db folder

MariaSemple:
### Intent

Addresses [Pro #2790](https://github.com/rstudio/rstudio-pro/issues/2790)

### Approach

Lower the log level from WARNING to DEBUG. This will ensure that these probably innocuous logs will not cause undue concern. The probably better long term solution here would be to have the installer remove the db/ files as a pre-install step, since all the needed files will be installed during the regular install step anyway.

### Automated Tests

N/A, logging change only.

### QA Notes

The logs should not appear unless the admin has enabled debug logging. Here is what it looks like when that is enabled:
image

### Checklist

### DIFF
--- a/src/cpp/core/Database.cpp
+++ b/src/cpp/core/Database.cpp
@@ -945,7 +945,7 @@ bool SchemaUpdater::parseVersionOfFile(const FilePath& file, SchemaVersion* pVer
    boost::split(split, fileStem, boost::is_any_of("_"));
    if (split.size() != 3)
    {
-      LOG_WARNING_MESSAGE("Unrecognized file in the db folder, it will be skipped: " + file.getAbsolutePath());
+      LOG_DEBUG_MESSAGE("Unrecognized file in /usr/lib/rstudio-server/db, it will be skipped: " + file.getAbsolutePath());
       return false;
    }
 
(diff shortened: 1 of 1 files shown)


src/cpp/core/Database.cpp
```diff
@@ -945,7 +945,7 @@ bool SchemaUpdater::parseVersionOfFile(const FilePath& file, SchemaVersion* pVer
    boost::split(split, fileStem, boost::is_any_of("_"));
    if (split.size() != 3)
    {
-      LOG_WARNING_MESSAGE("Unrecognized file in the db folder, it will be skipped: " + file.getAbsolutePath());
+      LOG_DEBUG_MESSAGE("Unrecognized file in /usr/lib/rstudio-server/db, it will be skipped: " + file.getAbsolutePath());
```
jeffvroom, line 948: Just a nit, but I think we'll already know the directory name from the absolute path right?  Or I suppose we could log the migrationsPath_ and use file.getFilename()?  It might be helpful to call out that this is an older schema file. How about something like this:

```suggestion
     if (split.size() == 2)
         LOG_DEBUG_MESSAGE("Not applying sql schema file from previous release: " + file.getAbsolutePath());
     else
         LOG_DEBUG_MESSAGE("Not applying unrecognized sql schema file: " + file.getAbsolutePath());
```

jeffvroom (approved):
LGTM!

src/cpp/core/Database.cpp
MariaSemple: Great Idea!

PR #9696, opened 2021-08-03
Feature/gwt callbacks craig

cm421:
### Intent

Connect electron's save file/open directory dialogs to GWT's requests for them. Also added open/show file/folder in the OS' file explorer cause (I thought) I needed them for debugging.

### Approach

Mirrored the existing open file functionality very closely.

### Automated Tests

I didn't add any tests, because this seems like something that's more tested as part of integrated testing for full system testing. If I'm mistaken I'd be happy to add some

### QA Notes

None
### Checklist

- [ x ] If this PR adds a new feature, or fixes a bug in a previously released version, it includes an entry in `NEWS.md` 
- [ x ] If this PR adds or changes UI, the updated UI meets [accessibility standards](https://github.com/rstudio/rstudio/wiki/Accessibility)
- [ x ] A reviewer is assigned to this PR (if unsure who to assign, check Area Owners list)
- [ x ] This PR passes all local unit tests

### DIFF
--- a/src/node/desktop/src/main/gwt-callback.ts
+++ b/src/node/desktop/src/main/gwt-callback.ts
@@ -18,7 +18,7 @@
 /* eslint-disable @typescript-eslint/no-empty-function */
 
 import { ipcMain, dialog, BrowserWindow, webFrameMain } from 'electron';
-import { IpcMainEvent, MessageBoxOptions, OpenDialogOptions } from 'electron/main';
+import { IpcMainEvent, MessageBoxOptions, OpenDialogOptions, SaveDialogOptions } from 'electron/main';
 
 import EventEmitter from 'events';
 
@@ -91,12 +91,22 @@ export class GwtCallback extends EventEmitter {
       }
     });
 
-    ipcMain.handle('desktop_get_save_file_name', (event, caption: string, label: string,
+    ipcMain.handle('desktop_get_save_file_name', async (event, caption: string, label: string,
       dir: string, defaultExtension: string, forceDefaultExtension: boolean,
       focusOwner: boolean
     ) => {
-      GwtCallback.unimpl('desktop_get_save_file_name');
-      return '';
+      const saveDialogOptions: SaveDialogOptions = {
+        title: caption,
+        defaultPath: dir,
+        buttonLabel: label
+      };
+
+      const focusedWindow = BrowserWindow.getFocusedWindow();
+      if (focusedWindow) {
+        return dialog.showSaveDialog(focusedWindow, saveDialogOptions);
+      } else {
+        return dialog.showSaveDialog(saveDialogOptions);
+      }
     });
 
     ipcMain.handle('desktop_get_existing_directory', (event, caption: string, label: string,
--- a/src/node/desktop/src/main/gwt-callback.ts
+++ b/src/node/desktop/src/main/gwt-callback.ts
@@ -101,6 +101,12 @@ export class GwtCallback extends EventEmitter {
         buttonLabel: label
       };
 
+      if (defaultExtension) {
+        saveDialogOptions['filters'] = [
+          {name: '', extensions: [defaultExtension.replace('.', '')]}
+        ];
+      }
+
       let focusedWindow = BrowserWindow.getFocusedWindow();
       if (focusOwner) {
         focusedWindow = this.getSender('desktop_open_minimal_window', event.processId, event.frameId).window;
@@ -200,10 +206,9 @@ export class GwtCallback extends EventEmitter {
     });
 
     ipcMain.on('desktop_show_folder', (event, path: string) => {
-      void shell.openPath(path)
-        .then((value) => {
-          if (value)
-            logger().logErrorMessage(value);
+      shell.openPath(path)
+        .catch((value) => {
+          logger().logErrorMessage(value);
         });
     });
 
--- a/src/node/desktop/src/main/gwt-callback.ts
+++ b/src/node/desktop/src/main/gwt-callback.ts
@@ -17,7 +17,7 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
 /* eslint-disable @typescript-eslint/no-empty-function */
 
-import { ipcMain, dialog, BrowserWindow, webFrameMain } from 'electron';
+import { ipcMain, dialog, BrowserWindow, webFrameMain, shell } from 'electron';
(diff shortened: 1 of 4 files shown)


src/node/desktop/src/main/gwt-callback.ts
```diff
@@ -173,11 +200,15 @@ export class GwtCallback extends EventEmitter {
     });
 
     ipcMain.on('desktop_show_folder', (event, path: string) => {
-      GwtCallback.unimpl('desktop_show_folder');
+      void shell.openPath(path)
+        .then((value) => {
+          if (value)
+            logger().logErrorMessage(value);
+        });
```
cm421, line 207: I'm not entirely sure what the `void` here is doing. This is what the linter automatically suggested to me, so I've left it

gtritchie: If you add on a `.catch` it isn't necessary to have the `void`. The `void` is just a hint to ESLint that you're aware you're not handling errors.

src/node/desktop/src/main/gwt-callback.ts
```diff
@@ -91,19 +91,46 @@ export class GwtCallback extends EventEmitter {
       }
     });
 
-    ipcMain.handle('desktop_get_save_file_name', (event, caption: string, label: string,
+    ipcMain.handle('desktop_get_save_file_name', async (event, caption: string, label: string,
       dir: string, defaultExtension: string, forceDefaultExtension: boolean,
```
gtritchie, line 95: You're not using `defaultExtension` or `forceDefaultExtension`; those are used in the current implementations and need to be reimplemented if possible on Electron. Probably involving `filters: FileFilter[]`, but I didn't look very closely. Be sure to checkout how it's done in both the Linux/Windows implementation (DesktopGwtCallback.cpp) and Mac (DesktopGwtCallbackMac.mm). The Qt docs are at https://doc.qt.io/qt-5/qfiledialog.html

Some of the resulting behavior may be easier to see/troubleshoot on Windows.

gtritchie (changes requested):

src/node/desktop/src/main/gwt-callback.ts
cm421: I couldn't see a clear way to implement the default extension, as the `filters:` option requires a `name:` in addition to the extension. The example from electron's documentation gives something like:

```
{
  filters: [
    { name: 'Images', extensions: ['jpg', 'png', 'gif'] },
    { name: 'Movies', extensions: ['mkv', 'avi', 'mp4'] },
    { name: 'Custom File Type', extensions: ['as'] },
    { name: 'All Files', extensions: ['*'] }
  ]
}
```

Since a `name` for the extension isn't included in this call from GWT, the best I could think of was to create some static filter object like the above that holds all filetypes/extensions we work with and sort them into categories, but that sounds tedious to create and even worse to maintain.

All of that seemed moot to me anyway, because when you have a document open, and trigger this "Save As..." dialog, the extension is already present, and left there by default (the filename, not including the file extension, is automatically highlighted by default, ready to be changed by typing), which is what I'm assuming the filter was supposed to do in Qt anyway:
Screen Shot 2021-08-05 at 10 03 56 AM

This seemed reasonable to me, so I didn't spend any more time on it. With this extra info, is leaving it as-is okay with you?

src/node/desktop/src/main/gwt-callback.ts
```diff
@@ -91,19 +91,46 @@ export class GwtCallback extends EventEmitter {
...
+
+      let focusedWindow = BrowserWindow.getFocusedWindow();
+      if (focusOwner) {
+        focusedWindow = this.getSender('desktop_open_minimal_window', event.processId, event.frameId).window;
+      }
+      if (focusedWindow) {
+        return dialog.showOpenDialog(focusedWindow, openDialogOptions);
+      } else {
+        return dialog.showOpenDialog(openDialogOptions);
+      }
```
cm421, line 133: I just noticed that if we call `showOpenDialog` with `focusedWindow`, a dialog shows up in the middle of the main electron window, and it cannot be moved. The only way to "move" it is to move the main electron window, and the dialog window will move with the main window.

Personally I don't like it when sub-windows are locked in place, is this the behavior we want? I suggest removing most this logic and replacing it with:
`return dialog.showOpenDialog(openDialogOptions);`

cm421: With that command, it creates the same dialog, but it's free to be moved around anywhere on the screen/other displays

gtritchie: That is a Mac-specific thing, mentioned here: https://www.electronjs.org/docs/latest/api/dialog#sheets

We do need these dialogs to be modal, if you make them modeless (by not supplying the owning BrowserWindow), then you can bring up the dialog and continue interacting with the main window, which puts the app in an undefined state.

For example, you can bring up multiple File Open dialogs at once, and you can put the main window over a file-open dialog, both of which are Bad Things.

On Windows and Linux, the modal is not attached to its parent window like it is on the Mac, but the parent window is disabled while the dialog is displayed, and cannot be put under the parent window.

Unfortunately, I don't see a way to tell Electron to not make these modal dialogs be sheets on MacOS. In Qt there was a flag for it (`Qt::Sheet`). I didn't look very hard, though.

src/node/desktop/src/main/gwt-callback.ts
gtritchie: Sorry, `filters` was a red herring (I wasn't looking at the code), as I do use it for `desktop_get_open_file_name`.

We definitely need the behavior of the current code to be ported over. Here's an example why:

* File / New File / R Script
* Save
* Type "hello" and hit Enter

In RStudio (shipped version) this will save as "hello.R". In the Electron version (without `defaultExtension` implementation), the file tries to save as plain `hello`, prompting a warning message.

This behavior is implemented AFTER the dialog is dismissed (again, sorry to confuse matters with mention of `filters`), here: https://github.com/rstudio/rstudio/blob/5dd6fdfb5c912e7afc0975c78dd4f313722bd5dd/src/cpp/desktop/DesktopGwtCallback.cpp#L298-L323

cm421: Ah! I see now. 

I guess `filters` wasn't entirely a red herring, because you need `filters` to get the `.R` extension to show up for a new, untitled document

gtritchie (approved):
LGTM! Thanks!

PR #9717, opened 2021-08-10
implement MainWindow.closeEvent (Electron)

gtritchie:
### Intent

Implement commented-out bits of MainWindow.closeEvent.

### Approach

Convert to TypeScript.

### Automated Tests

None specifically, did tweak a test's timeouts.

### QA Notes

None

### Checklist

### DIFF
--- a/src/node/desktop/src/main/main-window.ts
+++ b/src/node/desktop/src/main/main-window.ts
@@ -30,6 +30,7 @@ import { GwtWindow } from './gwt-window';
 import { appState } from './app-state';
 import { DesktopOptions } from './desktop-options';
 import { RemoteDesktopSessionLauncher } from './remote-desktop-session-launcher-overlay';
+import { CloseServerSessions } from './session-servers-overlay';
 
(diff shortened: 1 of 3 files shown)


cm421:
Beautiful! Seems to work really nicely on linux too.

I did notice that on the integration tests there's a new prompt that now appears during the first test (session restart, I believe) as we're shutting down:
Screen Shot 2021-08-10 at 12 38 45 PM

We'll want to add a line in the integration test to clear that dialog, otherwise it will exceed the timeout and the test fails. Not sure if that's for this issue or some future one

gtritchie:
Just to capture a Slack conversation:

Ah, that prompt about saving workspace image, that’s something we turn off on our server integration tests (it’s a user preference). Right now Playwright can’t drive desktop dialogs (including the File picking dialogs), so we can’t have those showing. That’s an Electron limitation, really. It might be doable in Electron 15 (naturally, there’s always something): https://github.com/electron/electron/pull/26102

Issue #9715, opened 2021-08-10
Log file permissions for rserver change after session start in open source version


jeffvroom:
### System details

    RStudio Edition : OS
    RStudio Version : latest

### Steps to reproduce the problem

  Enable log-level=debug in logging.conf, start a session in the open source version. Last message is: 

2021-08-09T19:26:21.929080Z [rserver] DEBUG Launched session process for user: mario: /usr/lib/rstudio-server/bin/rsession pid: 73651

   and no more messages will be logged after this point.

   Do ls -l of /var/log/rstudio/rstudio-server/rserver.log and see its now owned by the user who started the session. 

### Describe the problem in detail

Tracked this down to [this line of the function permanentlyDropPriv](https://github.com/rstudio/rstudio/blob/ff9bd9e4e9424e2be8408e07a04dd5daaffec5e9/src/cpp/core/system/PosixSystem.cpp#L2477) that is running after the fork, before the exec for the session. Since we haven't exec'd at this point, but are running as the session user, and have privs, the log destinations are for the parent not the child as expected.

I can see the temporarilyDropPriv does need this call because it's not about to call exec. All of the calls to permentlyDropPriv that I could see are between fork and exec so I think we should just remove that line.

astayleraz:
I verified in version 2021-08.0 Build 317-2 that with a clean install on a Linux machine, /var/lib/rstudio-server is owned and writable by rstudio-server. Closing issue.

PR #9718, opened 2021-08-10
Fix permissions for /var/lib/rstudio-server directory for sqlite 

jeffvroom:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/9715

### Approach

The sqlite database requires that write perms be available for the directory. Other parts of the code do this for Workbench but OS is missing some code to try and do the chown of the directory so I added it. 

### Automated Tests

Not necessary

### QA Notes

A clean install on a linux machine of the OS version should show /var/lib/rstudio-server owned and writable by rstudio-server

### Checklist

- [ x] This PR passes all local unit tests

### DIFF
--- a/src/cpp/server_core/ServerDatabase.cpp
+++ b/src/cpp/server_core/ServerDatabase.cpp
@@ -151,7 +151,23 @@ Error readOptions(const std::string& databaseConfigFile,
          // always ensure the database file user is correct, if specified
          error = databaseFile.changeOwnership(databaseFileUser.get());
          if (error)
+         {
+            error.addProperty("description", "Unable to change ownership of database file");
             return error;
+         }
+
+#ifndef _WIN32
+         error = databaseDirectory.changeOwnership(databaseFileUser.get());
+         if (error)
+         {
+            bool writable = false;
+            Error writableError = databaseDirectory.isWriteable(writable);
+            if (writableError || !writable)
+            {
+                LOG_ERROR_MESSAGE("Sqllite database directory: " + databaseDirectory.getAbsolutePath() + " must be writable by: " + databaseFileUser.get().getUsername());
+            }
+         }
+#endif
       }
 
       LOG_INFO_MESSAGE("Connecting to sqlite3 database at " + options.file);
--- a/src/cpp/server_core/ServerDatabase.cpp
+++ b/src/cpp/server_core/ServerDatabase.cpp
@@ -152,6 +152,7 @@ Error readOptions(const std::string& databaseConfigFile,
          error = databaseFile.changeOwnership(databaseFileUser.get());
          if (error)
          {
+            error.addProperty("server-user", databaseFileUser.get().getUsername());
             error.addProperty("description", "Unable to change ownership of database file");
             return error;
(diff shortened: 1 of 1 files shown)


src/cpp/server_core/ServerDatabase.cpp
```diff
@@ -151,7 +151,23 @@ Error readOptions(const std::string& databaseConfigFile,
          // always ensure the database file user is correct, if specified
          error = databaseFile.changeOwnership(databaseFileUser.get());
          if (error)
+         {
+            error.addProperty("description", "Unable to change ownership of database file");
```
MariaSemple, line 155: Should we include which user we're trying to change the owner to?

src/cpp/server_core/ServerDatabase.cpp
```diff
@@ -151,7 +151,23 @@ Error readOptions(const std::string& databaseConfigFile,
...
             return error;
+         }
+
+#ifndef _WIN32
+         error = databaseDirectory.changeOwnership(databaseFileUser.get());
```
MariaSemple, line 160: on non windows platforms, this line is duplicated (on successful tests) Should we move the `bool writable = false ...` block inside the first `if (error)` block and avoid calling it twice?

src/cpp/server_core/ServerDatabase.cpp
```diff
@@ -151,7 +151,23 @@ Error readOptions(const std::string& databaseConfigFile,
...
+         if (error)
+         {
+            bool writable = false;
+            Error writableError = databaseDirectory.isWriteable(writable);
+            if (writableError || !writable)
+            {
+                LOG_ERROR_MESSAGE("Sqllite database directory: " + databaseDirectory.getAbsolutePath() + " must be writable by: " + databaseFileUser.get().getUsername());
```
MariaSemple, line 167: Shouldn't we exit with an error after logging this message?

MariaSemple (changes requested):

src/cpp/server_core/ServerDatabase.cpp
jeffvroom: The first changeOwnership call is for databaseFile. This one is for databaseDirectory... due to some weird sqlite requirement that the directory itself be writable.

src/cpp/server_core/ServerDatabase.cpp
jeffvroom: For OS, it seems like things work pretty well even with a read-only database. I was concerned about adding a new "exit error" that might trip up some users that don't care about the DB.

src/cpp/server_core/ServerDatabase.cpp
MariaSemple: Ah! Such close names! Makes sense to me!

src/cpp/server_core/ServerDatabase.cpp
MariaSemple: I see, that seems reasonable.

MariaSemple (approved):
LGTM!

PR #9747, opened 2021-08-17
use web dialogs for input prompts w/Electron

kevinushey:
### Intent

Use the existing GWT `promptForText()` functions when prompting for text in Electron.

### Approach

Detect whether we're running within Electron, and if we are, use the web dialogs rather than desktop dialogs.

### Automated Tests

Should be covered by any existing automation that tests input prompts for RStudio Server.

### QA Notes

N/A

### Checklist

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/files/Files.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/files/Files.java
@@ -525,15 +525,11 @@ public class Files
       ArrayList<FileSystemItem> selectedFiles = view_.getSelectedFiles();
 
       // validation: some selection exists
-      if  (selectedFiles.size() == 0)
-         return;
-
-      // validation: no more than one file selected
-      if  (selectedFiles.size() > 1)
+      if (selectedFiles.size() != 1)
       {
          globalDisplay_.showErrorMessage(
-                           "Invalid Selection",
-                           "Please select only one file to rename");
+               "Invalid Selection",
+               "Please select a single file to rename.");
          return;
       }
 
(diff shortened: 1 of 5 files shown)


src/gwt/src/org/rstudio/studio/client/workbench/views/files/Files.java
```diff
@@ -525,15 +525,11 @@ void onRenameFile()
       ArrayList<FileSystemItem> selectedFiles = view_.getSelectedFiles();
 
       // validation: some selection exists
-      if  (selectedFiles.size() == 0)
```
gtritchie, line 528: In the interest of reducing risk for Ghost Orchid, I'd recommend not making any changes in GWT that aren't strictly needed by this Electron tweak, so I'd leave out these tweaks to Files.java.

Unlikely to cause any problems but it is a slight change in behavior which could break an integration test (didn't look to see if we have one for this code), generating more noise.

gtritchie (approved):
LGTM, except per my comment I'd not change Files.java given we're very late in the Ghost Orchid endgame. Unless it's needed by this Electron tweak, somehow, and I'm missing it.

src/gwt/src/org/rstudio/studio/client/workbench/views/files/Files.java
gtritchie: I'm being overly nitpicky, merging!

kevinushey: For context, I decided to make this change after seeing that clicking the Rename button without anything selected just silently did nothing, which is a somewhat confusing UI experience. This is marginally better, I think?

PR #9759, opened 2021-08-20
highlight color for yaml in python node

jjallaire:
### Intent

Match special highlighting for YAML chunk options in R mode within Python mode.

### Approach

Added a highlighting rule that runs before the default default highlighting rule for Python comments.


kevinushey:
LGTM!

PR #9787, opened 2021-08-27
Add a "Create New File" menu to the Files view. 

GCRev:
Issue: https://github.com/rstudio/rstudio/issues/1564

These commands create new empty files with the selected type, and opens them automatically

A new set of "touch" commands have been created, reminiscent of the touch command on unix systems. These commands make use of a new RPC function of the same name, which just wraps existing "ensure file" functionality, only with a check that the target file does not exist beforehand.

New RDoc and New RPres have been made unavailable, as both would require a lot more overhead. If we want to add those two back in to this menu, then we need further discussion about how the Files pane should interact with the rest of the application.

It looks like a lot of changes because of all the copied icons...

For QA:
Make sure to test on directories with restrictive permissions and also that a file with the same name as another cannot be created in the same folder.


src/gwt/src/org/rstudio/studio/client/workbench/views/files/Files.java
```diff
@@ -777,13 +838,75 @@ private void showFileInBrowser(FileSystemItem file)
...
+      FileSystemItem newTempFile = getDefaultFileName(fileType);
+      
+      // guard for reentrancy
+      if (inputPending_)
+         return;
+      
+      inputPending_ = true;
+
+      // prompt for new file name then execute the operation
+      globalDisplay_.promptForText("Create and Open New File",
```
jmcphers, line 865: Since you already have the TextFileType, we could add the description of the file type to this dialog (e.g. "Create and Open New R Markdown File", "Create and Open New JSON File", etc.) What do you think?

src/gwt/src/org/rstudio/studio/client/workbench/views/files/Files.java
```diff
@@ -777,13 +838,75 @@ private void showFileInBrowser(FileSystemItem file)
...
+            // no longer waiting for user to input
+            inputPending_ = false;
+
+            progress.onProgress("Creating file...");
+
+            String path = currentPath_.completePath(input);
+            final FileSystemItem newFile = FileSystemItem.createFile(path);
+
+            // execute on the server
+            server_.touchFile(newFile, new VoidServerRequestCallback(progress)
```
jmcphers, line 884: I think we should add an error handler here so we can show something better than a generic error message.

src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml
```diff
@@ -2334,6 +2334,126 @@ well as menu structures (for main menu and popup menus).
...
+   <cmd id="touchRDocumentationDoc"
+        menuLabel="R Doc_umentation..."
+        desc="Create and open a new Rd documentation file"
+        rebindable="false"/>
+
+   <cmd id="touchSqlDoc"
+        menuLabel="SQ_L Script"
+        desc="Create and open a new SQL script"
+        rebindable="false"/>
+
```
melissa-barca, line 2456: I think we should make the description more distinct from the original new file commands. If a user accesses these via the Command Palette it will be confusing when the only difference is the words "and open". Maybe something like "Create a new <type> file in open directory"?

src/gwt/src/org/rstudio/studio/client/workbench/views/files/model/FilesServerOperations.java
```diff
@@ -74,6 +74,10 @@ void moveFiles(ArrayList<FileSystemItem> files,
    void renameFile(FileSystemItem file,
                    FileSystemItem targetFile,
                    ServerRequestCallback<Void> serverRequestCallback);
+   
+   // touch file
```
melissa-barca, line 78: Nit: this comment doesn't add anything more than the function's name 
```suggestion
```

src/gwt/src/org/rstudio/studio/client/workbench/views/files/ui/FileCommandToolbar.java
```diff
@@ -38,6 +38,39 @@ public FileCommandToolbar(Commands commands, UserPrefs prefs)
...
+      newFileMenu.addItem(commands.touchPythonDoc().createMenuItem(false));
+      newFileMenu.addItem(commands.touchSqlDoc().createMenuItem(false));
+      newFileMenu.addItem(commands.touchStanDoc().createMenuItem(false));
+      newFileMenu.addItem(commands.touchD3Doc().createMenuItem(false));
+      newFileMenu.addSeparator();
+      newFileMenu.addItem(commands.touchSweaveDoc().createMenuItem(false));
+      newFileMenu.addItem(commands.touchRHTMLDoc().createMenuItem(false));
+
+      ToolbarMenuButton newFileButton = new ToolbarMenuButton(
+            "Create New File",
```
melissa-barca, line 66: Nit: To match the "New Folder" button should we call this "New File"?

melissa-barca (changes requested):

melissa-barca:
Overall, great work getting this together so quickly! I pulled the branch to play around with it and noticed a few things. 

1. When I try to a create a new R Notebook, the file gets created but not opened, and I see these errors: ​image
2. When creating a new R Markdown document, you don't see the "New R Markdown" widget that gives you different options for the template included in the file. Would it be significant work to add this?
3. When creating a Shiny Web App or Plumber API, you're given their unique widgets which include directory browsers. Could you auto-populate these with the active directory from the Files Pane?

This may be my personal preference, but I think the snake case in the new file name looks a little clunky. What do you think about switching it to camel case? Or perhaps upper camel case with the first word being "Untitled" instead of "new" to match the naming scheme of opening files from the drop down? e.g. UntitledCPPFile.cpp, UntitledRNWFile.rnw - you might even consider remove the word "File".

src/gwt/src/org/rstudio/studio/client/workbench/views/files/Files.java
GCRev: I like that. I'll implement this for the next revision.

src/gwt/src/org/rstudio/studio/client/workbench/views/files/Files.java
GCRev: On the RPC request or on the creation of the fileSystemItem itself?

jmcphers: On the RPC request.

src/gwt/src/org/rstudio/studio/client/workbench/views/files/ui/FileCommandToolbar.java
GCRev: I wanted to distinguish between the existing "New File" operation in the File menu, and this action. I'll go the consistency path -- also fewer words is less UI clutter.

GCRev:
@melissa-barca I didn't notice that R-markdown is another multi-stage file creation process, but I think that this would require modifying the dialog to support specifying a file path up-front, or finagling something to support that. I'll see what I can do.

For item 3, it was already proving unexpectedly difficult to copy the current working directory from the files pane into the creation dialogs, but I can take another crack at it.

GCRev:
After a discussion about how to proceed with regards to multi-step document creation workflows, we have decided that this functionality is here specifically to create blank files immediately, and open them for editing. The RNotebook file is represented using a .Rmd proxy file, which requires a template in order to open, so we have removed the command to touch a new blank notebook (.nb[.html]) file (because such a think doesn't really exist).

RMarkdowns can still be created, but they are created and opened as totally blank files without any templates or additional supporting files.

src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml
GCRev: Verbiage will be changed to: "Create a blank {fileType} in current directory"

src/gwt/src/org/rstudio/studio/client/workbench/views/files/model/FilesServerOperations.java
GCRev: This was done for consistency with the other comments for the other functions.

I will remove this, because I also agree that the function name is self-explanatory. Unless I hear otherwise, I won't remove the other comments -- only this one.

src/gwt/src/org/rstudio/studio/client/workbench/views/files/ui/FileCommandToolbar.java
GCRev: This has been changed to "New Blank File" to make it totally clear that it will create a blank file without applying any code templates.

src/gwt/src/org/rstudio/studio/client/workbench/views/files/Files.java
GCRev: I've overridden the default error display associated with the progress indicator with one that describes that the error occurred specifically in the file creation process.

This should help locate the error better, especially if the server returns a generic (or generic-looking) error string.

image

melissa-barca (approved):
LGTM!

jmcphers (approved):
LGTM!

PR #9834, opened 2021-09-10
Add commands to open selected files in columns or active editor

GCRev:
See: https://github.com/rstudio/rstudio/issues/7920

A couple of implementation notes about this, to keep in mind while reviewing why there are so many changes: 
 * There is an interesting constraint that files can only be opened in a column if the column exists beforehand. However, there are cases in which we don't know whether the file can open until _after_ we have attempted to open it. The changes I've made are just there to validate whether it is possible to create a new column.
 * Most of the code added is there simply to pre-screen the file selection to make sure that no extra columns are opened, or no blank columns are opened, particularly in the case of notebooks.
 * There is also a lot of code in place just to perform the Rmd document extraction ahead of time, to prevent opening new empty columns.
 * Adding a way to track which column to target when submitting files to the open-file queue was necessary, otherwise files would just open in the last available active column due to a race condition.

The behavior should still intuitive to the user: 
 * the columns are opened in the order that the files are listed in the file viewer, or however the getSelectedFiles() method returns them. 
 * If the user selects more files than there are remaining columns, then the remaining files that cannot be opened in columns are opened in the last column as tabs. 
 * A new column will not open if the user tries to open a file that is already open in another column, but that file will not count towards the column limit.

The new options are located in the "More" menu
image

After opening the previous selection there are three columns, where the final column contains two tabs:
image


melissa-barca:
Could you add some screenshots of the new behavior and menu to the PR's description?

src/gwt/src/org/rstudio/studio/client/workbench/views/files/Files.java
```diff
@@ -607,6 +622,176 @@ void onCopyFilesPaneCurrentDirectory()
...
+         }
+      }
+
+      view_.selectNone();
+   }
+
+   // the purpose of this function is specifically to pre-screen selected files for the following:
+   //  - the file must NOT be a directory
+   //  - the file must NOT already be open in a source editor 
+   //  - if the file is an RNotebook, then make sure that only ONE .RMarkdown file is opened for editing
```
melissa-barca, line 645: We typically use javadoc style comments to comment functions.

src/gwt/src/org/rstudio/studio/client/workbench/views/files/Files.java
```diff
@@ -607,6 +622,176 @@ void onCopyFilesPaneCurrentDirectory()
...
+               selectedFilePaths.add(item.getPath());
+         }
+      }
+
+      for (FileSystemItem notebook : notebooks)
+      {
+         commandList.addCommand(new SerializedCommand()
+               {
+                  @Override
+                  public void onExecute(final Command continuation)
```
melissa-barca, line 687: Nit here & elsewhere: typically we do not include the additional indentation when the line only contains a bracket. 
```suggestion
         {
            @Override
            public void onExecute(final Command continuation)
```

src/gwt/src/org/rstudio/studio/client/workbench/views/files/Files.java
```diff
@@ -607,6 +622,176 @@ void onCopyFilesPaneCurrentDirectory()
...
+                           {
+                              @Override
+                              public void onSuccess(SourceDocumentResult doc)
+                              {
+                                 // this means the operation succeeded; add ONLY the Rmd file to the list
+                                 // if it is not already open in a source editor
+                                 final FileSystemItem rmdFile = FileSystemItem.createFile(rmdPath);
+                                 if (!selectedFilePaths.contains(rmdPath) 
+                                       && !mgr.openFileAlreadyOpen(rmdFile, null))
+                                    selectedFilePaths.add(rmdPath);
```
melissa-barca, line 704:
```suggestion
                                 if (!selectedFilePaths.contains(rmdPath) &&
                                     !mgr.openFileAlreadyOpen(rmdFile, null))
                                    selectedFilePaths.add(rmdPath);
```
Nit: more often we put the operator at the beginning of the line rather than the end.

src/gwt/src/org/rstudio/studio/client/workbench/views/files/Files.java
```diff
@@ -607,6 +622,176 @@ void onCopyFilesPaneCurrentDirectory()
...
+
+                                 continuation.execute();
+                              }
+
+                              @Override
+                              public void onFailure(ServerError error)
+                              {
+                                 // fail silently?
```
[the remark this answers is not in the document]
GCRev: We have a couple options on how to handle an error (or multiple). I agree that we should show something on failure, but I see two ways to handle this: 
 1. we can cancel the entire process at the first error and pop that up
 2. we can aggregate the errors into a list and show all of them at the very end. 

The first approach is more immediate and provides the user with a smaller error message to read, which is good for reducing clutter, however, if they fix the file then attempt to open the same items again and encounter another error, they'll have to stop and resolve a second (or more) error.

Normally I would lean towards reducing clutter, but speaking from experience dealing with sequential errors, I'd much rather see all errors at once and know the scope of the problem up front, than have to resolve errors one after another.

src/gwt/src/org/rstudio/studio/client/workbench/views/files/ui/FilePathToolbar.java
```diff
@@ -119,6 +119,13 @@ public FilePathToolbar(Files.Display.NavigationObserver navigationObserver, bool
       CheckBoxHiddenLabel selectAllCheckBox = new CheckBoxHiddenLabel("Select all files",
                                                                       ElementIds.FP_SELECT_ALL);
       selectAllCheckBox.addStyleDependentName("FilesSelectAll");
+      
+      // TODO: there needs to be some way to two-way couple the selection state
+      // with the selection in the files list itself.  
+
```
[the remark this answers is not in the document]
GCRev: I'll remove this comment and look into filing an issue. It may not be worth it.

src/gwt/src/org/rstudio/studio/client/workbench/views/files/Files.java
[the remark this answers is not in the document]
melissa-barca: I agree with your assessment and think it makes the most sense to go with the second approach.

GCRev:
RNotebook opening errors are now aggregated and displayed all at once as such:
image

Any errors will stop the operation as a whole from completing. I have also wrapped the "Open Selected Files in Source Pane" command with the same helper function, so it performs the same checks and shows errors, only without opening in new columns.

melissa-barca (approved):
LGTM!

src/gwt/src/org/rstudio/studio/client/workbench/views/source/SourceColumnManager.java
```diff
@@ -2332,9 +2347,54 @@ public void onError(ServerError error)
...
+                             final SourceColumn targetColumn,
+                             final ResultCallback<EditingTarget, ServerError> resultCallback)
+   {
+      // construct path to .Rmd
+      final String rnbPath = rnbFile.getPath();
+      final String rmdPath = FilePathUtils.filePathSansExtension(rnbPath) + ".Rmd";
+      final FileSystemItem rmdFile = FileSystemItem.createFile(rmdPath);
+
+      // if we already have associated .Rmd file open, then just edit it
+      // TODO: should we perform conflict resolution here as well?
```
gtritchie, line 2360: Are you OK leaving this as a TODO, or should it be investigated further now?

gtritchie (approved):
One generic question on a TODO, but otherwise LGTM!

src/gwt/src/org/rstudio/studio/client/workbench/views/source/SourceColumnManager.java
melissa-barca: I believe this TODO already exists in our codebase and just looks new due to git's faulty diffing.

GCRev: Correct, this comment was left in, but because I moved some code around it looks like it was appended.

PR #9845, opened 2021-09-14
append the daily link if it's not found

jeffvroom:
Look for daily link in .htaccess and if not found append it.

I tested this locally to be sure it replaces and appends URLs for some example OS and pro links.


jmcphers (approved):

MariaSemple (approved):
LGTM!

PR #9876, opened 2021-09-22
Explain TTY Logging on Startup

jonyoder:
Connected rstudio/launcher#73.

### Intent

When configured to log to `file` or `syslog` but running in a TTY, adds a log like the following:
> Logging configuration loaded from '[path to logging.conf]'. Logging to '[path to log file | "syslog"]'. Displaying this since a TTY was detected.

This was requested in rstudio/launcher#73 to clarify why you sometimes see STDERR logs even when configured to log to a file or syslog.

### Approach

When initializing the main log writer, we check to see if we're running in a TTY and output a message to STDERR if (a) running in a TTY and (b) not configured to log to STDERR.

### Automated Tests

Unit tests didn't seem practical due to the complexity of mocking or simulating running in (or not in) a TTY. There are a couple of options, but none seem ideal, so I skipped unit tests:
* Update `stderrIsTerminal()` to look for a test env var that overrides the result. This clutters production code with test conditionals.
* Create a mockable interface that includes a new `stderrIsTerminal()` method. Inject an instance of this class, but use a mock implementation for unit tests. This is a bit complicated since we'd need to reference it from a number of methods.
* Use `Pseudoterminal` to launch a process during a unit teset. This also seems like overkill.

### QA Notes

* Configure `logging.conf` at various supported locations. The new STDERR message should include the correct path to `logging.conf` in any case.
* Configure `logging.conf` with an `[@rserver]` (or other binary) section. When running the process in a TTY, the new STDERR message should appear *unless* configured explicitly to log to STDERR. The message should indicate that you are configured to log to either `file` or `syslog` correctly.
* Configure `logging.confg` with an `[*]` section. When running the process in a TTY, the new STDERR message should appear *unless* configured explicitly to log to STDERR.

### Checklist

- [x] If this PR adds a new feature, or fixes a bug in a previously released version, it includes an entry in `NEWS.md` (n/a)
- [x] If this PR adds or changes UI, the updated UI meets [accessibility standards](https://github.com/rstudio/rstudio/wiki/Accessibility) (n/a)

### DIFF
--- a/src/cpp/core/LogOptions.cpp
+++ b/src/cpp/core/LogOptions.cpp
@@ -262,6 +262,23 @@ void LogOptions::initProfile()
 
 Error LogOptions::read()
 {
+   FilePath optionsFile = getLogConfigFile();
+
+   // if the options file does not exist, that's fine - we'll just use default values
+   if (!optionsFile.exists())
+      return Success();
+
+   Error error = profile_.load(optionsFile);
+   if (error)
+      return error;
+
+   setLowestLogLevel();
+
+   return Success();
+}
+
+FilePath LogOptions::getLogConfigFile() {
+
    // look for config file in a specific environment variable
    FilePath optionsFile(core::system::getenv(kLogConfEnvVar));
    if (!optionsFile.exists())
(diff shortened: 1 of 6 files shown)


colearendt:
FWIW, my vote on the message is:

> TTY detected. Printing informational message about logging configuration. Logging configuration loaded from '/etc/rstudio/logging.conf'. Logging to 'syslog'.

jonyoder:
@colearendt 
> TTY detected. Printing informational message about logging configuration. Logging configuration loaded from '/etc/rstudio/logging.conf'. Logging to 'syslog'.

I feel a bit odd about saying "Printing informational message..." and then printing the informational message. To me, it feels like the informational message is obvious enough that we don't need to say that we're going to print it. 🤣 But maybe I'm missing something important.

colearendt:
What if we said “Only printing informational message...” 

The key is that “the message I am printing here is special and is only getting printed because I detected a TTY.”

jmcphers (approved):
I think we could just print the informational message. There's no need to say we're about to print it, or even that we printed it due to stderr being a tty. But not work bikeshedding over the language; I don't think it'll actually confuse anyone.

PR #9877, opened 2021-09-22
Feature/db session metadata

AndrewMcClain:
### Intent

Lay the groundwork for storing session information in a database. This has the potential to resolve a number of bugs and pain points around session management.

### Approach

I've encapsulated the on disk session metadata storage to limit the knowledge that calling classes have about how the session metadata is stored. An interface and factory have been created so that in the future switching to other storage options will be opaque to the callers.

### Automated Tests

This does not include additional automated testing.

### QA Notes

There are two things to be checked by QA:

- The new licensed_users and active_session_metadata tables will exist in a fresh install, and migration from previous databases will create them. This should be true of both sqlite and postgresql
- That there are no regressions surrounding session metadata. Examples to test: where it's written to disk, make sure the files are being updated as expected.

### Checklist

### DIFF
--- a/src/cpp/core/Database.cpp
+++ b/src/cpp/core/Database.cpp
@@ -826,6 +826,7 @@ const std::map<std::string, int>& SchemaVersion::versionMap()
          {
             versions[""] = 0;
             versions["Ghost Orchid"] = 1;
+            versions["Prairie Trillium"] = 2;
          }
       }
       END_LOCK_MUTEX
@@ -868,7 +869,6 @@ Error SchemaUpdater::migrationFiles(std::vector<std::pair<SchemaVersion, FilePat
                         const std::pair<SchemaVersion, FilePath>& b)
    { return a.first > b.first; };
    std::sort(pMigrationFiles->begin(), pMigrationFiles->end(), comparator);
-
    return Success();
 }
 
@@ -1020,6 +1020,7 @@ Error SchemaUpdater::isUpToDate(bool* pUpToDate)
       return error;
 
    *pUpToDate = version >= migrationVersion;
+
    return Success();
 }
 
@@ -1039,6 +1040,10 @@ Error SchemaUpdater::update()
 
       SchemaVersion currentVersion;
       error = databaseSchemaVersion(&currentVersion);
+
+      LOG_DEBUG_MESSAGE("Database Schema Version:\t\t" + currentVersion.Date + " : " + currentVersion.Flower);
+      LOG_DEBUG_MESSAGE("Highest Migration Schema Version:\t" + migrationVersion.Date + " : " + migrationVersion.Flower);
+
       if (currentVersion < migrationVersion)
       {
          LOG_INFO_MESSAGE(
@@ -1135,6 +1140,13 @@ Error SchemaUpdater::updateToVersion(const SchemaVersion& maxVersion)
    for (const std::pair<SchemaVersion, FilePath>& migrationFile : files)
    {
 
+      // Is the migration file from a update that predates the current version?
+      // If so, skip it
+      if(migrationFile.first <= currentVersion)
+      {
+         continue;
+      }
+
       bool applyMigration = false;
 
       if (migrationFile.second.getExtensionLowerCase() == SQL_EXTENSION)
@@ -1168,6 +1180,7 @@ Error SchemaUpdater::updateToVersion(const SchemaVersion& maxVersion)
       error = connection_->executeStr(fileContents);
       if (error)
          return error;
+
    }
 
    transaction.commit();
--- a/src/cpp/core/Database.cpp
+++ b/src/cpp/core/Database.cpp
@@ -1156,6 +1156,8 @@ Error SchemaUpdater::updateToVersion(const SchemaVersion& maxVersion)
       if (!applyMigration)
          continue;
 
+      LOG_DEBUG_MESSAGE("Applying database schema alter file " + migrationFile.second.getAbsolutePath());
+
       // we are clear to apply the migration
       // load the file and execute its SQL contents
       std::string fileContents;
--- a/src/cpp/core/include/core/r_util/RActiveSessionStorage.hpp
+++ b/src/cpp/core/include/core/r_util/RActiveSessionStorage.hpp
@@ -1,0 +1,65 @@
+#ifndef CORE_R_UTIL_R_ACTIVE_SESSION_STORAGE
+#define CORE_R_UTIL_R_ACTIVE_SESSION_STORAGE
+
+#include <shared_core/FilePath.hpp>
+
+namespace rstudio {
+namespace core {
+namespace r_util {
+
+   class IActiveSessionStorage 
+   {
+   public:
+      Error virtual readProperty(const std::string& id, const std::string& name, std::string* pValue) = 0;
+      Error virtual writeProperty(const std::string& id, const std::string& name, const std::string& value) = 0;
+
+   protected:
+      virtual ~IActiveSessionStorage() = default;
+      IActiveSessionStorage() = default;
+   };
+
+   class LegacySessionStorage : public IActiveSessionStorage
+   {
+   public:
+      explicit LegacySessionStorage(const FilePath& location);
+
+      Error readProperty(const std::string& id, const std::string& name, std::string* pValue) override;   
+      Error writeProperty(const std::string& id, const std::string& name, const std::string& value) override;
+
+   private:
+      FilePath activeSessionsDir_;
+      const std::string propertiesDirName_ = "properites";
+      const std::string legacySessionDirPrefix_ = "session-";
+
+      FilePath buildPropertyPath(const std::string& id, const std::string& name);
+
+      static const std::string& getLegacyName(const std::string& name)
+      {
+         static const std::map<std::string, std::string> legacyNames = 
+         {
+            { "last_used" , "last-used" },
+            { "r_version" , "r-version" },
+            { "r_version_label" , "r-version-label" },
+            { "r_version_home" , "r-version-home" },
+            { "working_directory" , "working-dir" },
+            { "launch_parameters" , "launch-parameters" }
+         };
+
+         if (legacyNames.find(name) != legacyNames.end())
+            return legacyNames.at(name);
+
+         return name;
+      }
+   };
+
+   class ActiveSessionStorageFactory
+   {
+   public:
+      static std::shared_ptr<IActiveSessionStorage> getActiveSessionStorage();
+      static std::shared_ptr<IActiveSessionStorage> getLegacyActiveSessionStorage();
+   };
+} // namespace r_util
+} // namespace core
+} // namespace rstudio
+
+#endif // CORE_R_UTIL_ACTIVE_SESSIONS_STORAGE_HPP
--- a/src/cpp/core/include/core/r_util/RActiveSessions.hpp
+++ b/src/cpp/core/include/core/r_util/RActiveSessions.hpp
@@ -1,10 +1,10 @@
 /*
  * RActiveSessions.hpp
  *
- * Copyright (C) 2021 by RStudio, PBC
+ * Copyright (C) 2021 by Rstudio, PBC
  *
- * Unless you have received this program directly from RStudio pursuant
- * to the terms of a commercial license agreement with RStudio, then
+ * Unless you have received this program directly from Rstudio pursuant
+ * to the terms of a commercial license agreement with Rstudio, then
  * this program is licensed to you under the terms of version 3 of the
  * GNU Affero General Public License. This program is distributed WITHOUT
  * ANY EXPRESS OR IMPLIED WARRANTY, INCLUDING THOSE OF NON-INFRINGEMENT,
@@ -28,6 +28,7 @@
 
 #include <core/r_util/RSessionContext.hpp>
 #include <core/r_util/RProjectFile.hpp>
+#include <core/r_util/RActiveSessionStorage.hpp>
 
 namespace rstudio {
 namespace core {
@@ -36,27 +37,45 @@ namespace r_util {
 class ActiveSession : boost::noncopyable
 {
 private:
-   friend class ActiveSessions;
-   ActiveSession() {}
 
-   explicit ActiveSession(const std::string& id) : id_(id) 
+   friend class ActiveSessions;
+   explicit ActiveSession(std::shared_ptr<IActiveSessionStorage> storage) {}
+
+   ActiveSession(
+      std::shared_ptr<IActiveSessionStorage> storage,
+      const std::string& id) :
+         storage_(storage),
+         id_(id) 
    {
    }
 
-   explicit ActiveSession(const std::string& id, const FilePath& scratchPath)
-      : id_(id), scratchPath_(scratchPath)
+   ActiveSession(
+      std::shared_ptr<IActiveSessionStorage> storage,
+      const std::string& id,
+      const FilePath& scratchPath) : 
+         storage_(storage),
+         id_(id),
+         scratchPath_(scratchPath)
    {
       core::Error error = scratchPath_.ensureDirectory();
       if (error)
          LOG_ERROR(error);
-
-      propertiesPath_ = scratchPath_.completeChildPath("properites");
-      error = propertiesPath_.ensureDirectory();
-      if (error)
-         LOG_ERROR(error);
    }
 
-public:
+   std::string const kExecuting = "executing";
+   std::string const kInitial = "initial";
+   std::string const kLabel = "label";
+   std::string const kLastUsed = "last_used";
+   std::string const kProject = "project";
+   std::string const kSavePromptRequired = "save_prompt_required";
+   std::string const kSessionSuspendData = "suspended_session_data";
+   std::string const kRunning = "running";
+   std::string const kRVersion = "r_version";
+   std::string const kRVersionHome = "r_version_home";
+   std::string const kRVersionLabel = "r_version_label";
+   std::string const kWorkingDir = "working_directory";
+
+ public:
 
    bool empty() const { return scratchPath_.isEmpty(); }
 
@@ -66,37 +85,59 @@ public:
 
    std::string project() const
    {
+      std::string project;
       if (!empty())
-         return readProperty("project");
-      else
-         return std::string();
+      {
+         Error error = storage_->readProperty(id_, kProject, &project);
+         if (error)
+            LOG_ERROR(error);
+      }
+
+      return project;
    }
 
    void setProject(const std::string& project)
    {
       if (!empty())
-         writeProperty("project", project);
+      {
+         Error error = storage_->writeProperty(id_, kProject, project);
+         if (error)
+            LOG_ERROR(error);
+      }
    }
 
    std::string workingDir() const
    {
+      std::string value;
       if (!empty())
-         return readProperty("working-dir");
-      else
-         return std::string();
+      {
+         Error error = storage_->readProperty(id_, kWorkingDir, &value);
+         if (error)
+            LOG_ERROR(error);
+      }
+      return value;
    }
 
    void setWorkingDir(const std::string& workingDir)
    {
       if (!empty())
-         writeProperty("working-dir", workingDir);
+      {
+         Error error = storage_->writeProperty(id_, kWorkingDir, workingDir);
+         if (error)
+            LOG_ERROR(error);
+      }
    }
 
    bool initial() const
    {
       if (!empty())
       {
-         std::string value = readProperty("initial");
+         std::string value;
+         Error error = storage_->readProperty(id_, kInitial, &value);
+         
+         if (error)
+            LOG_ERROR(error);
+
          if (!value.empty())
             return safe_convert::stringTo<bool>(value, false);
          else
@@ -116,25 +157,32 @@ public:
       if (!empty())
       {
          std::string value = safe_convert::numberToString(initial);
-         writeProperty("initial", value);
+         Error error = storage_->writeProperty(id_, kInitial, value);
+         if (error)
+            LOG_ERROR(error);
       }
    }
 
    double lastUsed() const
    {
-      return timestampProperty("last-used");
+      return timestampProperty(kLastUsed);
    }
 
    void setLastUsed()
    {
-      setTimestampProperty("last-used");
+      setTimestampProperty(kLastUsed);
    }
 
    bool executing() const
    {
       if (!empty())
       {
-         std::string value = readProperty("executing");
+         std::string value = "";
+         Error error = storage_->readProperty(id_, kExecuting, &value);
+
+         if (error)
+            LOG_ERROR(error);
+
          if (!value.empty())
             return safe_convert::stringTo<bool>(value, false);
          else
@@ -149,7 +197,10 @@ public:
       if (!empty())
       {
          std::string value = safe_convert::numberToString(executing);
-         writeProperty("executing", value);
+         Error error = storage_->writeProperty(id_, kExecuting, value);
+
+         if (error)
+            LOG_ERROR(error);
       }
    }
 
@@ -157,7 +208,12 @@ public:
    {
       if (!empty())
       {
-         std::string value = readProperty("save_prompt_required");
+         std::string value = "";
+         Error error = storage_->readProperty(id_, kSavePromptRequired, &value);
+
+         if (error)
+            LOG_ERROR(error);
+
          if (!value.empty())
             return safe_convert::stringTo<bool>(value, false);
          else
@@ -172,7 +228,10 @@ public:
       if (!empty())
       {
          std::string value = safe_convert::numberToString(savePromptRequired);
-         writeProperty("save_prompt_required", value);
+         Error error = storage_->writeProperty(id_, kSavePromptRequired, value);
+
+         if (error)
+            LOG_ERROR(error);
       }
    }
 
@@ -181,7 +240,12 @@ public:
    {
       if (!empty())
       {
-         std::string value = readProperty("running");
+         std::string value = "";
+         Error error = storage_->readProperty(id_, kRunning, &value);
+
+         if (error)
+            LOG_ERROR(error);
+
          if (!value.empty())
             return safe_convert::stringTo<bool>(value, false);
          else
@@ -193,26 +257,42 @@ public:
 
    std::string rVersion()
    {
+      std::string value;
       if (!empty())
-         return readProperty("r-version");
-      else
-         return std::string();
+      {  
+         Error error = storage_->readProperty(id_, kRVersion, &value);
+
+         if (error)
+            LOG_ERROR(error);
+      }
+      return value;
    }
 
    std::string rVersionLabel()
    {
+      std::string value;
       if (!empty())
-         return readProperty("r-version-label");
-      else
-         return std::string();
+      {
+         Error error = storage_->readProperty(id_, kRVersionLabel, &value);
+
+         if (error)
+            LOG_ERROR(error);
+         
+      }
+      return value;
    }
 
    std::string rVersionHome()
    {
+      std::string value;
       if (!empty())
-         return readProperty("r-version-home");
-      else
-         return std::string();
+      {
+         Error error = storage_->readProperty(id_, kRVersionHome, &value);
+         
+         if (error)
+            LOG_ERROR(error);
+      }
+      return std::string();
    }
 
    void setRVersion(const std::string& rVersion,
@@ -221,26 +301,47 @@ public:
    {
       if (!empty())
       {
-         writeProperty("r-version", rVersion);
-         writeProperty("r-version-home", rVersionHome);
-         writeProperty("r-version-label", rVersionLabel);
+         Error error = storage_->writeProperty(id_, kRVersion, rVersion);
+
+         if (error)
+            LOG_ERROR(error);
+
+         error = storage_->writeProperty(id_, kRVersionHome, rVersionHome);
+
+         if (error)
+            LOG_ERROR(error);
+
+         error = storage_->writeProperty(id_, kRVersionLabel, rVersionLabel);
+
+         if (error)
+            LOG_ERROR(error);
       }
    }
 
    // historical note: this will be displayed as the session name
    std::string label()
    {
+      std::string value;
       if (!empty())
-         return readProperty("label");
-      else
-         return std::string();
+      {
+         Error error = storage_->readProperty(id_, kLabel, &value);
+         
+         if (error)
+            LOG_ERROR(error);
+      }
+      return value;
    }
 
    // historical note: this will be displayed as the session name
    void setLabel(const std::string& label)
    {
       if (!empty())
-         writeProperty("label", label);
+      {
+         Error error = storage_->writeProperty(id_, kLabel, label);
+
+         if (error)
+            LOG_ERROR(error);
+      }
    }
 
    void beginSession(const std::string& rVersion,
@@ -257,15 +358,6 @@ public:
       setLastUsed();
       setRunning(false);
       setExecuting(false);
-   }
-
-   uintmax_t suspendSize()
-   {
-      FilePath suspendPath = scratchPath_.completePath("suspended-session-data");
-      if (!suspendPath.exists())
-         return 0;
-
-      return suspendPath.getSizeRecursive();
    }
 
    core::Error destroy()
@@ -360,7 +452,7 @@ public:
       {
          double now = date_time::millisecondsSinceEpoch();
          std::string value = safe_convert::numberToString(now);
-         writeProperty(property, value);
+        storage_->writeProperty(id_, property, value);
       }
    }
 
@@ -368,7 +460,12 @@ public:
    {
       if (!empty())
       {
-         std::string value = readProperty(property);
+         std::string value = "";
+         Error error = storage_->readProperty(id_, property, &value);
+
+         if (error)
+            LOG_ERROR(error);
+
          if (!value.empty())
             return safe_convert::stringTo<double>(value, 0);
          else
@@ -386,20 +483,19 @@ public:
       if (!empty())
       {
          std::string value = safe_convert::numberToString(running);
-         writeProperty("running", value);
+         Error error = storage_->writeProperty(id_, kRunning, value);
+
+         if(error) 
+            LOG_ERROR(error);
       }
    }
 
-   void writeProperty(const std::string& name, const std::string& value) const;
-   std::string readProperty(const std::string& name) const;
-
-private:
+   std::shared_ptr<IActiveSessionStorage> storage_;
    std::string id_;
    FilePath scratchPath_;
    FilePath propertiesPath_;
    SortConditions sortConditions_;
 };
-
 
 class ActiveSessions : boost::noncopyable
 {
@@ -407,6 +503,7 @@ public:
    explicit ActiveSessions(const FilePath& rootStoragePath)
    {
       storagePath_ = storagePath(rootStoragePath);
+      storage_ = ActiveSessionStorageFactory::getActiveSessionStorage();
       Error error = storagePath_.ensureDirectory();
       if (error)
          LOG_ERROR(error);
@@ -440,10 +537,12 @@ public:
 
    FilePath storagePath() const { return storagePath_; }
 
-   static boost::shared_ptr<ActiveSession> emptySession(const std::string& id);
+   boost::shared_ptr<ActiveSession> emptySession(
+      const std::string& id) const;
 
 private:
    core::FilePath storagePath_;
+   std::shared_ptr<IActiveSessionStorage> storage_;
 };
 
 // active session as tracked by rserver processes
@@ -490,7 +589,8 @@ private:
    core::FilePath rootPath_;
 };
 
-void trackActiveSessionCount(const FilePath& rootStoragePath,
+void trackActiveSessionCount(std::shared_ptr<IActiveSessionStorage> storage,
+                             const FilePath& rootStoragePath,
                              const FilePath& userHomePath,
                              bool projectSharingEnabled,
                              boost::function<void(size_t)> onCountChanged);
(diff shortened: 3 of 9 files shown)


src/cpp/core/Database.cpp
```diff
@@ -1135,6 +1140,13 @@ Error SchemaUpdater::updateToVersion(const SchemaVersion& maxVersion)
    for (const std::pair<SchemaVersion, FilePath>& migrationFile : files)
    {
 
+      // Is the migration file from a update that predates the current version?
```
MariaSemple, line 1143: Nit:
```suggestion
      // Is the migration file from an update that predates the current version?
```

src/cpp/core/Database.cpp
```diff
@@ -1135,6 +1140,13 @@ Error SchemaUpdater::updateToVersion(const SchemaVersion& maxVersion)
...
+      // If so, skip it
+      if(migrationFile.first <= currentVersion)
+      {
+         continue;
+      }
```
MariaSemple, line 1148:
```suggestion
      if(migrationFile.first <= currentVersion)
         continue;
```

src/cpp/core/include/core/r_util/RActiveSessionStorage.hpp
```diff
@@ -0,0 +1,80 @@
...
+   {
+   public:
+      static std::shared_ptr<IActiveSessionStorage> getActiveSessionStorage();
+      static std::shared_ptr<IActiveSessionStorage> getLegacyActiveSessionStorage();
+   };
+} // namespace r_util
+} // namespace core
+} // namespace rstudio
+
+#endif // CORE_R_UTIL_ACTIVE_SESSIONS_STORAGE_HPP
```
MariaSemple, line 80:
```suggestion
#endif // CORE_R_UTIL_ACTIVE_SESSIONS_STORAGE_HPP

```

src/cpp/core/include/core/r_util/RActiveSessionStorage.hpp
```diff
@@ -0,0 +1,80 @@
...
+         return name;
+      }
+   };
+
+   class ActiveSessionStorageFactory
+   {
+   public:
+      static std::shared_ptr<IActiveSessionStorage> getActiveSessionStorage();
+      static std::shared_ptr<IActiveSessionStorage> getLegacyActiveSessionStorage();
+   };
```
MariaSemple, line 75:
```suggestion
   };

```

src/cpp/core/include/core/r_util/RActiveSessionStorage.hpp
```diff
@@ -0,0 +1,80 @@
...
+   protected:
+      virtual ~IActiveSessionStorage() = default;
+      IActiveSessionStorage() = default;
+   };
+
+   class LegacySessionStorage : public IActiveSessionStorage
+   {
+   public:
+      explicit LegacySessionStorage(const FilePath& location);
+
```
MariaSemple, line 40: I think we may also need an explicit instantiation of the virtual destructor here, since destructors are special
```suggestion

virtual ~LegacySessionStorage() = default;
```

src/cpp/core/include/core/r_util/RActiveSessions.hpp
```diff
@@ -1,10 +1,10 @@
 /*
  * RActiveSessions.hpp
  *
- * Copyright (C) 2021 by RStudio, PBC
+ * Copyright (C) 2021 by Rstudio, PBC
  *
- * Unless you have received this program directly from RStudio pursuant
- * to the terms of a commercial license agreement with RStudio, then
+ * Unless you have received this program directly from Rstudio pursuant
+ * to the terms of a commercial license agreement with Rstudio, then
```
MariaSemple, line 7: Looks like this got included by mistake

src/cpp/core/include/core/r_util/RActiveSessions.hpp
```diff
@@ -66,37 +85,59 @@ class ActiveSession : boost::noncopyable
...
    {
+      std::string project;
       if (!empty())
-         return readProperty("project");
-      else
-         return std::string();
+      {
+         Error error = storage_->readProperty(id_, kProject, &project);
+         if (error)
+            LOG_ERROR(error);
```
MariaSemple, line 93: To avoid duplication, I would put this implementation into `ActiveSession::readProperty` and leave these methods the same.

MariaSemple: The same for writing properties

src/cpp/core/include/core/r_util/RActiveSessionStorage.hpp
```diff
@@ -0,0 +1,80 @@
...
+            return legacyNames.at(name);
+
+         return name;
+      }
+   };
+
+   class ActiveSessionStorageFactory
+   {
+   public:
+      static std::shared_ptr<IActiveSessionStorage> getActiveSessionStorage();
```
MariaSemple, line 73: What's the purpose of returning shared pointers here, as opposed to unique pointers? With shared pointers, we'll need to consider multi-threading, but as I think about this more, maybe we need to consider multi-threading anyway, since the file can change on disk while another thread is reading it, even if you're not using shared pointers (i.e if two storages are in use for the same session).

What do you think about a minor refactor so that the session storage accepts an ID, the `IActiveSessionStorage` becomes `AbstractSessionStorage`, it keeps the ID in a protected member, and it uses a ReaderWriterMutex (there's a pre-existing class for this somewhere in core or shared_core), and calls into the private, inheritable read/write functions after locking?

Then the factory would keep a map of session id->shared_ptr (for which you will also need a mutex, but a regular one will do) and if the storage for that session already exists, return it instead of creating a new one. The only thing left to consider is how to clean unused sessions out of the map. The most self contained way would be to use [`std::shared_ptr::unique`](https://en.cppreference.com/w/cpp/memory/shared_ptr/unique) to check whether you're the only one referencing that object and delete it from the map, if so. That could either be done as part of any invocation of `getActiveSessionStorage` or `getLegacyActiveSessionStorage` or on a regular, timed callback.

src/cpp/core/r_util/RActiveSessionStorage.cpp
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    std::shared_ptr<IActiveSessionStorage> ActiveSessionStorageFactory::getLegacyActiveSessionStorage()
+    {
+        FilePath dataDir = ActiveSessions::storagePath(core::system::xdg::userDataDir());
+        return std::make_shared<LegacySessionStorage>(LegacySessionStorage(dataDir));
+    }
+} // namespace r_util
+} // namepsace core
+} // namespace rstudio
```
MariaSemple, line 80:
```suggestion
} // namespace rstudio

```

src/cpp/core/r_util/RActiveSessionStorage.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+        FilePath readPath = buildPropertyPath(id, legacyName);
+        if (readPath.exists())
+        {
+            Error error = core::readStringFromFile(readPath, pValue);
+            if (error)
+                return error;
+
+            boost::algorithm::trim(*pValue);
+        }
```
MariaSemple, line 51:
```suggestion
        }

```

src/cpp/core/r_util/RActiveSessionStorage.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+#include <core/r_util/RActiveSessionStorage.hpp>
+#include <core/r_util/RActiveSessions.hpp>
+#include <core/FileSerializer.hpp>
+#include <core/Log.hpp>
+#include <core/system/Xdg.hpp>
+
+
+
+namespace rstudio {
```
MariaSemple, line 24:
```suggestion
#include <core/system/Xdg.hpp>

namespace rstudio {
```

src/cpp/server/db/20210916132211194382021_Prairie-Trillium_AlterTables.postgresql
```diff
@@ -0,0 +1,88 @@
...
+      FOREIGN KEY (user_id) 
+         REFERENCES licensed_users(id) 
+         ON DELETE CASCADE
+);
+
+/* Remove previous schema version data */
+TRUNCATE schema_version;
+
+/* Update the version to the latest after all alter statements are applied. */
+INSERT INTO schema_version (current_version, release_name) VALUES ('20210916132211194382021', 'Prairie Trillium');
```
MariaSemple, line 88:
```suggestion
UPDATE schema_version SET current_version = '20210916132211194382021' , release_name = 'Prairie Trillium';

```

src/cpp/server/db/20210916132211194382021_Prairie-Trillium_AlterTables.postgresql
```diff
@@ -0,0 +1,88 @@
...
+   label text NOT NULL,
+
+   /* The parameters with which this session was launched */
+   launch_parameters text NOT NULL,
+
+   /* Whether or not a save prompt is required */
+   save_prompt_required text NOT NULL DEFAULT 'not_required',
+
+   /* The size of the session on disk, if it is curruently suspended */
+   suspended_size_kb integer,
```
MariaSemple, line 73: While I was working on this, I found that the suspended size isn't written to a file on disk - it's the actual size of the session when it's suspended on disk. Since we need to read/write to that file for session suspension/resumption, I think we should always go to the ultimate source of truth for this value.

Also, IIRC, the value doesn't come from the ActiveSession object, so the current implementation plan doesn't include a way to read/write this value from the Session itself.
```suggestion
```

src/cpp/server/db/20210916132211194382021_Prairie-Trillium_AlterTables.postgresql
```diff
@@ -0,0 +1,88 @@
...
+   
+   /* User entries should not be deleted, but if they are, remove all of that user's sessions */
+   CONSTRAINT fk_user
+      FOREIGN KEY (user_id) 
+         REFERENCES licensed_users(id) 
+         ON DELETE CASCADE
+);
+
+/* Remove previous schema version data */
+TRUNCATE schema_version;
```
MariaSemple, line 85: This should not be necessary - better to do it all in one step in the update
```suggestion```

File: src/cpp/server/db/20210916132211194382021_Prairie-Trillium_AlterTables.sqlite [hunk 15]
```diff
@@ -0,0 +1,81 @@
...
+   
+   /* User entries should not be deleted, but if they are, remove all of that user's sessions */
+   FOREIGN KEY (user_id) REFERENCES licensed_users(id) ON DELETE CASCADE
+);
+
+/* Remove old schema version */
+DELETE FROM schema_version;
+
+/* Update the version to the latest after all alter statements are applied. */
+INSERT INTO schema_version (current_version, release_name) VALUES ('20210916132211194382021', 'Prairie Trillium');
```
MariaSemple, line 81:
```suggestion
/* Update the version to the latest after all alter statements are applied. */
UPDATE schema_version SET current_version = '20210916132211194382021', release_name = 'Prairie Trillium';
```

src/cpp/server/db/20210916132211194382021_Prairie-Trillium_AlterTables.sqlite
```diff
@@ -0,0 +1,81 @@
...
+   label text NOT NULL,
+
+   /* The parameters with which this session was launched */
+   launch_parameters text NOT NULL,
+
+   /* Whether or not a save prompt is required */
+   save_prompt_required text NOT NULL DEFAULT 'not_required',
+
+   /* The size of the session on disk, if it is curruently suspended */
+   suspended_size_kb integer,
```
MariaSemple, line 69: See note above
```suggestion
```

src/cpp/server/db/CreateTables.sqlite
```diff
@@ -24,7 +24,83 @@ CREATE TABLE revoked_cookie(
...
+   label text NOT NULL,
+
+   /* The parameters with which this session was launched */
+   launch_parameters text NOT NULL,
+
+   /* Whether or not a save prompt is required */
+   save_prompt_required text NOT NULL DEFAULT 'not_required',
+
+   /* The size of the session on disk, if it is curruently suspended */
+   suspended_size_kb integer,
```
MariaSemple, line 95:
```suggestion
```

src/cpp/server/db/CreateTables.postgresql
```diff
@@ -26,7 +26,90 @@ CREATE TABLE revoked_cookie(
...
+   label text NOT NULL,
+
+   /* The parameters with which this session was launched */
+   launch_parameters text NOT NULL,
+
+   /* Whether or not a save prompt is required */
+   save_prompt_required text NOT NULL DEFAULT 'not_required',
+
+   /* The size of the session on disk, if it is curruently suspended */
+   suspended_size_kb integer,
```
MariaSemple, line 101:
```suggestion
```

MariaSemple (changes requested):
Great progress!

src/cpp/core/include/core/r_util/RActiveSessionStorage.hpp
AndrewMcClain: I'll take more time to look through and let LegacySessionStorage to become more complicated. In support of future database and rpc request activities, factory could produce a single instance of SessionStorage. This would necessitate that the session storage object concerns itself with thread safety.

LegacySessionStorage would need a mutex per session id. May not matter, session writes / server reads. Investigate
Think about this more.

src/cpp/core/include/core/r_util/RActiveSessionStorage.hpp
```diff
@@ -0,0 +1,81 @@
...
+   public:
+      Error virtual readProperty(const std::string& id, const std::string& name, std::string* pValue) = 0;
+      Error virtual writeProperty(const std::string& id, const std::string& name, const std::string& value) = 0;
+
+   protected:
+      virtual ~IActiveSessionStorage() = default;
+      IActiveSessionStorage() = default;
+   };
+
+   class LegacySessionStorage : public IActiveSessionStorage
```
jeffvroom, line 36: Could we call this FileSessionStorage?  It's a little more descriptive and a little less judgmental :)

src/cpp/core/include/core/r_util/RActiveSessionStorage.hpp
```diff
@@ -0,0 +1,81 @@
...
+      Error writeProperty(const std::string& id, const std::string& name, const std::string& value) override;
+
+   private:
+      FilePath activeSessionsDir_;
+      const std::string propertiesDirName_ = "properites";
+      const std::string legacySessionDirPrefix_ = "session-";
+
+      FilePath buildPropertyPath(const std::string& id, const std::string& name);
+
+      static const std::string& getLegacyName(const std::string& name)
```
jeffvroom, line 51:
```suggestion
      static const std::string& getPropertyFileName(const std::string& name)
```

src/cpp/core/include/core/r_util/RActiveSessions.hpp
AndrewMcClain: Reading and writing properties encapsulated to avoid repeated error handling and pivoting

src/cpp/core/include/core/r_util/RActiveSessionStorage.hpp
AndrewMcClain: Refactored from legacy to PropertyFile

src/cpp/core/include/core/r_util/RActiveSessionStorage.hpp
AndrewMcClain: I settled on FileActiveSessionStorage since the base is ActiveSessionStorage, but the spirit is the same, thanks!

src/cpp/core/include/core/r_util/RActiveSessionStorage.hpp
jeffvroom: Apologies for chiming in late here or repeating these comments. I thought I already made them but don't see them here now.

- As far as I can tell, the SessionStorage object itself is ok to be global/static - i.e. we only use it from the session and the fields never change. So I agree with Maria that we don't need the shared_ptr since that implies multiple clients.  
- I don't have a recommendation for fixing this now, but it's unfortunate from the performance perspective that we have readProperty/writeProperty calls, without any batching - i.e. no way to call readProperties and writeProperties.  With the DB version, each of these will turn into round-trips of the pipeline: rsession -> rserver -> postgres -> commit.  Right now, it's bad enough with rsession -> nfs-file and we are going to make it worse. If each hop is 10ms, 3 calls in a row is only 30ms but if latency for any hop gets high (like 200ms) it can really slow things down. Maybe the RPC apis should support multiple properties and we can make it easier for performance tuning?  I suppose it's also something we could address later...

AndrewMcClain: @jeffvroom I'm glad you brought that up, I've been meaning to ping you, but so many other things cropped up. One thing that I have hoped to do, and it's influencing the thinking around the FileActiveSessionStorage, is that database or rpc instances should be used by multiple clients, and encapsulate a batching mechanism to reduce the round trip calls, and the number of open connections to the database or the server. I wanted to ask you for your opinions, I'm worried about the latency, especially to the remote rpc, and the open connections to the database when we have multiple sessions involved.

The rpc apis already are specified to read or write multiple properties at a time, so that will reduce the number of round trips necessary, but it's currently only able to batch together changes per session, I don't know if that's enough batching, or if we need to go further.

src/cpp/core/Database.cpp
```diff
@@ -1039,6 +1040,10 @@ Error SchemaUpdater::update()
 
       SchemaVersion currentVersion;
       error = databaseSchemaVersion(&currentVersion);
+
+      LOG_DEBUG_MESSAGE("Database Schema Version:\t\t" + currentVersion.Date + " : " + currentVersion.Flower);
+      LOG_DEBUG_MESSAGE("Highest Migration Schema Version:\t" + migrationVersion.Date + " : " + migrationVersion.Flower);
```
MariaSemple, line 1045: Should we log this as all one message? Or maybe even just some minor wording changes
```suggestion
      LOG_DEBUG_MESSAGE("Current Database Schema Version:\t\t" + currentVersion.Date + " : " + currentVersion.Flower);
      LOG_DEBUG_MESSAGE("Highest Available Database Schema Version:\t" + migrationVersion.Date + " : " + migrationVersion.Flower);
```

src/cpp/core/include/core/r_util/RActiveSessions.hpp
```diff
@@ -36,67 +37,101 @@ namespace r_util {
...
+   std::string const kLabel = "label";
+   std::string const kLastUsed = "last_used";
+   std::string const kProject = "project";
+   std::string const kSavePromptRequired = "save_prompt_required";
+   std::string const kSessionSuspendData = "suspended_session_data";
+   std::string const kRunning = "running";
+   std::string const kRVersion = "r_version";
+   std::string const kRVersionHome = "r_version_home";
+   std::string const kRVersionLabel = "r_version_label";
+   std::string const kWorkingDir = "working_directory";
```
MariaSemple, line 76: Nit
```suggestion
   const std::string kExecuting = "executing";
   const std::string kInitial = "initial";
   const std::string kLabel = "label";
   const std::string kLastUsed = "last_used";
   const std::string kProject = "project";
   const std::string kSavePromptRequired = "save_prompt_required";
   const std::string kSessionSuspendData = "suspended_session_data";
   const std::string kRunning = "running";
   const std::string kRVersion = "r_version";
   const std::string kRVersionHome = "r_version_home";
   const std::string kRVersionLabel = "r_version_label";
   const std::string kWorkingDir = "working_directory";
```

src/cpp/core/include/core/r_util/RActiveSessions.hpp
```diff
@@ -36,67 +37,101 @@ namespace r_util {
...
+ public:
 
    bool empty() const { return scratchPath_.isEmpty(); }
 
    std::string id() const { return id_; }
 
    const FilePath& scratchPath() const { return scratchPath_; }
 
-   std::string project() const
+   std::string getProperty(std::string const propertyName) const
```
MariaSemple, line 86: I think you probably meant to pass const references here, rather than const copies. Also, to avoid added git history noise, can we stick with the original name? We'll have less differences further down in the file
```suggestion
   std::string readProperty(const std::string& propertyName) const
```

src/cpp/core/include/core/r_util/RActiveSessions.hpp
```diff
@@ -36,67 +37,101 @@ namespace r_util {
...
+         Error error = storage_->readProperty(id_, propertyName, &value);
+         if (error)
+            LOG_ERROR(error);
+      }
+
+      return value;
    }
 
-   void setProject(const std::string& project)
+   void setProperty(std::string const propertyName, std::string const value)
```
MariaSemple, line 99: Same as for `getProperty`
```suggestion
   void writeProperty(const std::string& propertyName, const std::string& value)
```

src/cpp/core/include/core/r_util/RActiveSessions.hpp
```diff
@@ -490,7 +466,8 @@ class GlobalActiveSessions : boost::noncopyable
    core::FilePath rootPath_;
 };
 
-void trackActiveSessionCount(const FilePath& rootStoragePath,
+void trackActiveSessionCount(std::shared_ptr<IActiveSessionStorage> storage,
```
MariaSemple, line 469:
```suggestion
void trackActiveSessionCount(const std::shared_ptr<IActiveSessionStorage>& storage,
```

src/cpp/server/db/20210916132211194382021_Prairie-Trillium_AlterTables.postgresql
```diff
@@ -0,0 +1,82 @@
...
+   
+   /* User entries should not be deleted, but if they are, remove all of that user's sessions */
+   CONSTRAINT fk_user
+      FOREIGN KEY (user_id) 
+         REFERENCES licensed_users(id) 
+         ON DELETE CASCADE
+);
+
+/* Update the version to the latest after all alter statements are applied. */
+UPDATE schema_version SET current_version = '20210916132211194382021' , release_name = 'Prairie Trillium';
```
MariaSemple, line 82:
```suggestion
UPDATE schema_version SET current_version = '20210916132211194382021' , release_name = 'Prairie Trillium';

```

src/cpp/core/r_util/RActiveSessions.cpp
```diff
@@ -181,17 +148,18 @@ boost::shared_ptr<ActiveSession> ActiveSessions::get(const std::string& id) cons
...
+                                                                id,
                                                                 scratchPath));
    else
       return emptySession(id);
 }
 
 
 boost::shared_ptr<ActiveSession> ActiveSessions::emptySession(
-      const std::string& id)
+   const std::string& id) const
```
MariaSemple, line 160: Nit: this probably doesn't need a line break

src/cpp/core/include/core/r_util/RActiveSessions.hpp
MariaSemple: Actually, this function isn't actually implemented with this change, because the definition now has different parameters. Because your code still compiles & links, I think we must not be using this. Can you try deleting this function (here and in the cpp) and see if everything still compiles?

src/cpp/core/r_util/RActiveSessionStorage.cpp
```diff
@@ -0,0 +1,114 @@
...
+            LOG_ERROR(error);
+    }
+
+    Error FileActiveSessionStorage::readProperty(const std::string& id, const std::string& name, std::string* pValue)
+    {
+        std::set<std::string> propertyName = {name};
+        std::map<std::string, std::string> propertyValue{};
+        *pValue = "";
+
+        Error error = readProperties(id, propertyName, &propertyValue);
```
MariaSemple, line 40: Nit (for less lines - is a debatable nit)
```suggestion
        std::map<std::string, std::string> propertyValue{};
        *pValue = "";

        Error error = readProperties(id, { name }, &propertyValue);
```

src/cpp/core/r_util/RActiveSessionStorage.cpp
```diff
@@ -0,0 +1,114 @@
...
+    Error FileActiveSessionStorage::writeProperties(const std::string& id, const std::map<std::string, std::string>& properties)
+    {
+        Error error;
+        for (const std::pair<std::string, std::string> &prop : properties)
+        {
+            FilePath writePath = buildPropertyPath(id, prop.first);
+            error = core::writeStringToFile(writePath, prop.second);
+            if (error)
+                return error;
+        }
```
MariaSemple, line 91: Nit
```suggestion
        for (const std::pair<std::string, std::string> &prop : properties)
        {
            FilePath writePath = buildPropertyPath(id, prop.first);
            Error error = core::writeStringToFile(writePath, prop.second);
            if (error)
                return error;
        }
```

src/cpp/core/r_util/RActiveSessionStorage.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+    Error FileActiveSessionStorage::writeProperties(const std::string& id, const std::map<std::string, std::string>& properties)
+    {
+        Error error;
+        for (const std::pair<std::string, std::string> &prop : properties)
+        {
+            FilePath writePath = buildPropertyPath(id, prop.first);
+            error = core::writeStringToFile(writePath, prop.second);
+            if (error)
+                return error;
```
MariaSemple, line 90: Right now it looks like if we fail writing a single property, we won't write any more. Should we just log the error here and try to continue on, maybe returning a group error at the end?

src/cpp/core/r_util/RActiveSessionStorage.cpp
```diff
@@ -0,0 +1,114 @@
...
+        {
+            const std::string& fileName = getPropertyFileName(name);
+            FilePath readPath = buildPropertyPath(id, fileName);
+            std::string value = "";
+
+            if (readPath.exists())
+            {
+                Error error = core::readStringFromFile(readPath, &value);
+                if (error)
+                    return error;
```
MariaSemple, line 66: Similar comment as in `writeProperties` - should we set any missing values as `""` and log an error for them, so we don't fail to read other properties?

src/cpp/core/include/core/r_util/RActiveSessions.hpp
```diff
@@ -36,67 +37,101 @@ namespace r_util {
...
+      {
+         Error error = storage_->writeProperty(id_, propertyName, value);
+         if (error)
+            LOG_ERROR(error);
+      }
+   }
+
+   std::string project() const
+   {
+      return getProperty(kProject);
```
MariaSemple, line 111:
```suggestion
      return readProperty(kProject);
```

src/cpp/core/include/core/r_util/RActiveSessions.hpp
```diff
@@ -36,67 +37,101 @@ namespace r_util {
...
+   }
+
+   void setProject(const std::string& project)
+   {
+      setProperty(kProject, project);
```
MariaSemple, line 116:
```suggestion
      writeProperty(kProject, project);
```

src/cpp/core/include/core/r_util/RActiveSessions.hpp
```diff
@@ -36,67 +37,101 @@ namespace r_util {
...
    }
 
    std::string workingDir() const
    {
-      if (!empty())
-         return readProperty("working-dir");
-      else
-         return std::string();
+      return getProperty(kWorkingDir);
```
MariaSemple, line 121:
```suggestion
      return readProperty(kWorkingDir);
```

src/cpp/core/include/core/r_util/RActiveSessions.hpp
```diff
@@ -36,67 +37,101 @@ namespace r_util {
...
    }
 
    void setWorkingDir(const std::string& workingDir)
    {
-      if (!empty())
-         writeProperty("working-dir", workingDir);
+      setProperty(kWorkingDir, workingDir);
```
MariaSemple, line 126:
```suggestion
      writeProperty(kWorkingDir, workingDir);
```

src/cpp/core/include/core/r_util/RActiveSessions.hpp
```diff
@@ -36,67 +37,101 @@ namespace r_util {
...
    }
 
    bool initial() const
    {
       if (!empty())
       {
-         std::string value = readProperty("initial");
+         std::string value = getProperty(kInitial);
```
MariaSemple, line 133:
```suggestion
         std::string value = readProperty(kInitial);
```

src/cpp/core/include/core/r_util/RActiveSessions.hpp
```diff
@@ -113,134 +148,97 @@ class ActiveSession : boost::noncopyable
 
    void setInitial(bool initial)
    {
-      if (!empty())
-      {
-         std::string value = safe_convert::numberToString(initial);
-         writeProperty("initial", value);
-      }
+      std::string value = safe_convert::numberToString(initial);
+      setProperty(kInitial, value);
```
MariaSemple, line 152:
```suggestion
      writeProperty(kInitial, safe_convert::numberToString(initial));
```

src/cpp/core/include/core/r_util/RActiveSessions.hpp
```diff
@@ -113,134 +148,97 @@ class ActiveSession : boost::noncopyable
...
    {
-      if (!empty())
-      {
-         std::string value = readProperty("executing");
-         if (!value.empty())
-            return safe_convert::stringTo<bool>(value, false);
-         else
-            return false;
-      }
+      std::string value = getProperty(kExecuting);
```
MariaSemple, line 167:
```suggestion
      std::string value = readProperty(kExecuting);
```

src/cpp/core/include/core/r_util/RActiveSessions.hpp
```diff
@@ -113,134 +148,97 @@ class ActiveSession : boost::noncopyable
...
 
    void setExecuting(bool executing)
    {
-      if (!empty())
-      {
-         std::string value = safe_convert::numberToString(executing);
-         writeProperty("executing", value);
-      }
+      std::string value = safe_convert::numberToString(executing);
+      setProperty(kExecuting, value);
```
MariaSemple, line 178:
```suggestion
      writeProperty(kExecuting, safe_convert::numberToString(executing));
```

src/cpp/core/include/core/r_util/RActiveSessions.hpp
```diff
@@ -113,134 +148,97 @@ class ActiveSession : boost::noncopyable
...
    {
-      if (!empty())
-      {
-         std::string value = readProperty("save_prompt_required");
-         if (!value.empty())
-            return safe_convert::stringTo<bool>(value, false);
-         else
-            return false;
-      }
+      std::string value = getProperty(kSavePromptRequired);
```
MariaSemple, line 183:
```suggestion
      std::string value = readProperty(kSavePromptRequired);
```

src/cpp/core/include/core/r_util/RActiveSessions.hpp
```diff
@@ -113,134 +148,97 @@ class ActiveSession : boost::noncopyable
...
    }
 
    void setSavePromptRequired(bool savePromptRequired)
    {
-      if (!empty())
-      {
          std::string value = safe_convert::numberToString(savePromptRequired);
-         writeProperty("save_prompt_required", value);
-      }
+         setProperty(kSavePromptRequired, value);
```
MariaSemple, line 194:
```suggestion
         writeProperty(kSavePromptRequired, value);
```

src/cpp/core/include/core/r_util/RActiveSessions.hpp
```diff
@@ -113,134 +148,97 @@ class ActiveSession : boost::noncopyable
...
    {
-      if (!empty())
-      {
-         std::string value = readProperty("running");
-         if (!value.empty())
-            return safe_convert::stringTo<bool>(value, false);
-         else
-            return false;
-      }
+      std::string value = getProperty(kRunning);
```
MariaSemple, line 200:
```suggestion
      std::string value = readProperty(kRunning);
```

src/cpp/core/include/core/r_util/RActiveSessions.hpp
```diff
@@ -113,134 +148,97 @@ class ActiveSession : boost::noncopyable
...
          return false;
    }
 
    std::string rVersion()
    {
-      if (!empty())
-         return readProperty("r-version");
-      else
-         return std::string();
+      return getProperty(kRVersion);
```
MariaSemple, line 210:
```suggestion
      return readProperty(kRVersion);
```

src/cpp/core/include/core/r_util/RActiveSessions.hpp
```diff
@@ -113,134 +148,97 @@ class ActiveSession : boost::noncopyable
...
    }
 
    std::string rVersionLabel()
    {
-      if (!empty())
-         return readProperty("r-version-label");
-      else
-         return std::string();
+      return getProperty(kRVersionLabel);
```
MariaSemple, line 215:
```suggestion
      return readProperty(kRVersionLabel);
```

src/cpp/core/include/core/r_util/RActiveSessions.hpp
```diff
@@ -113,134 +148,97 @@ class ActiveSession : boost::noncopyable
...
    }
 
    std::string rVersionHome()
    {
-      if (!empty())
-         return readProperty("r-version-home");
-      else
-         return std::string();
+      return getProperty(kRVersionHome);
```
MariaSemple, line 220:
```suggestion
      return readProperty(kRVersionHome);
```

src/cpp/core/include/core/r_util/RActiveSessions.hpp
```diff
@@ -113,134 +148,97 @@ class ActiveSession : boost::noncopyable
...
    {
-      if (!empty())
-      {
-         writeProperty("r-version", rVersion);
-         writeProperty("r-version-home", rVersionHome);
-         writeProperty("r-version-label", rVersionLabel);
-      }
+         setProperty(kRVersion, rVersion);
+         setProperty(kRVersionHome, rVersionHome);
+         setProperty(kRVersionLabel, rVersionLabel);
```
MariaSemple, line 229:
```suggestion
         writeProperty(kRVersion, rVersion);
         writeProperty(kRVersionHome, rVersionHome);
         writeProperty(kRVersionLabel, rVersionLabel);
```

src/cpp/core/include/core/r_util/RActiveSessions.hpp
```diff
@@ -113,134 +148,97 @@ class ActiveSession : boost::noncopyable
...
    }
 
    // historical note: this will be displayed as the session name
    std::string label()
    {
-      if (!empty())
-         return readProperty("label");
-      else
-         return std::string();
+      return getProperty(kLabel);
```
MariaSemple, line 235:
```suggestion
      return readProperty(kLabel);
```

src/cpp/core/include/core/r_util/RActiveSessions.hpp
```diff
@@ -113,134 +148,97 @@ class ActiveSession : boost::noncopyable
...
    }
 
    // historical note: this will be displayed as the session name
    void setLabel(const std::string& label)
    {
-      if (!empty())
-         writeProperty("label", label);
+      setProperty(kLabel, label);
```
MariaSemple, line 241:
```suggestion
      writeProperty(kLabel, label);
```

src/cpp/core/include/core/r_util/RActiveSessions.hpp
```diff
@@ -356,57 +345,42 @@ class ActiveSession : boost::noncopyable
...
    {
-      if (!empty())
-      {
-         double now = date_time::millisecondsSinceEpoch();
-         std::string value = safe_convert::numberToString(now);
-         writeProperty(property, value);
-      }
+      double now = date_time::millisecondsSinceEpoch();
+      std::string value = safe_convert::numberToString(now);
+      setProperty(property, value);
```
MariaSemple, line 350:
```suggestion
      writeProperty(property, value);
```

src/cpp/core/include/core/r_util/RActiveSessions.hpp
```diff
@@ -356,57 +345,42 @@ class ActiveSession : boost::noncopyable
...
    {
-      if (!empty())
-      {
-         std::string value = readProperty(property);
-         if (!value.empty())
-            return safe_convert::stringTo<double>(value, 0);
-         else
-            return 0;
-      }
+      std::string value = getProperty(property);
```
MariaSemple, line 355:
```suggestion
      std::string value = readProperty(property);
```

src/cpp/core/include/core/r_util/RActiveSessions.hpp
```diff
@@ -356,57 +345,42 @@ class ActiveSession : boost::noncopyable
...
 
 
    void setRunning(bool running)
    {
-      if (!empty())
-      {
          std::string value = safe_convert::numberToString(running);
-         writeProperty("running", value);
-      }
+         setProperty(kRunning, value);
```
MariaSemple, line 367:
```suggestion
         writeProperty(kRunning, value);
```

MariaSemple (changes requested):

src/cpp/core/include/core/r_util/RActiveSessionStorage.hpp
```diff
@@ -51,25 +53,33 @@ namespace r_util {
       const std::string propertiesDirName_ = "properites";
       const std::string fileSessionDirPrefix_ = "session-";
 
+      FilePath buildPropertyDir(const std::string& id);
       FilePath buildPropertyPath(const std::string& id, const std::string& name);
```
MariaSemple: Nit: In order to contrast more with `buildPropertyDir` I might change `buildPropertyPath` to `buildPropertyFile` although I might change them both to `get` instead of `build`. Also, is it possible to make them `const`?

PR #9885, opened 2021-09-24
run GWT (Java) build step after C++ builds complete

gtritchie:
### Intent

Addresses: #7660

When doing a full build (e.g. package build), the GWT portion builds in parallel with the C++ code, and this can cause thrashing and resource exhaustion that leads to a GWT build failure. Our answer has been to constrain number of cores allowed for building (`-j#`), e.g. for Mac `make-package` we set `-j2` because of this.

### Approach

Make the GWT code dependent (cmake-wise) on the major C++ components; rsession, rstudio (desktop), and/or rserver so it waits for them to complete, then can give Java all available resources.

Once this gets through a round of builds or two, I'll go back and bump up the `-j` value for Mac a bit (after finding out how many cores the builder actually has...).

### Automated Tests

None, purely a build-time thing.

I tested by doing a local Mac desktop build via make-package, and a Bionic Server build via docker.

### QA Notes

Nothing to test here; problems would manifest as build failures.

### Checklist

### DIFF
--- a/src/gwt/CMakeLists.txt
+++ b/src/gwt/CMakeLists.txt
@@ -61,6 +61,19 @@ if(GWT_BUILD)
 
    # invoke ant to build
    add_custom_target(gwt_build ALL)
+
+   # wait for major C++ work to complete before running the GWT build;
+   # the Java compiler often runs out of resources when a parallel C++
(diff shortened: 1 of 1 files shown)


jmcphers (approved):
I've wished for this so many times but always thought it'd be difficult. Nice work!!

Issue #7868, opened 2020-09-23
RStudio failing to find libR.so


gavinsimpson:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.3.1073 (provided by Fedora's repos)
    OS Version      : Linux (Fedora, 32)
    R Version       : 4.0.2

### Steps to reproduce the problem

Copy the script below to `~/bin/R` (or somewhere on your path, or wherever user scripts live) and make it executable
Start RStudio
(it runs `~/bin/R` see below)
RStudio fails to find `libR.so`
Raises Dialogue box
RStudio closes when you click OK

### Describe the problem in detail

I start R via a shell script `~/bin/R` with an extra note to indicate which BLAS is being used and to preload that BLAS when starting R (This is the suggested way to swap out the BLAS on Fedora so I can use other BLAS options when developing or doing research). The script I use is:

```
#!/bin/sh

echo "Starting R with Atlas BLAS support"
LD_PRELOAD=/lib64/atlas/libsatlas.so.3 \
/usr/bin/R  ${args} "${@}"
```

When I start RStudio it fails to locate `libR.so`, produces a dialogue box with the following information:

>    Unable to find libR.so in expected locations within R Home directory Starting R with Atlas BLAS support
>    /usr/lib64/R; R home path (Starting R with Atlas BLAS support
>    /usr/lib64/R) not found

If I comment out the line `echo "Starting R with Atlas BLAS support"` in the script RStudio starts up normally.

### Describe the behavior you expected

I expect RStudio to start normally and find R

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ ] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting a RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
RStudio tries to query information about installed versions of R by invoking the moral equivalent of:

```
R --vanilla -s -e <...>
```

and then attempts to parse that output. I'm guessing interleaving your own messages is causing issues with that process.

Enchufa2:
[...]

PR #9893, opened 2021-09-27
Add check for the LibR.so that rsession believes it's using

cm421:
### Intent

Addresses https://github.com/rstudio/rstudio-pro/issues/2477 by adding a check on the `libR.so` that `rsession` believes it is running. If there are issues with loading that shared object, generate an error.

Note that this does not fix the problem. Instead it creates an error that can be used to find and fix the cause of the problem, which is a system environment issue

### Approach

The linked issue describes the problem in detail, but essentially while linux is loading the `rsession` binary, if there's an issue with `libR.so`, it may load a different `libR.so` without `rsession`'s knowledge. To address this, we check for issues manually once `rsession` is fully loaded by calling `dload()` with `RTLD_NOLOAD`. Any issues Linux had with loading the shared object will become apparent with that call, and the error can then be forwarded on to an error.

This is linux-only code; Windows and MacOS won't perform this check.

### Automated Tests

There are no unit tests included. Integration tests will require:

- A linux environment with at least 2 `R` installations
  - One system `R` install
  - One custom `R` install
- Ability to "disable" the custom `R` version's `libR.so` before the test by `chmod`ing out permissions so it can't be opened, and then restoring the permissions after the test is done

The test should include

- Start an `rsession` with the default system `R` install
  - No error should be generated
- Switch `R` version to the broken custom version
  - Error should be generated 

### QA Notes

None, see testing section above for how to reproduce/test

### Checklist

### DIFF
--- a/src/cpp/core/system/PosixLibraryLoader.cpp
+++ b/src/cpp/core/system/PosixLibraryLoader.cpp
@@ -53,6 +53,32 @@ Error loadLibrary(const std::string& libPath, void** ppLib)
    }
 }
 
+/**
+ * Verify that we can load a library (without actually loading it)
+ * by passing RTLD_NOLOAD to dlopen
+ */
+Error verifyLibrary(const std::string& libPath, void** ppLib)
+{
+   *ppLib = nullptr;
+   *ppLib = ::dlopen(libPath.c_str(), RTLD_LAZY | RTLD_NOLOAD);
+   if (*ppLib == nullptr)
+   {
+      // dlopen returns a null pointer on error rather than an error number.
+      // using -1 here as a generic non-zero placeholder
+      Error error = Error("LibraryVerifyFailed",
+                          -1,
+                          "Attempt to verify library failed",
+                          ErrorLocation());
+      error.addProperty("lib-path", libPath);
+      addLastDLErrorMessage(&error);
+      return error;
+   }
+   else
+   {
+      return closeLibrary(*ppLib);
+   }
+}
+
 Error loadSymbol(void* pLib, const std::string& name, void** ppSymbol)
 {
    *ppSymbol = nullptr;
--- a/src/cpp/core/system/PosixLibraryLoader.cpp
+++ b/src/cpp/core/system/PosixLibraryLoader.cpp
@@ -57,11 +57,11 @@ Error loadLibrary(const std::string& libPath, void** ppLib)
  * Verify that we can load a library (without actually loading it)
  * by passing RTLD_NOLOAD to dlopen
  */
-Error verifyLibrary(const std::string& libPath, void** ppLib)
+Error verifyLibrary(const std::string& libPath)
 {
-   *ppLib = nullptr;
-   *ppLib = ::dlopen(libPath.c_str(), RTLD_LAZY | RTLD_NOLOAD);
-   if (*ppLib == nullptr)
+   void* pLib = nullptr;
+   pLib = ::dlopen(libPath.c_str(), RTLD_LAZY | RTLD_NOLOAD);
+   if (pLib == nullptr)
    {
       // dlopen returns a null pointer on error rather than an error number.
       // using -1 here as a generic non-zero placeholder
@@ -75,7 +75,7 @@ Error verifyLibrary(const std::string& libPath, void** ppLib)
    }
    else
    {
-      return closeLibrary(*ppLib);
+      return closeLibrary(pLib);
    }
 }
 
--- a/src/cpp/session/SessionMain.cpp
+++ b/src/cpp/session/SessionMain.cpp
@@ -322,9 +322,6 @@ bool s_destroySession = false;
 // did we fail to coerce the charset to UTF-8
 bool s_printCharsetWarning = false;
 
-// (linux only) can we load the libR.so shared library
-std::string s_libRSoWarning = "";
-
 void handleINT(int)
 {
    rstudio::r::exec::setInterruptsPending(true);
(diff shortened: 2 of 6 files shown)


src/cpp/core/system/PosixLibraryLoader.cpp
```diff
@@ -53,6 +53,32 @@ Error loadLibrary(const std::string& libPath, void** ppLib)
    }
 }
 
+/**
+ * Verify that we can load a library (without actually loading it)
+ * by passing RTLD_NOLOAD to dlopen
+ */
+Error verifyLibrary(const std::string& libPath, void** ppLib)
```
MariaSemple, line 60: Since you never use `ppLib`, I don't think you should call it here. You're also never closing the handle on success, but I'm not sure if that matters.

src/cpp/core/system/PosixLibraryLoader.cpp
```diff
@@ -53,6 +53,32 @@ Error loadLibrary(const std::string& libPath, void** ppLib)
...
+   *ppLib = ::dlopen(libPath.c_str(), RTLD_LAZY | RTLD_NOLOAD);
+   if (*ppLib == nullptr)
+   {
+      // dlopen returns a null pointer on error rather than an error number.
+      // using -1 here as a generic non-zero placeholder
+      Error error = Error("LibraryVerifyFailed",
+                          -1,
+                          "Attempt to verify library failed",
+                          ErrorLocation());
+      error.addProperty("lib-path", libPath);
```
MariaSemple, line 72:
```suggestion
      error.addProperty("path", libPath);
```

src/cpp/session/SessionMain.cpp
```diff
@@ -699,6 +704,12 @@ Error rInit(const rstudio::r::session::RInitInfo& rInitInfo)
    if (s_printCharsetWarning)
       rstudio::r::exec::warning("Character set is not UTF-8; please change your locale");
 
+   if (!ensureLibRSoValid())
+     LOG_ERROR_MESSAGE(s_libRSoWarning +
+                       ". Linux may have loaded a different libR.so than requested."
+                       " R_HOME: " + rsession::module_context::rHomeDir() +
+                       ", R Version: " + rsession::module_context::rVersion());
+
```
MariaSemple, line 712: Rather than use a static which isn't thread safe, we should log this error message from within the `ensureLibRSoValid()` method.

MariaSemple: Additionally, should we exit here? Or at the very list emit a similar error to the R Console so that the user can be informed? IIRC, this error message will only go to our log files, and a big problem with this behaviour is that the user is unware that they're using the wrong version of R. We should probably include a note about contacting your system administrator if the versions don't match.

MariaSemple (changes requested):
Elegant solution! Very excited to see this improve for users.

MariaSemple (approved):
LGTM!

abalter:
Is this really fixeed?

https://github.com/rstudio/rstudio/issues/7868

abalter:
This is really not fixed. I'm still having the same problem. I've tried setting the path to R and R libraries in all of the places the docs say. For instance, `/etc/rstudio/rsession-profile` and `/etc/rstudio/rserver.conf`. I've made sure that root is using the same R as I am.

error while loading shared libraries: libR.so - R Admins - Posit Community
https://community.rstudio.com/t/error-while-loading-shared-libraries-libr-so/101389/13

r - RStudio Server (community) can't find libR.so even though it is in path - Stack Overflow
https://stackoverflow.com/questions/74299104/rstudio-server-community-cant-find-libr-so-even-though-it-is-in-path

/usr/local/lib64/R/lib/libR.so: cannot open shared object file: No such file or directory - Stack Overflow
https://stackoverflow.com/questions/69013147/usr-local-lib64-r-lib-libr-so-cannot-open-shared-object-file-no-such-file-or

r - Error with shared library with RStudio Server - Stack Overflow
https://stackoverflow.com/questions/33897998/error-with-shared-library-with-rstudio-server

r - error while loading shared libraries: libR.so - Stack Overflow
https://stackoverflow.com/questions/37967497/error-while-loading-shared-libraries-libr-so

error while loading shared libraries: libR.so: cannot open shared object file
https://groups.google.com/g/epacts/c/5JNFkpmyuSU

ubuntu - libR.so not found in R Rstudio - Stack Overflow
https://stackoverflow.com/questions/60893387/libr-so-not-found-in-r-rstudio

R shared library (/usr/local/lib/R/lib/libR.so) not found. If this is a custom build of R, was it built with the --enable-R-shlib option? - RStudio IDE - Posit Community
https://community.rstudio.com/t/r-shared-library-usr-local-lib-r-lib-libr-so-not-found-if-this-is-a-custom-build-of-r-was-it-built-with-the-enable-r-shlib-option/126418

Rstudio error "R shared library (/usr/local/lib/R/lib/libR.so) not found." - RStudio IDE - Posit Community
https://community.rstudio.com/t/rstudio-error-r-shared-library-usr-local-lib-r-lib-libr-so-not-found/62598

[R] R shared library (/usr/lib64/R/lib/libR.so) not found.
https://stat.ethz.ch/pipermail/r-help/2018-August/455984.html

RStudio Server (community) can't find libR.so even though it is in path. · Issue #12281 · rstudio/rstudio
https://github.com/rstudio/rstudio/issues/12281

abalter:
How about just adding a switch to `rstudio-server`? Then I could run `sudo rstudio-server start --r_path=/home/balter/micromamba/bin/R r_libs=/home/balter/micromamba/lib/R/lib`

PR #9907, opened 2021-10-01
Menus now load keyboard shortcuts for App Commands.

GCRev:
See: https://github.com/rstudio/rstudio/issues/5256

### Intent

Note: For the sake of brevity and ease-of-review, this PR addresses menus only. Displaying custom shortcuts in other locations will have to be evaluated on a case-by-case basis. We can have a discussion about whether performing some kind of event trigger-based update to force tooltips to refresh is worth the deep-dive. There are a couple of ways to design that.

An issue (5256) was identified where custom shortcuts set through the "Modify Keyboard Shortcuts" do not appear next to their commands in menus.

The title of the original ticket mentions the command palette as well, but that has been previously addressed and fixed.

### Approach

I've added a custom keybinding shortcut property to the `AppCommand` class which override the display of the default shortcut. Bindings loaded from the config file are considered "custom," which do not overwrite the default shortcut value on the AppCommand itself (but they still change the binding from default). Tooltips still do not immediately update, as they are rendered in when the pane itself is rendered.

Refreshing the page works with varying success, but the browser caches most of the DOM. I could only get source editor tooltips to update by closing and then re-opening a source editor, for example. This will need a more in-depth approach to updating using more event listeners, and that could complicate things.

An earlier approach I took worked well for customizing AppCommands that had no shortcut to begin with, but would override commands with an initial value, such that resetting would no longer work. The current approach makes it possible to reset correctly.

### Automated Tests

An automated browser test would be required to input the keys and then check the visual outcome in a menu.

### QA Notes

**Verify Constraints:**

- The command must appear in a menu
- The command be available to assign a custom shortcut to. Some commands have custom shortcuts disabled.

**Test:**

1. Test changing commands
  - Open the Keyboard Shortcut editor and make changes to a blank and non-blank (with defaults) command.
  - Apply the changes
  - Verify that the items in their menus display the new shortcuts
  - Verify that the app command palette shows the correct new values

2. Test resetting
 - Reset changes
 - Verify that the items have returned to their previous states, blank or otherwise
 - Verify that the app command palette shows the correct values

I've also fixed a minor issue related to the visual display of changing commands in the Shortcuts Editor, so keep an eye out for similar tickets about that: Keyboard shortcuts are now compared correctly, so the editor no longer displays a highlighted row just by clicking in and clicking out again. Similarly, if you input the same command before you made changes, it resets the changed state visually by removing the highlight.

1. Verify current behavior
  - Open the Keyboard Shortcut editor and click in to, then out of any row
  - The row should be highlighted in green as if it has a change, even if there is no change
  - Change any existing keyboard shortcut, overriding the default, and then change it back manually
  - The row should remain highlighted as if has a change, even though the value matches the old value

2. Verify fixed behavior
  - Open the Keyboard Shortcut editor and click in to, then out of any row
  - The row should not be highlighted, correctly showing no change
  - Change any existing keyboard shortcut, overriding the default, and then change it back manually
  - The row should not be highlighted, correctly showing no change from the original value

### Checklist


NEWS.md
```diff
@@ -11,6 +11,7 @@
 ### RStudio Workbench
 
 * Added support for setting the `subPath` on Kubernetes sessions using `KubernetesPersistentVolumeClaim` mounts in `/etc/rstudio/launcher-mounts` (Pro #2976).
+* Fixed custom shortcuts not appearing correctly in menus
```
gtritchie, line 14: I recommend opening a new issue to track the remaining behaviors not addressed by this fix (e.g. tooltips). Then narrow down the existing issue to only cover the menus, and reference it here in the news entry. Also QA only tests issues, not PRs, so having an issue that covers what this PR fixes while tracking the unfixed parts elsewhere is ideal.

src/gwt/src/org/rstudio/core/client/widget/ModifyKeyboardShortcutsWidget.java
```diff
@@ -1450,6 +1466,8 @@ public Object getKey(KeyboardShortcutEntry item)
    private List<KeyboardShortcutEntry> originalBindings_;
    private Pair<Integer, Integer> lastSelectedIndices_;
    private boolean swallowNextKeyUpEvent_;
+
+   private ThemedButton applyButton_;
```
gtritchie, line 1470: This change (having the apply button be disabled initially) causes a change in behavior that I suspect will result in negative feedback from some users. Previously you could type a new shortcut sequence and immediately click the Apply button; now you have to either hit the Enter key or click outside the text field to enable the Apply button. What you did is reasonable, but I'd be inclined to leave this behavior the way it was.

gtritchie (changes requested):
This does not update the shortcuts shown on RStudio Desktop menus, only RStudio Server. Ideally we'd want to do both, though you could open (yet another) issue tracking the Desktop menu issue and tackle it separately.

melissa-barca (approved):
Nice job figuring this one out and coming up with a succinct solution. Pending the change's from Gary's feedback - LGTM!

NEWS.md
GCRev: I have opened new issue https://github.com/rstudio/rstudio/issues/9915 to track this, and changed the PR description to mention it.

src/gwt/src/org/rstudio/core/client/widget/ModifyKeyboardShortcutsWidget.java
GCRev: This behavior has been reverted, but I have left in the reference to `applyButton_` in case we ever change our minds about this.

GCRev:
@gtritchie since the custom shortcuts are loaded upon refresh on the GWT side, and they are loaded _after_ the UI loads, the only way to actually set custom shortcuts is by doing so through a desktop menu callback. This means any changes the user makes to the shortcuts will be reflected immediately in the desktop menus... but the solution seems like a bit of a hack to me.

I also can't tell if the `binders_` vector will also require cleaning in the same way that the `actions_` vector does... in theory it shouldn't because the menu bindings should only be added once, uniquely. I was unable to tell if that is actually the case, however. Either way, there needed to be some method to associate the custom shortcut with the menu binding so that it would load the correct shortcut key sequence upon opening the menu.

src/cpp/desktop/DesktopMenuCallback.cpp
```diff
@@ -202,8 +201,10 @@ QAction* MenuCallback::duplicateAppMenuAction(QString commandToDuplicate,
       menuStack_.top()->addAction(pAction);
 
       auto* pBinder = new MenuActionBinder(menuStack_.top(), pAction);
-      connect(pBinder, SIGNAL(manageCommand(QString, QAction * )), this, SIGNAL(manageCommand(QString, QAction * )));
+      connect(pBinder, SIGNAL(manageCommand(QString, QAction*)),
+              this, SIGNAL(manageCommand(QString, QAction*)));
       connect(pAction, SIGNAL(triggered()), this, SLOT(actionInvoked()));
+      binders_[commandId].push_back(QPointer<MenuActionBinder>(pBinder));
```
melissa-barca, line 207:
```suggestion
      binders_[commandId].push_back(pBinder);
```
Nit: I don't think you need to cast here

src/cpp/desktop/DesktopMenuCallback.cpp
```diff
@@ -283,9 +284,10 @@ void MenuCallback::addCommand(QString commandId,
       if (checkable)
          pAction->setCheckable(true);
 
-      auto * pBinder = new MenuActionBinder(menuStack_.top(), pAction);
-      connect(pBinder, SIGNAL(manageCommand(QString,QAction*)),
-              this, SIGNAL(manageCommand(QString,QAction*)));
+      auto* pBinder = new MenuActionBinder(menuStack_.top(), pAction);
+      connect(pBinder, SIGNAL(manageCommand(QString, QAction*)),
+              this, SIGNAL(manageCommand(QString, QAction*)));
+      binders_[commandId].push_back(QPointer<MenuActionBinder>(pBinder));
```
melissa-barca, line 290:
```suggestion
      binders_[commandId].push_back(pBinder);
```

src/cpp/desktop/DesktopMenuCallback.cpp
```diff
@@ -385,13 +395,22 @@ MenuActionBinder::MenuActionBinder(QMenu* pMenu, QAction* pAction) : QObject(pAc
    connect(pMenu, SIGNAL(aboutToHide()), this, SLOT(onHideMenu()));
    pAction_ = pAction;
    keySequence_ = pAction->shortcut();
+   customSequence_ = QKeySequence();
```
melissa-barca, line 398: Nit: since this object has a default constructor, you shouldn't need to explicitly initialize it. 
```suggestion
```

melissa-barca (approved):

gtritchie (changes requested):
I figured out the problem on Mac, and will send you a patch with the fix, plus minimal changes needed to keep the Electron desktop working with this change applied.

Issue #9953, opened 2021-10-13
Custom shortcuts not reflected in "Keyboard Shortcuts Help"


GCRev:
### System details

```shell
    RStudio Edition : Desktop/Server
    RStudio Version : "Prairie Trillium" Daily (99999999, 2021-10-05)
    OS Version      : Ubuntu Hirsute 21.04
    R Version       : 4.1.1
```

### Steps to reproduce the problem

This behavior is described in https://github.com/rstudio/rstudio/issues/5256, but this issue is focused only on "Keyboard Shortcuts Help"

### Describe the problem in detail

![image](https://user-images.githubusercontent.com/5544965/137218278-b0cee0e2-eb55-49b3-b332-ff7c3e7d710f.png)

![image](https://user-images.githubusercontent.com/5544965/137218341-46482182-55cd-4700-b2fa-b653b2e5689a.png)

Customized keyboard shortcuts do not appear in the Keyboard Shortcuts Help overlay.

### Describe the behavior you expected

Custom shortcuts should print out in the Keyboard Shortcuts Help overlay, including when they are cleared out.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] ~If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).~
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

astayleraz:
Verified in 2022.01.0-daily+199.pro1 remapping works for a command with and without a default command and added automation coverage for those cases. Closing issue.

PR #9954, opened 2021-10-13
Correctly show custom shortcuts in Electron app menus and "Keyboard Shortcuts Help"

GCRev:
See: https://github.com/rstudio/rstudio/issues/9953

### Intent

Keyboard shortcuts update properly in Server menus and Desktop QT menus, so this commit adds support for make those changes in Electron menus. However, updating shortcuts in Electron menus is no small task. Shortcuts are the only property that cannot be updated with a simple write, so an entire menu rebuild must be performed in order to make that change. 

I also went ahead and addressed the custom shortcuts not appearing in "Keyboard Shortcuts Help" (https://github.com/rstudio/rstudio/issues/9953) because it is related. The only remaining issue to fix for https://github.com/rstudio/rstudio/issues/5256 is tooltips.

### Approach

What we had to do was what I would call a hack, but we have to work around the constraints that Electron applies.

The update approach for shortcuts (called "Accelerators" in Electron) is to queue up changes from GWT and then have GWT commit them once all updates have been submitted. Upon commit, the entire app menu is copied to a new menu, including submenus, with the changes from the queue applied. In order to do this properly, the templates for each menu item are stored for re-use, so that they can be used to re-build the new menu without manually copying keys between `MenuItems` and their constructors.

### Automated Tests

There are no automated tests yet.

### QA Notes

Verification is the same as in https://github.com/rstudio/rstudio/pull/9907, only Electron is the Desktop environment to use rather than QT.

Additionally, verify the following:
* Verify that the Command is visible in Keyboard Shortcuts Help.
* Set a custom keyboard shortcut for one of the commands (can be blank).
* Verify that shortcut listed in Keyboard Shortcuts Help reflects all changes made.
* Reset all changes in the Modify Keyboard Shortcuts dialog
* Verify that the reset has worked, and that Keyboard Shortcuts Help reflects this

### Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,6 +17,7 @@
 
 * Added support for setting the `subPath` on Kubernetes sessions using `KubernetesPersistentVolumeClaim` mounts in `/etc/rstudio/launcher-mounts` (Pro #2976).
 * Fixed custom shortcuts not appearing correctly in menus (#9915)
+* Fixed custom shortcuts not appearing correctly in "Keyboard Shortcuts Help" and Electron menus. (#9953)
 
 ### Misc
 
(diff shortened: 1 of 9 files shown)


melissa-barca (approved):
LGTM!

PR #9957, opened 2021-10-14
Support building Mac Electron Desktop via make-package

gtritchie:
### Intent

Step towards building the Electron-based desktop application on Mac and producing the DMG; this change enables building it entirely via `make-package` on Mac (Intel-only at the moment).

No changes made to Jenkinsfiles, and the intent of this change is to be a no-op for our current builds.

#### Usage

To build a redistributable Electron desktop DMG on Mac:

```
cd rstudio/package/osx
./make-package clean --rstudio-target=Electron
```

A DMG will be produced: `src/node/desktop/out/make/RStudio-0.0.1-x64.dmg`.

#### How it works

First, build is configured with `-DRSTUDIO_TARGET=Electron` causing all the same things to be built as Qt desktop (rsession, GWT, various support things), MINUS the Qt desktop itself. The Electron app itself is not built during this phase.

After building, cmake is run again with the `install` target. This was already being done in the existing Mac build. In the Electron case, it lays out the structure for all the non-Electron bits needed in the app image.

Finally, `yarn make` is executed. This builds and packages the Electron app itself. A node script (`import-resources.ts`) is invoked by the electron-forge postPackage hook (see `forge.config.js`) and it copies the files from the cmake-generated install, into the Electron application. Then a DMG is produced (as part of the `yarn make` step).

A downside of this is an extra step (`yarn make`) is needed beyond running the cmake build. I will look at whether I can hook the `yarn make` step as a final step in the cmake `install` target so it is only necessary to run a single build step.

#### Next steps

- entitlements and notarization
- plumbing for versioning instead of current hardcoded 0.0.1
- support building native M1 version
- Linux and Windows (most of this work is reusable for those, but some additional script plumbing is required)
- (eventually) Jenkinsfile tweaks to actually build, run tests, and publish

### Automated Tests

Purely build-time.

### QA Notes

None.

### Checklist


package/osx/make-package
```diff
@@ -81,6 +82,15 @@ find-program ANT ant            \
    "${HOME}/homebrew/arm64/bin" \
    "${HOME}/opt/bin"
 
+# find node
+find-program NODE node \
+   "../../dependencies/common/node/${RSTUDIO_NODE_VERSION}/bin" \
+   "/opt/rstudio-tools/dependencies/common/node/${RSTUDIO_NODE_VERSION}/bin"
```
MariaSemple, line 88: Does this mean we need to have node.js installed even if we want to build server or QT-based desktop?

gtritchie: Yes, but this was already true. The visual editor stuff requires node.js at build time. Only difference is I need to invoke it from cmake; whereas visual editor invokes it via ant (build.xml).

Issue #8208, opened 2020-10-27
Scrolling an unfocused source window leaves column headings stationary


wgrundlingh:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.4.953
    OS Version      :  Windows
    R Version       : 4.0.3

### Steps to reproduce the problem

1. Add multiple code/source windows to your display
2. Open a data frame (or `View` it) with more columns that can be displayed in the code/source window. Here's code to generate an example data frame:

```
library(tidyverse)

df <- tibble(
  letters = LETTERS,
  values = 1:26
) %>% pivot_wider(
  names_from = letters,
  values_from = values
)
```

3. Change the focus to a code/source window that is not the `View` of the data frame.
4. Scroll left/right to view the data; the headings/variables remain fixed.

### Describe the problem in detail

Upon a focus change, scrolling left/right within a object display window leaves the headers unchanged. Here's a visual of the stationary column headers/variable names within `df`:

![temp](https://user-images.githubusercontent.com/1776164/97371700-181e2f80-186f-11eb-80a9-34bf94ea7c77.gif)

### Describe the behavior you expected

The column/variable names should scroll with the data, regardless of whether that element has the focus or not.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting a RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
@wgrundlingh Thank you for raising the issue! I can reproduce this in MacOS 11, as well, in RStudio Desktop 1.4.969. We'll review it as we continue development of RStudio.

mikebessuille:
Interesting bug, thanks!   At this point in 1.4 we're only looking at critical failures, so moving this to 1.5.

PR #9964, opened 2021-10-15
Fix horizontal scroll behavior of data viewer tables in non-active frames

GCRev:
See https://github.com/rstudio/rstudio/issues/8208

### Intent

Solve the issue described in https://github.com/rstudio/rstudio/issues/8208 in which focusing a source panel in a different column causes a strange scrolling behavior in the data viewer.

Investigation revealed that this was a previously-implemented Firefox compatibility bugfix, in which all of the scroll event listeners are purged when focus is lost in a source window. The intent of that was to fix a vertical scrolling issue, but I haven't been able to replicate this in current versions of Firefox. 

The problem with the previous fix is that removing all scroll listeners also breaks horizontal scrolling, because both vertical and horizontal scrolling listens to the same `scroll` event. JQuery scrolls the table headers programmatically using a `scroll` event listener, so scroll events cannot be removed because we cannot assume we know which scroll listeners is which.

### Approach

I have commented out the code that removes the scroll listeners, and the code that adds them back. I have left in the code that stores and reloads the vertical scroll positions, however, as it is harmless.

### Automated Tests

There are no automated tests that I can think of that would be able to test this.

### QA Notes

* Ensure that **Firefox** is used to test, as the original code was added to solve a Firefox-specific bug
* Open a regular source pane and a new source column, so there are two focusable windows present on screen (like shown in https://github.com/rstudio/rstudio/issues/8208)
* Make sure there are files loaded in each source pane.
* Run a `View(mtcars)` to load in a dataset that is both long-enough and wide-enough to create scroll bars on both the vertical and horizontal axes
* Click between the source panels and verify that the column headers continue to scroll with the table data in either state
* Verify that no vertical scrolling artifacts appear as a result of these changes

### Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,6 +17,7 @@
 
 * Added support for setting the `subPath` on Kubernetes sessions using `KubernetesPersistentVolumeClaim` mounts in `/etc/rstudio/launcher-mounts` (Pro #2976).
 * Fixed custom shortcuts not appearing correctly in menus (#9915)
+* Fixed header scrolling in data viewer tables not following table contents in unfocused windows (#8208)
 
 ### Misc
 
--- a/src/cpp/session/resources/grid/dtviewer.js
+++ b/src/cpp/session/resources/grid/dtviewer.js
@@ -320,9 +320,11 @@
     var scrollBody = $(".dataTables_scrollBody");
     if (scrollBody) {
       // reattach handlers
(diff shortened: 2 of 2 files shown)


jmcphers (approved):
LGTM!

jeffvroom (approved):
LGTM!

PR #9972, opened 2021-10-18
Add Element ID's to the Filter field and Reset button for automation

GCRev:
### Intent

Add Element ID's to the filter and reset buttons in the Keyboard Shortcuts Editor so that automation testing can find these elements and make use of them.

### Approach

Add a reference to the reset button so that GWT can attach an ID to it. Declare two new `private static final String` constants that fix the ID values.

### Automated Tests

The purpose of this branch is to enable automated testing on the other Keyboard shortcut-related issues.

### QA Notes

Verify that you get elements when you run the following queries on the document:
* Filter element: `document.getElementById('rstudio_kybrd_shrtcts_fltr')` 
  Keep in mind that this does not point directly to the `<input/>`, but to the parent `<div/>` of the filter Widget. If you want the `<input/>` then you have to add a `querySelectorAll('input')` to find it in the widget hierarchy.
* Reset button: `document.getElementById('rstudio_kybrd_shrtcts_rst')` this returns the `<button/>` element directly.

### Checklist

- [x] ~If this PR adds a new feature, or fixes a bug in a previously released version, it includes an entry in `NEWS.md`~
- [x] ~If this PR adds or changes UI, the updated UI meets [accessibility standards](https://github.com/rstudio/rstudio/wiki/Accessibility)~

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/widget/ModifyKeyboardShortcutsWidget.java
+++ b/src/gwt/src/org/rstudio/core/client/widget/ModifyKeyboardShortcutsWidget.java
@@ -369,6 +369,8 @@ public class ModifyKeyboardShortcutsWidget extends ModalDialogBase
          }
          
       });
+
+      filterWidget_.getElement().setId(FILTER_WIDGET_ID);
       
       filterWidget_.addValueChangeHandler(new ValueChangeHandler<String>()
       {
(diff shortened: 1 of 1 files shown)


src/gwt/src/org/rstudio/core/client/widget/ModifyKeyboardShortcutsWidget.java
```diff
@@ -1465,6 +1470,8 @@ public Object getKey(KeyboardShortcutEntry item)
    private Pair<Integer, Integer> lastSelectedIndices_;
    private boolean swallowNextKeyUpEvent_;
 
+   private static final String RESET_BUTTON_ID = "rstudio_kybrd_shrtcts_rst";
```
gtritchie, line 1473: For consistency, should define the constant in `ElementIds.java` and use the appropriate `assignElementId()` overload to apply it.

Someday we should decentralize this so it's not all in this single source file but until we do let's follow the same pattern.

PR #9985, opened 2021-10-20
Implement scroll workaround to RMD files using iframes to show content

GCRev:
See: https://github.com/rstudio/rstudio/issues/2202

### Intent

Solve an annoying UX issue in which users scrolling through RMarkdown files with sub-content (like kables), will be be stopped once their mouse enters the content area.

The "real" solution would be to not use `<iframe/>` anymore, as `<iframe/>` can cause unpredictable behavior.

### Approach

The solution is a hack using a native function - by default the rendered output `<iframe/>` elements are "disabled" from user interaction until the user moves their mouse inside the sub-content. This will "enable" the `<iframe/>`, and it will begin accepting pointer events. Moving the mouse out of the content "re-disables" user interaction so that scroll behavior becomes predictable again.

This makes it possible to scroll vertically through a document, uninterrupted, assuming you do not move your mouse at all, without being interrupted by sub-content. 

### Automated Tests

I am not sure if this is possible to test with automation, as it is so specific to mouse movement and scroll inputs.

### QA Notes

Use [kable_scolling.rmd](https://github.com/rstudio/rstudio/files/1727946/kable_scolling.txt) as an example
- Run all code blocks to generate sub-content that both scrolls and does not scroll
- Verify that you can scroll from the top of the document to the bottom without encountering a pause, provided you keep the mouse pointer perfectly still.
- Verify that moving the mouse over sub-content allows you to scroll its content
- Verify that removing the mouse from the sub-content area disables the sub-content again, and scroll behavior on the main document is restored.

### Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -25,6 +25,7 @@
 * Fixed a bug where Slurm Launcher jobs with standard error would never be written to the output file (#203)
 * Fixed a bug where Slurm Launcher jobs that exited due to a signal would not show the exit code as 128+signal (#203)
 * Fixed a bug where Launcher log files could be stuck being owned by the root user (#9728)
+* Fixed scrolling past long sub-content (like kables) in RMD files. User must interact with sub-content in order to scroll through it (#2202)
 
 ### Misc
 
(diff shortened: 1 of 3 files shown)


melissa-barca (approved):
LGTM!

PR #10028, opened 2021-10-30
Feature/session suspend info

cm421:
### Intent

The Open Source PR for https://github.com/rstudio/rstudio-pro/issues/2618. This code tracks IDE operations that block a session from timing out and suspending, and reports these "blocking ops" to the user when a session attempts to suspend, but cannot.

### Approach

The code that manages suspension is scattered through several locations. Wherever was possible, the suspend-related logic was moved into `SessionSuspend.hpp/cpp`. Internally `SessionSuspend` will track any operations that can block timeout suspension to later be reported if and when the session attempts to suspend because of a timeout. It is the responsibility of external code to report its own suspend-blocking operations to `SessionSuspend` via `SessionSuspend::[add/remove/check]BlockingOps()` calls.

~~To minimize noise to the user an indicator that the session attempted to suspend, but was blocked, is only displayed once a suspend attempt was made.~~ When a user starts an activity that will block automatic suspension, an icon will appear in the R console toolbar within 5 seconds to indicate that the session will remain active until the task completes, if ever. The indicator will remain visible until all blocking operations have stopped, updating dynamically as blocking ops start or stop. As long as the indicator is visible, the session will not attempt to suspend regardless the amount of time that passes. Once all blocking operations have stopped the indicator will disappear, the timeout will restart to whatever value is set by the `session-timeout-minutes` option, and the process repeats.

To minimize noise to the logs ~~a warning~~ an informational-level log that a session attempted to suspend is only emitted once ~~per session suspend attempt~~ when both conditions are true:

1. Auto suspension is actively being blocked by some user activity
2. The above activity ran longer than whatever value is set by `session-timeout-minutes`

Aka, if the session would have normally tried to auto suspend, but couldn't because auto suspension was blocked. The warning lists the blocking operations as well as the session id, username, and the value of `session-timeout-minutes`. 

None of this additional logic will interfere with session suspend request RPCs or forced session suspends.

A "Session Suspended" Icon was added to the R console toolbar. It will display whenever the session is suspended, similar to how the memory usage widget grays out when a session suspends.

### Automated Tests

This PR doesn't include automated testing. A basic selenium testing suit might include:

1. Start a new R session
2. Create one of the conditions that will prevent R sessions from suspending
3. Wait for session timeout
4. Session does not suspend
5. Mouse over/title of suspend blocked icon should match the condition we created in step 2
6. Stop/remove the blocking op
7. suspend blocked icon should disappear

We will likely want to test each of the suspend-blocking operations separately, as well as some(or all) together to make sure multiple reasons a session won’t suspend are also displayed.

### QA Notes

See the testing section for how to reproduce. Other things to consider:

- Requesting a session to suspend from the GUI (Session->Suspend R Session) should still force the session to suspend, even if the suspend blocked icon is displaying
- This should not change any functionality on the desktop app, as the desktop app does not suspend

### Checklist


gtritchie (changes requested):
A person with limited or no vision using a screen reader would have no practical way to know that a session has suspended or failed to suspend and why. They could, in theory, navigate to the new icon to hear its text, but how would they know to do this?

The simplest fix is to give those new icons the "alert" role, this will cause a screen reader to announce their titles when set. This should be gated via a new accessibility announcement option so users can control which announcements they hear.

I'll send you details on how to do this.

jmcphers (changes requested):
I think this might still need a little more UI to make it address the primary motivating use case. If you're an rstudio.cloud user, and you intend to let your session go to sleep, how do you know if it will or not? As written this is only going to tell you _after_ we have already tried to put it to sleep, likely long after you've moved on to other tasks. 

Some ideas: 
- We could add a "Suspend Now" command that would immediately suspend the session (if suspension is allowed), or explain why suspension is disallowed.
- We could poll for suspension status every (say) 30 seconds and show the result in the UI. 

A related problem is that if the session didn't go to sleep, and you close the tab, then (much later) reopen it, there's no way to tell whether (a) the session woke up when you reopened the tab, or (b) the session never went to sleep. I think this is probably the most common use case so we should find a way to make sure the status is visible in this case.

GCRev:
I would consider changing the "three z's" icon to a "pause" icon because it'll read better at small scales. We only have 15x15 pixels to work with, which means you have space for one letter at most, and it has to take up the entire area. For usability purposes, It's best to consider how you can use the largest possible shapes in an icon to make it read, even at very small scales.

Something filled-in like this would be extremely visible (from the Noun Project).
24x24: pause
15x15: 

As for the "blocked" state, I suppose you could throw a diagonal line across the middle.

src/gwt/src/org/rstudio/studio/client/workbench/views/console/Console.java
```diff
@@ -55,6 +58,9 @@
       IsWidget getConsoleInterruptButton();
       IsWidget getConsoleClearButton();
       IsWidget getProfilerInterruptButton();
+      IsWidget getSuspendBlockedIcon();
+      IsWidget setSuspendBlockedIcon(SessionSuspendBlockedEvent event);
+      IsWidget getSuspendedIcon();
```
GCRev, line 63:
```suggestion
      void onSuspendedBlockedEvent(SessionSuspendBlockedEvent event);
      void onSerializationEvent(SessionSerializationEvent event);
```
I suggest replacing these specific methods with more general ones that feed the event/event type through to the view class, so that the view class can handle the visibility logic internally. This way you don't have to "export" these elements out from the view class, and if you want to add more logic/elements, you can modify just the view class without worrying about the model.

src/gwt/src/org/rstudio/core/client/theme/res/ThemeResources.java
```diff
@@ -297,4 +297,10 @@
    
    @Source("browsePackage_2x.png")
    ImageResource browsePackage2x();
+
+   @Source("suspended.png")
+   ImageResource suspended();
+
+   @Source("SuspendBlocked.png")
+   ImageResource suspendBlocked();
```
GCRev, line 305: Check file name here. The added .png is lower case for me -- so this won't load anything.
```suggestion
   @Source("suspendBlocked.png")
   ImageResource suspendBlocked();
```

src/gwt/src/org/rstudio/studio/client/workbench/views/console/ConsolePane.java
```diff
@@ -121,6 +122,20 @@ public IsWidget getProfilerInterruptButton()
...
+   public IsWidget setSuspendBlockedIcon(SessionSuspendBlockedEvent event) {
+      consoleSuspendBlockedIcon_.setVisible(event.isBlocked());
+      consoleSuspendBlockedIcon_.setTitle(event.getMsg());
+
+      return consoleSuspendBlockedIcon_;
+   }
+
+   @Override
+   public IsWidget getSuspendedIcon() { return consoleSuspendedIcon_; }
+
```
GCRev, line 138: NOTE: these require imports, and I have not added those to the suggestions specifically.

Example of how this would change based on the other changes:
```suggestion
   @Override
   public void onSerializationEvent(SessionSerializationEvent event) {
      int eventType = event.getAction().getType();
      switch (eventType) {
         case SessionSerializationAction.SUSPEND_SESSION:
            consoleSuspendBlockedIcon_.setVisible(false);
            consoleSuspendedIcon_.setVisible(true);
            break;
         case SessionSerializationAction.RESUME_SESSION:
            consoleSuspendBlockedIcon_.setVisible(false);
            consoleSuspendedIcon_.setVisible(false);
            break;
      }
   }

   @Override
   public void onSuspendedBlockedEvent(SessionSuspendBlockedEvent event) {
      consoleSuspendBlockedIcon_.setVisible(event.isBlocked());
      consoleSuspendBlockedIcon_.setTitle(event.getMsg());
   }
```

GCRev (changes requested):
Some organization opportunities in the UI code. Please be aware that I have not included changes to the import blocks.

src/gwt/src/org/rstudio/studio/client/workbench/views/console/Console.java
```diff
@@ -174,6 +180,23 @@ else if (StringUtil.equals(type, ReticulateEvent.TYPE_REPL_BUSY))
...
+         else if (event.getAction().getType() == SessionSerializationAction.RESUME_SESSION)
+         {
+            view.getSuspendBlockedIcon().asWidget().setVisible(false);
+            view.getSuspendedIcon().asWidget().setVisible(false);
+         }
+      });
+
+      events.addHandler(SessionSuspendBlockedEvent.TYPE, event -> {
+         view.setSuspendBlockedIcon(event);
+      });
```
GCRev, line 199: Oops, I forgot to point this one out:
```suggestion
      events.addHandler(SessionSerializationEvent.TYPE, event -> {
         view.onSerializationEvent(event);
      });

      events.addHandler(SessionSuspendBlockedEvent.TYPE, event -> {
         view.onSuspendedBlockedEvent(event);
      });
```

src/gwt/src/org/rstudio/studio/client/application/events/SessionSuspendBlockedEvent.java
```diff
@@ -0,0 +1,95 @@
...
+         if (this.hasOwnProperty('edit-completion'))
+            msg += '\nWaiting for edit completion';
+         if (this.hasOwnProperty('choose-file-completion'))
+            msg += '\nWaiting for Choose File completion';
+         if (this.hasOwnProperty('locator-completion'))
+            msg += '\nWaiting for Locator completion';
+         if (this.hasOwnProperty('unsaved-handler-completion'))
+            msg += '\nWaiting for Unsaved Work Prompt completion';
+         if (this.hasOwnProperty('user-prompt-completion'))
+            msg += '\nWaiting for User Prompt completion';
```
MariaSemple, line 56: I think you could take a similar approach here, with a map of names to a map of descriptions and a loop.

MariaSemple (changes requested):

src/cpp/session/SessionSuspend.cpp
```diff
@@ -41,15 +46,222 @@ bool s_suspendedFromTimeout = false;
...
+}
+
+void resetSuspendTimeout()
+{
+   s_suspendTimeoutTime = timeoutTimeFromNow();
+}
+
+core::json::Object blockingOpsToJson()
+{
+   core::json::Object blockingOps;
```
MariaSemple, line 122: Nit: Rather than name this the reverse order of the other array, I'd name it almost the same thing but with `Json` at the end, as it more clearly explains their difference

GCRev:
pause_octagon 


pause_slash 


stop_circle 


line_circle 



Here are some icon ideas. We can iterate on these if we need to. I haven't varied the color yet so that we can nail down a suitable shape first.

GCRev (approved):
Code changes to GWT LGTM!

cm421:
Pro-side companion PR: https://github.com/rstudio/rstudio-pro/pull/3116

src/cpp/session/SessionSuspend.cpp
```diff
@@ -122,6 +427,73 @@ void suspendIfRequested(const boost::function<bool()>& allowSuspend)
...
+std::string getResumedMessage()
+{
+   if (s_timeoutState == kWaitingForInactivity)
+   {
+      std::string xAmountOfTimeAgo = mostSignificantTimeAgo(s_blockingTimestamp);
+      if (xAmountOfTimeAgo.empty())
+         return "";
+
+      std::string blockedTime = boost::posix_time::to_simple_string(s_blockingTimestamp);
+      return "Welcome Back! Session has been running in the background since "
```
jmcphers, line 468:
```suggestion
      return "Connected to your session in progress, last used "
```

src/cpp/session/SessionSuspend.cpp
```diff
@@ -122,6 +427,73 @@ void suspendIfRequested(const boost::function<bool()>& allowSuspend)
...
+   }
+   else
+   {
+      boost::posix_time::ptime suspensionTimestamp = module_context::activeSession().suspensionTime();
+      std::string xAmountOfTimeAgo = mostSignificantTimeAgo(suspensionTimestamp);
+      if (xAmountOfTimeAgo.empty())
+         return "";
+
+      std::string suspendedTime = boost::posix_time::to_simple_string(suspensionTimestamp);
+      return "Session restored from your saved work on "
```
jmcphers, line 482: I'm worried that these will wind up being just a little noisy in the R console, but we can get them in as is and see how it feels in real-world usage.

jmcphers (changes requested):

src/cpp/session/SessionSuspend.cpp
cm421: I thought the same, and have it configured to at least not display the message when they just refresh the page. Hopefully that reduces any annoying repeats

src/gwt/src/org/rstudio/studio/client/workbench/views/console/ConsolePane.java
```diff
@@ -121,6 +127,27 @@ public IsWidget getProfilerInterruptButton()
...
+   @Override
+   public void onSerializationEvent(SessionSerializationEvent event) {
+      int eventType = event.getAction().getType();
+      switch (eventType) {
+         case SessionSerializationAction.SUSPEND_SESSION:
+            consoleSuspendBlockedIcon_.setVisible(false);
+            consoleSuspendedIcon_.setVisible(true);
+            break;
+         case SessionSerializationAction.RESUME_SESSION:
+            consoleSuspendBlockedIcon_.setVisible(false);
```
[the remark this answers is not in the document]
cm421: This was just to make sure it's always hidden as the session is suspending/resuming. It will re-appear later on if it's supposed to be visible

src/gwt/src/org/rstudio/studio/client/application/events/SessionSuspendBlockedEvent.java
```diff
@@ -0,0 +1,76 @@
...
+   }
+   public static final Type<Handler> TYPE = new Type<>();
+
+   public SessionSuspendBlockedEvent(Data data)
+   {
+      data_ = data;
+   }
+
+
+   public String getMsg() {
```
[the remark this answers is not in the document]
cm421: That's a little frustrating! IntelliJ IDEA formats it so it looks like the whole thing is on one line, even though in the actual source code it's formatted like this. And here I thought I was just matching the formatting I was seeing in rest of the file!

I'll be turning that "feature" off

src/cpp/session/SessionSuspend.cpp
```diff
@@ -42,15 +47,231 @@ bool s_suspendedFromTimeout = false;
...
+}
+
+void resetBlockingTimestamp()
+{
+   s_blockingTimestamp = boost::posix_time::not_a_date_time;
+}
+
+bool shouldNotify()
+{
+   return s_blockingTimestamp + boost::posix_time::seconds(5) < boost::posix_time::second_clock::universal_time();
```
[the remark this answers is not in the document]
cm421: I like that. Would this be better as a Global Option set in the client, or a session.conf option? My first thought is an option set inside the client, but how does that value from the client get passed back to the session?

src/gwt/src/org/rstudio/studio/client/workbench/views/console/ConsolePane.java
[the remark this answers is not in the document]
jmcphers: 👍

gtritchie (approved):
Approved from accessibility side.

src/cpp/session/resources/schema/user-prefs-schema.json
```diff
@@ -1054,6 +1054,18 @@
             "title": "Double click to select in the Console",
             "description": "Whether double-clicking should select a word in the Console pane."
         },
+        "console_suspend_blocked_notice": {
+            "type": "boolean",
+            "default": false,
```
jmcphers, line 1059:
```suggestion
            "default": true,
```

I think this should default to _true_ to maximize exposure during development. We can switch the default to false later if we need to based on feedback.

jmcphers (changes requested):

PR #10032, opened 2021-11-01
remove duplicate command for serveQuartoSite

gtritchie:
### Intent

`Commands.cmd.xml` has two almost identical entries for `serveQuartoSite`. This hasn't been a problem (apparently) but some incoming work for adding localizability to commands and menus breaks if there are duplicates, so want this in before that.

The original is here:

https://github.com/rstudio/rstudio/blob/7dcb20a2cc2ca8ba2a83387e38e23f222d7d4db0/src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml#L1122-L1125

The newer one here (only difference is addition of the "... in current directory" to the description):

https://github.com/rstudio/rstudio/blob/7dcb20a2cc2ca8ba2a83387e38e23f222d7d4db0/src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml#L2441-L2444

### Approach

Kept the original, got rid of the newer one.

### Automated Tests

None, this should be a no-op from an end-user and automation standpoint.

### QA Notes

The command in question is shown on the Build menu when a Quarto project is currently open, and the name is dynamically set at runtime (e.g. to "Serve Website" for a Quarto website project).

### Checklist

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml
@@ -2437,11 +2437,6 @@ well as menu structures (for main menu and popup menus).
         menuLabel="_Quarto Doc..."
         desc="Create a blank Quarto document in current directory"
         rebindable="false"/>
-        
-   <cmd id="serveQuartoSite"
-        menuLabel="_Serve Site"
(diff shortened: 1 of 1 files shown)


GCRev (approved):
Based on where this appears, I probably copied this by accident.
LGTM!

PR #10069, opened 2021-11-10
Fix for Verify Installation

AndrewMcClain:
RActiveSessionStorage was making assumptions about the context when a
ActiveSessionStorage was created, using our XDG to discover the storage
directory.

Launcher creates some sessions to do verification but the XDG is not
populated, so it was searching in the / of the filesystem.

Removing unused or easily substitutable constructors for ActiveSession
to increase clarity.

Switched to the FileActiveSessionStorage always taking a directory,
which will be the scratch path (session-<id>) path. Gets
FileActiveStorage out of the storage discovery game.

Removed unused factory methods, we can reintroduce them later, but they were leading us astray here.
The id will still be necessary for DB calls to read and write properties

### Intent

Address issue [3084](https://github.com/rstudio/rstudio-pro/issues/3084) When using verify-installation, and certain other activities, rserver or rlauncher throws out
```
ERROR system error 13 (Permission denied) [path: /root/.local/share/rstudio/sessions/active, target-dir: ]; OCCURRED AT rstudio::core::Error rstudio::core::FilePath::createDirectory(const string&) const src/cpp/shared_core/FilePath.cpp:838; LOGGED FROM: rstudio::core::r_util::FileActiveSessionStorage::FileActiveSessionStorage(const rstudio::core::FilePath&) src/cpp/core/r_util/RActiveSessionStorage.cpp:51
```
This is caused by the no argument use of the getActiveSessionStorage() which then attempts to discover the storage directory. In some contexts the appropriate user dir isn't populated, or is populated with root's info.

### Approach

Stop using no arg getActiveSessionStorage(). We now use the getFileActiveStorage(scratchPath) to construct a FileActiveStorage for a specific scratchpath, instead of attempting discovery.
Refactor FileActiveStorage to hold the scratch path, not the storage path, so that changes the paths to the property files.

Removed unused factory methods.

Removed unused ActiveSession constructors which just muddied the waters.

### Automated Tests

Without running as an integration test this will only either perform trivial tests, or be untestable.

### QA Notes

The above error message should no longer appear when running verify-installation in launcher configuration.

### Checklist

### DIFF
--- a/src/cpp/core/r_util/RActiveSessionStorage.cpp
+++ b/src/cpp/core/r_util/RActiveSessionStorage.cpp
@@ -107,7 +107,7 @@ namespace
 
     Error FileActiveSessionStorage::readProperties(const std::string& id, std::map<std::string, std::string>* pValues)
     {
-        FilePath propertyDir = getPropertyDir(id);
+        FilePath propertyDir = getPropertyDir();
         std::vector<FilePath> files{};
         std::vector<FilePath> failedFiles{};
         pValues->empty();
@@ -156,7 +156,7 @@ namespace
                 failedFiles, ERROR_LOCATION);
     }
 
-    FilePath FileActiveSessionStorage::getPropertyDir(const std::string& id) const
+    FilePath FileActiveSessionStorage::getPropertyDir() const
     {
         FilePath propertiesDir = scratchPath_.completeChildPath(propertiesDirName_);
         propertiesDir.ensureDirectory();
@@ -165,7 +165,7 @@ namespace
 
     FilePath FileActiveSessionStorage::getPropertyFile(const std::string& id, const std::string& name) const
     {
-        FilePath propertiesDir = getPropertyDir(id);
+        FilePath propertiesDir = getPropertyDir();
         const std::string& fileName = getPropertyFileName(name);
         return propertiesDir.completeChildPath(fileName);
     }
--- a/src/cpp/core/r_util/RActiveSessionStorage.cpp
+++ b/src/cpp/core/r_util/RActiveSessionStorage.cpp
@@ -43,10 +43,10 @@ namespace
     }
 } // anonymous namespace
 
-    FileActiveSessionStorage::FileActiveSessionStorage(const FilePath& activeSessionsDir) :
-        activeSessionsDir_ (activeSessionsDir)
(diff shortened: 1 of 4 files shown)


src/cpp/core/r_util/RActiveSessionStorage.cpp
```diff
@@ -158,7 +158,7 @@ namespace
 
     FilePath FileActiveSessionStorage::getPropertyDir(const std::string& id) const
     {
-        FilePath propertiesDir = activeSessionsDir_.completeChildPath(fileSessionDirPrefix_ + id + "/" + propertiesDirName_);
+        FilePath propertiesDir = scratchPath_.completeChildPath(propertiesDirName_);
```
MariaSemple, line 161: Does the `id` variable not need to be used here somewhere?

AndrewMcClain: The id variable does not need to be used, as the FileActiveSessionStorage constructor takes in the scratchPath, which is the session/active/session-<id> path. That is the ID is already baked in at initialization time. We only do the getPropertyDir to grab the `properites` path because presumabley in the future we'll be getting more types of session data than just the properties.

This does point out that the parameter isn't being used, so I'll remove that.

MariaSemple: When is the ID baked in? I also see the parameter is still being passed in the getPropertyFile method but it's now unused there as well

AndrewMcClain: The ActiveSession constructor has the ID and scratchPath, and when it calls the storage factory getFileActiveSessionStorage(<FilePath>) it passes the scratchPath for the session which is passed along to the FileActiveSessionStorage constructor. The scratch path is `session/active/session-<session-id>/` path. By constructing with the session specific path from the beginning, the ID isn't needed when we're constructing the properties path.

The ID also isn't necessary for the read/writeProperties methods, but it will be needed for later DB connections, so i decided it could stay.

MariaSemple (approved):
LGTM!

melissa-barca (approved):
LGTM!

This bug is another issue that rstudio/rstudio-ide-automation#100 would've identified.

PR #10082, opened 2021-11-15
Update SessionDataViewer.R

iseppo:
Fixes issue https://github.com/rstudio/rstudio/issues/9771
Checks if the colLabels is named vector and uses the named value. Otherwise uses idx'th element of the vector colLabels.

### Intent

The sessionDataViewer currently shows wrong labels of variables. It assumes the variable label of n-th variable is also n-th value of variable.labels attribute, but this is not guaranteed. This change will take the variable label from the named vector of variable.labels, matching the variable label with the name of the variable.

### Approach

There is first a test if the colLabels is named vector. If it is, it finds the matching value from named vector. If colLabels is not a named vector it returns to old behaviour and sets the n-th value in colLabels to be the name of n-th variable.

Note that it does not currently check if the matching value exists. Would there be any circumstance that variable.labels would be a named vector but the names would be something else than variable names?  The check can be added.

### Automated Tests

No test but a visual one described in the issue and in the QA Note.

### QA Notes

```r
df <- data.frame(a=1, b=2, c=3)
attr(df, "variable.labels") <- c(a = "a", b = "b", c = "c")
View(df) #variable labels showed correctly
library(dplyr)
View(select(df, -b)) #wrong variable labels showed, should show the correct ones
```

### Checklist

I have sent a signed contributor agreement.

### DIFF
--- a/src/cpp/session/modules/SessionDataViewer.R
+++ b/src/cpp/session/modules/SessionDataViewer.R
@@ -132,7 +132,13 @@
       if (is.character(label))
          col_label <- label
       else if (idx <= length(colLabels))
+      {
+      # check if colLabels is a named vector, if yes, use the named value, if not use the idx'th element
+      if (!is.null(names(colLabels)))
+         col_label <- colLabels[[col_name]]
+      else
          col_label <- colLabels[[idx]]
+      }
       else 
          col_label <- ""
       


src/cpp/session/modules/SessionDataViewer.R
```diff
@@ -132,7 +132,13 @@
       if (is.character(label))
          col_label <- label
       else if (idx <= length(colLabels))
+      {
+      # check if colLabels is a named vector, if yes, use the named value, if not use the idx'th element
+      if (!is.null(names(colLabels)))
+         col_label <- colLabels[[col_name]]
```
kevinushey, line 138: We should verify that `col_name` is actually a name that exists within `names(colLabels)`.

kevinushey (changes requested):

kevinushey:
Thanks for the pull request! We require a [Contributor License Agreement](https://www.rstudio.com/wp-content/uploads/2014/06/RStudioIndividualContributorAgreement.pdf) before we can accept pull requests on RStudio repositories. Would you be willing to fill one out, and send it to the e-mail indicated in the form? (Please also reference this PR in your e-mail.)

romainfrancois:
I'm wondering if this should rather be dealt with higher up, in: 

```r
   colLabels <- attr(x, "variable.labels", exact = TRUE)
   if (!is.character(colLabels)) 
   {
      colLabels <- character()
   }
```

romainfrancois:
But this might still need to be addressed in `dplyr` for e.g. cases like `select(df, a = b)`, in https://github.com/tidyverse/dplyr/issues/6136

```r
library(dplyr, warn.conflicts = FALSE)

df <- data.frame(a=1, b=2, c=3)
attr(df, "variable.labels") <- c(a = "a", b = "b", c = "c")
purrr::map(.rs.describeCols(select(df,  a = b)), ~.$col_label)
```

gives : 

```
> purrr:: map(.rs.describeCols(select(df,  a = b)), ~.$col_label)
[[1]]
[1] ""
attr(,"class")
[1] "rs.scalar"

[[2]]
[1] "a"
attr(,"class")
[1] "rs.scalar"
```

image

kevinushey:
Superseded by https://github.com/rstudio/rstudio/pull/10335.

PR #10105, opened 2021-11-17
changed + to - in regex to fix latest links

jeffvroom:
Looks like the rsp-session will have the daily in the path now so I will delete that other PR.


MariaSemple (approved):
LGTM!

PR #10136, opened 2021-11-25
Fixed issue with Electron showErrorBox function

### LINKED ISSUE #10114: Electron dev mode startup fails to show config error dialogs, then hangs
### System details

    RStudio Edition : Electron Desktop
    RStudio Version : main branch
    OS Version      : macOS Monterey (haven't tried other macOS versions)
    R Version       : Any

This is a dev-environment-only issue.

Build the native code using `-DRSTUDIO_TARGET=Server`. Specifically, don't build the Qt Desktop.

Then try to start the Electron app using `yarn start`.

The Electron icon appears, but no UI, and you can't quit the Electron app. Have to force-kill it.

There should be an error dialog displaying saying the `rdesktop-dev.conf` file wasn't found. Currently you have to build Qt Desktop to create this file.

We will get rid of this dependency, but more concerned with why showing a dialog with `dialog.showErrorBox` isn't working. In this case, it happens here:

https://github.com/rstudio/rstudio/blob/2853a5d4aa92f323856ec944db77f70d528cc173/src/node/desktop/src/main/application.ts#L117

It is supposed to be safe to use this API even during app startup before `ready`, and it was previously working. Probably either an Electron update, or an electron-forge update somehow broke this.

If we can't make this work then we should switch to just showing the error via `logger().logErrorMessage`, but it would be nice to be able to show error dialogs during startup!

- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).

matheustavaresdev:
### Intent

Solved issue that would make `showErrorBox` function from Electron Dialogs not to show an error dialog and keeping electron's process alive after the app being force closed when an error would occur on startup.

### Approach

A function called `createStandaloneErrorDialog` was created that handles a creation of an error dialog and waiting for it to be closed before quitting the electron app if needed.

### Automated Tests

None

### QA Notes

Please refer to issue #10114

### Checklist

Resolves #10114


gtritchie (approved):
Looks good!

PR #9485, opened 2021-06-09
Add barebones internationalization support

ca-scribner:
### Intent

See #9083 for more context on i18n efforts.

This PR enables i18n in RStudio development and "translates" some content of the AboutDialog in a backwards compatable way.

### Approach

* Enable i18n features by modifying RStudio.gwt.xml to accept only "en" locale and redirect all other locales to "en" by default.  This allows any available translations to be live without changing function for the user.  
* To make development easier, add "dev" locale to RStudioSuperDevMode.gwt.xml and RStudioDesktopSuperDevMode.gwt.xml.  This lets developers add `*_dev.properties` files with obvious changes to see if i18n changes worked without exposing them to and real users
* Add create_dev_locale.sh which creates a "dev" locale from any existing "en" properties files.  The "dev" locale is a copy of "en" properties files with "@" prepended to the strings to make it easy to see if i18n features worked during development
* Add i18n for part of the AboutDialog as a non-breaking proof of concept.  Frame Title, "OK" button, and license message are  "translated" using GWT i18n features to English as proof-of-concept translation.  AboutDialogConstants.java and AboutDialogMessages.java provide the interfaces for translation
* Add "*_dev.properties" to .gitignore to avoid committing any development locales

### Automated Tests

No additional testing is added.  I'm interested in feedback on what testing would be appropriate.

### QA Notes

This change is believed to be fully working without changing the user experience, but feedback is welcome as I am not an expert RStudio or GWT.  To confirm i18n is working for AboutDialog, while running RStudio in devmode you can do one of:
* Modify AboutDialogConstants_en.properties and AboutDialogMessages_en.properties and refresh RStudio with ant in devmode.  The following fields should change (if you have a license, the license message should also change)
image
* Run `create_dev_locale.sh` to clone all `_en` properties files, adding a prefix of "@" to each constant and message.  Then open RStudio using `http://localhost:8787/?locale=dev` and browse to the AboutDialog

### Checklist

- [ ] If this PR adds a new feature, or fixes a bug in a previously released version, it includes an entry in `NEWS.md` 
- [x] If this PR adds or changes UI, the updated UI meets [accessibility standards](https://github.com/rstudio/rstudio/wiki/Accessibility)
- [ ] A reviewer is assigned to this PR (if unsure who to assign, check Area Owners list)
- [ ] This PR passes all local unit tests

cc @gtritchie

PR #10197, opened 2021-12-07
feat(i18n): views-environment

Jose-Matsuda:
Implement i18n for classes under workbench.views.environment.

### Intent

This PR tackles hard-coded strings to i18n-ify in org.rstudio.studio.client.workbench.views.source (excluding false-positives).
Issue: https://github.com/StatCan/daaas/issues/728

### Approach

This pull-request and future ones will tackle i18nify classes at package-level (e.g. org.rstudio.core) instead of the per-component approach (like we did for all-options-menu)

Implementation follows the same pattern from #9485 as described in gwt/tools/i18n-helpers/README.md. Running the create_dev_locale.sh will create all the new "dev" locale from any new "en" properties files for easier visual confirmation. Refer to README for more information.

### Automated Tests

### QA Notes

Modify an entry in one of the new _en.properties files, and refresh RStudio with ant in devmode.
Run create_dev_locale.sh to clone all _en properties files, which will add a prefix of "@" to each constant and message. Then open RStudio using http://localhost:8787/?locale=dev and browse to the Options->* menus. You can also view in desktop mode by swapping the locale element value attribute from value="en" to value="dev" in RStudio.gwt.xml.

### Checklist

- [x] Signed the contributor agreement

cc @gtritchie


src/gwt/src/org/rstudio/studio/client/workbench/views/environment/EnvironmentPane.java
```diff
@@ -100,7 +101,7 @@ public EnvironmentPane(Commands commands,
                           UserPrefs prefs,
                           DependencyManager dependencyManager)
    {
-      super("Environment", events);
+      super(constants_.environmentCapitalized(), events);
```
gtritchie, line 104: There may be non-obvious side effects of changing this. I need to do further research. Might be best to leave this one string unchanged for now until I can look into it. See similar feedback in this PR: #10196

> I need to do some research on this. I suspect this may prove problematic without some additional work and this is very sensitive / fragile code with some non-obvious assumptions.

src/gwt/src/org/rstudio/studio/client/workbench/views/environment/EnvironmentTab.java
```diff
@@ -68,7 +69,7 @@ public EnvironmentTab(final Shim shim,
                          Commands commands,
                          Session session)
    {
-      super("Environment", shim);
+      super(constants_.environmentCapitalized(), shim);
```
gtritchie, line 72: There may be non-obvious side effects of changing this. I need to do further research. Might be best to leave this one string unchanged for now until I can look into it. See similar feedback in this PR: #10196

> I need to do some research on this. I suspect this may prove problematic without some additional work and this is very sensitive / fragile code with some non-obvious assumptions.

gtritchie (changes requested):

src/gwt/src/org/rstudio/studio/client/workbench/views/environment/EnvironmentTab.java
Jose-Matsuda: Ahh that one eh, I was wondering what would happen here. I was wondering about a possible interaction with the 
`eventBus_.fireEvent(new ActivatePaneEvent(constants_.environmentCapitalized()));` present in the EnvironmentPresenter.

Should that be changed as well?

gtritchie: For this PR, just put back to original hardcoded ("Environment") in both EnvironmentPane and EnvironmentTab. There are potentially some other problems that I need to research before we change this (and a bunch of similar things for other tabs and panes).

gtritchie (approved):

PR #10206, opened 2021-12-08
feat(i18n): views-vcs

Jose-Matsuda:
Implement i18n for classes under workbench.views.vcs

### Intent

This PR tackles hard-coded strings to i18n-ify in org.rstudio.studio.client.workbench.views.vcs (excluding false-positives).
Issue: https://github.com/StatCan/daaas/issues/732

### Approach

This pull-request and future ones will tackle i18nify classes at package-level (e.g. org.rstudio.core) instead of the per-component approach (like we did for all-options-menu)

Implementation follows the same pattern from #9485 as described in gwt/tools/i18n-helpers/README.md. Running the create_dev_locale.sh will create all the new "dev" locale from any new "en" properties files for easier visual confirmation. Refer to README for more information.

### Automated Tests

### QA Notes

Modify an entry in one of the new _en.properties files, and refresh RStudio with ant in devmode.
Run create_dev_locale.sh to clone all _en properties files, which will add a prefix of "@" to each constant and message. Then open RStudio using http://localhost:8787/?locale=dev and browse to the Options->* menus. You can also view in desktop mode by swapping the locale element value attribute from value="en" to value="dev" in RStudio.gwt.xml.

### Checklist

- [x] Signed the contributor agreement

cc @gtritchie 

### NOTE
Left **4 TODO** comments regarding some things I were unsure about internationalizing or not


src/gwt/src/org/rstudio/studio/client/workbench/views/vcs/common/diff/LineTableView.java
```diff
@@ -148,8 +149,8 @@ public void render(Context context, ChunkOrLine value, SafeHtmlBuilder sb)
                      sb,
                      RES.cellTableStyle().lineActions(),
                      selectionModel_.getSelectedSet().size() > 1
-                     ? " selection"
-                     : " line");
+                     ? constants_.selectionSuffix("")
```
gtritchie, line 152: Is this technique (passing an empty string to placeholder) required to have a leading space at the start of the localizable string?

`selectionSuffix={0} selection`

Guessing you can't do this?

`selectionSuffix= selection`

src/gwt/src/org/rstudio/studio/client/workbench/views/vcs/common/diff/UnifiedParser.java
```diff
@@ -187,6 +187,7 @@ else if (directive != StringUtil.charAt(diffLine,i))
 
    private boolean isNewFileLine(String nextLine)
    {
+      //TODO unsure about this check with Gary. I did change some SizeWarningWidgets to use `constants_`
```
gtritchie, line 190: Do not localize these. They are for parsing output of git commands, and not shown in the UI.

src/gwt/src/org/rstudio/studio/client/workbench/views/vcs/git/GitStatusRenderer.java
```diff
@@ -98,6 +98,7 @@ private String descForStatus(char c)
    {
       switch (c)
       {
+         // TODO Unsure about these 'case's, might need to leave be for the time
```
gtritchie, line 101: You can localize the return strings here. Don't localize the case values. These strings show up in the Git UI if you hover the mouse over the status graphic (e.g. [M] in screenshot below).

Lower in this file, the `imgForStatus()` function is returning those graphics which contain a letter, such as the [M]. We aren't localizing those graphics, but having the tooltip localized would be good.

Screen Shot 2021-12-08 at 16 52 06

src/gwt/src/org/rstudio/studio/client/workbench/views/vcs/svn/SVNStatusRenderer.java
```diff
@@ -92,6 +92,7 @@ private String descForStatus(String str)
       
       switch (c)
       {
+         // TODO Similar to GitStatusRender unsure about cases
```
gtritchie, line 95: Same answer as for GitStatusRenderer. Localize the returns strings, but not the case values. These strings show up as tooltips in the Subversion UI.

I'd show a screenshot but then I'd have to remember how to setup svn.

gtritchie (changes requested):

src/gwt/src/org/rstudio/studio/client/workbench/views/vcs/common/diff/LineTableView.java
Jose-Matsuda: Yup that's correct, doing `selectionSuffix= selection` will just have `selectionSuffix` resolving to `selection` with no leading space. If you would prefer, we could change it to be `" " + constants_.selectionSuffix()`.

gtritchie: It's fine as-is. I've not worked with Java property files before so wasn't sure.

gtritchie (approved):

Issue #10221, opened 2021-12-09
make it easier to report bugs in the RStudio IDE


kevinushey:
The GitHub issue template has a large number of fields which normally need to be filled out by hand (e.g. the RStudio version, the R version, and so on). It'd be useful to have a tool (both internally, and for users) for reporting of bugs.

AndrewMcClain:
On Ubuntu 20.04.3 LTS I installed `rstudio-2022.02.0-daily-318-amd64.deb` and when invoking `.rs.bugReport()` I receive the following error. Is there a missing dependency or something?

`Error: Clipboard on X11 requires 'xclip' (recommended) or 'xsel'.`

![Screenshot from 2021-12-14 14-11-57.png](https://codetreeusercontent.s3.amazonaws.com/uploads/b590ad51-7393-4bbc-91cd-d19aa0eb1412/Screenshot+from+2021-12-14+14-11-57.png)

kevinushey:
Thanks for catching that -- I should fix this to use RStudio Desktop's internal tools for reading the clipboard.

AndrewMcClain:
Confirmed working in desktop and server as of 2022.02.0-daily+326

PR #10248, opened 2021-12-14
allow session to copy text to clipboard

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/10221.

Previously, we were relying on the `clipr` package being available to write to the clipboard; however, this can be unreliable if the command line tools it relies on aren't available.

This PR makes it possible for the session request the front-end write some text to the clipboard.

### Approach

Fairly straightforward addition of a clipboard module + wiring.

### Automated Tests

N/A

### QA Notes

Test that `.rs.bugReport()` works as expected on RStudio Desktop with Ubuntu 20.04.

### Checklist

### DIFF
--- a/src/cpp/r/R/Tools.R
+++ b/src/cpp/r/R/Tools.R
@@ -1357,9 +1357,35 @@ environment(.rs.Env[[".rs.addFunction"]]) <- .rs.Env
    
    rendered <- sprintf(template, rstudioVersion, rstudioEdition, osVersion, rVersion, rInfoText)
    
-   # try to copy the text to the clipboard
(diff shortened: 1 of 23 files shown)


jmcphers (approved):
I can't put my finger on it, but something feels a little insecure about opening a channel on RStudio Server for R code to write arbitrary content to the user's clipboard in the browser. How would you feel about either (a) making this code desktop only, or (b) having the client display a notification containing a portion of the copied text, so it isn't possible to sneak something in?

kevinushey:
I think you're right. I'll preserve that behavior (write the content to a file and ask the user to copy + paste) for RStudio Server.

PR #10256, opened 2021-12-16
support for R graphics engine 15 (#10058)

jgutman:
### Intent

IDE support for R graphics engine 15 used in R version 4.2 (not yet released -- latest devel build)
Addresses issue #10058.
This should be roughly similar to the [PR for graphics engine 14](https://github.com/rstudio/rstudio/pull/9283)

### Approach

Followed approach outlined in [wiki](https://github.com/rstudio/rstudio-pro/wiki/RStudio-Graphics-Device)

### Automated Tests

None

### QA Notes

TBD

### Checklist

### DIFF
--- a/src/cpp/r/session/graphics/RGraphicsDevDesc.cpp
+++ b/src/cpp/r/session/graphics/RGraphicsDevDesc.cpp
@@ -1380,7 +1380,7 @@ SEXP capabilities(SEXP cap)
     SEXP (*callback)(SEXP cap) = nullptr;
 
     if (callback != nullptr)
-    return callback(cap);
+        return callback(cap);
 
     return R_NilValue;
 }
--- a/src/cpp/r/session/graphics/RGraphicsDevDesc.cpp
+++ b/src/cpp/r/session/graphics/RGraphicsDevDesc.cpp
@@ -228,11 +228,47 @@ pDevDesc allocate(const RSDevDesc& devDesc)
 
       return (pDevDesc) pDD;
    }
+
+   case 15:
+   {
+       DevDescVersion15* pDD =
+               allocAndInitCommonMembers<DevDescVersion15>(devDesc);
+
+       pDD->path = devDesc.path;
+       pDD->raster = devDesc.raster;
+       pDD->cap = devDesc.cap;
+       pDD->eventEnv = devDesc.eventEnv;
+       pDD->eventHelper = devDesc.eventHelper;
+
+       pDD->holdflush = devDesc.holdflush;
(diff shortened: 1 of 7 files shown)


kevinushey (commented):
Nice work! I think the only missing pieces here are bumping the supported graphics device version here:

https://github.com/rstudio/rstudio/blob/7ab9cf87fc53bbb06d8c146e4fb9b27bbc39daac/src/cpp/session/session-options.json#L552-L558

https://github.com/rstudio/rstudio/blob/7ab9cf87fc53bbb06d8c146e4fb9b27bbc39daac/src/cpp/r/include/r/session/RSession.hpp#L65-L78

That was missing from the wiki, so I've since updated it.

src/cpp/r/session/graphics/RGraphicsDevDesc.cpp
```diff
@@ -1194,15 +1255,136 @@ void releaseMask(SEXP ref, pDevDesc dd)
...
+    if (callback != nullptr)
+        callback(path, rule, gc, dd);
+}
+
+SEXP capabilities(SEXP cap)
+{
+    SEXP (*callback)(SEXP cap) = nullptr;
+
+    if (callback != nullptr)
+    return callback(cap);
```
kevinushey, line 1383: nit: indention here?

kevinushey: It also looks like we're not reading the value of the `callback` pointer similar to what we're doing in other methods; does that need to be added back here as well?

jgutman: Yeah good call, I was not exactly sure what to do here because in 
[the R repo code](https://github.com/wch/r-source/blob/68f448578b206175d5d10eb4bd1c55b4ca4a0227/src/include/R_ext/GraphicsDevice.h#L787), the `capabilities` signature doesn't take `pDevDesc dd` like all the other methods do, so wasn't sure what to do in place of 
```
switch (engineVersion)
    {
        case 15:
        default:
            callback = ((DevDescVersion15*)dd)->capabilities;
            break;
    }
```
given that `dd` is not being passed in

kevinushey (approved):
LGTM!

src/cpp/r/session/graphics/RGraphicsDevDesc.cpp
kevinushey: Good question... I think we can visit this in a separate PR; having this be a no-op for now is probably fine.

PR #10278, opened 2021-12-22
Fix HTML in marker message

### LINKED ISSUE #9010: sourceMarkers does not evaluate HTML
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.4.1106
    OS Version      : macOS Mojave 10.14.6        
    R Version       : R version 4.0.3 (2020-10-10)

``` r
foo <- shiny::HTML('<div style="color:red;">I am red</div>')
bar <- shiny::HTML('<p style="color:red;">I am red</p>')
inherits(foo, "html")
#> [1] TRUE
inherits(bar, "html")
#> [1] TRUE

markers <- list(
  list(
    type = "error",
    file = getwd(),
    line = 145,
    column = 1,
    message = foo),
  list(
    type = "info", 
    file = getwd(),
    line = 145,
    column = 1,
    message = bar))

rstudioapi::sourceMarkers("Test Name", markers)
```

<sup>Created on 2021-02-23 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessioninfo::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value                       
#>  version  R version 4.0.3 (2020-10-10)
#>  os       macOS Mojave 10.14.6        
#>  system   x86_64, darwin17.0          
#>  ui       X11                         
#>  language (EN)                        
#>  collate  en_US.UTF-8                 
#>  ctype    en_US.UTF-8                 
#>  tz       Europe/Berlin               
#>  date     2021-02-23                  
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package     * version date       lib source        
#>  assertthat    0.2.1   2019-03-21 [1] CRAN (R 4.0.2)
#>  backports     1.2.1   2020-12-09 [1] CRAN (R 4.0.2)
#>  cli           2.2.0   2020-11-20 [1] CRAN (R 4.0.2)
#>  crayon        1.3.4   2017-09-16 [1] CRAN (R 4.0.2)
#>  digest        0.6.27  2020-10-24 [1] CRAN (R 4.0.2)
#>  ellipsis      0.3.1   2020-05-15 [1] CRAN (R 4.0.2)
#>  evaluate      0.14    2019-05-28 [1] CRAN (R 4.0.1)
#>  fansi         0.4.2   2021-01-15 [1] CRAN (R 4.0.2)
#>  fastmap       1.0.1   2019-10-08 [1] CRAN (R 4.0.2)
#>  fs            1.5.0   2020-07-31 [1] CRAN (R 4.0.2)
#>  glue          1.4.2   2020-08-27 [1] CRAN (R 4.0.2)
#>  highr         0.8     2019-03-20 [1] CRAN (R 4.0.2)
#>  htmltools     0.5.0   2020-06-16 [1] CRAN (R 4.0.2)
#>  httpuv        1.5.4   2020-06-06 [1] CRAN (R 4.0.2)
#>  knitr         1.30    2020-09-22 [1] CRAN (R 4.0.2)
#>  later         1.1.0.1 2020-06-05 [1] CRAN (R 4.0.2)
#>  lifecycle     0.2.0   2020-03-06 [1] CRAN (R 4.0.2)
#>  magrittr      2.0.1   2020-11-17 [1] CRAN (R 4.0.2)
#>  mime          0.9     2020-02-04 [1] CRAN (R 4.0.2)
#>  pillar        1.4.7   2020-11-20 [1] CRAN (R 4.0.2)
#>  pkgconfig     2.0.3   2019-09-22 [1] CRAN (R 4.0.2)
#>  promises      1.1.1   2020-06-09 [1] CRAN (R 4.0.2)
#>  purrr         0.3.4   2020-04-17 [1] CRAN (R 4.0.2)
#>  R6            2.5.0   2020-10-28 [1] CRAN (R 4.0.2)
#>  Rcpp          1.0.5   2020-07-06 [1] CRAN (R 4.0.2)
#>  reprex        1.0.0   2021-01-27 [1] CRAN (R 4.0.2)
#>  rlang         0.4.10  2020-12-30 [1] CRAN (R 4.0.3)
#>  rmarkdown     2.5     2020-10-21 [1] CRAN (R 4.0.3)
#>  rstudioapi    0.13    2020-11-12 [1] CRAN (R 4.0.2)
#>  sessioninfo   1.1.1   2018-11-05 [1] CRAN (R 4.0.2)
#>  shiny         1.5.0   2020-06-23 [1] CRAN (R 4.0.2)
#>  stringi       1.5.3   2020-09-09 [1] CRAN (R 4.0.2)
#>  stringr       1.4.0   2019-02-10 [1] CRAN (R 4.0.2)
#>  styler        1.3.2   2020-02-23 [1] CRAN (R 4.0.2)
#>  tibble        3.0.5   2021-01-15 [1] CRAN (R 4.0.2)
#>  vctrs         0.3.6   2020-12-17 [1] CRAN (R 4.0.2)
#>  withr         2.3.0   2020-09-22 [1] CRAN (R 4.0.2)
#>  xfun          0.19    2020-10-30 [1] CRAN (R 4.0.2)
#>  xtable        1.8-4   2019-04-21 [1] CRAN (R 4.0.2)
#>  yaml          2.2.1   2020-02-01 [1] CRAN (R 4.0.2)
#> 
#> [1] /Library/Frameworks/R.framework/Versions/4.0/Resources/library
```

</details>

grafik

I already filed [an issue at rstudioapi](https://github.com/rstudio/rstudioapi/issues/221) yet have been able to track down the problem now.

The help of ?rstudioapi::sourceMarkers states:

> Note that if the message field is of class "html" (i.e. inherits(message, "html") == TRUE) then its contents will be treated as HTML.

However, when running the Code above, the text is evaluated as is, not as html.

 I actually expected this when providing text of class HTML:
grafik

I was able to track down and solve the issue by just switching [these two lines](https://github.com/rstudio/rstudio/blob/1da8dca94c68fbf2dace6e5977f8bbb2355c5487/src/cpp/r/R/Api.R#L246-L247). The `.rs.scalar()` function overwrites the class of an object to `"rs.scalar"`. Therefore, the following line, which checks whether the `message` string inherits the `"html"` class, is **determined** to be `FALSE`. Changing the order of those lines would solve the issue.

```
# .rs.scalar() function:
function (obj) {
    if (!is.null(obj)) 
        class(obj) <- "rs.scalar"
    return(obj)
}
```

timtmok:
### Intent
Addresses #9010
Allow HTML in the marker message
image

### Approach
Unescapes text when setting innerHTML for marker. The marker has escaped text so it would never be processed as HTML in the marker pane.

### Automated Tests
none

### QA Notes
Marker text containing HTML will need to be escaped or it will be rendered as HTML.
Setting the text to `&lt;div style=&quot;color:red;&quot;&gt;I am red&lt;&#x2F;div&gt;` should be treated as literal text when showing up in the marker pane.

### Checklist


kevinushey:
Do we need to worry about XSS here?

timtmok:
I accidentally closed this!

I don't think XSS is an issue but I only tried setting a marker message with `<script>alert("test");</script>` and didn't get an alert. It does show up in the HTML source even though the marker pane doesn't show the actual text. It is concerning that there would be hidden HTML and I think there is potential for an exploit.

Maybe it makes more sense to allow certain HTML tags or build options for setting text colours?

kevinushey:
> Maybe it makes more sense to allow certain HTML tags or build options for setting text colours?

I think this is the right call. Or even, since we already support ANSI escapes... we could allow (and document) the use of ANSI escapes for colored text, and ask users to use things like `cli::col_red()` to color their text.

Screen Shot 2022-01-03 at 2 01 04 PM

If we go that route, I think it would suffice to have the text "rendered" by a virtual console. I think we do something similar e.g. here:

https://github.com/rstudio/rstudio/blob/9a64952561b434959ffe7af70edb592a73943c8f/src/gwt/src/org/rstudio/studio/client/common/debugging/ui/ConsoleError.java#L66-L68

timtmok:
I've managed to switch over to using the VirtualConsole to render the ANSI codes. I was a bit stuck for a bit until I realized the trim was removing the ANSI escape characters. I have tried to inject HTML into the marker text and it's escaped as expected.

I've retargeted to merge into spotted wakerobin since the docs will need updating.

image

kevinushey (approved):
Very nice!

The next thing we could do is update the documentation in the `rstudioapi` package, which provides the user-facing interface to source marker generation. That documentation lives here:

https://github.com/rstudio/rstudioapi/blob/96fad1d75b5cacd9a0fe9ff7f58d140f5cf4300b/R/stubs.R#L160-L200

Would you be willing to update the documentation there, giving a note that ANSI escapes are supported? (And so, colors as generated via the `cli` package can be used.)

kevinushey:
We can also add a stub to the relevant NEWS file for SW?

timtmok:
I've reset the NEWS file and added in the change. The doc change is over here: https://github.com/rstudio/rstudioapi/pull/242

gadenbuie:
What types of text and content can the VirtualConsole render using ANSI escapes?

timtmok:
Any of the color and background color codes could be used (ie. `cli::col_red` or `cli::bg_white`). The styles allow inverting the background and foreground colors, italic, bold, and underline from what I've tried. It should be any of the codes as defined here: https://en.wikipedia.org/wiki/ANSI_escape_code#SGR_(Select_Graphic_Rendition)_parameters. However, I think some wouldn't work like blink (uses CSS that browsers don't support anymore) and super/sub script (doesn't appear to be handled in VirtualConsole).

gadenbuie:
Thanks for the clarification @timtmok. 

I understand the need to avoid exploits but I think that the colored or styled text in #9010 was primarily to demonstrate the issue and not actually how users would like to the feature. 

Personally, I was looking at using source markers with rich text as a way to provide better feedback to users who are learning R. Being able to include basic HTML, including things like links, lists, formatted text, etc (essentially the HTML produced by strict markdown) made this feature attractive for that purpose, especially since the feedback could be linked to a specific line. This would be even more useful when combined with https://github.com/rstudio/rstudio/issues/10062. There are also a few other IDEs that have this feature and show rich text in linting-related popovers.

kevinushey:
> Personally, I was looking at using source markers with rich text as a way to provide better feedback to users who are learning R. Being able to include basic HTML, including things like links, lists, formatted text, etc (essentially the HTML produced by strict markdown) made this feature attractive for that purpose, especially since the feedback could be linked to a specific line. This would be even more useful when combined with #10062. There are also a few other IDEs that have this feature and show rich text in linting-related popovers.

For what it's worth, @romainfrancois's work in https://github.com/rstudio/rstudio/pull/10299 should make it possible to create custom hyperlinks here as well. The only thing that might not fit here would be lists.

If we wanted to support this case while remaining safe, we could (in theory) allow for strict Markdown text, which we'd then render to HTML before display.

I think it'd be helpful to have an illustrative example of what kind of error you might expect an R user to see, and what text / output you'd like to be able to display, + what interactive features that could have.

gadenbuie:
Strict markdown would definitely work for me, or even HTML with a subset of allowed tags. 

One example might be the [goodpractice](http://mangothecat.github.io/goodpractice/) package, where source markers could be used to provide package authoring advice, e.g.

> ~/source/DESCRIPTION
> `Line 13` **Importing packages with "Depends" is not recommended**
> 
> Avoid using "Depends" in `DESCRIPTION`, as it can cause name clashes, and poor interaction with other packages. Use "Imports" instead.
>
> Read more about package dependency best practices in [Chapter 13: Namespaces of _R Packages_](https://r-pkgs.org/namespace.html#imports).

Another example would be auto-grading messages for example through [gradethis](https://pkgs.rstudio.com/gradethis). In this scenario, a grading script could read an R script or R Markdown document and produce teaching-related content or advice to a learner. Here's an example of the feedback we might want to provide:

> ~/source/student.R
> `Line 13` **Incorrect**
> 
> I expected you to call [`mean()`](https://rdrr.io/r/base/mean.html) where you called [`max()`](https://rdrr.io/r/base/Extremes.html). Try it again. I have a good feeling about this.

This example brings up another idea. It would be extra awesome if there were a mechanism to interact with the IDE in basic ways. The one that comes to mind specifically would be to be able to open a help topic in the Help pane -- rather than linking to external documentation as in the example above.

Issue #10274, opened 2021-12-22
conda packages fail to install in Windows Terminal


kevinushey:
```
> conda install numpy
Collecting package metadata (current_repodata.json): failed

CondaHTTPError: HTTP 000 CONNECTION FAILED for url <https://repo.anaconda.com/pkgs/main/win-64/current_repodata.json>
Elapsed: -

An HTTP error occurred when trying to retrieve this URL.
HTTP errors are often intermittent, and a simple retry will get you on your way.

If your current network has https://www.anaconda.com blocked, please file
a support request with your network engineering team.

'https://repo.anaconda.com/pkgs/main/win-64'
```

This is likely because we're not putting some requisite Anaconda libraries on the PATH, which Windows uses when resolving DLLs. We should try and mock this process; we'll need to put these directories on the PATH:

```
env
env\Library\mingw-w64\bin
env\Library\usr\\bin
env\Library\bin
env\Scripts
env\bin
```

`reticulate` likely needs to do something similar as well.

kevinushey:
In addition, Git Bash doesn't source `.bash_profile` on startup in RStudio Terminals -- rather, it only sources `.bashrc`.

kevinushey:
This may be because we tell the default shell to act like a login shell on not-Windows:

https://github.com/rstudio/rstudio/blob/902bf7f95b1e2ff83521f9d8e2656002272ca129/src/cpp/session/modules/SessionTerminalShell.cpp#L323-L328

We should do the same with Git Bash for parity on Windows.

kevinushey:
To test, do the following on Windows:

1. Ensure that you have Anaconda or Miniconda installed,
2. Ensure that you have the `reticulate` package installed,
3. Try to activate the default conda environment with `reticulate::use_condaenv("r-reticulate", required = TRUE)`,
4. Verify that you've loaded that environment with `py_config()`,
5. Open a new terminal using Git Bash,
6. Try running `conda install numpy`, and confirm that you can install it without issue.

PR #10283, opened 2021-12-23
place conda utilities on PATH for windows

kevinushey:
Addresses https://github.com/rstudio/rstudio/issues/10274.

### DIFF
--- a/src/cpp/session/modules/SessionTerminalShell.cpp
+++ b/src/cpp/session/modules/SessionTerminalShell.cpp
@@ -78,6 +78,7 @@ void scanAvailableShells(std::vector<TerminalShell>* pShells)
 {
 #ifdef _WIN32
    std::vector<std::string> args;
+   std::vector<std::string> bashLoginArgs = { "-l" };
 
    const std::string sys32("system32\\");
    const std::string cmd("cmd.exe");
@@ -92,6 +93,7 @@ void scanAvailableShells(std::vector<TerminalShell>* pShells)
       LOG_ERROR(err);
       return;
    }
+
    std::string progfiles;
    err = core::system::expandEnvironmentVariables("%ProgramFiles%\\", &progfiles);
    if (err)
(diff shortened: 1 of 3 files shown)


CMakeGlobals.txt
```diff
@@ -145,6 +145,9 @@ endif()
 # set session32 if specified
 if(RSTUDIO_TARGET STREQUAL "SessionWin32")
    set(RSTUDIO_SESSION_WIN32 TRUE)
+   add_definitions(-D_X86_)
```
kevinushey, line 148: This seems to be needed to ensure we compile for the right architecture when running on ARM64 Windows.

src/cpp/session/modules/SessionTerminalShell.cpp
```diff
@@ -78,6 +78,7 @@ void scanAvailableShells(std::vector<TerminalShell>* pShells)
 {
 #ifdef _WIN32
    std::vector<std::string> args;
+   std::vector<std::string> bashLoginArgs = { "-l" };
```
kevinushey, line 81: This is required as otherwise the shell isn't treated as a login shell, and the expected profile isn't sourced. (Note that the "default" shell profile _does_ include `-l` so this is specific to Windows bash discovery.)

src/cpp/session/resources/terminal/hooks
```diff
@@ -3,8 +3,39 @@
...
+
+	# on Windows, we need to put a number of locations
+	# on the PATH for conda utilities to function
+	#
+	# in particular, conda's crypto libraries need to
+	# be on the PATH for 'conda install' to function
+	case "$(uname -s)" in
+	MINGW32*)
+		
+		_CONDA_PREFIX=$(cygpath "${CONDA_PREFIX}")
```
kevinushey, line 21: We might need something similar on WSL; note that `cygpath` isn't available by default there.

PR #10301, opened 2022-01-03
use libcurl with windows and R 4.2+

jgutman:
### Intent

wininet for secure download is soft-deprecated in R 4.2. It was previously the default download method on Windows. Users would see a deprecation warning message whenever installing packages on Windows using R-devel (R 4.2+)

We are replacing wininet with libcurl as the default method for secure download for Windows where R is version 4.2+ and libcurl is available.

It deals with how Windows handles downloading any file from the internet (http or https), which most importantly is part of every call to install.packages

You will see that currently in the daily builds if you use R-devel (4.2) on Windows, and run `install.packages(pkg)` you will get a warning message
```
Warning in install.packages :
  the 'wininet' method is deprecated for http:// and https:// URLs
```
And if you run `getOption("download.file.method")` you will get `"wininet"`

This branch should 1) remove the warning you get every time you call `install.packages` and 2) change `getOption("download.file.method")`  to `"libcurl"` for R-devel. 
3) On R 4.1 or below, there should be no warning, and download.file.method will still be "wininet"

### Approach

Use wininet on Windows with R versions 3.2 to 4.1. Use libcurl with R versions 4.2+
No change to linux or Mac

### Automated Tests

### QA Notes

Using R-devel (4.2), ran on Windows 11 VM. Ran `install.packages("remotes")` and verified that package installed successfully, and warning message that was previously there is no longer there.

### Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -38,7 +38,7 @@
 
 * RStudio now supports the experimental UTF-8 UCRT builds of R (#9824)
 * Preliminary support for R graphics engine version 15 in R 4.2.0. (#10058)
-* Default file download method in Windows for R 4.2 and above changed from `wininet` to `libcurl` (#10301)
+* Default file download method in Windows for R 4.2 and above changed from `wininet` to `libcurl` (#10163)
 
 #### Misc
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -36,6 +36,7 @@
 
 #### R
 
+* RStudio now supports the experimental UTF-8 UCRT builds of R (#9824)
 * Preliminary support for R graphics engine version 15 in R 4.2.0. (#10058)
 * Default file download method in Windows for R 4.2 and above changed from `wininet` to `libcurl` (#10301)
 
@@ -52,11 +53,6 @@
 * Added support for Amazon Linux 2 (Pro #2474)
 * Treat Alt and Caption fields differently depending on file type (#9713)
 * Fixed shortcut conflict on German keyboard (#9276)
-
-#### R
-
-* RStudio now supports the experimental UTF-8 UCRT builds of R (#9824)
-* Preliminary support for R graphics engine version 15 in R 4.2.0. (#10058)
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


jgutman:
If the user wishes to revert back to wininet for downloads, they could continue to do so by adding the following line to their `.Rprofile` file: `options(download.file.method = "wininet")`. We could potentially include this documentation somewhere, but upon review of the documentation in [?download.file](https://github.com/wch/r-source/blob/f09d3d7fa4af446ad59a375d914a0daf3ffc4372/src/library/utils/man/download.file.Rd), it sounds like support for libcurl in Windows is no longer optional in R 4.2+, and also that methods which access `https://`, like `install.packages` will try to verify the site certificates. There is much more information about this in the download.file help page, but I'm not sure that anything additional is actually needed here.

melissa-barca (approved):
LGTM!

atocharnaud:
Hi,

We are running R on Windows. Access to the internet is controlled by a security proxy doing deep packet inspection. I couldn't get our certificates to be accepted by commands like download.file or install.packages which are now relying on libcurl rather than wininet even after setting
CURL_CA_BUNDLE=c:\temp\OECDPKIG2PEM_All.txt

download.file('https://cran.rstudio.com/bin/windows/contrib/4.2', destfile =$
trying URL 'https://cran.rstudio.com/bin/windows/contrib/4.2'
Error in download.file("https://cran.rstudio.com/bin/windows/contrib/4.2",  :
  cannot open URL 'https://cran.rstudio.com/bin/windows/contrib/4.2'
In addition: Warning message:
In download.file("https://cran.rstudio.com/bin/windows/contrib/4.2",  :
  URL 'https://cran.rstudio.com/bin/windows/contrib/4.2': status was 'SSL peer certificate or SSH remote key was not OK'
Could you confirm R 4.2.0 implementation of libcurl on windows does properly honor the CURL_CA_BUNDLE environment variable ?

Merci

Arnaud Atoch

PR #10315, opened 2022-01-06
Prevent session resume messages from appearing too often and persisting in console history

cm421:
Improve session resumed message so only the most recent message appears on resume
Add timestamps when a client last connected to a session
Improve detecting of connecting to a running session vs. a suspended session
Remove parenthetical plural from mostSignificantTimeAgo()

### Intent

Addresses https://github.com/rstudio/rstudio-pro/issues/3158. Also fixes https://github.com/rstudio/rstudio-pro/issues/3158

### Approach

Added timestamps that record when a client connects to a session. This "resumed" timestamp, when used in conjunction with the "suspended" timestamp is used to determine if a client is connecting to a session that was awake (has a valid resumed timestamp, but an invalid suspended timestamp) or that was suspended (has a valid suspended timestamp).

Unfortunately, I couldn't come up with a way to determine the difference between connecting to a session that's been active for hours vs a simple browser refresh. To accommodate this, no session resume message is printed if the last timestamp was less than 2 minutes prior. Unfortunately this still means if a user uses the session for 2 or more minutes and then refreshes the browser, this will generate a session resume message.

To prevent the session resumed messages from being added to the console history and stacking up indefinitely, the message is manually added into the client_init message when a client connects to a session, circumnavigating the console history entirely. This results in only the most recent resumed message ever showing.

### Automated Tests

There are already multiple automated tests around this functionality. These changes are likely covered in existing changes and should not break the current tests

### QA Notes

N/A

### Checklist

### DIFF
--- a/src/cpp/core/include/core/r_util/RActiveSessions.hpp
+++ b/src/cpp/core/include/core/r_util/RActiveSessions.hpp
@@ -76,6 +76,9 @@ private:
    const std::string kRVersionHome = "r_version_home";
    const std::string kRVersionLabel = "r_version_label";
    const std::string kWorkingDir = "working_directory";
+   const std::string kLastResumed = "last_resumed";
+   const std::string kSuspendTimestamp = "suspend_timestamp";
+   const std::string kBlockingSuspend = "blocking_suspend";
 
  public:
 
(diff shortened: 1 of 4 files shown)


MariaSemple (approved):
LGTM!

AndrewMcClain:
LGTM!

AndrewMcClain (approved):

Issue #9931, opened 2021-10-07
ragg backend rendering doesn't work in RMarkdown inline execution


Gedevan-Aleksizde:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2021.09.0 build 351
    OS Version      : MacOS Catalina 10.15.7
    R Version       : 4.1.1

### Steps to reproduce the problem

1. set `AGG` as RStudio's backend graphic device
2. write any plot code in a Rmd file, e.g., 
    ````
	---
	title: "Minimal Example"
	output:
	  html_document:
            dev: ragg_png
	---

	```{r}
    plot(1, main = "あいう ABC", xlab = "あいう ABC")
	```
    ````
3. run the chunk inline (click "run current chunk" or press `cmd` + `shift`  + ` return`)

### Describe the problem in detail

The font is not fallbacking and the Japanese letters are not displayed correctly (see the image). So the output graphic seemed not to be rendered by ragg device. 

<img width="1427" alt="スクリーンショット 2021-10-08 1 19 35" src="https://user-images.githubusercontent.com/8291043/136424533-932eb7db-9f34-461b-bd9e-f6f0e0cf5669.png">

NOTE 1: the versions of related packages and libraries are the latest:

* rmarkdown: 2.11
* knitr: 1.36
* ragg: 1.1.3
* XQuartz: 2.8.1

NOTE 2: I confirmed that ragg's font fallbacking works correctly when running the same code in the console (see the Plots pane in the image).

NOTE 3: This problem may be related with issue #7313. But in my cases, there is no explicit error or warning.

NOTE 4: I added extra line to specify Japanese font like the following code (`YuMincho` is one of the Japanese standard font families in MacOS).

````
```{r}
par(family = "YuMincho")
plot(1, main = "あいう ABC", xlab = "あいう ABC")
```
````

Then the following warnings return and the output has no text:

<details>
  <summary>click to expand</summary>

```
Warning in axis(side = side, at = at, labels = labels, ...) :
  no font could be found for family “YuMincho”
Warning in axis(side = side, at = at, labels = labels, ...) :
  no font could be found for family “YuMincho”
Warning in axis(side = side, at = at, labels = labels, ...) :
  no font could be found for family “YuMincho”
Warning in axis(side = side, at = at, labels = labels, ...) :
  no font could be found for family “YuMincho”
Warning in axis(side = side, at = at, labels = labels, ...) :
  no font could be found for family “YuMincho”
[...]

PR #10318, opened 2022-01-06
use graphics device backend type when set, to allow for ragg rendering of NB plot

jgutman:
### Intent

Address the issue in #9931. 

```
plot(1, main = "あいう ABC", xlab = "あいう ABC")
```
should be able to render properly with Japanese characters when ragg is the graphics device, whether the plot is being rendered in the console or inline as a notebook chunk.

### Approach

`.rs.createNotebookGraphicsDevice` function should respect and pass along the graphics device type, if it has been set. We can read it in directly from `getOption("RStudioGD.backend")`. 

The user can set the default device by either adding `options(RStudioGD.backend='ragg')` to their .Rprofile, or in the IDE by going to Tools > Global Options > Graphics and selecting AGG under Backend (after installing `ragg`). 

If the device has been set in the knitr chunk options, it will render using that device when knit, but to render properly inline, the user must set the RStudio graphics device backend explicitly.

### Automated Tests

None

### QA Notes

Should test that 

```{r}
plot(1, main = "あいう ABC", xlab = "あいう ABC")
```
renders both inline and in the console when the graphics device is set to `ragg` (see above) and that Japanese characters are missing when the graphics device is left as Default or any other option

### Checklist

### DIFF
--- a/src/cpp/session/modules/NotebookPlots.R
+++ b/src/cpp/session/modules/NotebookPlots.R
@@ -30,13 +30,6 @@
       res      = 96 * pixelRatio
    )
 
-   gdBackend <- getOption("RStudioGD.backend")
-   if (!identical(gdBackend, "default"))
-   {
-     # pass along graphics device backend if set
-     args$type <- gdBackend
-   }
-
    if (nchar(extraArgs) > 0)
    {
       # trim leading comma from extra args if present
@@ -67,6 +60,15 @@
       )
       
       return(device)
+   }
+
+   gdBackend <- getOption("RStudioGD.backend")
+   if (!identical(gdBackend, "default"))
+   {
+     # pass along graphics device backend if set
+     # we allow this option to be temporarily set by the knitr chunk option while the notebook
+     # chunk is executing; this takes precedence over the device passed via extraArgs
+     args$type <- gdBackend
    }
    
    
--- a/src/cpp/session/modules/NotebookPlots.R
+++ b/src/cpp/session/modules/NotebookPlots.R
@@ -29,7 +29,14 @@
       units    = units,
       res      = 96 * pixelRatio
    )
-   
+
+   gdBackend <- getOption("RStudioGD.backend")
+   if (gdBackend != "default")
+   {
+     # pass along graphics device backend if set
+     args$type <- gdBackend
+   }
+
    if (nchar(extraArgs) > 0)
    {
       # trim leading comma from extra args if present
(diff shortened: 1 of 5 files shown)


src/cpp/session/modules/NotebookPlots.R
```diff
@@ -29,7 +29,14 @@
       units    = units,
       res      = 96 * pixelRatio
    )
-   
+
+   gdBackend <- getOption("RStudioGD.backend")
+   if (gdBackend != "default")
```
kevinushey, line 34:
```suggestion
   if (!identical(gdBackend, "default"))
```

Just to guard against cases where `gdBackend` might not be a string. (this shouldn't happen but better safe than sorry)

kevinushey (commented):
We should also see if we can support chunks which set the `dev` chunk option here.

jgutman:
I think this doesn't quite work, but hopefully is on the right track. The remaining issue is that, as of now, sometimes the device backend is already getting passed on to `.rs.createNotebookGraphicsDevice` via the `extraArgs` command. For example, if the default backend device is set to Cairo, then `, type = \"cairo\"` gets passed into `extraArgs`, so we risk setting the `args$type` twice.

jgutman:
Can test this works with the following file:

~~~
---
title: "Minimal Example"
output:
  html_document
---

```{r setup}
getOption("RStudioGD.backend")
#knitr::opts_chunk$set(dev="ragg_png")
```

```{r debug_chunk, dev='ragg_png'}
getOption("RStudioGD.backend")
plot(1:10, 1:10, main = "あいう ABC", xlab = "あいう ABC")
```

```{r}
getOption("RStudioGD.backend")
plot(1:10, 1:10, main = "あいう ABC", xlab = "あいう ABC")
```

```{r dev='quartz_png', warning=FALSE}
getOption("RStudioGD.backend")
plot(1:10, 1:10, main = "あいう ABC", xlab = "あいう ABC")
```

```{r}
getOption("RStudioGD.backend")
```
~~~


Japanese characters render properly when the RStudio backend device is set to `ragg` any of the following ways:
- via Tools > Global Options > Graphics > Backend -> `AGG`
- via adding `dev='ragg_png'` to a knitr header chunk and executing that chunk inline
- via adding `knitr::opts_chunk$set(dev='ragg_png')` to the setup chunk to execute all chunks with ragg

Japanese characters do not render and display as empty boxes when the RStudio backend device is set to anything else, including default, quartz, and Cairo.

Chunk options override the global RStudio option, so the `dev='quartz_png'` will not render the characters even when the global option is set to ragg, and the `dev='ragg_png'` chunk will render the characters even when the global option is set to default

kevinushey (approved):
LGTM!

I think this PR also warrants an update to `NEWS.md` -- would you mind updating that before merge?

Issue #10192, opened 2021-12-07
session process supervisor should require callbacks run on main thread by default


kevinushey:
Distilled from a conversation with @jjallaire and @jeffvroom.

In general, code within the R session assumes it is running on the main thread; however, some recent changes to allow for handling of offline RPC requests have led to some cases where that same code is executed on a background thread. One example of this is child processes; if the main thread is busy, the background thread might run process callbacks; this is something that should be opt-in rather than opt-out.

We have a process supervisor singleton used by the session here:

https://github.com/rstudio/rstudio/blob/da714bd16c03755efc8ac1856e9f6e8a5484c3ae/src/cpp/session/SessionModuleContext.cpp#L2353-L2357

We should provide a mechanism for constructing that in a way that assumes (by default) callbacks will run on the main thread, but with the ability to override on a case-by-case basis if appropriate.

jeffvroom:
This will be an easy thing to add to another issue I have for PT.

jeffvroom:
@jjallaire and @kevinushey - when I changed the default value of callbacksRequireMainThread to false, it caused the runProgram call in rserver of rserver-acls to hang - and led to [this bug](https://github.com/rstudio/rstudio-pro/issues/3202). The code that calls runProgram to start rserver-acls uses its own ProcessSupervisor, and AsyncChildProcess and requires processSupervisor.poll() to be made for it to process I/O.  In this case, it poll() is called in a scheduled thread and not the "main thread" of rserver.  I'll create a PR that sets this flag to false for all uses of AsyncChildProcess in rserver, but let me know if you think we should handle this some other way. I could see adding an "enforceMainThread" global flag that is only true for rsession and then just only doing this test there so we don't change the default behavior for our APIs.

kevinushey:
Your proposed solution sounds like the right way forward to me.

> I could see adding an "enforceMainThread" global flag that is only true for rsession and then just only doing this test there so we don't change the default behavior for our APIs.

I still think this should probably be done on a case-by-case, since (in theory) other processes not using R might not to be thread-safe by default. (unless I might be misunderstanding your proposal?)

jeffvroom:
[...]

PR #10319, opened 2022-01-06
Fixed errors when opening/saving Rmd files when R is busy

jeffvroom:
### Intent

Fix problems introduced in Ghost Orchid to avoid R code from running from the offline thread (and causing spurious errors)
Addresses https://github.com/rstudio/rstudio/issues/9868 and https://github.com/rstudio/rstudio/issues/10192

### Approach

- Went through the offlineable requests and found a few places
where R code could be invoked by onDocUpdated, onDocAdded, and onDocPendingRemoved  For those code paths, schedule the R code to run in a delayed callback that will later be run on the main thread before handling the next rpc request 

- Make sure that rmarkdownAvailable() can be called from the offline thread. This will get initialized when the session starts, before it's possible for the offline thread to handle any requests. Same thing for reticulatePython()

- Changed a few LOG_ERROR calls to debug messages. This eliminates some spurious errors when a file gets removed that's open in the editor, and when interrupting an R operation that's running

- Changed default so that process supervisor callbacks are only run on the main thread unless explicitly overridden.  Enable offline thread operation for these child processes for the terminal callbacks since they only do I/O and don't depend on R. 

### Automated Tests

I will run all automated tests with 

--- rsession.conf:
session-handle-offline-timeout-ms=0

this will ensure all offlineable requests are runnable in the offline thread without errors.

### QA Notes

- Without setting, session-handle-offline-timeout-ms=0, another way to validate this is working is to run:

> system("sleep 60")

then open, save and create new files:  R, rmd, qmd, and cpp will cover the affected code paths. 

### Checklist


src/cpp/session/modules/data/DataViewer.cpp
```diff
@@ -869,6 +869,35 @@ Error getGridData(const http::Request& request,
    return Success();
 }
 
+Error removeRCachedData(const std::string& cacheKey);
+
+void removeRCachedDataCallback(const std::string& cacheKey)
+{
+   Error error = removeRCachedData(cacheKey);
+   if (error)
+     LOG_ERROR(error);
```
kevinushey, line 878:
```suggestion
      LOG_ERROR(error);
```

kevinushey (approved):
LGTM!

Do you think it's possible (or worthwhile) to capture this pattern into some generic code? E.g. instead of writing:

```
void run()
{
   if (core::thread::isMainThread())
   {
      // do the thing
   }
   else
   {
      module_context::scheduleDelayedWork(
                           boost::posix_time::milliseconds(1),
                           boost::bind(run),
                           false);
   }
}
```

We'd instead be able to just use something like:

```
module_context::executeOnMainThread(run);
```

and that `executeOnMainThread()` helper would take care of running immediately if on main thread, or scheduling for later if not.

jeffvroom:
Thanks for the quick review and nice idea for that utility method. It's in there now and will make it a little easier to find and maintain these code paths.

PR #10320, opened 2022-01-06
Update how quarto is bundled in macOS package

cm421:
### Intent

Places the `quarto/share` folder under `Rstudio.app/Contents/Resources/quarto/` to prevent signing errors caused by a folder inside of `share` called `deno.land`. The period in the folder name throws off the macOS signing code.

### Approach

And Apple-specific switch case was added to copy the quarto directory as normal _except_ for the `share/` directory. The `share/` directory is then copied into `Rstudio.app/Contents/Resources/quarto/`

### Automated Tests

None. This should fix the broken builds by allowing the package to be signed correctly

### QA Notes

N/A

### Checklist


cm421:
Note that this approach is still using the `--deep` flag when signing the package. We should do away with that eventually, but wanted to get this fix in for now so builds can start being signed again

matheustavaresdev (approved):
Looks good to me!

Issue #9880, opened 2021-09-23
Remove r-base from `recommends` package


melissa-barca:
### System details

    RStudio Edition : All
    RStudio Version : Any
    OS Version      :  Ubuntu/Debian
    R Version       : Any

### Steps to reproduce the problem
On a machine with R installed, install the RStudio debian package  (`.deb`) with `apt` or `apt-get`. Note that `r-base` is installed.

### Describe the problem in detail
R is getting installed even once it's already installed, which can lead to the user running an unexpected version of R (worst case) or having an extra version of R they don't use (best case).  The [Admin guide](https://docs.rstudio.com/ide/server-pro/latest/r_versions/installing_r.html) recommends install our R binaries, not the vendor versions.  In the past we've recommended users install RStudio using `gdebi`, which this does **not** happen with.

The issue stems from `apt` installing recommended dependencies while `gdebi` does not. When running with `---no-install-recommends`, R will not be installed. 

### Describe the behavior you expected
We should remove `r-base` from the recommended depencies: https://github.com/rstudio/rstudio/blob/main/package/linux/CMakeLists.txt#L206

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

timtmok:
Verified on an x86_64 VM. `r-base` shows up in the suggested list now.

PR #10334, opened 2022-01-10
Remove r-base from recommended dependencies for linux

jgutman:
### Intent

> Remove r-base from the recommended dependencies for linux builds so we don't re-install the wrong version of R. Addresses #9880

### Approach

> Remove r-base from recommends and change it to suggests in `package/linux/CMakeLists.txt`

### Automated Tests

> NA

### QA Notes

- On Ubuntu/Debian system, install R using [instructions here](https://docs.rstudio.com/resources/install-r/). Verify R installation and version with `/opt/R/${R_VERSION}/bin/R --version`.  Then, install the RStudio debian package (.deb) with apt or apt-get. Verify that R does not get re-installed and the base R version doesn't change. 

- If installing RStudio and base R has not yet been installed, it sshould still suggest installing R

### Checklist


kevinushey:
What about using `CPACK_DEBIAN_PACKAGE_SUGGESTS` instead here?

My main worry is that some users on Linux installing the open-source version of RStudio desktop might not realize they need to install R separately, since many users in those environments will rely on using `r-base` from Debian.

melissa-barca (approved):
LGTM! 

Can you update the QA Notes section in the description with a short note about how QA can test this? Probably something similar to the "Steps to reproduce the problem" section in the original issue.

melissa-barca (approved):

PR #10342, opened 2022-01-11
Add check for "not-a-date-time" values in session property timestamps

cm421:
Improve error message, and reduce severity, when session property timestamps can't be read

### Intent

When reading timestamps from a string/file, a valid value is `not-a-date-time`, which can be generated by `boost::posix_time::to_iso_extended_string`. However, the companion function `boost::posix_time::from_iso_extended_string` does not handle the `not-a-date-time` case and will throw an error:

```
 2022-01-11T20:00:23.749984Z [rsession-rstudio] ERROR Failed to read property suspend_timestamp with value: 'not-a-date-time' error: bad lexical cast: source type value could not be interpreted as target; LOGGED FROM: rstudio_boost::posix_time::ptime rstudio::core::r_util::ActiveSession::ptimeTimestampProperty(const string&) const src/cpp/session/SessionSuspend.cpp:433
```

### Approach

Added a check to handle the `not-a-date-time` case manually
Improved the error message to also print out the value that was read, if any
Lowered the severity of the error, since the code can largely recover without impacting stability, even when the value cannot be read or cast into a `ptime` object

### Automated Tests

None, this is more a convenience change to improve the logs

### QA Notes

None 

### Checklist

### DIFF
--- a/src/cpp/core/include/core/r_util/RActiveSessions.hpp
+++ b/src/cpp/core/include/core/r_util/RActiveSessions.hpp
@@ -414,10 +414,15 @@ private:
    {
       if (!empty())
       {
+         std::string value = "Value Not Read";
          try
          {
-            std::string value = readProperty(property);
+            value = readProperty(property);
             if (value.empty())
+               return boost::posix_time::not_a_date_time;
+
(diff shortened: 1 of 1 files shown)


jeffvroom (approved):
LGTM!

PR #10347, opened 2022-01-12
Avoid offline requests before the session is initialized

jeffvroom:
This fixes a problem that showed up when running the test suite
with session-handle-offline-timeout-ms=0 - basically, running all
offlineable requests on the offline thread. Every once and a while it
would try to run an offline request before the session was initialized.
That would initialize the session on the wrong thread, getting a ton of
errors and eventually a crash.

I don't think it's likely without that setting but just to
be safe I also added the session init test to the offline thread as well.

### Intent

Part of the fix for: https://github.com/rstudio/rstudio/issues/9868

### Approach

Check if session has been initialized before all "offlineable" request handling. 

### Automated Tests

### QA Notes

Tested as part of #9868 

### Checklist


kevinushey (approved):
LGTM!

PR #10348, opened 2022-01-12
Ensure that onWarning onMessage handlers actually get respected during chunk execution

jgutman:
### Intent

Addresses #9436. 

### Approach

Handle message and warning chunk options with rs.notebookConditions.onWarning and rs.notebookConditions.onMessage handlers. These functions handle whether or not warnings and messages should be output to the console, but they weren't actually being called. Assigning the handler to a temporary variable in the global environment, rather than the rstudio::tools environment fixes the issue.

### Automated Tests

NA

### QA Notes

Run the following notebook chunks inline and ensure warnings and messages are shown/not shown as expected.
~~~
---
title: "message"
output: html_document
---

```{r, message=FALSE}
message("You will NOT see the message.")
```

```{r}
message("You WILL see the message.")
```

```{r, message=TRUE}
message("You WILL see the message.")
```


```{r, warning=FALSE}
warning("You will NOT see the warning")
```

```{r}
warning("You WILL see the warning")
```
~~~

### Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,7 @@
 - Added support for using the AGG renderer (as provided by the ragg package) as a graphics backend for inline plot execution; also added support for using the backend graphics device requested by the knitr `dev` chunk option (#9931)
 
 ### Fixed
+- Fixed notebook execution handling of knitr `message=FALSE` chunk option to suppress messages if the option is set to FALSE.
 
 ### Breaking
 
--- a/src/cpp/session/modules/NotebookConditions.R
+++ b/src/cpp/session/modules/NotebookConditions.R
@@ -42,13 +42,8 @@
 
 .rs.addFunction("notebookConditions.connectImpl", function()
 {
-   # NOTE: because the body of this function will be evaluated in the
-   # global environment, we need to avoid defining variables here.
-   #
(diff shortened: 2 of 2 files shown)


kevinushey (approved):
LGTM!

appropiate:
@jgutman, how can I assign "the handler to a temporary variable in the global environment, rather than the rstudio::tools environment"? I get warnings and message when not supposed when running the above chunks. Thanks!

jgutman:
The issue is already resolved when using RStudio 2022.07.0 or higher

appropiate:
Not on my end looks like
Screenshot from 2022-07-31 14-11-47

jgutman:
It looks like you probably have the setting "Show chunk output inline"
turned off. These settings only work when chunk output is displayed inline,
not in the console. The issue of chunk options when output is sent to the
console is a separate issue we are tracking elsewhere, and this resolved PR
is not the best place to comment on it

appropiate:
My apologies. Since the issue I show above happens only when the setting "Show chunk output inline" is turned off, I thought perhaps it was related to that setting somehow. Will try to find the right PR for this then, thanks!

PR #10349, opened 2022-01-12
Add retry and timeout logic for sentry upload

jeffvroom:
(disabled the old retry logic in sentry-upload.sh)

this is primarily to get around the a problem when sentry can hang up the build indefinitely.  Note that previously we were retrying 5 times with 30 seconds of sleep in between. I just disabled that logic in favor of 2 retries with a timeout.


MariaSemple (approved):
LGTM!

PR #10357, opened 2022-01-14
Export PDF option fixes

### LINKED ISSUE #9185: Export PDF Size Wrong
### System details

    RStudio Edition : Desktop
    RStudio Version :  1.4.1106
    OS Version      :  OSX Catalina
    R Version       :  4.0.3

Export any image in PDF format: `plots` panel > `Save as PDF`

image

The default height and width are reversed by default. Should be `5.26 x 7.85` instead of `7.85 x 5.26`. The exported figures are in landscape mode, but portrait canvas with height > width (should be landscape canvas too).

timtmok:
### Intent
Fixes #9185
Fix orientation to match portrait or landscape selection
Fix setting paper size name to custom when setting a custom size

It's confusing for a user to set a size where the dimensions create a landscape aspect ratio and the landscape option is set. It is assumed that the custom dimensions have a width less than the height (portrait). So setting custom dimensions where width is greater than height (landscape), results in a portrait output when landscape is selected.

This change is more intuitive for the user since it will export the plot according to the set orientation and use the dimensions as the paper size.

Also, the paper size name when setting a custom size has been fixed. It never updated the UI to `(Custom)` despite the model updating the available paper sizes.

### Approach
If portrait is selected, set the width to the smaller value and height to the greater value.
If landscape is selected, set the width to the greater value and the height to the smaller value.

The paper size is respected and it takes the guesswork out of setting the plot's orientation.

When setting a custom paper size, the UI is updated with the `(Custom)` label.

### Automated Tests
None

### QA Notes
Any of the paper sizes custom or not should always orient the plot as is selected with the orientation option. There should be no change in behaviour for the suggested paper sizes. For the `(Device size)`, this is based on the size of the Plots pane and will orient the plot to match the selected option. `(Custom)` will replace `(Device size)` when the dimension values are edited.

Also, equal values for the dimensions (square) should render the same for portrait and landscape options.

### Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,7 @@
 
 ### Fixed
 - Fixed notebook execution handling of knitr `message=FALSE` chunk option to suppress messages if the option is set to FALSE (#9436)
+- Fixed plot export to PDF options (#9185)
 
 ### Breaking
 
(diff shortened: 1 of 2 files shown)


kevinushey (approved):
LGTM!

Issue #9813, opened 2021-09-04
find + replace whole word shows whole word only but replaces partial matches


AltfunsMA:
### System details

    RStudio Edition : Server
    RStudio Version : 1.4.1717
    OS Version      :  Ubuntu 20.04
    R Version       :  4.1.0

### Steps to reproduce the problem

1. Create at least one file with the text "proc" and one with "processed" (for example)
2. Search for "proc" with 'whole word' checked in the Find in Files menu
3. Only the one 'proc' will be displayed
4. Replace 'proc' with 'extracted'
5. Result will be 'extracted' in one file and 'extractedessed' in the other

### Describe the problem in detail

The whole word option is ignored during replacement. Even though the GUI displays the number of successes corresponding to the actual count of "proc" it has also replaced all partial matches, and shows them, too.

In the real life case for me this was the result:

![image](https://user-images.githubusercontent.com/26325512/132082070-85c9aaad-fc54-4e45-bf47-874c9313c64c.png)

### Describe the behavior you expected

Only full word matches (equivalent to a regex like "\bproc\b") to be replaced. Alternatively, a clear warning that this option will not work for "replace".

- [ x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [ x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
@AltfunsMA Thank you for raising this!

Reproducible
- RStudio Desktop 1.4.1717 and 2022.02.0+364 on MacOS 12.1
- RStudio Server 2021.09.0+351 on SLES 15 SP3 via Chrome on MacOS 12.1

timtmok:
Working well on Mac and Windows!

PR #10362, opened 2022-01-14
Change regex flags and add unit tests for word boundaries

jgutman:
### Intent

Addresses both #9344 and #9813. 

### Approach

Apparently, `grep` flavor regexes within the `boost::regex_replace` function don't process `\b` as word boundaries, even though grep on the command line, which we use to find the matches in the first place, works just fine. [This page](https://www.boost.org/doc/libs/1_76_0/libs/regex/doc/html/boost_regex/syntax/basic_syntax.html) indicates that the `egrep` flavor which supports POSIX ERE, and is equivalent to `grep -E` in the command line, resolves the original issue.

By adding support for ERE, we have now changed which characters do and do not need to be escaped. In general, only escape sequences such as `\b` need to be escaped.  Characters with special regex meaning

```
.[{}()\*+?|^$
```
will now be treated as literals if escaped with a backslash, and will retain their special regex meaning if unescaped.

- Quantifiers (such as `?` and `+`) were previously supported but were treated as literals if unescaped, and needed to be escaped with `\` to be treated as quantifiers. They will now be treated as quantifiers if unescaped, and will need to be escaped to be treated as literals.
- Parentheses and curly braces `()` and `{}` were previously supported but were treated as literals if unescaped, and needed to be escaped with to be treated as their special regex meaning (e.g. marked sub-expressions, bounded repeats). They will now be parsed with their regex meaning if unescaped, and will need to be escaped to be treated as literals.
- The same is true for other characters such as `|`, `^`, and `$` which previously needed to be escaped and were treated as literals if unescaped. They now do not need to be escaped and will be treated as literals only if escaped.
- Extended regex escape sequences, such as `\w`, `\W`, `\b`, `\B`, `\d`, `\D` that are part of the POSIX ERE standard, are now supported, during find, preview, and replace. See [this page](https://www.boost.org/doc/libs/1_71_0/libs/regex/doc/html/boost_regex/syntax/basic_extended.html#boost_regex.syntax.basic_extended.single_character_character_class) for more examples

See also [this](https://en.wikibooks.org/wiki/Regular_Expressions/POSIX-Extended_Regular_Expressions#Table_of_metacharacters) for a table of the differences in escaping between BRE and ERE.

### Automated Tests

Existing CPP unit tests in [SessionFindTests.cpp](https://github.com/rstudio/rstudio/blob/12bc87d051519f11c096024023cc8996616b138f/src/cpp/session/modules/SessionFindTests.cpp) all pass. I've also added two new test cases to specifically test the word boundary handling

### QA Notes

Should be QA'd on at least Windows and either Mac or Linux, since the find part of Find in Files is handled slightly differently on Windows. 

Should test that when "Whole Word" is checked under Find in Files, the find matches, preview replacements, and final replacements match whole word matches only and do not replace partial word matches (e.g. `map` replaces the `map` in `map this!` and `map(1:5)` but not `mapping`)

Should test that when Regular expression is checked, and special characters such as `.[{}()\*+?|^$` are escaped with a single backslash, they are matched literally.

Should test that when Regular expression is checked and and special characters such as `.[{}()\*+?|^$` are not escaped with a single backslash, they are treated as special regex characters (e.g. `+` means match the preceding item 1 or more times, `{n}` means match the preceding item exactly n times, etc, using the POSIX extended regular expression standard.

Should test that escape sequences such as `\b` (word boundary), `\w` (word-like character) `\d` (digit), `\s` (whitespace characters) and their inverses (e.g. `\B` `\W` `\D`) match expected characters.

### Checklist

### DIFF
--- a/src/cpp/session/modules/SessionFind.cpp
+++ b/src/cpp/session/modules/SessionFind.cpp
@@ -1753,7 +1753,8 @@ core::Error Replacer::replaceRegexIgnoreCase(size_t matchOn, size_t matchOff,
 {
    try
    {
-      boost::regex find(findRegex, boost::regex::grep | boost::regex::icase);
+      // emacs flavor regex includes POSIX BRE
+      boost::regex find(findRegex, boost::regex::emacs | boost::regex::icase);
       core::Error error = completeReplace(find, replaceRegex, matchOn, matchOff, pLine,
          pReplaceMatchOff);
       return error;
@@ -1778,7 +1779,7 @@ core::Error Replacer::replaceRegexWithCase(size_t matchOn, size_t matchOff,
 {
    try
    {
-      boost::regex find(findRegex, boost::regex::grep);
+      boost::regex find(findRegex, boost::regex::emacs);
       core::Error error = completeReplace(find, replaceRegex, matchOn, matchOff, pLine,
          pReplaceMatchOff);
       return error;
--- a/src/cpp/session/modules/SessionFind.cpp
+++ b/src/cpp/session/modules/SessionFind.cpp
@@ -1404,7 +1404,8 @@ core::Error runGrepOperation(const GrepOptions& grepOptions, const ReplaceOption
       cmd << "-c" << "grep.fullName=false";
       cmd << "-C";
       cmd << string_utils::utf8ToSystem(dirPath.getAbsolutePath());
-      cmd <<  "grep";
+      cmd << "grep";
+      cmd << "-E"; // use extended-grep (egrep) for Extended Regular Expressions
       cmd << "-I"; // ignore binaries
       cmd << "--untracked"; // include files not tracked by git...
       cmd << "--exclude-standard"; // but exclude gitignore
@@ -1753,8 +1754,9 @@ core::Error Replacer::replaceRegexIgnoreCase(size_t matchOn, size_t matchOff,
 {
    try
    {
-      // emacs flavor regex includes POSIX BRE
-      boost::regex find(findRegex, boost::regex::emacs | boost::regex::icase);
+
+      // egrep (grep -E) flavor regex includes POSIX ERE
+      boost::regex find(findRegex, boost::regex::egrep | boost::regex::icase);
       core::Error error = completeReplace(find, replaceRegex, matchOn, matchOff, pLine,
          pReplaceMatchOff);
       return error;
@@ -1779,7 +1781,7 @@ core::Error Replacer::replaceRegexWithCase(size_t matchOn, size_t matchOff,
 {
    try
    {
-      boost::regex find(findRegex, boost::regex::emacs);
+      boost::regex find(findRegex, boost::regex::egrep);
       core::Error error = completeReplace(find, replaceRegex, matchOn, matchOff, pLine,
          pReplaceMatchOff);
       return error;
--- a/src/cpp/session/modules/SessionFind.cpp
+++ b/src/cpp/session/modules/SessionFind.cpp
@@ -1757,7 +1757,9 @@ core::Error Replacer::replaceRegexIgnoreCase(size_t matchOn, size_t matchOff,
    {
 
       // egrep (grep -E) flavor regex includes POSIX ERE
-      boost::regex find(findRegex, boost::regex::egrep | boost::regex::icase);
+      boost::regex_constants::syntax_option_type flags = boost::regex::egrep & boost::regex::bk_vbar;
+      flags = flags | boost::regex::icase;
+      boost::regex find(findRegex, flags);
       core::Error error = completeReplace(find, replaceRegex, matchOn, matchOff, pLine,
          pReplaceMatchOff);
       return error;
@@ -1782,7 +1784,8 @@ core::Error Replacer::replaceRegexWithCase(size_t matchOn, size_t matchOff,
 {
    try
    {
-      boost::regex find(findRegex, boost::regex::egrep);
+      boost::regex_constants::syntax_option_type flags = boost::regex::egrep & boost::regex::bk_vbar;
+      boost::regex find(findRegex, flags);
       core::Error error = completeReplace(find, replaceRegex, matchOn, matchOff, pLine,
          pReplaceMatchOff);
       return error;
--- a/src/cpp/session/modules/SessionFind.cpp
+++ b/src/cpp/session/modules/SessionFind.cpp
@@ -1432,6 +1432,7 @@ core::Error runGrepOperation(const GrepOptions& grepOptions, const ReplaceOption
    }
    else
    {
+      cmd << "-E"; // use extended-grep (egrep) for Extended Regular Expressions
       cmd << "--binary-files=without-match";
       cmd << "-rHn";
       cmd << "--color=always";
--- a/src/cpp/session/modules/SessionFind.cpp
+++ b/src/cpp/session/modules/SessionFind.cpp
@@ -1405,7 +1405,6 @@ core::Error runGrepOperation(const GrepOptions& grepOptions, const ReplaceOption
       cmd << "-C";
       cmd << string_utils::utf8ToSystem(dirPath.getAbsolutePath());
       cmd << "grep";
-      cmd << "-E"; // use extended-grep (egrep) for Extended Regular Expressions
       cmd << "-I"; // ignore binaries
       cmd << "--untracked"; // include files not tracked by git...
       cmd << "--exclude-standard"; // but exclude gitignore
@@ -1417,7 +1416,9 @@ core::Error runGrepOperation(const GrepOptions& grepOptions, const ReplaceOption
       // escaping double quotes, etc.
       cmd << "-f";
       cmd << tempFile;
-      if (!grepOptions.asRegex())
+      if (grepOptions.asRegex())
+         cmd << "-E"; // use extended-grep (egrep) for Extended Regular Expressions
+      else
          cmd << "-F";
       addDirectoriesToCommand(
          grepOptions.packageSourceFlag(), grepOptions.packageTestsFlag(), dirPath, &cmd);
@@ -1432,7 +1433,6 @@ core::Error runGrepOperation(const GrepOptions& grepOptions, const ReplaceOption
    }
    else
    {
-      cmd << "-E"; // use extended-grep (egrep) for Extended Regular Expressions
       cmd << "--binary-files=without-match";
       cmd << "-rHn";
       cmd << "--color=always";
@@ -1445,7 +1445,9 @@ core::Error runGrepOperation(const GrepOptions& grepOptions, const ReplaceOption
       // escaping double quotes, etc.
       cmd << "-f";
       cmd << tempFile;
-      if (!grepOptions.asRegex())
+      if (grepOptions.asRegex())
+         cmd << "-E"; // use extended-grep (egrep) for Extended Regular Expressions
+      else
          cmd << "-F";
       for (std::string arg : grepOptions.includeArgs())
          cmd << arg;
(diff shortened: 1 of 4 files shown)


kevinushey:
I think rather than `emacs` we want to just use the default (`perl`). See e.g. https://www.boost.org/doc/libs/1_50_0/libs/regex/doc/html/boost_regex/syntax/perl_syntax.html

My guess is that we chose to use `boost::regex::grep` here because we wanted to be sure the regular expression "flavor" used by the `grep` utility matched what we were trying to use in Boost. I think we wanted to use PCREs with `grep`, but the version of `grep` bundled on Windows is too old:

https://github.com/rstudio/rstudio/issues/3572

So, in order to "fully" fix this I think we should also explore updating the bundled copy of grep we use on Windows.

jgutman:
Yeah, that makes sense to me, this still doesn't give us matching behaviors between the find and replace for things like `\+` and `\?` either so I don't particularly love the solution

kevinushey:
It also looks like we aren't actually trying to request extended regular expressions (or perl regular expressions) here:

https://github.com/rstudio/rstudio/blob/01ef6e05e3a5e02f39741e78cc6519b6b074b20d/src/cpp/session/modules/SessionFind.cpp#L1433-L1454

I wonder if adding `-E` or `-P` (and also updating our usages of `boost::regex` in this PR) would suffice?

src/cpp/session/modules/SessionFind.cpp
```diff
@@ -1404,7 +1404,8 @@ core::Error runGrepOperation(const GrepOptions& grepOptions, const ReplaceOption
       cmd << "-c" << "grep.fullName=false";
       cmd << "-C";
       cmd << string_utils::utf8ToSystem(dirPath.getAbsolutePath());
-      cmd <<  "grep";
+      cmd << "grep";
+      cmd << "-E"; // use extended-grep (egrep) for Extended Regular Expressions
```
kevinushey, line 1408: We should only set this when `grepOptions.asRegex()` is true.

src/cpp/session/modules/SessionFind.cpp
```diff
@@ -1431,6 +1432,7 @@ core::Error runGrepOperation(const GrepOptions& grepOptions, const ReplaceOption
    }
    else
    {
+      cmd << "-E"; // use extended-grep (egrep) for Extended Regular Expressions
```
kevinushey, line 1435: Same as above.

kevinushey: (You might want to combine with https://github.com/rstudio/rstudio/pull/10362/files?w=1#diff-94062616737b3005a9fa89d60aa0596694818c97563995b50cba4661956d69dbR1448-R1449)

src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/FindOutputPresenter.java
```diff
@@ -302,8 +302,12 @@ public void execute()
                         for (String pattern : dialogState_.getExcludeFilePatterns())
                            excludeFilePatterns.push(pattern);
 
-                        server_.completeReplace(dialogState_.getQuery(),
-                                                dialogState_.isRegex(),
+                        String serverQuery = dialogState_.getQuery();
```
jgutman, line 305: Previously, we were padding with `\b` in the find operation but not in the complete replace operation, which is why the find and preview respected the whole word option, but complete replace did not. This fixes that issue. 

This bit could have been separated out into its own PR, since it was strictly a failure to pass the correct args to the RPC, but for purposes of testing I've included it together with the other changes.

kevinushey: Nice catch!

src/cpp/session/modules/SessionFindTests.cpp
```diff
@@ -42,11 +42,16 @@ const size_t rMatchOff = 10;
 const size_t caseMatchOn = 21;
 const size_t caseMatchOff = 27;
 
-const std::string kFindRegex("\\([a-z]\\)\\1\\{2\\}\\([a-z]\\)\\2\\{2\\}");
+const std::string kFindRegex("([a-z])\\1{2}([a-z])\\2{2}");
```
melissa-barca, line 45: Were the original escaping patterns here not actually needed or do your changes affect how the user needs to enter their queries? If the ladder, can you include the differences in the description of this PR?

src/cpp/session/modules/SessionFindTests.cpp
```diff
@@ -246,6 +275,36 @@ TEST_CASE("SessionFind")
...
+      const size_t matchOn = 2;
+      const size_t matchOff = 7;
+      size_t replaceMatchOff;
+
+      replacer.replaceRegex(matchOn, matchOff, regex, replacement, &line, &replaceMatchOff);
+      CHECK(line.compare("heLO!") == 0);
+      CHECK(replaceMatchOff == 4);
+   }
+
+   SECTION("Replace regex with quantifiers")
```
melissa-barca, line 293:
```suggestion
   SECTION("Replace regex with quantifiers and backreference")
```
We should use unique section names for these tests so that it's obvious which test is failing when there is an issue.

melissa-barca (changes requested):
Really awesome job with this! Could you list in the PR what functionality we are gaining with the switch to extended regexes? We should add unit test coverage for any new functionality. 
Could you fill in the QA notes section as well? It will be very important to test this thoroughly on all OSes (windows vs mac vs linux) as well as any new functionality or changes to the way the user escapes queries/replace text. It'd may be helpful to include the section title list of unit tests we have coverage for under the "Automated Tests" section.

src/cpp/session/modules/SessionFindTests.cpp
jgutman: Yes, I think that with ERE enabled, and the regex option selected, the user does not need to escape parentheses and curly braces. If they want a literal parenthesis or curly brace, or if they want a literal `?` or `+`, they will now escape it. Will do!

src/cpp/session/modules/SessionFind.cpp
```diff
@@ -1753,7 +1757,11 @@ core::Error Replacer::replaceRegexIgnoreCase(size_t matchOn, size_t matchOff,
 {
    try
    {
-      boost::regex find(findRegex, boost::regex::grep | boost::regex::icase);
+
+      // egrep (grep -E) flavor regex includes POSIX ERE
+      boost::regex_constants::syntax_option_type flags = boost::regex::egrep & boost::regex::bk_vbar;
+      flags = flags | boost::regex::icase;
```
jgutman, line 1763: I wasn't 100% sure about the difference between `&` and `|` in terms of these flags, how they're combined, and their precedence, but this seemed in line with other usages I saw for these flags

melissa-barca (approved):
LGTM!

src/cpp/session/modules/SessionFind.cpp
kevinushey: These are bitflags; those are usually combined with `|`. `&` is usually used to check if a particular bitflag is "active" in a set.

melissa-barca: Hm, using `&` here means we're not actually passing either flag to our regex and are relying on defaults (perl regexes). You'd get the same result if you only set `boost::regex::icase`.

kevinushey: That is surprising to me. You can see how flags are defined and combined here, for example, in the boost.regex source code:

https://github.com/boostorg/regex/blob/a9f18c2c03a5e64745ec8fcdb8247dbb2a84e9df/include/boost/regex/v4/regbase.hpp#L93-L104

I wonder if we want to start with "normal" and then add some flags to turn off PCRE-specific features? (Or, just use `perl` anyway?)

jgutman:
I think the reason we decided not to use perl was that we'd then need grep
-P for the find to truly be aligned between the find and the replace, and
grep -P is not available on Mac system's grep. But maybe that's not
actually a concern. I guess I'm not sure then exactly what's failing when I
use `flags = boost::regex::egrep | boost::regex::bk_vbar;`. The original
suggestion I saw was to use `flags = boost::regex::egrep  &
~boost::regex::no_bk_refs;` but I'm not sure if that's correct

src/cpp/session/modules/SessionFind.cpp
melissa-barca: Relying on PCRE seems reasonable to me, especially since either way we were announcing a change in behavior and all of our unit tests pass on all platforms. Is that what R uses?

jgutman: I think not by default -- R's grep functions all take a `perl` argument, but it's FALSE by default

melissa-barca:
The tilde is bitwise NOT so that suggestion is equivalent to `flags = boost::regex::egrep;`

jgutman:
> The tilde is bitwise NOT so that suggestion is equivalent to `flags = boost::regex::egrep;`

No but I think it's a double negative, ie `boost::regex::bk_vbar` (enable backrefernces) and `~boost::regex::no_bk_refs` (turn off disabling backreferences) should be the same. Either way, I don't seem to be able to combine them with `egrep` using an `|` and I'm not clear on why

kevinushey:
I think all (or at least most) extended regular expressions are also valid perl regular expressions, so it might be simplest to just use `normal` (or `perl`) for our Boost regular expressions.

melissa-barca:
Ah - I didn't realize the double negative/variable name change, thank you.

From boost's regbase.hpp it looks like no_bk_refs in specific to perl and you'll notice the `egrep` is derived from `extended` which specifies `no_bk_refs`. 

```
   58       // options specific to perl group:
   59       //
   60       no_bk_refs = 1 << 8,                        // \d not allowed
   61       no_perl_ex = 1 << 9,                        // disable perl extensions
   62       no_mod_m = 1 << 10,                         // disable Perl m modifier
   63       mod_x = 1 << 11,                            // Perl x modifier
   64       mod_s = 1 << 12,                            // force s modifier on (overrides match_not_dot_newline)
   65       |o_mod_s = 1 << 13,                         // force s modifier off (overrides match_not_dot_newline)
   66
   67       //
   68       // options specific to basic group:
   69       //
   70       no_char_classes = 1 << 8,                   // [[:CLASS:]] not allowed
   71       no_intervals = 1 << 9,                      // {x,y} not allowed
   72       bk_plus_qm = 1 << 10,                       // uses \+ and \?
   73       bk_vbar = 1 << 11,                          // use \| for alternatives
   74       emacs_ex = 1 << 12,                         // enables emacs extensions
   75
   76       //
   77       // options common to all groups:
   78       //
   79       no_escape_in_lists = 1 << 16,                     // '\' not special inside [...]
   80       newline_alt = 1 << 17,                            // \n is the same as |
   81       no_except = 1 << 18,                              // no exception on error
   82       failbit = 1 << 19,                                // error flag
   83       icase = 1 << 20,                                  // characters are matched regardless of case
   84       nocollate = 0,                                    // don't use locale specific collation (deprecated)
   85       collate = 1 << 21,                                // use locale specific collation
   86       nosubs = 1 << 22,                                 // don't mark sub-expressions
   87       save_subexpression_location = 1 << 23,            // save subexpression locations
   88       no_empty_expressions = 1 << 24,                   // no empty expressions allowed
   89       optimize = 0,                                     // not really supported
   90
   91
   92
   93       basic = basic_syntax_group | collate | no_escape_in_lists,
   94       extended = no_bk_refs | collate | no_perl_ex | no_escape_in_lists,
   95       normal = 0,
   96       emacs = basic_syntax_group | collate | emacs_ex | bk_vbar,
   97       awk = no_bk_refs | collate | no_perl_ex,
   98       grep = basic | newline_alt,
   99       egrep = extended | newline_alt,
```

It might be worth further investigating if there are any scenarios where find doesn't align with replace in the current scenario.

melissa-barca:
> I think all (or at least most) extended regular expressions are also valid perl regular expressions, so it might be simplest to just use `normal` (or `perl`) for our Boost regular expressions.

Yes, as far as I can tell ERE are a subset of PCRE so searching with ERE and replacing with PCRE shouldn't be a problem (while searching with PCRE and replacing with ERE would cause problems).

PR #10368, opened 2022-01-17
Discard pending input on error

timtmok:
### Intent
Addresses #3014
Stops further execution of selected code when encountering an error. The last output in the console will be the last statement executed so that it is easier to see what occurred during execution.

### Approach
Add callback for console reset to clear console input. This appears to be called when an error occurs. Any pending input can be cleared. The callback is only set when executing a code selection. It is not used when submitting the entire script for execution.

### Automated Tests
None

### QA Notes
This only addresses Linux/MacOS. R does not have a console reset callback for Windows.

The code selection sends each line of code as a separate console input. If executing one of the lines produces an error, subsequent lines should not execute. Below, the call to `stop` should be the end when executing these lines. The call to `install.packages` stylizes output like an error but should not halt execution.

```R
print("65")
install.packages("rmarkdown")
stop("something went wrong")
print("3")
```

### Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,7 @@
 
 ### New
 - Source marker `message` can contain ANSI SGR codes for setting style and color (#9010)
+- Linux/MacOS: Executing a code selection that encounters an error will stop execution of remaining code (#3014)
 
 #### R
 - Added support for using the AGG renderer (as provided by the ragg package) as a graphics backend for inline plot execution; also added support for using the backend graphics device requested by the knitr `dev` chunk option (#9931)
--- a/src/cpp/r/include/r/session/RSession.hpp
+++ b/src/cpp/r/include/r/session/RSession.hpp
@@ -188,6 +188,7 @@ struct RCallbacks
    boost::function<void(const std::string&, core::FilePath&, bool)> showFile;
    boost::function<void(const std::string&, int)> consoleWrite;
    boost::function<void()> consoleHistoryReset;
+   boost::function<void()> consoleReset;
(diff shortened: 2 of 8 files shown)


kevinushey (approved):
Awesome work; looks good to me!

Is it worth providing a user preference to enable / disable this feature? (Do you think any users would want to revert to the old behavior?)

I'm going to file an issue with R Core to see if they might be receptive to making this callback available on Windows.

timtmok:
> Awesome work; looks good to me!
> 
> Is it worth providing a user preference to enable / disable this feature? (Do you think any users would want to revert to the old behavior?)

As a new user to RStudio, it seems odd to continue executing code when an error is encountered. I would expect selecting all lines of code to execute similarly to passing the script to R. I would not expect a user to want execution to continue and write R code that expected that.

kevinushey:
I could see one potential reason to keep it: someone trying to demonstrate how error messages are presented in RStudio (e.g. a teacher?) might want to write code of the form:
```
stop("This is error number one")
stop("This is error number two")
stop("This is error number three")
```

They might be surprised / frustrated to find that submitting this code halts after the first `stop()`.

Having a UI preference here also gives us an escape hatch in case we find out that the console is being reset too eagerly in some cases. Since this is a low-level change touching a lot of execution paths, I think it's worthwhile to have a user preference just in case some users need a quick workaround to recover the old behavior.

kevinushey:
Thanks -- I'm going to merge this and create an issue tracking whether we want a user preference.

PR #10379, opened 2022-01-18
OS companion commit to fix defaults for callbacksRequireMainThread for rsession

jeffvroom:
companion to: https://github.com/rstudio/rstudio-pro/pull/3207

This one should have gone into OS first since almost all of the files touched are here. It was reviewed by Kevin in rstudio-pro.


kevinushey (approved):
LGTM!

PR #10384, opened 2022-01-19
Use `reticulate::py_len()` instead of `builtins$len()`

t-kalinowski:
### Intent

Currently, the IDE calls Python's builtin `len()` on defined Python objects, and wraps the call in an R `tryCatch()` in case the object doesn't define a `__len__` method. While the exception is handled on the R side, it is unfortunately allowed to persist in the Python session, resulting in a less-than-useful value for `reticulate::py_last_error()` in interactive sessions.

Related: https://github.com/rstudio/reticulate/issues/1134, cc @kevinushey 

### Approach

A recent patch to reticulate introduced the R function `py_len()`, which can optionally restore the state of the Python session exception global variables.

### Automated Tests

`py_len()` is tested in reticulate.

### Checklist

### DIFF
--- a/src/cpp/session/modules/SessionReticulate.R
+++ b/src/cpp/session/modules/SessionReticulate.R
@@ -121,8 +121,12 @@
       .rs.reticulate.usePython(python)
       
       if (packageVersion("reticulate") >= "1.23")
-         .rs.addFunction("reticulate.describeObjectLength",
-                         function(object) reticulate::py_len(object, -1L))
+      {
+         .rs.addFunction("reticulate.describeObjectLength", function(object)
+         {
+            reticulate::py_len(object, -1L)
+         })
+      }
    })
    
 })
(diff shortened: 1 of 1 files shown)


src/cpp/session/modules/SessionReticulate.R
```diff
@@ -119,6 +119,10 @@
    {
       python <- .rs.readUiPref("python_path")
       .rs.reticulate.usePython(python)
+      
+      if (packageVersion("reticulate") >= "1.23")
+         .rs.addFunction("reticulate.describeObjectLength",
+                         function(object) reticulate::py_len(object, -1L))
```
kevinushey, line 125:
```suggestion
      if (packageVersion("reticulate") >= "1.23")
      {
         .rs.addFunction("reticulate.describeObjectLength", function(object)
         {
            reticulate::py_len(object, -1L))
         })
      }
```

kevinushey (approved):

kevinushey (approved):
LGTM; thanks!

PR #10393, opened 2022-01-19
implement Electron packaging step for macOS

kevinushey:
### Intent

This PR adds the Electron build of RStudio into our build steps.

### Approach

This PR basically adds some more CMake goo to our existing build setup, with a standalone script to build the RStudio Electron pieces, and then merge those together with the other parts of our RStudio bundle.

We should probably talk through whether this is the right approach.


kevinushey:
Note: need to finalize copying of Electron bits to install directory.

gtritchie:
Oops didn't mean to mark this as ready for review. Touched something when looking at this in the github app on my phone!

src/node/desktop/src/main/utils.ts
```diff
@@ -188,14 +202,22 @@ function rsessionNotFoundError(): Error {
...
+
   // determine paths to config file, rsession, and desktop scripts
   let confPath: FilePath = new FilePath();
   let sessionPath: FilePath = new FilePath();
 
-  const binRoot = new FilePath(app.getAppPath());
   if (app.isPackaged) {
+
+    // resolve session path
+    if (process.platform === 'darwin') {
```
kevinushey, line 213: Rather than make this change, we should make sure that we package things into the expected Electron application folder (`Resources/app`; double-check the "real" location)

matheustavaresdev: Do you remember that I said yesterday that if the app is the binary, `app.getAppPath()` will resolve to `Resources/app` and if we are running the app in debug / live reload mode for example, it will resolve to another location? (`src/node/desktop`). 

Altough I haven't ever seen `rsession` inside `src/node/desktop`, maybe this block of code aims on not forcing the code to find `rsession` inside `src/node/desktop` when we are running the app in debug / live reload mode

src/node/desktop/CMakeLists.txt
```diff
@@ -16,35 +16,39 @@
...
-# install mac-terminal script on apple
-if(APPLE)
-   configure_file(${CMAKE_CURRENT_SOURCE_DIR}/mac-terminal.in
-                  ${CMAKE_CURRENT_BINARY_DIR}/mac-terminal)
+# build electron components
+add_custom_target(
+   electron_build ALL
+   DEPENDS rsession diagnostics
+   WORKING_DIRECTORY "${ELECTRON_BUILD_DIR}"
+   COMMAND ${CMAKE_COMMAND} -E env "PATH=${MODIFIED_PATH}" "PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1" ${YARN} make
```
kevinushey, line 48: We probably want `yarn pkg` or something else that avoids the package-building step (as we haven't yet copied in the session components required at packaging time).

matheustavaresdev: We can just remove the DMG and signing steps from `yarn make`. do you want me to do it?

package/osx/make-package
```diff
@@ -313,20 +311,8 @@ if [ "${build_package}" = "1" ]; then
...
-      cd "${ELECTRON_SRC_DIR}"
-      ./scripts/update-json-version.sh "$RSTUDIO_VERSION_FULL"
-      yarn make
-      cd "${PKG_DIR}"
-      PATH="${OLDPATH}"
+   if [ "${build_dmg}" = "1" ]; then
 
-   elif [ "${build_dmg}" = "1" ]; then
       # move to target directory
       cd "${INSTALL_DIR}"
```
kevinushey, line 317: Two approaches to consider here:

- Copy the session components into the Electron build tree, and use Electron forge to produce the final package build;
- Copy the session components into an Electron-specific sub-directory, merge the Electron build together with those session bits, and produce the built DMG ourselves. (Basically, mimicing the Electron packaging process, but still using our existing CMake tools to produce a DMG)

matheustavaresdev: I think 2nd approach is better due to not having to deal with some issues that electron could bring when copying before building the binary...

kevinushey: Sounds good! This is what @gtritchie and I were leaning towards, as well.

src/node/desktop/CMakeLists.txt
kevinushey: Sounds good!

src/node/desktop/src/main/utils.ts
kevinushey: Thanks! I recall that now. I'll try and use that approach to produce a package with the Electron-expected layout here.

kevinushey:
Writing this down because I get dizzy trying to keep it all in my head... the ultimate goal here is to use our existing CMake build workflow, but to be able to merge the Desktop build together with the regular "session" build tree.

The approach I'm taking here:

1. RStudio Desktop is now built as part of the "regular" build pipeline. We can produce both x86 and arm64 desktop builds.

2. The CMake install step prepares the session pieces of our bundle. When we detect an Electron build, we install into `RStudio.app/Contents/Resources/app` instead of the regular location.

3. We then use https://github.com/electron/universal, in `merge-electron.mjs`, to merge the x86 and arm64 Electron builds into a single universal build.

4. Finally, we use `rsync` to "merge" the Universal Electron application into the "install" build tree (where the session bits have already been installed).

5. Now that the package is effectively ready, we run our final post-install steps (fix library paths; notarize).

Still need to tie up some loose ends to make this happen, but this is the approach we've taken. Note that this means we might need to move some of the DMG-related steps that were added to our Electron forge workflow back to CMake.

kevinushey:
Phew! Finally have a working universal Electron build of RStudio produced locally.

```
kevin@MBP-P2MQ:~/rstudio/package/osx/install/RStudio.app/Contents/MacOS [feature/electron-installer]
$ file RStudio
RStudio: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit executable x86_64] [arm64:Mach-O 64-bit executable arm64]
RStudio (for architecture x86_64):      Mach-O 64-bit executable x86_64
RStudio (for architecture arm64):       Mach-O 64-bit executable arm64
```

And it indeed appears to run and do the right thing. It also seems (subjectively) to launch a lot faster.

package/osx/scripts/merge-electron.mjs
```diff
@@ -0,0 +1,26 @@
+
+import 'process';
+import { execSync } from 'child_process';
+import { makeUniversalApp } from '@electron/universal';
```
kevinushey, line 4: @matheustavaresdev, what's the right way to run this script in a way that ensures `@electron/universal` is available? I'm executing it in CMake here:

https://github.com/rstudio/rstudio/blob/6358e490e8ac73e10af12af2a76f50e76d0d8243/package/osx/cmake/merge-electron.cmake#L21-L30

Would it suffice to just call `npm install @electron/universal` before running this script?

package/osx/scripts/merge-electron.mjs
```diff
@@ -0,0 +1,26 @@
...
+// merge the two builds together
+await makeUniversalApp({
+  x64AppPath: x64AppPath,
+  arm64AppPath: arm64AppPath,
+  outAppPath: tmpPath,
+  force: true,
+});
+
+// use rsync to move them into the final install path
+execSync(`rsync -azvhP "${tmpPath}/" "${outPath}/"`, { stdio: 'inherit' });
```
kevinushey, line 22: We can remove the `stdio` piece if we want output to be a little less noisy.

jmcphers (approved):
This LGTM for dev machines but as far as I can tell the resulting packages are not uploaded anywhere yet. If you want to tackle that in this change, use `electron` for the product name you pass to `publish-build.sh` (in the `Jenkinsfile.macos`) -- the dailies page is already primed to pick that up and display Electron builds.

gtritchie (approved):
LGTM!

PR #10380, opened 2022-01-18
Update GNU Grep for Windows to Grep 3.0, after new upload to S3

jgutman:
### Intent

Addresses #3572. Bundle GNU Grep 3.0 instead of Grep 2.5.4 in Windows. New zipped binary uploaded to s3://rstudio-buildtools/gnugrep-3.0.zip

### Approach

Not having a modern grep on Windows is beginning to cause some issues. Here we point install dependencies script for Windows to download s3://rstudio-buildtools/gnugrep-3.0.zip, unzip to `dependencies/windows/gnugrep/3.0` and install to `bin/gnugrep`

### Automated Tests

> Indicate whether this change includes unit tests or integration tests, or link a work item or pull request to add those tests to another repo. If the change cannot or will not be covered by a test, indicate why.

### QA Notes

> Add additional information for QA on how to validate the change, paying special attention to the level of risk, adjacent areas that could be affected by the change, and any important contextual information not present in the linked issues. 

### Checklist

- [ ] If this PR adds a new feature, or fixes a bug in a previously released version, it includes an entry in `NEWS.md` 
- [ ] If this PR adds or changes UI, the updated UI meets [accessibility standards](https://github.com/rstudio/rstudio/wiki/Accessibility)
- [ ] A reviewer is assigned to this PR (if unsure who to assign, check Area Owners list)
- [ ] This PR passes all local unit tests

PR #10410, opened 2022-01-21
Gnugreppath windows -- fix default path

jgutman:
### Intent
As a follow-up to #10380 (addressing #3572), we updated the installation path for grep (GNU grep 3.0) to `bin/grep/3.0`. Previously lived at `bin/grep`.  Forgot to update the default value of `gnugrepPath` in the previous PR, which is necessary for Find in Files functionality in Windows


jgutman:
Find in Files on 2022.06.0+21 wasn't working, failed to find any matches. With this change, building locally on Windows, find in files finds matches again as expected.

kevinushey (approved):
LGTM!

Issue #10381, opened 2022-01-18
unexpected label changes after internationalization


kevinushey:
### System details

    RStudio Edition : 2022.1.999.999
    RStudio Version : Server [Open Source]
    OS Version      : macOS Monterey 12.1
    R Version       : R version 4.1.2 (2021-11-01)

### Steps to reproduce the problem

A number of labels have lost their trailing colon, e.g.

<img width="611" alt="Screen Shot 2022-01-18 at 1 14 20 PM" src="https://user-images.githubusercontent.com/1976582/150019991-a44b75b5-3b81-4b67-bde2-ecae0bebfba8.png">

Compare with RStudio 2021.09.2 Build 382

<img width="605" alt="Screen Shot 2022-01-18 at 1 18 51 PM" src="https://user-images.githubusercontent.com/1976582/150020397-67b157ec-e1a2-46b2-b370-a1a5848d637c.png">

The internationalization of our UI shouldn't change the associated English text, and should preserve colons and other text.

---

<details>

<summary>Session Info</summary>

``` r
R version 4.1.2 (2021-11-01)
Platform: aarch64-apple-darwin20 (64-bit)
Running under: macOS Monterey 12.1

Matrix products: default
LAPACK: /Library/Frameworks/R.framework/Versions/4.1-arm64/Resources/lib/libRlapack.dylib

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

loaded via a namespace (and not attached):
[1] compiler_4.1.2 tools_4.1.2   
```

</details>

gtritchie:
Related to #10249

ronblum:
Verified that both the colons and original language are restored for
- All Global Options—all screens and tabs—in RStudio Desktop 2022.02.0-preview-398 on MacOS 12.1.
- Diagnostics tabs as an additional spot check in RStudio Server 2022.02.0-preview-398 on Ubuntu 20.04.3 via Chrome on MacOS 12.1.

PR #10418, opened 2022-01-24
Fix labels on Global Options dialog to match what they were before localization

gtritchie:
### Intent

Addresses #10381

NOTE: A couple of the property files were saved using Windows line ending; I changed that back to Linux, so you'll want to ignore whitespace when reviewing.

Labels on most controls in Global Options / Code preferences UI were changed as a result of the loc work (#10249). We want the English text to stay the same, so this addresses that.

### Approach

Controls that previously had custom labels (different than what is defined in the preferences schema) now have those labels again. For French, I just reused the existing labels from the preferences schema so this should be a no-op in French (we aren't currently testing French, just FYI).

### Automated Tests

None

### QA Notes

Verify that the English text shown in Global Options / Code (Editing/Display/Saving/Completion/Diagnostics matches what it was before the loc change for preferences was applied ().

You can also check all other Global Preferences UI panes; I didn't find label differences outside the Code pane, but if you do, please open a new issue.

The last build with the old labels was `2022.02.0-daily-357`. The first one with the incorrect labels was `2022.02.0-daily-359`.

### Checklist


jmcphers (approved):

kevinushey (approved):
LGTM!

Issue #10424, opened 2022-01-25
R runtime is accessed from offline thread in onDetectChanges callback


jeffvroom:
### System details

    RStudio Edition : All
    RStudio Version : Ghost orchid and above

### Steps to reproduce the problem

After each RPC request, even when not on the main thread, the onDetectChanges event is sent. This event is handled by the SessionEnvironment.cpp callback that calls an R runtime method: rstudio::r::sexp::listEnvironment(), that uses R_lsInternal - an internal R function that should only be used on the main thread. 

Unfortunately this happens to run without errors most of the time, but intermittently when I run all offlineable requests on a separate thread, and for various automated tests I see this message from R:

`*** recursive gc invocation`

Presumably, this call is made while the main thread is in the midst of a GC and the R_in_gc flag is set that triggers an error when an attempt to access the heap is made from the offline thread.

Most recently it took two runs of the test suite: test_VisualMarkdownEditor to hit this problem (it showed up during different tests).  

### Describe the behavior you expected

* We should not try to refresh the environment from R if not on the main thread. 
* The listEnvironment function (and probably others in RSexp.cpp) should have an ASSERT_MAIN_THREAD directive so we catch problems like this more quickly
* Need to go through other onDetectChanges event listeners to see if any of them can access the main thread

MariaSemple:
Verified fixed.

I was not able to reproduce with 4 test runs on `2022.02.0-preview+401.pro1`, and @ronblum was able to reproduce on `2022.02.0-preview+386.pro2`

PR #10426, opened 2022-01-25
Fixed main thread issues with onDetectChanges event

jeffvroom:
### Intent

Addresses issue: https://github.com/rstudio/rstudio/issues/10424

### Approach

- SessionEnvironment calls an internal R function and so should not run off the main thread
- SessionPlots uses the graphics display object and since we are only running offline RPCs
none of those operations will affect the plots or data viewer
- SessionPackrat need the main thread

Added asserts to the listEnvironments apis

### Automated Tests

Reproduced with the existing automated tests with this custom setting to ensure all offlineable code runs in a thread other than main:

--- /etc/rstudio/rsession.conf:

session-handle-offline-timeout-ms=0

### QA Notes

Even with session-handle-offline-timeout-ms=0, this does not reproduce every time. When I run the test_VisualMarkdownEditor tests the "recursive gc invocation" error would show up in the console in one of the 13 tests every other time I ran it. Since the tests clear the console, and that message is not logged, it's hard to tell it's happening.  I was only able to catch it by setting this environment variable in rsession-run 

     _R_GC_FAIL_ON_ERROR_=true
     export _R_GC_FAIL_ON_ERROR_

With this setting, R will exit after printing that message. It calls the rsession's exit callback and there I was able to log the stack trace to figure out where it was coming from.

### DIFF
--- a/src/cpp/r/RSexp.cpp
+++ b/src/cpp/r/RSexp.cpp
@@ -35,6 +35,7 @@
 #include <core/Macros.hpp>
 #include <core/Log.hpp>
 #include <core/DateTime.hpp>
+#include <core/Thread.hpp>
 
 #include <r/RExec.hpp>
 #include <r/RErrorCategory.hpp>
@@ -374,6 +375,9 @@ void listEnvironment(SEXP env,
                      Protect* pProtect,
                      std::vector<Variable>* pVariables)
 {
+   if (!ASSERT_MAIN_THREAD())
+      return;
+
    // reset passed vars
    pVariables->clear();
    
@@ -425,6 +429,9 @@ void listEnvironment(SEXP env,
 
 void listNamedAttributes(SEXP obj, Protect *pProtect, std::vector<Variable>* pVariables)
 {
+   if (!ASSERT_MAIN_THREAD())
+      return;
+
    // reset passed vars
    pVariables->clear();
 
--- a/src/cpp/session/modules/SessionPackrat.cpp
+++ b/src/cpp/session/modules/SessionPackrat.cpp
@@ -847,6 +847,9 @@ void detectReposChanges()
 
 void onDetectChanges(module_context::ChangeSource source)
 {
+   if (!core::thread::isMainThread())
+      return;
+
    if (source == module_context::ChangeSourceREPL)
       detectReposChanges();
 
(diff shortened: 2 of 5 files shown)


kevinushey:
Do we know why these functions are being called on the offline thread? What do those stack traces look like?

I also wonder if these `detectChanges()` calls should be re-scheduled to run on the main thread, so they aren't silently dropped.

jeffvroom:
Yes, good point.  I should have included the stacktrace.  

The onDetectChanges event is sent after almost all RPC request so I don't think we need to schedule another one. As soon as the RPC that's running on the main thread completes, it will send another onDetectChanges() event and for stuff like Plots, the environment and data viewer panels, that's when we'd want them to update.

Because I'm using the offline timeout of 0, the handleConnection call in this trace come from the HttpConnectionListenerImpl::enqueConnection() method, right on the HTTP acceptor thread. The same thing could happen in SessionOfflineService's call to handleConnection though if R was busy long enough and an offlineable request was made.  

Here's the stacktrace with that _R_GC_FAIL_ON_ERROR_ env var set:

rSuicide(..)
... gc_error(...) - because the flag for gc in progress is set
... inside of libR.so - probably starting with the function: R_lsInternal
rstudio::r::sexp::listEnvironment()
rstudio::session::modules::environment::EnvironmentMonitor::listEnv()
rstudio::session::modules::environment::EnvironmentMonitor::checkForChanges()
rstudio::session::modules::environment::(anonymous namespace)::onDetectChanges()
.... boost signal stuff ...
rstudio::session::(anonymous namespace)::endHandleRpcRequestDirect
rstudio::session::rpc::handleRpcRequest
rstudio::session::http_methods::handleConnection
rstudio::session::HttpConnectionListenerImpl<rstudio_boost::asio::local::stream_protocol>::enqueConnection()

BTW -  I used core::backtrace::printBacktrace and gdb's "info symbol" command to translate the rsession symbols but don't know how to get the symbols for addresses in libR.so.

kevinushey:
> The onDetectChanges event is sent after almost all RPC request so I don't think we need to schedule another one. As soon as the RPC that's running on the main thread completes, it will send another onDetectChanges() event and for stuff like Plots, the environment and data viewer panels, that's when we'd want them to update.

Good point. In that case, I don't think we need to bother re-scheduling here. Thanks!

> BTW - I used core::backtrace::printBacktrace and gdb's "info symbol" command to translate the rsession symbols but don't know how to get the symbols for addresses in libR.so.

For cases like this, I usually try reproducing with my own debug builds of R + RStudio, so that we can get symbols etc. when debugging without too much trouble.

If you're using a Debian build of R, you can also add the relevant debug symbols repository and install those symbols. https://wiki.debian.org/HowToGetABacktrace might be helpful.

PR #10434, opened 2022-01-26
Editing NEWS item for Find in File's move to extended grep

jgutman:
Per [this comment](https://github.com/rstudio/rstudio/issues/9344#issuecomment-1021624714), adding comparison to R's native grep functions

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,7 +12,7 @@
 - Fixed plot export to PDF options (#9185)
 - `.rs.formatDataColumnDispatch()` iterates through classes of `x` (#10073)
 - Fixed Find in Files whole-word replace option, so that when "Whole word" is checked, only file matches containing the whole word are replaced, as displayed in the preview (#9813)
-- Adds support for POSIX extended regular expressions with backreferences in Find in Files find and replace modes, so that special regex characters such as `+` and `?`, `|`, `(`, etc do not need to be escaped, and other ERE escape sequences such as `\b`, `\w`, and `\d` are now supported (#9344)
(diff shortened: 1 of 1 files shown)


NEWS.md
```diff
@@ -12,7 +12,7 @@
 - Fixed plot export to PDF options (#9185)
 - `.rs.formatDataColumnDispatch()` iterates through classes of `x` (#10073)
 - Fixed Find in Files whole-word replace option, so that when "Whole word" is checked, only file matches containing the whole word are replaced, as displayed in the preview (#9813)
-- Adds support for POSIX extended regular expressions with backreferences in Find in Files find and replace modes, so that special regex characters such as `+` and `?`, `|`, `(`, etc do not need to be escaped, and other ERE escape sequences such as `\b`, `\w`, and `\d` are now supported (#9344)
+- Adds support for POSIX extended regular expressions with backreferences in Find in Files find and replace modes, so that special regex characters such as `+` and `?`, `|`, `(`, etc do not need to be escaped, and other ERE escape sequences such as `\b`, `\w`, and `\d` are now supported (#9344). This is in line with the flavor of regular expressions supported by native R's `grep` function, except that Find in Files uses single backslashes rather than double backslashes
```
kevinushey, line 15:
```suggestion
- Adds support for POSIX extended regular expressions with backreferences in Find in Files find and replace modes, so that special regex characters such as `+` and `?`, `|`, `(`, etc do not need to be escaped, and other ERE escape sequences such as `\b`, `\w`, and `\d` are now supported (#9344). This matches the behavior of R's own `grep()` function, but note that backslashes do not need to be escaped (as they typically are in R strings).
```

Issue #10103, opened 2021-11-17
RStudio does not detect conda environments


psads-git:
As the title of this issue tells, my RStudio does not detect conda environments:

![conda](https://user-images.githubusercontent.com/75945439/142244741-52c86e79-a0e7-4f40-bf3d-86a301dc8e8a.png)

This is a follow up of this issue already reported:

https://github.com/rstudio/keras/issues/1291

The RStudio version I am using is the following:

```
RStudio 2021.09.1+372 "Ghost Orchid" Release (8b9ced188245155642d024aa3630363df611088a, 2021-11-11) for Fedora release 35 (Thirty Five)
Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) QtWebEngine/5.15.6 Chrome/87.0.4280.144 Safari/537.36
```

More information below:

```
$ find -L /opt ~ -type f -executable -name python |& grep -v 'find: .* Permission denied'
/home/x/.local/share/r-miniconda/envs/r-reticulate/bin/python
/home/x/.local/share/r-miniconda/pkgs/python-3.6.13-hb7a2778_2_cpython/bin/python
/home/x/.local/share/r-miniconda/pkgs/python-3.7.11-h12debd9_0/bin/python
/home/x/.local/share/r-miniconda/pkgs/python-3.9.5-h12debd9_4/bin/python
```

```
> Sys.getenv()
BASH_FUNC_ml%%          () { module ml "$@" }
BASH_FUNC_module%%      () { eval `/usr/bin/tclsh
                        /usr/share/Modules/libexec/modulecmd.tcl bash
                        "$@"`; _mlstatus=$?; return $_mlstatus }
BASH_FUNC_scl%%         () { if [ "$1" = "load" -o "$1" = "unload" ];
                        then eval "module $@"; else /usr/bin/scl "$@";
                        fi }
BASH_FUNC_which%%       () { ( alias; eval ${which_declare} ) |
                        /usr/bin/which --tty-only --read-alias
                        --read-functions --show-tilde --show-dot "$@" }
CLICOLOR_FORCE          1
CONDA_SHLVL             0
CVS_RSH                 ssh
DBUS_SESSION_BUS_ADDRESS
                        unix:path=/run/user/1000/bus
DEBUGINFOD_URLS         https://debuginfod.fedoraproject.org/
DESKTOP_SESSION         xfce
DISPLAY                 :0.0
EDITOR                  /usr/bin/nano
[...]

PR #10435, opened 2022-01-26
Error message propagate to console when conda cannot find environments

jgutman:
### Intent

Addresses #10103.

### Approach
Take the error message from conda system command and send to console via stop. They might also hit this error via the UI (via Tools > Global Options > Python > Select python interpreter), so it could be worth showing the error in the UI as well instead of just
image

### Automated Tests

NA

### QA Notes

Would be best to have a way to trigger an error upon running `conda env list --json --quiet` in the shell in order to be able to properly test the error handling

### Checklist

### DIFF
--- a/src/cpp/session/modules/SessionPythonEnvironments.R
+++ b/src/cpp/session/modules/SessionPythonEnvironments.R
@@ -639,7 +639,7 @@
    if (attr(output, "status") != 0) {
      errors <- paste(readLines(tmp), collapse = "\n")
      
-     .rs.stopf("Error while executing %s %s:\n%s", conda, paste(args, collapse=" "), errors)
+     .rs.stopf("Error executing %s %s:\n%s", conda, paste(args, collapse = " "), errors)
    }
    json <- .rs.fromJSON(paste(output, collapse = "\n"))
    envList <- unlist(json$envs)
--- a/src/cpp/session/modules/SessionPythonEnvironments.R
+++ b/src/cpp/session/modules/SessionPythonEnvironments.R
@@ -634,7 +634,13 @@
    
    # ask it for environments
    args <- c("env", "list", "--json", "--quiet")
-   output <- system2(conda, args, stdout = TRUE, stderr = FALSE)
+   tmp <- tempfile()
+   output <- system2(conda, args, stdout = TRUE, stderr = tmp)
+   if (attr(output, "status") != 0) {
+     errors <- paste(readLines(tmp), collapse = "\n")
+     
+     stop(conda, args, ":\n", errors)
+   }
    json <- .rs.fromJSON(paste(output, collapse = "\n"))
    envList <- unlist(json$envs)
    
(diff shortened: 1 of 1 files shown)


src/cpp/session/modules/SessionPythonEnvironments.R
```diff
@@ -639,7 +639,7 @@
    if (attr(output, "status") != 0) {
      errors <- paste(readLines(tmp), collapse = "\n")
      
-     stop(conda, args, ":\n", errors)
+     .rs.stopf("Error while executing %s %s:\n%s", conda, paste(args, collapse=" "), errors)
```
kevinushey, line 642:
```suggestion
     .rs.stopf("Error executing %s %s:\n%s", conda, paste(args, collapse = " "), errors)
```

kevinushey (approved):
LGTM!

jgutman:
> LGTM!

Do we want to get this error back to the UI? Is there a way to do that?

kevinushey:
Good question... I had thought it might be available on the error object being logged here:

https://github.com/rstudio/rstudio/blob/cedab12df1b312a0ca1301b3442412c0c776d6a9/src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/PythonPreferencesPaneBase.java#L111-L120

Our R RPC methods get invoked here:

https://github.com/rstudio/rstudio/blob/cedab12df1b312a0ca1301b3442412c0c776d6a9/src/cpp/r/RJsonRpc.cpp#L126-L139

So if the error message is missing, I wonder if there's an appropriate place to tag it on the generated error object.

kevinushey:
Or, I wonder if part of the issue is that this is just a warning, not an error, so it doesn't fail the RPC call? E.g. I see in the console:

```
Warning in system2(conda, args, stdout = TRUE, stderr = FALSE) :
  running command ''/opt/local/bin/conda' env list --json --quiet 2>/dev/null' had status 1
```

and I don't think that causes the rest of the code execution in this RPC to halt. Maybe that's fine?

kevinushey:
> But that's the old code. I changed it to actually raise an error with stop (.rs.stopf)

Sorry, you're right. Not enough sleep / coffee last night...

Let me take a closer look.

kevinushey:
It looks like we do not tag our R code execution errors with the underlying R error message, which is unfortunate:

Screen Shot 2022-01-26 at 11 58 15 AM

Getting that to work properly is going to be a larger change, I think.

jgutman:
Okay. It's not too bad, the dialog box will be unhelpful, but then when the user closes the dialog box, they'll see the actual error message in their console. But seems like we should maybe make an issue for that, because it seems like at some point being able to pass, or categorize, error messages from R to the RPC could be helpful

kevinushey:
> Okay. It's not too bad, the dialog box will be unhelpful, but then when the user closes the dialog box, they'll see the actual error message in their console. But seems like we should maybe make an issue for that, because it seems like at some point being able to pass, or categorize, error messages from R to the RPC could be helpful

Would you mind filing an issue for that?

jgutman:
Note: to reproduce the error for further testing, navigate to `PYTHON_PREFIX/lib/pythonX.Y/site-packages/conda/_vendor/toolz/` and rename `itertoolz.py` to `XXXitertoolz.py`. You should then see an error when running `conda env list --json --quiet`. Be sure to rename it back when you're done.

PR #10436, opened 2022-01-26
Fix `yarn start` for Electron

timtmok:
### Intent
Running Electron in dev on M1 Macs exits with an error that it cannot find the rsession executable. When building arm64, it creates the binary named `rsession` but it is renamed in the packaging step to `rsession-arm64`. In dev, it's build and run (no packaging) so the binary is not named as expected.

### Approach
Check if `app.isPackaged` in addition to arm64 before using `rsession-arm64` as the executable name.

### Automated Tests
None

### QA Notes
This should affect development flows only. When packaged, it should always be using `rsession-arm64` on M1 Macs.

### Checklist


kevinushey:
LGTM; thanks for catching that!

PR #10446, opened 2022-01-26
fixups for code signing on macOS

kevinushey:
### Intent

Ensure that the macOS Electron package build is properly signed.

### Approach

Implement our own helper script for codesigning, to ensure that we properly recursively sign everything in our created package. Also validate the signature after we finish the signature.

### Automated Tests

The build system is the automated test. :-)

### QA Notes

N/A

### Checklist

### DIFF
--- a/package/osx/cmake/codesign-package.cmake
+++ b/package/osx/cmake/codesign-package.cmake
@@ -13,14 +13,12 @@
 #
 #
 
+cmake_minimum_required(VERSION 3.19)
+
 # CMake's message is suppressed during install stage so just use echo here
 function(echo MESSAGE)
 	execute_process(COMMAND echo "-- ${MESSAGE}")
 endfunction()
-
-# don't follow symlinks in GLOB_RECURSE
-cmake_policy(SET CMP0009 NEW)
-cmake_policy(SET CMP0011 NEW)
 
 # flags to pass to codesign executable
 set(CODESIGN_FLAGS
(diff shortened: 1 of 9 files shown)


package/osx/cmake/codesign-package.cmake
```diff
@@ -45,20 +43,15 @@ else()
    list(APPEND CODESIGN_FLAGS -s -)
 endif()
 
-list(APPEND CODESIGN_TARGETS "${CMAKE_INSTALL_PREFIX}/RStudio.app")
```
kevinushey, line 48: I should gate-keep these changes to Electron only, just to avoid potentially breaking the existing desktop builds.

src/node/desktop/forge.config.js
```diff
@@ -65,32 +25,14 @@ const config = {
...
-            path: '/Applications',
-          },
-        ],
-      },
-    },
+      // we don't actually use the created zipfile; this is just a hack
+      // to make sure electron forge does the other packaging steps we
+      // want it to do without complaining we don't have any packaging
+      // steps defined
+      name: '@electron-forge/maker-zip',
```
gtritchie, line 32: I _think_ our builds could switch to just running "electron-forge package" instead of "make" and eliminate this section altogether (thus skipping creation of the unused zipfile).

Conceptually "package" is put the electron bits in the standard layout for a given platform, including minifying the Javascript and all that fun.

"make" is "create a redistributable like a DMG or setup program from that created package" (which we are now doing outside of electron-forge so don't need to bother with)

These phases are described here:

https://www.electronforge.io/cli

Up to you if you want to try this as part of this PR, or do that separately.

src/node/desktop/package.json
```diff
@@ -10,7 +10,7 @@
   "scripts": {
     "clean": "ts-node scripts/clean.ts",
     "lint": "eslint ./src ./test",
-    "make": "wrap () { set -e ; yarn ; yarn run clean ; electron-forge make $*; ./scripts/rename-dmg.sh;}; wrap",
+    "make": "wrap () { set -e; yarn; yarn run clean; electron-forge make $*; }; wrap",
```
gtritchie, line 13: See other comment about potentially just using `electron-forge package` and skipping `make` altogether.

gtritchie (changes requested):

src/node/desktop/forge.config.js
kevinushey: Aha! Thanks; I'll definitely do that as part of this PR. And thanks for the pointer to the documentation.

PR #10452, opened 2022-01-27
Properly handle successful conda execution, status attr will be NULL

jgutman:
Previously only checked if status was not equal to 0, however, if the command executes successfully, it will generally not have a status attribute. Followup to #10435 to fix that.

### DIFF
--- a/src/cpp/session/modules/SessionPythonEnvironments.R
+++ b/src/cpp/session/modules/SessionPythonEnvironments.R
@@ -636,8 +636,9 @@
    args <- c("env", "list", "--json", "--quiet")
    tmp <- tempfile()
    output <- system2(conda, args, stdout = TRUE, stderr = tmp)
-   exit_code <- attr(output, "status")
-   if ( !is.null(exit_code) && !identical(exit_code, 0)) {
+   
+   status <- .rs.nullCoalesce(attr(output, "status", exact = TRUE), 0L)
+   if (!identical(status, 0L)) {
      errors <- paste(readLines(tmp), collapse = "\n")
      .rs.stopf("Error executing %s %s:\n%s", conda, paste(args, collapse = " "), errors)
    }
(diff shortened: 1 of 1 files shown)


src/cpp/session/modules/SessionPythonEnvironments.R
```diff
@@ -636,9 +636,9 @@
    args <- c("env", "list", "--json", "--quiet")
    tmp <- tempfile()
    output <- system2(conda, args, stdout = TRUE, stderr = tmp)
-   if (attr(output, "status") != 0) {
+   exit_code <- attr(output, "status")
+   if ( !is.null(exit_code) && !identical(exit_code, 0)) {
```
kevinushey, line 640: I'm not sure if this is correct. For example:

```
> identical(0, 0L)
[1] FALSE
```

And

```
> status <- system("true")
> status
[1] 0
> identical(status, 0L)
[1] TRUE
> identical(status, 0)
[1] FALSE
```

jgutman: Ah okay. Yeah good point I can switch it to `==`. Or possibly just remove it -- AFAIK if you use `system2` (note we're not using system) with stdout = TRUE, it will not actually set status to 0 to indicate success. 
From the help: 
> If stdout = TRUE or stderr = TRUE...If command runs but gives a non-zero exit status this will be reported with a warning and in the attribute "status" of the result
> In other cases _[meaning, when neither stdout nor stderr is TRUE]_, the return value is an error code (0 for success)

Currently merged version is incorrect because attr(output, "status") will be NULL when the command succeeds, and `NULL != 0` is TRUE

jgutman: I see elsewhere in the code we used 

```
status <- .rs.nullCoalesce(attr(result, "status", exact = TRUE), 0L)
   if (!identical(status, 0L))
```
which looks correct so I'm going to change to that

kevinushey (approved):
LGTM!

PR #10459, opened 2022-01-27
Fix French strings

timtmok:
### Intent
Fixes failure to run `make-package` because an error when building GWT.

### Approach
Escape the apostrophes in the string since it's a special character.

### Automated Tests
None

### QA Notes
The breaking commit is not in `main` so this is only to fix `spotted-wakerobin`

### Checklist


gtritchie (approved):
Thanks for catching and fixing that!

Issue #10451, opened 2022-01-27
Rstudio-2021.09.2-382 version, dir() can't display Chinese characters normally.


wangbinzjcc:
### System details

    RStudio Edition :  Desktop
    RStudio Version :  RStudio-2021.09.2-382
    OS Version      :  Windows 10 x64
    R Version       :  R version 4.1.2

### Steps to reproduce the problem

> dir("D:/test", full.names = T)
[1] "D:/test/涓枃"     "D:/test/涓枃.txt"

### Describe the problem in detail

In the version of RStudio-2021.09.1-372, the file path with Chinese characters can be displayed normally by using dir () in the console.
However, In the versions of RStudio-2021.09.2-382 and RStudio-2022.02.0-preview-392,  the file path with Chinese characters can't be displayed normally by using dir () in the console.

[dir_chinese_path.txt](https://github.com/rstudio/rstudio/files/7946909/dir_chinese_path.txt)

### Describe the behavior you expected

dir("D:/test", full.names = T)
[1] "D:/test/中文" "D:/test/中文.txt"

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
Thanks for the bug report. I'm not entirely sure what's going on, but I'm 99% sure it'll be fixed when R 4.2.0 is released, with UTF-8 locale support.

kevinushey:
As a workaround, you can do something like:

```
files <- dir("D:/test", full.names = TRUE)
Encoding(files) <- "UTF-8"
files
```

and you should hopefully see the correct Chinese text.

wangbinzjcc:
@kevinushey Thank you. The workaround method is feasible. Look forward to the arrival of version R 4.2.0.

wangbinzjcc:
The temporary solution still can't solve the Chinese path problem.

For example: dir("D:/test/中文", full.names = TRUE)

Run directly in R 4.1.2, or use Rstudio-2021.09.1-372 version,  the correct results can be obtained.
[...]

PR #10464, opened 2022-01-28
dodge path re-encoding issues on Windows in list.files() and friends

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/10451.

### Approach

The fix here is somewhat heavy-handed: we overwrite R's implementation of `list.files()`, `list.dirs()`, and `dir()`, with our own versions that better handle different combinations of {system encoding} vs {path encoding}.

This is necessary because RStudio now declares itself to be compatible with the UTF-8 codepage 65001, and so paths returned from R's `list.files()` are actually UTF-8 encoded. However, R does not mark the returned files as such, and so they're mis-interpreted in the native encoding, which may not be UTF-8 for older versions of R.

We still make use of the same underlying internal `list.files()` implementation; we just massage the file encodings + how these functions are invoked in a way that makes sure they behave regardless of the current locale.

### Automated Tests

Tests included, but since they're not run on Windows we'll need extra QA testing.

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/10451. This _definitely_ needs extra scrutiny since we're messing with a base R function!

### Checklist

### DIFF
--- a/src/cpp/session/modules/SessionFiles.cpp
+++ b/src/cpp/session/modules/SessionFiles.cpp
@@ -1300,6 +1300,18 @@ void handleFileExportRequest(const http::Request& request,
    }
 }
 
+SEXP rs_getACP()
+{
+   int codepage = 0;
+
+#ifdef _WIN32
+   codepage = GetACP();
+#endif
+
+   r::sexp::Protect protect;
+   return r::sexp::create(codepage, &protect);
+}
+
 SEXP rs_pathInfo(SEXP pathSEXP)
 {
    try
@@ -1397,8 +1409,9 @@ Error initialize()
    // subscribe to events
    events().onClientInit.connect(bind(onClientInit));
 
-   RS_REGISTER_CALL_METHOD(rs_readLines, 1);
-   RS_REGISTER_CALL_METHOD(rs_pathInfo, 1);
+   RS_REGISTER_CALL_METHOD(rs_getACP);
+   RS_REGISTER_CALL_METHOD(rs_pathInfo);
+   RS_REGISTER_CALL_METHOD(rs_readLines);
 
    // install handlers
    using boost::bind;
--- a/src/cpp/session/modules/SessionFiles.cpp
+++ b/src/cpp/session/modules/SessionFiles.cpp
@@ -1429,10 +1429,10 @@ Error initialize()
       (bind(registerRpcMethod, "move_files", moveFiles))
       (bind(registerRpcMethod, "rename_file", renameFile))
       (bind(registerRpcMethod, "touch_file", touchFile))
-      (bind(registerUriHandler, "/files", handleFilesRequest))
-      (bind(registerUploadHandler, "/upload", handleFileUploadRequestAsync))
-      (bind(registerUriHandler, "/export", handleFileExportRequest))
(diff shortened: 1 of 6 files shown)


MariaSemple:
@kevinushey did you intend to merge this into the PT release? If so, can you rebase it to `prairie-trillium`?

jgutman:
Since Kevin mentions that this requires "extra QA testing" and "extra scrutiny" maybe it's best to keep it in main for Spotted Wakerobin?

kevinushey:
I was also wondering whether we should just provide our own implementation (e.g. built on top of Boost.Filesystem) that tries to mimic R's own implementations as closely as possible.

src/cpp/session/modules/SessionFiles.cpp
```diff
@@ -1397,8 +1409,9 @@ Error initialize()
    // subscribe to events
    events().onClientInit.connect(bind(onClientInit));
 
-   RS_REGISTER_CALL_METHOD(rs_readLines, 1);
-   RS_REGISTER_CALL_METHOD(rs_pathInfo, 1);
+   RS_REGISTER_CALL_METHOD(rs_getACP);
```
jgutman, line 1412: Is there a reason for removing the 1 from the RS_REGISTER_CALL_METHOD calls?

jgutman (approved):
LGTM, although I'm curious to know what alternative approaches are possible.

Also, can you clarify, when we bind our functions to these base R functions, what does the user see if they try to look at the source code for the original function with `View()` or the help page with `?` (not that I'd expect the source code to be particularly helpful since it's just `.Internal`)

src/cpp/session/modules/SessionFiles.cpp
kevinushey: Those are sort of unneeded historical artifacts. Back when we were compiling against C++98, we didn't have a clean way of figuring out how many parameters each call method received, and so we have to explicitly declare it. That was a bit error-prone, as we occasionally forgot to update our registration code whenever we updated / extended an existing `rs_` method.

Now that we can figure it out using C++11 tooling (https://github.com/rstudio/rstudio/blob/89d90af57611a1d5e42b3dfdd2abf3565d029ea1/src/cpp/r/include/r/RRoutines.hpp#L80) I've been slowly cleaning up some of those usages as I see them.

kevinushey:
> LGTM, although I'm curious to know what alternative approaches are possible.

The other thing we could do is try to implement our own "list files" functionality, that tries to behave as similarly as R's own internal implementation as possible. I wasn't sure if we'd be able to match all of the nuances of R's implementation ourselves, so I tried this approach first (which has its own caveats; e.g. I'm not sure if we can be sure we can always `setwd()`

> Also, can you clarify, when we bind our functions to these base R functions, what does the user see if they try to look at the source code for the original function with View() or the help page with ? (not that I'd expect the source code to be particularly helpful since it's just .Internal)

In this case, they'd see our implementation text here: https://github.com/rstudio/rstudio/pull/10464/files?w=1#diff-c6b73983b11a05d33626cb51249e44dfa5c6171f65975ba52439cf4a9c379abaR39-R51

kevinushey:
Superceded by https://github.com/rstudio/rstudio/pull/10481.

PR #10481, opened 2022-02-01
implement own list.files list.dirs methods built on Boost.Filesystem

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/10451. Alternate to https://github.com/rstudio/rstudio/pull/10464.

### Approach

Implement our own file listing routines, using Boost.Filesystem. Might be worth discussing if we like this more than https://github.com/rstudio/rstudio/pull/10464.

### Automated Tests

Extensive tests are included.

### QA Notes

Note that these changes should only effect Windows. Test something like the following:

```
# restore the original R bindings
.rs.files.restoreBindings()

# try comparing R's output to ours
lhs <- list.files("~", recursive = TRUE)
rhs <- list.files("~", recursive = TRUE)

# hopefully TRUE
all.equal(lhs, rhs)
```

### Checklist

### DIFF
--- a/src/cpp/r/RSexp.cpp
+++ b/src/cpp/r/RSexp.cpp
@@ -1517,9 +1517,8 @@ SEXP createUtf8(const std::string& data, Protect* pProtect)
    SEXP strSEXP;
    pProtect->add(strSEXP = Rf_allocVector(STRSXP, 1));
 
-   SEXP charSEXP;
-   pProtect->add(charSEXP = Rf_mkCharLenCE(data.c_str(), data.size(), CE_UTF8));
-
+   // protection not required as long as we immediately assign into protected STRSXP
+   SEXP charSEXP = Rf_mkCharLenCE(data.c_str(), data.size(), CE_UTF8);
    SET_STRING_ELT(strSEXP, 0, charSEXP);
    return strSEXP;
 }
@@ -1536,8 +1535,8 @@ SEXP createUtf8(const std::vector<std::string>& data, Protect* pProtect)
    
    for (std::size_t i = 0, n = data.size(); i < n; i++)
    {
-      SEXP charSEXP;
-      pProtect->add(charSEXP = Rf_mkCharLenCE(data[i].c_str(), data[i].size(), CE_UTF8));
+      // protection not required as long as we immediately assign into protected STRSXP
+      SEXP charSEXP = Rf_mkCharLenCE(data[i].c_str(), data[i].size(), CE_UTF8);
       SET_STRING_ELT(strSEXP, i, charSEXP);
    }
    
--- a/src/cpp/r/RSexp.cpp
+++ b/src/cpp/r/RSexp.cpp
@@ -1529,6 +1529,21 @@ SEXP createUtf8(const FilePath& filePath, Protect* pProtect)
    return createUtf8(filePath.getAbsolutePath(), pProtect);
 }
 
+SEXP createUtf8(const std::vector<std::string>& data, Protect* pProtect)
+{
+   SEXP strSEXP;
+   pProtect->add(strSEXP = Rf_allocVector(STRSXP, data.size()));
+   
+   for (std::size_t i = 0, n = data.size(); i < n; i++)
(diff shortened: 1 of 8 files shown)


jgutman:
Still reviewing, but you might want to rebase against prairie-trillium branch and switch the base branch so this can get in

src/cpp/tests/testthat/test-files.R
```diff
@@ -73,3 +73,61 @@ test_that("file listings are correct", {
...
+   
+   dir.create("empty")
+   
+   dir.create("hasEmptyDir")
+   dir.create("hasEmptyDir/empty")
+   
+   nihao <- enc2utf8("\u4f60\u597d")  # 你好
+   dir.create(nihao)
+   file.create(paste(nihao, "file", sep = "/"))
+   file.create(paste(nihao, "R", sep = "."))
```
jgutman, line 101: Nice tests!

jgutman (approved):
I prefer this implementation, if only because I feel squishy about trying to change the working directory and change it back

src/cpp/r/RSexp.cpp
```diff
@@ -1517,9 +1517,8 @@ SEXP createUtf8(const std::string& data, Protect* pProtect)
    SEXP strSEXP;
    pProtect->add(strSEXP = Rf_allocVector(STRSXP, 1));
 
-   SEXP charSEXP;
-   pProtect->add(charSEXP = Rf_mkCharLenCE(data.c_str(), data.size(), CE_UTF8));
-
+   // protection not required as long as we immediately assign into protected STRSXP
+   SEXP charSEXP = Rf_mkCharLenCE(data.c_str(), data.size(), CE_UTF8);
```
jgutman, line 1521: Do you have any resources to better understand what this is doing here?

src/cpp/session/modules/SessionFiles.cpp
```diff
@@ -1347,6 +1364,281 @@ SEXP rs_pathInfo(SEXP pathSEXP)
...
+      const ListFilesOptions& options,
+      F&& accept,
+      std::vector<boost::filesystem::path>* pResult)
+{
+   // iterate through other files
+   for (auto&& path : paths)
+   {
+      // check for existence (swallow other errors)
+      boost::system::error_code ec;
+      if (boost::filesystem::exists(path, ec))
```
jgutman, line 1474: Did you have in your tests a test for a non-existent path? I didn't notice it, but it could be helpful?

src/cpp/session/modules/SessionFiles.cpp
```diff
@@ -65,8 +68,22 @@
 #include "SessionFilesListingMonitor.hpp"
 #include "SessionGit.hpp"
 
+#ifdef BOOST_WINDOWS_API
+# define kEmptyString L""
+# define kDotPath L"."
+# define kDotDotPath L".."
+#else
```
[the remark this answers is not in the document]
kevinushey: The `L` prefix is used to denote so-called "wide" strings; see https://docs.microsoft.com/en-us/cpp/cpp/string-and-character-literals-cpp?view=msvc-170 for an introduction.

Boost uses "wide" strings on Windows for its filesystem strings, so we need to similarly use wide strings ourselves when interacting with Boost here.

See also https://docs.microsoft.com/en-us/windows/win32/learnwin32/working-with-strings for some of the background around how wide strings are used in Windows + Windows APIs.

src/cpp/r/RSexp.cpp
kevinushey: The available documentation here is unfortunately kind of sparse; see e.g.

https://cran.r-project.org/doc/manuals/r-release/R-ints.html#index-mkCharLenCE
https://github.com/hadley/r-internals/blob/master/strings.md

src/cpp/session/modules/SessionFiles.cpp
kevinushey: Good idea! I'll add that.

Issue #10415, opened 2022-01-22
Support R_ResetConsole on Windows


kevinushey:
I wonder if we could just redefine R_ResetConsole and force R to use our implementation on Windows?

kevinushey:
This was implemented with R 4.2.0 + RStudio 2022.02.

PR #10496, opened 2022-02-03
support R_ResetConsole on Windows (R >= 4.2.0)

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/10415.

### Approach

Take advantage of newly-provided `R_ResetConsole` hook on Windows.

### Automated Tests

Would be worth testing output of the form:

```
x <- 1
stop("an error occurred")
x <- 2
```

and confirm that `x <- 2` is not executed. (Only works with recent versions of R-devel, which will be released as R 4.2.0).

### QA Notes

See above.

### Checklist

### DIFF
--- a/src/cpp/r/session/REmbeddedWin32.cpp
+++ b/src/cpp/r/session/REmbeddedWin32.cpp
@@ -49,16 +49,61 @@
 #include <Rembedded.h>
 #include <graphapp.h>
 
+#ifdef ReadConsole
+# undef ReadConsole
+#endif
+
+#ifdef WriteConsole
+# undef WriteConsole
+#endif
+
 extern "C" void R_ProcessEvents(void);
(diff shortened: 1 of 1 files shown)


timtmok:
@kevinushey Was this meant to merge into PT? My other change for reseting the console is in SW.

kevinushey:
Thanks for catching that! It was -- currently, changes in R-devel mean that RStudio needs to set the `ResetConsole` hook; otherwise, RStudio risks crashing on startup. (I'm working with a member of R Core to see if we can relax this requirement, since that change means that older versions of RStudio could crash when using the next version of R.)

However, for Prairie Trillium, I've made further changes to ensure we set the `ResetConsole` hook to `NULL` explicitly on Windows, since we won't be supporting that in PT.

PR #10503, opened 2022-02-04
Fix blank recent file/project menu items in Electron

timtmok:
### Intent
Fixes the blank menu items in the recent files/projects submenus.

### Approach
Electron creates menu items visible by default and the GWT commands for the recent items are not visible to start. The `addCommand` to `DesktopMenuCallback` doesn't specify visibility so Electron and GWT are out-of-sync. Adding visibility to `addCommand` resolves the bug.

The changes to the C++ side is for compatibility with Qt RStudio.

### Automated Tests
I'd like to add some kind of automated test around this but an integrated test might not be possible. I'm getting familiar with the unit tests and thinking of a way that it could be verified.

### QA Notes
The recent items should update immediately when opening a file or project. The item order should also update when opening new files or projects. Clearing the list should clear immediately and restarting should still show a cleared list (if a project was already open, it would add that item back).

The Qt version could have an impact but there should be no change to how the recent items work.

All menu items should also behave the same as before. Visible menu items by default should still be visible.

### Checklist

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/MRUList.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/MRUList.java
@@ -19,6 +19,7 @@ import org.rstudio.core.client.DuplicateHelper;
 import org.rstudio.core.client.command.AppCommand;
 import org.rstudio.core.client.command.AppMenuItem;
 import org.rstudio.core.client.command.CommandHandler;
+import org.rstudio.core.client.command.impl.DesktopMenuCallback;
 import org.rstudio.core.client.widget.OperationWithInput;
 
 import java.util.*;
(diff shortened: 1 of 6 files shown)


src/gwt/src/org/rstudio/studio/client/workbench/MRUList.java
```diff
@@ -128,6 +129,7 @@ private void updateCommands()
       if (hideClearOnEmpty_)
          clearCommand_.setVisible(clearCommand_.isEnabled());
       manageCommands(mruEntries_, mruCmds_);
+      DesktopMenuCallback.commitCommandShortcuts();
```
timtmok, line 132: This is required to force Electron to update the menu item labels. Updating the menu item isn't enough and it has to be rebuilt. Better to do it here so it catches all menu items that may have changed visibility, labels, and description.

gtritchie (approved):
LGTM!

PR #10524, opened 2022-02-08
produce separate UTF-8 rsession binary for ucrt builds

kevinushey:
### Intent

As discovered in https://github.com/rstudio/rstudio/issues/10507, requesting the use of a UTF-8 code page will cause issues with older versions of R. Unfortunately, because the application code page must be set at build time, the only apparent solution is to build and bundle two separate copies of the `rsession.exe` executable -- one for older builds of R, and one for newer (UCRT) builds of R.

### Approach

Use CMake to produce two separate `rsession` binaries, one using a "plain" manifest and one using a UTF-8 manifest.

### Automated Tests

N/A

### QA Notes

To test, when running RStudio on Windows, try running something like:

```
str(l10n_info())
system(paste("tasklist /FI \"PID eq", Sys.getpid(), "\""))
```

On older versions of R (< 4.2.0), you should see:

```
> str(l10n_info())
List of 4
 $ MBCS    : logi FALSE
 $ UTF-8   : logi FALSE
 $ Latin-1 : logi TRUE
 $ codepage: int 1252
> system(paste("tasklist /FI \"PID eq", Sys.getpid(), "\""))

Image Name                     PID Session Name        Session#    Mem Usage
========================= ======== ================ =========== ============
rsession.exe                  6008 Console                    2    158,744 K
```

On newer versions of R (>= 4.2.0), you should see:

```
> str(l10n_info())
List of 5
 $ MBCS           : logi TRUE
 $ UTF-8          : logi TRUE
 $ Latin-1        : logi FALSE
 $ codepage       : int 65001
 $ system.codepage: int 65001
> system(paste("tasklist /FI \"PID eq", Sys.getpid(), "\""))

Image Name                     PID Session Name        Session#    Mem Usage
========================= ======== ================ =========== ============
rsession-utf8.exe            11980 Console                    2    130,096 K
```

### Checklist


dpprdan:
FYI `l10n_info()` should report a `system.codepage` starting from R v4.1.0. From its help page in R 4.1.2:

> `system.codepage`   integer: the Windows system/ANSI codepage (the codepage Windows is using). Added in R 4.1.0.

src/cpp/desktop/DesktopSessionLauncher.cpp
```diff
@@ -555,15 +557,73 @@ void SessionLauncher::onLaunchFirstSession()
    }
 }
 
+namespace {
+
+#ifdef _WIN32
+
+FilePath resolveSessionPathWin32(const FilePath& defaultPath)
```
gtritchie, line 564: We'll need to port this over to Electron as well (in `main`). Perhaps open an issue to track doing that?

gtritchie (approved):
LGTM!

src/cpp/desktop/DesktopSessionLauncher.cpp
kevinushey: Will do!

PR #10528, opened 2022-02-08
rotate news files

kevinushey:
For the PT branch.


jgutman:
Should we put these all in /version/news as suggested in Slack discussion?

kevinushey:
Yes, I think so. I'll make that change.

jgutman (approved):
Not sure if NEWS-2020.02.0-prairie-trillium.md should be in the subfolder too or that should wait for release -- I've moved it into that subfolder on `main`, but at least this way we won't end up with conflicts or unintentional overrides

kevinushey:
I think it should be in `version/news/` as well -- only `NEWS.md` should be present at the top level, and that should be collecting NEWS items for the development branch. I'll make that change.

PR #10530, opened 2022-02-08
NEWS for Spotted Wakerobin only

jgutman:
Fix the NEWS file, remove all Prairie Trillium issues.


kevinushey (approved):
LGTM!

timtmok:
I think the wiki states that `main` should have a new NEWS file and the release branch (PT) would have the old NEWS file renamed to the release name.

PR #10531, opened 2022-02-08
Don't use the global environment as the caller env for rs API functions

jgutman:
### Intent

Addresses #8031. 

We use the globalenv() as the caller env for all .rs.api functions, which is not true of other internal .rs non-API functions.
This makes the behavior of these functions brittle, as whether they work properly or not can depend on what's in the user's global environment, and whether or not it masks a base function we might be relying on.

### Approach

Create a new environment for evaluation of .rs.api functions. This environment should have base packages like `base` and `stats` on its search path, but should not inherit from the global environment.

### DIFF
--- a/src/cpp/r/R/Tools.R
+++ b/src/cpp/r/R/Tools.R
@@ -27,6 +27,15 @@ assign(".rs.toolsEnv", function()
    .rs.Env
 }, envir = .rs.Env)
 
+# target environment for symbol lookup, with all necessary base packages
+.rs.ApiEnv <- new.env(parent = as.environment("package:stats"))
+
+# allow access to api env from helper function
+assign(".rs.apiEnv", function()
+{
+   .rs.ApiEnv
+}, envir = .rs.ApiEnv)
+
 #' Add a function to the 'tools:rstudio' environment.
 #' 
 #' This environment is placed on the search path, and so is accessible and
@@ -62,7 +71,7 @@ assign(".rs.addFunction", function(name, FN, attrs = list(), envir = .rs.toolsEn
 # force helper function to also execute in tools environment
 environment(.rs.Env[[".rs.addFunction"]]) <- .rs.Env
 
-# add a global (non-scoped) variable to the tools:rstudio environment
+# add a glob    al (non-scoped) variable to the tools:rstudio environment
 .rs.addFunction("addGlobalVariable", function(name, var)
 {
    envir <- .rs.toolsEnv()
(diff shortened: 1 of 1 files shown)


jgutman:
Test 1. Run `package_version <- function(...) { stop("oops") }`,  and `.rs.api.versionInfo()`
No error, returns expected output.

Since we removed the hotfix, this should fail if we use the global environment as the evaluation env
image

Test 2. Run `rlang::env_parents(.rs.symbolLookupEnv(), last=emptyenv())` to see the full environment stack for the new env.
Confirm it does not contain the global env
image

Test 3. 
Run `dog <- "May"; library(dplyr)` to add some symbols to the global env. Then run `debug(.rs.api.versionInfo)` and `.rs.api.versionInfo()` to step into the function. Try to evaluate `dog`; the object is not found. Try to evaluate `filter`, it will come from `namespace:stats`. If you try the same on a current build, `dog` will be found, and `filter` will come from namespace `dplyr`.

jgutman:
Are there rstudioapi functions which rely on the ability to access and edit the global env? Do we need to have the global env on the search path, but with lower precedence? I'd like to understand why it was set to use the global env in the first place

romainfrancois (approved):
LGTM.

romainfrancois:
> Are there rstudioapi functions which rely on the ability to access and edit the global env? Do we need to have the global env on the search path, but with lower precedence? I'd like to understand why it was set to use the global env in the first place

I don't think so, because at the end of the day, they all end up calling their `.rs.api.*` function, but I don't know enough about it to be sure.

jgutman:
> 

Yes but the .rs.api.* functions have, up till now, been evaluated in the global environment, which is the problem we're solving here -- I'm just not sure if that possibly breaks anything else

src/cpp/r/R/Tools.R
```diff
@@ -27,6 +27,12 @@ assign(".rs.toolsEnv", function()
    .rs.Env
 }, envir = .rs.Env)
 
+# target environment for symbol lookup, with all necessary base packages
+assign(".rs.symbolLookupEnv", function()
+{
+   new.env(parent = .rs.toolsEnv())
```
jgutman, line 33: Not clear to me whether this should inherit from .rs.toolsEnv or just the base packages. I don't think it really needs .rs.toolsEnv, but I'm not sure if there's harm in including it either. 

Also there may be a better name for this environment. `.rs.apiEnv` was another consideration, but I thought it might be confusing to have it start with the `.rs.api` prefix

kevinushey:
Regarding the worries about evaluation no longer being in the global environment, I think we're early enough in SW that we can take this PR now and if any issues were to pop up, we have ample time to catch them. That said, I don't think any API functions depend on things being available in the global environment, except perhaps functions in the `utils` / `tools` namespaces.

PR #10532, opened 2022-02-08
Check if load from base is masked before calling load workspace

jgutman:
### Intent

Addresses #10089

Check if `base::load` is masked. If yes, call `base::load` in console. If not masked, call `load` in console


kevinushey (approved):
LGTM!

PR #10533, opened 2022-02-08
Fix duplicate menu separators

### LINKED ISSUE #9948: Menu display visual issue on RStudio Desktop + Ubuntu Electron
### System details

```shell
    RStudio Edition : Desktop
    RStudio Version : "Prairie Trillium" Daily (99999999, 2021-10-05)
    OS Version      : Ubuntu Hirsute 21.04
    R Version       : 4.1.1
```

Hover over any menu in the top-bar. "Build" is a good example, assuming that you do not have build tools set up, because every option should be disabled except for "Configure Build Tools..." 

image

As seen above, some menus in the Electron version of RStudio Desktop on Ubuntu display separators for empty sections. This does not occur on Windows or Mac.

image

Above is an example of a non-buildable project on a Mac, which displays the menu properly without displaying several separators first.

- [x] ~If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).~

timtmok:
### Intent
Addresses #9948 

Change the eslint config to address the errors about prettier at the top of every file.

### Approach
This uses similar logic to how Electron handles menu separators. The difference is that Electron only removes separators if they are at the beginning/end or the previous item is also a separator. Our menus have many hidden items so check doesn't work for us. Even not adding the hidden items still means the separators have to be marked as hidden.

I've added in a check to not add hidden items to menus. The caveat is shortcuts won't work for that command. Although, I don't see that as an issue since the item is likely hidden for a reason. The second part of the fix is using the same logic as Electron to filter out unnecessary separators.

This also runs on MacOS but has no difference since it already hides unnecessary separators. I think it's fine to leave in just in case that ever changes and to maintain the same code across all platforms.

### Automated Tests
Added unit tests:
* check that hidden items also remove any trailing separators
* check that separators at beginning/end are removed
* check that multiple separators in a row are condensed to one separator

Unit tests pass on all platforms

### QA Notes
The Build menu was most noticeable with this bug. Using Configure Build Tools to change the type will update the Build menu.

### Checklist

### DIFF
--- a/src/node/desktop/.eslintrc.js
+++ b/src/node/desktop/.eslintrc.js
@@ -14,9 +14,9 @@ module.exports = {
     sourceType: 'module',
   },
 
-  plugins: ['@typescript-eslint'],
+  plugins: ['@typescript-eslint', 'prettier'],
 
-  extends: ['eslint:recommended', 'plugin:@typescript-eslint/recommended', 'prettier'],
+  extends: ['eslint:recommended', 'plugin:@typescript-eslint/recommended'],
 
   rules: {
     'no-empty-function': 1, // will only show a warning
--- a/src/node/desktop/src/main/menu-callback.ts
+++ b/src/node/desktop/src/main/menu-callback.ts
@@ -257,7 +257,7 @@ export class MenuCallback extends EventEmitter {
         const itemTemplate = this.menuItemTemplates.get(item);
         if (!itemTemplate) {
           // no itemTemplate found (separators, for example)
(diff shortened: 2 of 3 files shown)


gtritchie (approved):
LGTM! Please modify the PR comment to reword "fixes ####" to "addresses ####" before merging so the issue isn't auto-closed by github.

timtmok:
Good to know about the fixes/addresses

PR #10535, opened 2022-02-08
ensure desktop contains server sources

kevinushey:
Speculative fix for https://github.com/rstudio/rstudio-pro/issues/3248.


cm421 (approved):
LGTM!

cm421:
Not sure if this will completely fix the issue, but it definitely makes a step in the right direction testing on my machine

Issue #10107, opened 2021-11-17
Implement versioning for Electron Mac package builds


gtritchie:
RStudio Electron for Mac built with `make-package --rstudio-target=Electron` needs version information to be embedded in the Electron application and in the produced package name. Currently it is hardcoded.

The version information is passed to the script with four variables:

```
RSTUDIO_VERSION_MAJOR
RSTUDIO_VERSION_MINOR
RSTUDIO_VERSION_PATCH
RSTUDIO_VERSION_SUFFIX
```

For testing, you can run make-package via something like:

```
RSTUDIO_VERSION_MAJOR=1 RSTUDIO_VERSION_MINOR=7 RSTUDIO_VERSION_PATCH=1 RSTUDIO_VERSION_SUFFIX=-daily+123 ./make-package --electron
```

The version information should show up in:

- the name of the produced package (*.dmg); should follow same naming convention as existing builds
- in the Help / About dialog inside RStudio
- in the Electron About dialog (from the App menu)

More on versioning scheme: https://github.com/rstudio/rstudio-pro/wiki/RStudio-Build-Versioning

gtritchie:
This was regressed with the work to standup real Mac builds. The Help About dialog is still correct:

![Screen Shot 2022-02-08 at 11 13 11](https://user-images.githubusercontent.com/10569626/153050814-15df6fc5-b89b-4994-8853-33de8a3dd839.png)

But the Electron About dialog (on Mac, RStudio / About) is back to the hardcoded values:

![Screen Shot 2022-02-08 at 11 10 05](https://user-images.githubusercontent.com/10569626/153050891-f7aecf78-2875-4e5a-b912-cba5ba8f19b0.png)

gtritchie:
Confirmed we are correctly applying version info to builds.

PR #10536, opened 2022-02-09
stamp version into Electron app (macOS)

gtritchie:
### Intent

Issue #10107 was partially regressed by the work to enable Mac builds. The Electron-part of the program was not being stamped with the build version but with hardcoded 99.9.9 version.

### Approach

Put back (and slightly rework) setting the build version in package.json when doing an Electron build. I also restored package.json back to initial value at end of the build, just to reduce hassle when testing make-package builds locally.

Originally this used a shell script, but since that isn't going to work on Windows, I'm just invoking the command directly from `make-package` and will eventually do the same for Windows and Linux.

### Automated Tests

None

### QA Notes

Check that the version shown in the RStudio / About RStudio dialog (this is an Electron-supplied dialog, different from our Help / About RStudio dialog) isn't hardcoded to 99.9.9.

### Checklist


matheustavaresdev (approved):
lgtm

Issue #10447, opened 2022-01-26
Implement navigation toolbar for Electron secondary windows


gtritchie:
RStudio desktop (Qt) has a "secondary window" implementation that can show a toolbar with basic web-browser navigation controls (back/forward/refresh). This is NYI in Electron Desktop.

To see an example, load a help topic in the Help pane (`?cat` in console), the click the "Show in New Window" button on the help pane toolbar. Here's what Qt desktop shows:

![Screen Shot 2022-01-26 at 14 29 58](https://user-images.githubusercontent.com/10569626/151250180-c0146c96-4c63-4da0-bede-4de0ac576484.png)

Electron currently shows without this toolbar.

The stubbed code for this in Electron is here [desktop-browser-window.ts](https://github.com/rstudio/rstudio/blob/50602bb32574b0922d52677376b60912fa3228ca/src/node/desktop/src/main/desktop-browser-window.ts#L200-L203), and corresponds to this C++ code: [DesktopBrowserWindow.cpp](https://github.com/rstudio/rstudio/blob/50602bb32574b0922d52677376b60912fa3228ca/src/cpp/desktop/DesktopBrowserWindow.cpp#L175-L179).

The actual population of the toolbar happens here in the C++ code: [DesktopSecondaryWindow.cpp](https://github.com/rstudio/rstudio/blob/50602bb32574b0922d52677376b60912fa3228ca/src/cpp/desktop/DesktopSecondaryWindow.cpp#L54-L67). The corresponding Electron class is: [secondary-window.ts](https://github.com/rstudio/rstudio/blob/spotted-wakerobin/src/node/desktop/src/main/secondary-window.ts).

gtritchie:
Verified these are present and working.

PR #10541, opened 2022-02-09
Electron Toolbar Support

matheustavaresdev:
### Intent

Added toolbar support for secondary windows on Electron

### Approach

Created a HTML toolbar that will have support on all platforms with a consistent look.

### Automated Tests

None

### QA Notes

Help can be opened on an external window to test for this toolbar.

### Checklist

Addresses #10447


src/node/desktop/src/main/toolbar-manager.ts
```diff
@@ -0,0 +1,160 @@
...
+      ';' +
+      this.addButtonsJsAsText(toolbarData);
+
+    try {
+      await window.webContents.executeJavaScript(jsScript);
+
+      // eslint-disable-next-line @typescript-eslint/no-implicit-any-catch
+    } catch (err: any) {
+      if (err.message !== 'An object could not be cloned.') {
+        console.log('Error message: ', err.message);
```
gtritchie, line 47: Can we use the logging api here, instead of `console.log`? Look for examples of the form:

`logger().log*`

src/node/desktop/src/assets/img/test.html
```diff
@@ -0,0 +1,169 @@
+<!DOCTYPE html>
```
gtritchie, line 1: Is this file needed?

src/node/desktop/src/main/desktop-browser-window.ts
```diff
@@ -13,7 +13,8 @@
  *
  */
 
-import { BrowserWindow, shell, WebContents } from 'electron';
+import { BrowserWindow, nativeImage, shell, WebContents } from 'electron';
+// import { nativeImage } from 'electron/common';
```
gtritchie, line 17: Delete if not needed

src/node/desktop/src/main/toolbar-manager.ts
```diff
@@ -0,0 +1,160 @@
...
+    return `
+      const styleSheet = document.createElement('style');
+      styleSheet.innerText = \`${styles}\`;
+      document.head.appendChild(styleSheet);
+      `;
+  }
+
+  addButtonsJsAsText(toolbarData: ToolbarData) {
+    let addButtonsFn = `
+      const createButton = (iconPath, tooltip, onClick) => {
```
gtritchie, line 126: The buttons need to be annotated so they work correctly with screen readers, and be usable via the keyboard.

Strongly recommend using semantic html and making the outer "div" be a `<button>`, instead. That should give most of the necessary behavior automatically.

* each button needs to be reachable via the Tab key (tabindex="0")
* when given focus via the keyboard, a focus rectangle must be displayed (we use the `focus-visible` polyfill to accomplish this; we don't want focus rectangle shown if you just click on the button)
* when a button has focus you can activate it via keyboard (spacebar and enter key)
* each button must have descriptive text for screen readers (the alt text on the image might be sufficient, I would need to test once the rest of the work is done)

gtritchie: More on buttons:

* https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role
* https://www.w3.org/TR/wai-aria-practices/#button

src/node/desktop/src/main/toolbar-manager.ts
```diff
@@ -0,0 +1,160 @@
...
+        console.log('---------------------------');
+        console.log('JS Script for Toolbar: ');
+        console.log(jsScript);
+      }
+    }
+  }
+
+  addToolbarJsAsText() {
+    const toolbar = ` 
+      const toolbar = document.createElement('div');
```
gtritchie, line 58: Toolbar needs to be annotated as such for accessibility.

At a minimum the outer div needs `role="toolbar"` and `aria-label="Navigation"`.

Officially a toolbar should support moving between buttons with arrow keys, but we don't support that elsewhere yet, so as long as you can move between buttons with the tab keys that is fine.

gtritchie: https://www.w3.org/TR/wai-aria-practices/#toolbar

gtritchie (changes requested):
Changes requested related to accessibility. Once this is close, I will build it myself and do some testing with a screen reader.

src/node/desktop/src/main/toolbar-manager.ts
matheustavaresdev: Done in https://github.com/rstudio/rstudio/pull/10541/commits/6deede4405d4f6cfc08f5c0d2759f7ae4a7d7c6e

src/node/desktop/src/main/toolbar-manager.ts
matheustavaresdev: Done in https://github.com/rstudio/rstudio/pull/10541/commits/6deede4405d4f6cfc08f5c0d2759f7ae4a7d7c6e

gtritchie (approved):
Nice, works very well with keyboard and screen reader now. Thanks!

PR #10568, opened 2022-02-10
Fix missing custom markers in gutter

timtmok:
### Intent
Addresses #10062 
Custom source markers wouldn't show in the editor's gutter.

### Approach
The `isDiagnostics` flag wasn't initialized for `SourceMarkerSet` so some instances worked but others did not. Setting an initial value fixed the issue. I've also set an initial value for `isCustom` in `SourceMarker` where I had added a flag to distinguish customer source markers so that on the GWT side it can choose the correct way to render the content.

### Automated Tests
None

### QA Notes
I was only able to reproduce using a packaged build and with x86 R.

* Clear any custom source marker sets
* Use RS API to add source markers
* Go to one of the files and allow the linter to run

### Checklist

### DIFF
--- a/src/cpp/session/include/session/SessionModuleContext.hpp
+++ b/src/cpp/session/include/session/SessionModuleContext.hpp
@@ -834,7 +834,8 @@ struct SourceMarker
    };
 
    SourceMarker()
-      : type(Empty)
+      : type(Empty),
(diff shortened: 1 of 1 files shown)


MariaSemple (approved):
LGTM!

PR #10597, opened 2022-02-14
adding axis label and title to Shiny template files

tracykteal:
### Intent

Previously the Shiny template file included x-axis label and a title. That isn't included in the current version, and it makes the template example harder to work with, because 1) it's difficult to understand what the plot is about, because what the labels should be isn't immediately obvious and 2) it's a standard thing that people want to do, so it's nice to have an example of how it's done in the code. 

### Approach

For the shiny/app.R, shiny/server.R and shiny.qmd files, I added in the code to add the x-axis label and title. 

### Automated Tests

I'm not sure how this works with automated tests. I did run locally and it worked fine. 

### QA Notes

Those labels may have been removed for reasons I'm not aware of, including accessibility standards. If this is the case, that rationale likely supersedes template legibility. 

### Checklist

For purposes of the CONTRIBUTING file information, I'm an RStudio employee.

### DIFF
--- a/src/cpp/session/resources/templates/shiny.qmd
+++ b/src/cpp/session/resources/templates/shiny.qmd
@@ -18,7 +18,7 @@ output$distPlot <- renderPlot({
    x <- faithful[, 2]  # Old Faithful Geyser data
    bins <- seq(min(x), max(x), length.out = input$bins + 1)
    hist(x, breaks = bins, col = 'darkgray', border = 'white',
-        xlab = "Waiting time to next eruption (in mins)",
-        main = "Histogram of waiting times")
+        xlab = 'Waiting time to next eruption (in mins)',
+        main = 'Histogram of waiting times')
 })
 ```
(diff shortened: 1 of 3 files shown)


src/cpp/session/resources/templates/shiny.qmd
```diff
@@ -17,6 +17,8 @@ plotOutput("distPlot")
 output$distPlot <- renderPlot({
    x <- faithful[, 2]  # Old Faithful Geyser data
    bins <- seq(min(x), max(x), length.out = input$bins + 1)
-   hist(x, breaks = bins, col = 'darkgray', border = 'white')
+   hist(x, breaks = bins, col = 'darkgray', border = 'white',
+        xlab = "Waiting time to next eruption (in mins)",
```
kevinushey, line 21: Nit: it looks like we're using single quotes above, and double quotes here. Can we change these to be consistent?

I think the tidyverse style guide recommends using double quotes, so we could also just change the other usages in these documents to be double quotes as well.

kevinushey (approved):
LGTM, thanks!

Issue #10485, opened 2022-02-01
Implement Electron cleanClipboard callback


gtritchie:
In the Electron IDE, invoke "Copy" (e.g. in the Console) via right-click menu, or keyboard shortcut, or Edit / Copy, and you'll get a dialog saying: `desktop_clean_clipboard callback NYI`.

This needs to be implemented here: https://github.com/rstudio/rstudio/blob/b2a98ae29ca3a7218fd4d4f8ba19fcc763a2fd23/src/node/desktop/src/main/gwt-callback.ts#L410-L412

The C++ implementation (Qt) for MacOS is here:
https://github.com/rstudio/rstudio/blob/b2a98ae29ca3a7218fd4d4f8ba19fcc763a2fd23/src/cpp/desktop/DesktopGwtCallback.cpp#L1041-L1108

For other platforms this callback is a no-op.

kevinushey:
QA Notes -- note that this PR affects macOS only.

- Select some text in the Console history (e.g. using the mouse, and then clicking and dragging to create a selection),
- Press Ctrl + C to copy the text to the clipboard,
- Open the Pasteboard Viewer application (https://apps.apple.com/us/app/pasteboard-viewer/id1499215709?mt=12), and confirm that only a single entry is present, with the copied text.

<img width="619" alt="Screen Shot 2022-02-22 at 2 43 31 PM" src="https://user-images.githubusercontent.com/1976582/155232044-8fff0aba-5221-4d2a-971b-9d00a0923151.png">

gtritchie:
Verified with build 2022.06.0-daily+244 on MacOS 12.3.

PR #10618, opened 2022-02-16
example using node-addon-api for native module

kevinushey:
Draft PR just demonstrating the process.


src/node/desktop/binding.gyp
```diff
@@ -0,0 +1,14 @@
+{
+  "targets": [
+    {
+      "target_name": "macos",
+      "cflags!": [ "-fno-exceptions" ],
+      "cflags_cc!": [ "-fno-exceptions" ],
+      "sources": [ "src/native/macos/hello.cc" ],
```
kevinushey, line 7: It would be worth discussing if this structure makes sense -- that is, we place our native bindings in `src/native`, with platform-specific sub-directories as appropriate.

kevinushey: https://github.com/node-ffi/node-ffi/blob/d9b843c03b35b99f205d001ebcbedd1b6c63b3d0/binding.gyp#L16-L21 for an example of platform-specific sources.

src/node/desktop/package.json
```diff
@@ -66,11 +66,14 @@
     "typescript": "4.5.2"
   },
   "dependencies": {
+    "@types/bindings": "^1.5.1",
```
gtritchie, line 69: @matheustavaresdev recommends we pin specific versions so we can decide when to update a package and try it out very carefully... So no "^" in front.

src/node/desktop/src/main/desktop-browser-window.ts
```diff
@@ -26,6 +26,10 @@ import { showContextMenu } from './context-menu';
 import { ToolbarData, ToolbarManager } from './toolbar-manager';
 import { executeJavaScript, isSafeHost } from './utils';
 
+// TODO: how do we make this play well with typescript import?
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const desktop = require('../../build/Release/desktop.node');
```
kevinushey, line 31: Kind of ugly, but it seems like thanks to Webpack this "Just Works".

src/node/desktop/binding.gyp
```diff
@@ -0,0 +1,26 @@
...
+            "cflags!": [ "-fno-exceptions" ],
+            "cflags_cc!": [ "-fno-exceptions" ],
+            "sources": [ "src/native/desktop.cc" ],
+            "include_dirs": [
+                "<!@(node -p \"require('node-addon-api').include\")"
+            ],
+            "defines": [ "NAPI_DISABLE_CPP_EXCEPTIONS" ]
+        },
+        {
+            "target_name": "copy",
```
kevinushey, line 14: This target copies the native node module back into the source tree, which makes it easier to reference in sources (and allows us to provide an accompanying `.d.ts` file for types)

kevinushey:
I'm going to try and handle https://github.com/rstudio/rstudio/issues/10485 as part of this PR just to have a real proof-of-concept that this works.

src/node/desktop/package.json
```diff
@@ -11,7 +11,7 @@
     "clean": "ts-node scripts/clean.ts",
     "lint": "eslint ./src ./test",
     "package": "yarn && electron-forge package",
-    "start": "electron-forge start",
+    "start": "yarn && electron-forge start",
```
kevinushey, line 14: This was necessary to ensure native modules are rebuilt when starting RStudio.

src/node/desktop/package.json
```diff
@@ -11,7 +11,7 @@
     "clean": "ts-node scripts/clean.ts",
     "lint": "eslint ./src ./test",
     "package": "npm ci && electron-forge package",
-    "start": "electron-forge start",
+    "start": "npm install && electron-forge start",
```
kevinushey, line 14: Necessary to ensure that native node modules are rebuilt.

gtritchie (approved):
LGTM!

PR #10620, opened 2022-02-17
stop using yarn for Electron

gtritchie:
### Intent

Getting `yarn` installed reliably on the Linux containers is proving troublesome. Rather than continuing that battle, switch to npm for package management. We already know how to find that.

### Approach

* "yarn" becomes "npm install"
* "yarn foo" (where "foo" is a script in package.json) becomes "npm run foo"
* "yarn bar" (where "bar" is a command installed in node_modules) becomes "npx bar"

### Automated Tests

Build stuff

### QA Notes

Build stuff

### Checklist


package/osx/cmake/merge-electron.cmake
```diff
@@ -18,10 +18,11 @@ function(echo MESSAGE)
    execute_process(COMMAND echo "-- ${MESSAGE}")
 endfunction()
 
-# run yarn to ensure required packages are installed
+# run npm to ensure required packages are installed
 execute_process(
 	COMMAND
-		"@YARN@"
+		"@NPM@"
+		"install"
```
GCRev, line 25: Consider using `npm ci` instead of `npm install` in scripts. The `package-lock.json` file should be committed as part of the git history, so that `npm ci` can be used to guarantee that the packages installed are 100% the same across machines. The `ci` command means "clean-install," and only reads out of the `package-lock.json` file, which is faster and safer than performing dependency resolution with `npm install`. The `install` command is better for adding packages or updating minor versions of existing packages. This also prevents against a constantly-changing `package-lock.json` file with each build.

gtritchie: Yup! Matheus gave same feedback on Slack (some threads about it in Electron channel).

Issue #9881, opened 2021-09-23
"Find in Files" does not find files if non-ASCII characters are contained in the path name


aryoda:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.4.1717
    OS Version      : Windows NT 10.0; Win64; x64
    R Version       : 4.1.0 (2021-05-18) on Platform x86_64-w64-mingw32/x64 (64-bit) Running under Windows 10x64 (build 19042)

### Steps to reproduce the problem

Open RStudio and a new R file. Put this code into it but don't save it:

```
folder <- "~/search äöü test/subfolder" 
dir.create(folder)
setwd(folder)
x <- "hello world"
write.csv(x, "hello.txt")
dir()
```

Execute the code to create the folders and file.

Now use "Edit > Find in Files", change the "Search in" path to the above folder and search for "hello".

On Windows no file will be found even though "hello.txt" contains the search text!

### Describe the problem in detail

If the 2nd or higher parent folder contains a non-ASCII character like a german umlaut or french accent
then "Find in Files" does not find a text file with a matching content (if the file is contained at least one folder
deeper than the path component containing a non-ASCII character!).

The file will not be shown in the result list tab "Find in Files" on Windows OS.

On Linux it works!

Saving the R code with different encodings and recreate the folder structure does not help.

I am using a NTFS formatted file system.

### Describe the behavior you expected

The file "hello.txt" must be found if I search in "Find in Files"

- [ x ] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ x ] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [ x ] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
@aryoda Thank you for raising this! We unfortunately run into variations of path and encoding bugs, which we've been trying to fix. I'll add this to the list. Thanks!

jgutman:
Note: This still succeeds on Mac and fails on Windows using R 4.0.0, 4.1.2, and 4.2.0 using RStudio Desktop 2022.06.0+70

jgutman:
[...]

PR #10627, opened 2022-02-17
handle UTF-8 paths with grep in Find in Files

kevinushey:
Addresses https://github.com/rstudio/rstudio/issues/9881.

The general strategy here -- instead of passing the paths directly to `grep` (which might have unicode characters, and so may be mis-interpreted either by `cmd.exe` or `grep.exe` itself on Windows), just switch the working directory to the requested path, and then use `.`to reference the current working directory (thereby dodging extra encoding issues).

As part of this PR, we'll also be switching to the Rtools-bundled version of grep (3.0), since we've discovered various issues with the other more recent builds of grep on Windows.

### DIFF
--- a/src/cpp/core/r_util/RTokenizer.cpp
+++ b/src/cpp/core/r_util/RTokenizer.cpp
@@ -560,7 +560,7 @@ wchar_t RTokenizer::peek(std::size_t lookahead)
    // iterators here we need to make sure we don't construct an iterator
    // that points outside of the string from which it was derived
    // this is a safer way of checking pos_ + lookahead >= data_.end()
-   if (UNLIKELY(data_.end() - pos_ < lookahead))
+   if (UNLIKELY(data_.end() - pos_ <= lookahead))
       return 0;
 
    return *(pos_ + lookahead);


src/cpp/session/modules/SessionFind.cpp
```diff
@@ -68,6 +68,22 @@ const size_t MAX_COUNT = 1000;
...
 
+bool debugging()
+{
+   return !core::system::getenv("RSTUDIO_GREP_DEBUG").empty();
+}
+
+FilePath gnuGrepPath()
+{
+   // allow override, for testing
+   std::string overridePath = core::system::getenv("RSTUDIO_GREP_PATH");
```
jgutman, line 79: Neat idea

jgutman (commented):
Cool let's try it

jgutman:
LGTM pending the additional changes for git grep

jgutman (approved):

kevinushey:
I'm not sure if this is a change in the behavior of git grep, but the test strings don't quite match what I see output from `git grep`. I've tweaked the regular expressions to allow some color escapes to be missing, and that seems to accommodate the tests + manual verification.

PR #10630, opened 2022-02-17
extra precautions to not get out of index error in R tokenizer

jgutman:
Solves the painful iterator indexer error on startup (for debug builds)
Screen Shot 2022-02-15 at 12 42 07 PM

### DIFF
--- a/src/cpp/core/r_util/RTokenizer.cpp
+++ b/src/cpp/core/r_util/RTokenizer.cpp
@@ -556,7 +556,7 @@ wchar_t RTokenizer::peek()
 
 wchar_t RTokenizer::peek(std::size_t lookahead)
 {
-   if ((pos_ + lookahead) >= data_.end())
+   if (lookahead < data_.size() && pos_ >= data_.end() - lookahead)
       return 0;
    else
       return *(pos_ + lookahead);


src/cpp/core/r_util/RTokenizer.cpp
```diff
@@ -556,7 +556,7 @@ wchar_t RTokenizer::peek()
 
 wchar_t RTokenizer::peek(std::size_t lookahead)
 {
-   if ((pos_ + lookahead) >= data_.end())
+   if (lookahead < data_.size() && pos_ >= data_.end() - lookahead)
```
jgutman, line 559:
```suggestion
   if ((lookahead < data_.size()) && (pos_ >= data_.end() - lookahead))
```

jgutman: Double checking here -- should this be 
`(lookahead < data_.size()) || (pos_ >= data_.end() - lookahead)`

src/cpp/core/r_util/RTokenizer.cpp
```diff
@@ -556,7 +556,7 @@ wchar_t RTokenizer::peek()
 
 wchar_t RTokenizer::peek(std::size_t lookahead)
 {
-   if ((pos_ + lookahead) >= data_.end())
+   if (lookahead < data_.size() && pos_ >= (data_.end() - lookahead))
```
jgutman, line 559: Wouldn't we also want to return 0 if `lookahead >= data_.size()` ?

jgutman:
Superseded by #10627

PR #10646, opened 2022-02-22
Electron Packages Update

matheustavaresdev:
### Intent

Fix an issue that started to happen after the new updates for Electron: Node v14 to v16 and to use `npm` instead of `yarn`.

When removing the `package-lock.json`, an error was presented when installing `eslint` due to an incompatibility with `eslint-config-standard@16.0.3`.

### Approach

Downgraded the `eslint` package and regenerated the `package-lock.json` file. As ESLint was worked with no issues on v8 with `yarn`, I decided to keep the notation to use the latest v7 for `eslint` (^7.12.1 notation).

### Automated Tests

None

### QA Notes

None

### Checklist

### DIFF
--- a/src/node/desktop/package.json
+++ b/src/node/desktop/package.json
@@ -44,7 +44,7 @@
     "css-loader": "6.5.1",
     "electron": "16.0.7",
     "electron-mocha": "10.1.0",
-    "eslint": "8.2.0",
+    "eslint": "^7.12.1",
     "eslint-config-prettier": "8.3.0",
     "eslint-config-standard": "16.0.3",
     "eslint-plugin-import": "2.25.3",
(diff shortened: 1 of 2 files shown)


gtritchie (approved):

src/node/desktop/package.json
```diff
@@ -44,7 +44,7 @@
     "css-loader": "6.5.1",
     "electron": "16.0.7",
     "electron-mocha": "10.1.0",
-    "eslint": "8.2.0",
+    "eslint": "^7.12.1",
```
timtmok, line 47: Should using `^` be standard for all other new packages or is it just this case because all the minor updates are expected to be stable?

kevinushey (approved):

src/node/desktop/package.json
matheustavaresdev: Yes! All minor updates for that one should be stable as there is a Major version already released!

PR #10647, opened 2022-02-22
console pane should refresh logo/version on session restart

jgutman:
Addresses #10458 and #10579 

Whenever the R session restarts, we should update the label in the console to reflect 1) whether we are in a python or R session (R session restarts should always trigger a new R session, even if we were in python mode before restarting) and the current R version (major, minor, patch), which may have changed during the session restart.


src/gwt/src/org/rstudio/studio/client/workbench/views/console/ConsoleInterpreterVersion.java
```diff
@@ -169,7 +169,7 @@ else if (StringUtil.equals(type, ReticulateEvent.TYPE_REPL_TEARDOWN))
    }
    
    
-   private String rVersionLabel()
+   public String rVersionLabel()
```
jgutman, line 172: this will need to be changed back -- just made public to help debugging

src/gwt/src/org/rstudio/studio/client/workbench/views/console/ConsolePane.java
```diff
@@ -387,12 +405,13 @@ private void initJobToolbar()
...
    private final Commands commands_;
    private Shell shell_;
+   private ApplicationServerOperations server_;
    private Session session_;
    private AriaLiveService ariaLive_;
    private Label workingDir_;
    private ToolbarButton consoleInterruptButton_;
    private ToolbarButton consoleClearButton_;
-   private ConsoleInterpreterVersion consoleInterpreterVersion_;
+   public ConsoleInterpreterVersion consoleInterpreterVersion_;
```
jgutman, line 414: also made public just for debugging

src/cpp/session/modules/SessionAbout.cpp
```diff
@@ -21,8 +21,6 @@
 
 #include <session/SessionModuleContext.hpp>
 
-#include <string>
```
jgutman, line 24: unused import

src/gwt/src/org/rstudio/studio/client/workbench/views/console/ConsoleInterpreterVersion.java
```diff
@@ -167,24 +174,23 @@ else if (StringUtil.equals(type, ReticulateEvent.TYPE_REPL_TEARDOWN))
...
-      return "R " + version;
+         @Override
+         public void onResponseReceived(RVersionSpec versionSpec)
+         {
+            label_.setText("R " + versionSpec.getVersion());
+         }
+         @Override
+         public void onError(ServerError error)
+         {
+            Debug.logError(error);
```
gtritchie, line 191: Although unlikely to actually get an error here, maybe should set something in the label if it does? Otherwise could end up stuck showing an out-of-date label and we'd not know it was happening. Thinking leave the logError and also add `label_.setText("Error fetching R version") or something.

gtritchie (changes requested):
Looks good! A couple minor suggestions.

gtritchie (approved):
LGTM!

PR #10651, opened 2022-02-22
Implement undo/redo menu actions in Electron

timtmok:
### Intent
The undo/redo menu actions were being handled by Electron. Electron can perform these actions for input elements but not on the ACE editor. This allows editor actions to be undone/redone using the menu.

### Approach
The menu actions are implemented without the `undo` and `redo` roles. Instead, they emit the event for GWT to handle. If GWT detects Electron then it handles the action differently. For focus on an ACE editor, it will invoke the ACE undo/redo. For everything else, it calls back to Electron where it will use `webContents.undo()`.

### Automated Tests
None

### QA Notes
Technically, the menu and keyboard shortcut should be the same now that the shortcut is tied directly to the menu item.

The actions are the same in the editor and the console as both use ACE. Everywhere else should be handled by Electron calling `webContents.undo()`/`redo()`.

All platforms behaved slightly differently before but this change should make the actions work the same now.

### Checklist

### DIFF
--- a/src/node/desktop/src/main/menu-callback.ts
+++ b/src/node/desktop/src/main/menu-callback.ts
@@ -361,9 +361,9 @@ export class MenuCallback extends EventEmitter {
     } else if (cmdId === 'pasteWithIndentDummy') {
       menuItemOpts.role = 'pasteAndMatchStyle';
     } else if (cmdId === 'undoDummy') {
-      menuItemOpts.role = 'undo';
+      menuItemOpts.accelerator = 'CommandOrControl+Z';
     } else if (cmdId === 'redoDummy') {
-      menuItemOpts.role = 'redo';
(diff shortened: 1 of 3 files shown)


src/node/desktop/src/main/menu-callback.ts
```diff
@@ -361,9 +361,9 @@ export class MenuCallback extends EventEmitter {
     } else if (cmdId === 'pasteWithIndentDummy') {
       menuItemOpts.role = 'pasteAndMatchStyle';
     } else if (cmdId === 'undoDummy') {
-      menuItemOpts.role = 'undo';
```
kevinushey, line 364: Do you know why this change is necessary?

kevinushey (approved):
LGTM!

PR #10658, opened 2022-02-23
No-op for desktop_clean_clipboard event

timtmok:
### Intent
Handle the `desktop_clean_clipboard` event when copying from the IDE.

### Approach
The C++ code would check the clipboard for MacOS. It converts UTF-16 to UTF-8 and does some HTML stripping. It didn't appear to be necessary after trying to copy sample UTF-16 text and HTML from editors and console output.

But I do wonder if the HTML processing is required. MacOS appeared to behave the same as Windows and Linux.

### Automated Tests
None
Something might be possible with Playwright using keyboard shortcuts

### QA Notes
I used this text snippets to copy into and out of editors
```
ʕノ•ᴥ•ʔノ┬─┬
```
```html
<!DOCTYPE html>
<html>
<body>

<h1>My First Heading</h1>

<p>My first paragraph.</p>

</body>
</html>
```

### Checklist

### DIFF
--- a/src/node/desktop/src/main/gwt-callback.ts
+++ b/src/node/desktop/src/main/gwt-callback.ts
@@ -58,7 +58,7 @@ export class GwtCallback extends EventEmitter {
     super();
     this.owners.add(mainWindow);
 
-    ipcMain.on('desktop_browse_url', (event, url: string) => {
+    ipcMain.on('desktop_browse_url', (_event, url: string) => {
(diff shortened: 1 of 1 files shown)


timtmok:
Closing since Kevin's change will be the same behaviour as with Qt. This was only some cleanup that I wanted to do since I noticed the event triggering on Linux when working on the copy menu action.

Issue #10607, opened 2022-02-15
Editor windows not loading (Electron)


gtritchie:
### System details

    RStudio Edition : Electron Desktop
    RStudio Version : 2022.06.0-daily+109 "Spotted Wakerobin" 
    OS Version      : macOS 12.12.1 (also observed on Windows-10 with Electron build)
    R Version       : 4.1.2

### Steps to reproduce the problem

- start Electron desktop
- File / New File / R Script
- Click on arrow in editor to move to a separate window

### Describe the problem in detail

- window displays but never finishes loading (spinner)

![Screen Shot 2022-02-15 at 10 15 11](https://user-images.githubusercontent.com/10569626/154114771-811a4ff3-73cb-4b15-93bd-b599126b1c86.png)

### Describe the behavior you expected

- editor loads

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

gtritchie:
Upgrading Electron from v15 to v16 broke this. Working on the fix.

Related to https://github.com/rstudio/rstudio/pull/9737

kevinushey:
Verified fixed on Ubuntu 20.04 Build 242, as well as recent daily builds on macOS.

Issue #10606, opened 2022-02-15
git diff (Review Changes) window fails to load


kevinushey:
For Electron. It looks like it's trying to query `window.desktopInfo`, but that object isn't populated in windows other than the main window.

gtritchie:
This was working in early January 2022 (at 60e4de7c0c). Working to identify the source of the regression.

gtritchie:
Upgrading Electron from v15 to v16 broke this. Working on the fix.

Related to https://github.com/rstudio/rstudio/pull/9737

gtritchie:
Note to self: https://www.electronjs.org/docs/latest/api/window-open/

kevinushey:
Verified fixed on Ubuntu 20.04 with RStudio 2022.06.0 build 242.

PR #10663, opened 2022-02-23
Fixes for secondary/satellite window creation (Electron)

gtritchie:
### Intent

Upgrade from Electron 15 to 16 broke creation of secondary windows, such as the git window, and editor windows.

Addresses: #10606 #10607

### Approach

The code was relying on undocumented behaviour for how preloads were invoked on windows created via window.open, and this broke in newer Electrons.

Now, when window.open happens, directly supply the path to the preload, and which callbacks to include (desktop, desktopInfo), instead of relying on it just magically happening as it used to.

I upgraded us to Electron 17.0.1 as part of this PR. We plan to keep on current Electron builds until we have our first public previews, then we'll stay on the same major version and only take minor point releases.

### Automated Tests

None

### QA Notes

Test display and basic functionality of the git window, and of editor windows, in Electron desktop.

### Checklist


src/node/desktop/src/main/secondary-window.ts
```diff
@@ -34,4 +34,25 @@ export class SecondaryWindow extends DesktopBrowserWindow {
...
     this.window.close();
   }
+
+  /**
+   * 
+   * @returns Window creation request response
+   */
+  static windowOpening(width: number, height: number):
+    | { action: 'deny' }
+    | { action: 'allow'; overrideBrowserWindowOptions?: Electron.BrowserWindowConstructorOptions | undefined } {
```
kevinushey, line 44: 🤢

kevinushey (approved):
LGTM!

PR #10665, opened 2022-02-23
unset MAKEFLAGS when running npm install

kevinushey:
`node-gyp` barfs if you try to build its targets in parallel.


gtritchie (approved):
LGTM!

Issue #10610, opened 2022-02-15
Secondary windows showing menubar (Electron)


gtritchie:
### System details

    RStudio Edition : Electron Desktop
    RStudio Version : 2022.06.0-daily-111
    OS Version      : Windows-10 or Linux
    R Version       : Any

### Steps to reproduce the problem

- In Electron on Linux or Windows, bring up a secondary window in the IDE, such as an editor window or showing Help in another window

### Describe the problem in detail

- The secondary window gets the menubar just like the main window

### Describe the behavior you expected

- No menubar on anything but the main window

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

gtritchie:
Upgrading Electron from v15 to v16 broke this. Working on the fix.

Related to https://github.com/rstudio/rstudio/pull/9737

kevinushey:
Verified on Ubuntu 20.04 with RStudio 2022.06.0 Build 243.

<img width="667" alt="Screen Shot 2022-03-28 at 4 39 49 PM" src="https://user-images.githubusercontent.com/1976582/160504683-91688dab-9dc8-4091-817b-7e475ccc2b11.png">

PR #10668, opened 2022-02-24
Hide menubar on secondary windows

gtritchie:
### Intent

Addresses #10610

Secondary windows such as popped-out-editors or the git window were showing the menubar on Windows and Linux, which they didn't do on the Qt desktop.

### Approach

We load secondary windows in the same renderer process as the main window. This is for efficiency and also because we rely on the pattern `window.opener.doSomething()` and that can't work if each window is in its own renderer process.

As a result each window shares the same application menu. Using the autohide feature to hide the menu by default on secondary windows. It can still be displayed by tapping the Alt key.

This problem of having the menu on secondary windows is also true on Mac QtWebEngine desktop, so this actually isn't new. Just more noticable, perhaps, on Windows/Linux where menus are attached directly to the window instead of at the top.

On all three platforms the menu triggers actions in the main window (typically), even when it doesn't currently have focus. Not always great, but this isn't a new problem.

### Automated Tests

None

### QA Notes

Verify that secondary windows such as source control, editors, help content, zoomed plots, etc. don't show a menu by default on Windows and Linux.

### Checklist

### DIFF
--- a/src/node/desktop/test/unit/main/window-tracker.test.ts
+++ b/src/node/desktop/test/unit/main/window-tracker.test.ts
@@ -42,7 +42,7 @@ describe('window-tracker', () => {
   });
   it('tracks and returns a window by name', () => {
     const tracker = new WindowTracker();
-    const oneWin = new DesktopBrowserWindow(false, false, 'some name');
+    const oneWin = new DesktopBrowserWindow(false, false, false, 'some name');
     tracker.addWindow('one', oneWin);
     assert.equal(tracker.length(), 1);
     const result = tracker.getWindow('one');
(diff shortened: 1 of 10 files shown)


src/node/desktop/test/unit/main/window-tracker.test.ts
```diff
@@ -64,17 +64,17 @@ describe('window-tracker', () => {
...
+    const twoWin = new TestGwtWindow(false, false, false, 'the gwt window');
     tracker.addWindow('one', twoWin);
     assert.equal(tracker.length(), 1);
     const result = tracker.getWindow('one');
     assert.deepEqual(result, twoWin);
   });
   it('delete window removes it from map', async function () {
     const tracker = new WindowTracker();
-    const oneWin = new DesktopBrowserWindow(false, false, 'some name');
+    const oneWin = new DesktopBrowserWindow(false, false, false, 'some name');
```
kevinushey, line 77: Not for this PR, but it might be nice to define an interface for creating these objects with named parameters, so one could write:

```
new DesktopBrowserWindow({
  param1: false,
  param2: false,
});
```

Right now, it's hard to tell at a glance what the parameters actually mean or do in their usages.

kevinushey (approved):
LGTM!

src/node/desktop/test/unit/main/window-tracker.test.ts
kevinushey: (discussed a bit in https://www.typescriptlang.org/docs/handbook/2/functions.html#parameter-destructuring)

gtritchie: Agree, in C++ I usually comment calls like that as a cheap workaround: `(false/*foo*/, true/*bar*/) but better to do like you said.

Issue #10679, opened 2022-02-25
RStudio's list.files(pattern = ) is not compatible with the original list.files()


yutannihilation:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2022.06.0 Build 139
    OS Version      :  Windows 11
    R Version       :  R-devel (r81781) / R 4.1.2

### Steps to reproduce the problem

```
d <- tempfile()
dir.create(d)
file.create(file.path(d, "foo.csv"))
list.files(d, pattern = "*.csv")
```

### Describe the problem in detail

#10481 implemented the RStudio's original version of `list.files()`. But, it seems the behavior of the regex are different. 

On bare R session:

``` r
list.files(d, pattern = "*.csv")
#> [1] "foo.csv"
```

But, on RStudio:

``` r
list.files(d, pattern = "*.csv")
#> NULL
```

This particular pattern "*.csv" comes from the confusion with glob pattern, but the behavior itself is consistent with `egrep`. 

``` console
$ echo 'foo.csv' | grep -E '*.csv'
foo.csv
```

It seems the rule is that null character is repeated when `*` (or other repetition like `+` and `{n}`) is placed at the head of the pattern, though I couldn't find any source about the specification.

However, unfortunately, Boost implements it differently.

> It is an error to use a repeat operator, if the preceding construct can not be repeated, for example:
> 
>     a(*)
> 
> Will raise an error, as there is nothing for the `*` operator to be applied to.

So, I have no idea what we should do... Honestly, the patterns like `*.csv` should fail to let the users notice that it's NOT the glob pattern. On the other hand, it's annoying that `list.files()` behaves differently on RStudio and on other places.

### Describe the behavior you expected

(See above)

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
[...]

PR #10682, opened 2022-02-25
fixups for list.files on Windows

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/10679.
Addresses https://github.com/rstudio/rstudio/issues/10672.

### Approach

Use R's own `grepl()` rather than `boost::regex`, to avoid incompatibilities between the two regular expression engines.

Also make sure we expand tilde-prefixed paths before attempting to list files. 😭 

### Automated Tests

Tests included and updated.

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/10679.

### Checklist

### DIFF
--- a/src/cpp/session/modules/SessionFiles.cpp
+++ b/src/cpp/session/modules/SessionFiles.cpp
@@ -1611,6 +1611,16 @@ std::vector<boost::filesystem::path> initializePaths(SEXP pathSEXP)
          continue;
 
       const char* utf8Path = Rf_translateCharUTF8(charSEXP);
+
+      // ensure paths are expanded -- note that R does not
+      // preserve the tilde prefix in e.g.
+      //
+      //    list.files("~/hello", full.names = TRUE)
+      //
+      // rather, the expanded path is used when filling
+      // the names of listed files
+      utf8Path = R_ExpandFileName(utf8Path);
+
 #ifdef BOOST_WINDOWS_API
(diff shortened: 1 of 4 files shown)


src/cpp/session/modules/SessionFiles.cpp
```diff
@@ -1391,26 +1391,35 @@ class ListFilesAcceptAll
...
+      preserver_.add(callSEXP_);
    }
    
    bool operator()(const boost::filesystem::path& path)
    {
-      return boost::regex_search(path.native(), pattern_);
+      // construct call
+      r::sexp::Protect protect;
+      SEXP xSEXP = r::sexp::create(path.string(), &protect);
+      SETCADDR(callSEXP_, xSEXP);
```
jgutman, line 1413: does this just set xSEXP to the second argument because that one happens to be R_NilValue?

jgutman (approved):
LGTM!

src/cpp/session/modules/SessionFiles.cpp
kevinushey: That's right. The R call object is basically like a list of elements ["grepl", <pattern>, <x>, <ignore.case>]; to set the 3rd element in that list (the second argument), we use `SETCADDR()`.

kevinushey:
Holding on merge as we may want to take for PT and then merge back to SW.

PR #10685, opened 2022-02-26
use objects as constructor args (electron)

gtritchie:
### Intent

The window classes ported straight from C++ include long list of constructor arguments, leading to things like:

```
    const win = new DesktopBrowserWindow(false, false, false, 'some name');
```

### Approach

Use a more TypeScripty object approach so arguments at call sites are more self-documenting, i.e.:

```
    const win = new DesktopBrowserWindow({
      showToolbar: false,
      adjustTitle: false,
      autohideMenu: false,
      name: 'some name',
    });
```

Still kinda ugly and could use more tidying / consolidation / defaults, but I think this is somewhat better.

### Automated Tests

Updated existing tests, no new tests.

### QA Notes

Internal code cleanup, nothing specifically to test.

### Checklist

### DIFF
--- a/src/node/desktop/src/main/desktop-browser-window.ts
+++ b/src/node/desktop/src/main/desktop-browser-window.ts
@@ -31,6 +31,21 @@ import { executeJavaScript, isSafeHost } from './utils';
 // whether you're running in development or production).
 declare const MAIN_WINDOW_PRELOAD_WEBPACK_ENTRY: string;
 
+export interface WindowConstructorOptions {
+  showToolbar: boolean;
+  adjustTitle: boolean;
+  autohideMenu: boolean;
+  name: string;
+  baseUrl?: string;
+  parent?: DesktopBrowserWindow;
+  opener?: WebContents;
+  /** Default is `false` */
+  allowExternalNavigate?: boolean;
+  addApiKeys?: string[];
+  /** Attach to this `BrowserWindow` instead of creating a new one */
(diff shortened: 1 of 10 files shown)


kevinushey (approved):
LGTM! This feels like a nice readability improvement.

The only other change I'd suggest is that optional parameters could probably have their defaults set early on in the constructor, rather than at time of use.

It's also possible we could avoid the use of a separate `options` member with destructuring, as in https://medium.com/@rileyhilliard/es6-destructuring-in-typescript-4c048a8e9e15, but I think this is already enough of an improvement.

https://github.com/microsoft/TypeScript/issues/5326 also discusses some other missing pieces that could make this a bit nicer.

gtritchie:
> It's also possible we could avoid the use of a separate `options` member with destructuring, as in https://medium.com/@rileyhilliard/es6-destructuring-in-typescript-4c048a8e9e15, but I think this is already enough of an improvement.

I tried that aproached first. Couldn't quite make it build cleanly. As I understand it `Object.assign` isn't type-checked, and that seems to "concern" TypeScript about whether the non-optional class members are being initialized.

That article's example exhibits the same warning I was hitting. I could assign defaults to the non-optional class members, but that's an extra step that should be unnecessary. Also some concern about needing to make sure class members and those being assigned via Object.assign are spelled the same, the compiler can't check that.

Screen Shot 2022-02-26 at 11 04 29

Issue #10704, opened 2022-03-01
Localized IDE version can't return properly to no project when closing a project


cderv:
### System details

    RStudio Edition : Desktop [Open Source]
    RStudio Version : 2022.6.0.160
    OS Version      : Windows 10 x64 (build 22000)
    R Version       : R version 4.1.2 (2021-11-01)

**On a French Computer so I get localization activated in the IDE**

### Steps to reproduce the problem

* Open RStudio IDE **in a project**
* Close the project to go back to the equivalent of project: (None) -> in French translated to Projet: (Aucun)

### Describe the problem in detail

This will cause an issue in a loop - not possible to use RStudio IDE : 

* server 
  ![image](https://user-images.githubusercontent.com/6791940/156204135-3116200b-5f3b-429a-a368-df79fc879ce5.png)

* Desktop
  ![image](https://user-images.githubusercontent.com/6791940/156204624-c97ec00e-d6a1-4c7f-8bee-805a9feac297.png)

This is probably due to the translation because "none" is tested explicitly

https://github.com/rstudio/rstudio/blob/c96e01c1cde195dcdbafc1172072acfb0b0817dd/src/cpp/core/r_util/RSessionContext.cpp?q=kProjectNone#L54-L75

### Describe the behavior you expected

IDE can be opened without issue :)

PR #10707, opened 2022-03-01
fix project-related strings that were overlocalized

gtritchie:
### Intent

Addresses: #10704

Case where strings were localized in the UI code (GWT) but the C++ code has them hardcoded and requires that they match.

### Approach

Put back hardcoded strings.

Problem here is we have UI strings that should be localizable but are used internally as enum constants, existing in both the C++ and GWT code. This is objectively horrible, but not going to try and unravel that now.

### Automated Tests

None

### QA Notes

Test scenario described in the issue. You'll need to be running UI in French.

### Checklist


kevinushey (approved):
LGTM

Issue #10463, opened 2022-01-28
Ace editor preview doesn't get monospace font


kevinushey:
### System details

    RStudio Edition : 2022.6.0.40
    RStudio Version : Desktop [Open Source]
    OS Version      : macOS Monterey 12.1
    R Version       : R version 4.1.2 (2021-11-01)

---

<img width="608" alt="Screen Shot 2022-01-27 at 4 06 37 PM" src="https://user-images.githubusercontent.com/1976582/151463954-ef07df17-843c-48e4-a24f-5bda85aa2e66.png">

gtritchie:
Related to: #10438

timtmok:
I'll take a look at this while looking at #10438

kevinushey:
Verified fixed in RStudio 2022.06.0-daily+242 "Spotted Wakerobin" Daily (65af00457f533e066883e4a96833153a68c4aa4b, 2022-03-28) for macOS
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) RStudio/2022.06.0-daily+242 Chrome/98.0.4758.109 Electron/17.1.2 Safari/537.36.

<img width="607" alt="Screen Shot 2022-03-28 at 4 17 28 PM" src="https://user-images.githubusercontent.com/1976582/160502832-30a3d43e-5d38-4e88-b7f3-b50157ffd9ae.png">

Issue #10438, opened 2022-01-26
Electron font-selection logic


gtritchie:
The fonts used by Electron are currently hardcoded, e.g. here: https://github.com/rstudio/rstudio/blob/spotted-wakerobin/src/node/desktop/src/renderer/desktop-info-bridge.ts

Need to implement logic for selecting these; at a minimum replicate what we have in the Qt version, but also consider #4134 and #8659.

timtmok:
https://github.com/rstudio/rstudio/issues/10801 will address validating the values written.

kevinushey:
Almost looks ready. One thing I noticed is that one cannot select different font styles for a particular installed font on macOS; only the "regular" font is available.

---

## Ubuntu 20.04

### Electron

<img width="185" alt="Screen Shot 2022-03-28 at 4 48 42 PM" src="https://user-images.githubusercontent.com/1976582/160505443-fa68b52b-d1b7-4f95-b01c-8efe81e6bed1.png">

### Qt

[TBA]

## macOS

### Electron
<img width="187" alt="Screen Shot 2022-03-28 at 4 50 03 PM" src="https://user-images.githubusercontent.com/1976582/160505577-2ca0987a-c8e9-401b-890a-72482f785c1c.png">

### Qt

<img width="193" alt="Screen Shot 2022-03-28 at 4 50 48 PM" src="https://user-images.githubusercontent.com/1976582/160505648-add49bac-283c-49bd-b594-a31af70b2172.png">

## Windows

### Electron

<img width="186" alt="Screen Shot 2022-03-28 at 4 57 58 PM" src="https://user-images.githubusercontent.com/1976582/160506372-88ad568c-93a2-4b1e-9b4c-6bd48f422ab4.png">

### Qt

<img width="189" alt="Screen Shot 2022-03-28 at 4 58 43 PM" src="https://user-images.githubusercontent.com/1976582/160506428-4c792ce8-dff9-4868-8dab-8ce788ba43a6.png">

timtmok:
I think the native implementation provided in `node-system-fonts` could be patched to return fonts with their style options. I'm checking Apple docs to see if it's possible to get a font list similar to what Qt was providing.

PR #10711, opened 2022-03-01
Electron font selection

timtmok:
### Intent
Part one to address #10438 and #10463 for font selection and seeing the font in the preview. Setting the font doesn't actually change the editor's font yet since it doesn't write or read a persisted preference.

### Approach
Changes the preload so it sends an event to the main process. On startup, the main process will get all the system fonts. On receiving the event, it can then return the list of fonts or the default monospace/proportional font. It's a synchronous event call from the renderer to the main process since the existing GWT code does not expect a promise. Given that the font list is generated on startup, the call to retrieve the fonts is retrieving the cache.

I did notice that the old code would fallback to Qt for querying system fonts. There is no fallback here if `node-system-fonts` did not discover any fonts.

### Automated Tests
None.

### QA Notes
There should be no duplicate fonts in the list and all font options should change the preview. The editor does not apply the font selection at this time. 

### Checklist


src/node/desktop/NOTICE.electron
```diff
@@ -258,4 +258,28 @@ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
 AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
 LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-THE SOFTWARE.
\ No newline at end of file
+THE SOFTWARE.
```
gtritchie, line 261: Also add mention of this dependency in the list at the top of the file.

src/node/desktop/src/main/gwt-callback.ts
```diff
@@ -57,6 +57,14 @@ export class GwtCallback extends EventEmitter {
     super();
     this.owners.add(mainWindow);
 
+    const fontDescriptors = [
+      ...new Map<string, IQueryFontDescriptor>(findFontsSync({}).map((fd) => [fd.family, fd])).values(),
+    ].sort((a, b) => a.family?.localeCompare(b.family ?? '') ?? 0);
+    const monospaceFonts = fontDescriptors.filter((fd) => fd.monospace).map((font) => font.family);
+    const proportionalFonts = fontDescriptors.filter((fd) => !fd.monospace).map((font) => font.family);
+
+    // this.monospaceFonts = [...new Set(fontDescriptors)].sort((a, b) => a.family.localeCompare(b.family));
```
gtritchie, line 66: Either remove this commented out line or add a comment why you want to keep it around.

gtritchie (changes requested):
Couple very minor tweaks requested but otherwise LGTM!

src/node/desktop/src/main/gwt-callback.ts
```diff
@@ -441,17 +448,54 @@ export class GwtCallback extends EventEmitter {
...
 
-    ipcMain.handle('desktop_get_fixed_width_font', () => {
-      GwtCallback.unimpl('desktop_get_fixed_width_font');
-      return '';
+    ipcMain.on('desktop_get_fixed_width_font', (event) => {
+      // TODO: Read user preference for font
+
+      let defaultFonts: string[];
+      if (process.platform === 'darwin') {
+        defaultFonts = ['Monaco'];
```
kevinushey, line 460: Nit: the default monospace font on macOS is Menlo nowadays; is it worth changing that here?

src/node/desktop/src/renderer/desktop-info-bridge.ts
```diff
@@ -13,78 +13,19 @@
...
-      chromiumDevtoolsPort: 0,
-    };
-  }
+  return {
+    platform: '',
+    version: '',
+    scrollingCompensationType: '',
+    fixedWidthFontList: ipcRenderer.sendSync('desktop_get_fixed_width_font_list'),
+    fixedWidthFont: ipcRenderer.sendSync('desktop_get_fixed_width_font'),
+    proportionalFont: ipcRenderer.sendSync('desktop_get_proportional_font'),
```
kevinushey, line 26: One thing to be aware of -- historically, users (especially on Linux) have found that listing fonts can be very slow when there are a large number of fonts installed on the system. We tried to dodge that on the desktop side by making use of the font cache on Linux. See, for example:

https://github.com/rstudio/rstudio/blob/466d03b0ed61d66f3a80e7c2e530fb386818d9b4/src/cpp/desktop/DesktopInfo.cpp#L105-L146

I don't think it needs to be addressed in this PR, but I think it's worth adding an issue for QA testing later.

kevinushey (approved):
LGTM; main worry is around attempts to synchronously list available fonts, which has historically been slow for some users.

src/node/desktop/src/main/gwt-callback.ts
timtmok: Sorry, I missed this one when cleaning everything up

src/node/desktop/src/main/gwt-callback.ts
timtmok: I did not realize that! It's probably worth changing and I gave it a quick try. It still defaulted to Monaco because the style was overridden by Ace's default style. I'll have to look at what needs to be done to fix this. I think it needs a call to `preview_.setFont` since I don't see that being done anywhere when the appearance preferences pane is created.

FYI: This is also an issue in Prairie Trillium. The selected font in the dropdown is not the one used in the preview until the selection is changed.

src/node/desktop/src/renderer/desktop-info-bridge.ts
timtmok: Good point and I agree. I think this could be improved. The main process should asynchronously retrieve the fonts, which would be fine for the monospace font list since it is used for the appearance options. 

The proportional font needs to be discovered for the UI. Maybe we define default fonts for each platform and update the font once the system fonts have been retrieved.

For the selected fixed width font, that definitely needs changing once it reads the user's preference. There's nothing implemented so far to read the preference file that stores this setting and the file read would be an async call.

It's probably better if all of these font calls using the context bridge were changed to async. At the moment, it is awkward to call back into GWT from what I've seen. Maybe I'm not quite understanding  the correct class to use to handle promises in GWT.

PR #10713, opened 2022-03-01
update Search label on preview replace event

jgutman:
### Intent

Addresses #10626.

### Approach

When the previewReplace event fires (only when isRegex is True), the searchLabel is cleared (with stopAndClear) but not repopulated, as is done when the completeReplace event fires. Adding that bit when the RPC response is received appears to fix the clearing out of the search label in regex preview results

### Automated Tests

NA, just ran RStudio Desktop locally.

### QA Notes

See example in the #10626. Run Find in Files, with "Regular expression" checked, and in the preview results, enter a replacement. The search label at the top of the Find in Files pane should update from `Replace results for /hello/` to `Replace results for /hello/ with /bye/` rather than clearing out.

### Checklist


melissa-barca (approved):
LGTM!

PR #10727, opened 2022-03-02
Update initial editor preview with selected font

timtmok:
### Intent
Address #10463 

The editor preview wouldn't use the selected font when the appearance pane is first viewed. Switching to another font and back would show the correct font. I'm not sure when this broke but it's also an issue in Prairie Trillium.

Also, sets the MacOS default font to Menlo then Monaco if unavailable.

### Approach
When retrieving the selected font, it comes back in quotes. This has `!important` and `monospace` appended before setting the style element. The problem was adding another set of quotes (`""Menlo", monospace" !important`) that resulted in an invalid font name. This will properly set the font attribute to `"Menlo", monospace !important`.

### Automated Tests
None

### QA Notes
This fixes RStudio web, Qt, and Electron. Although, this was probably unnoticed since the invalid attribute meant falling back to Ace's default monospace list. It just happens to use the same font as what we would have wanted (on most OS'es). I only noticed this now because changing the default to Menlo showed Monaco (Ace's default monospace).

### Checklist

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/theme/ThemeFonts.java
+++ b/src/gwt/src/org/rstudio/core/client/theme/ThemeFonts.java
@@ -34,7 +34,7 @@ public class ThemeFonts
 
    public static String getFixedWidthFont()
    {
-      return fontLoader.getFixedWidthFont() + ", monospace";
+      return fontLoader.getFixedWidthFont();
    }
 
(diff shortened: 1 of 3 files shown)


kevinushey (approved):
LGTM; thanks for running that down!

PR #10739, opened 2022-03-04
Electron preference management for font selection

timtmok:
### Intent
Addresses #10438 so that it can read the existing preferences that Qt writes. This allows checking if there is an existing font selected. The editor will now show the selected font and the appearance options will have the font pre-selected.

Also fixed an issue with Linux loading the font list. `node-system-fonts` has a bug with how it uses Fontconfig to query all available fonts. It sometimes incorrectly marks a font as monospace. The bug is not present when providing query options.

### Approach
For MacOS, it uses Electron's `systemPreferences`. The `CFBundleIdentifier` needed to be properly set for this to work. In local dev, this is not set so it reads from `com.github.Electron`.

Windows and Linux both read from `desktop.ini`. A package has been added that can parse ini files. Both platforms use the same code but use the platform specific location for `desktop.ini`.

There is an interface for the preference manager. It uses Electron's preference value types. This might be able to be cleaned up a bit to use something more agnostic to all platforms. For Linux/Windows, it will always be a `string`.

### Automated Tests
WIP

### QA Notes
 Linux should only show monospace fonts in the options. All platforms should load the font specified in the preference store. Setting a font is still unsupported.

### Checklist


src/node/desktop/src/main/preferences/mac-preferences.ts
```diff
@@ -0,0 +1,19 @@
+import { systemPreferences, UserDefaultTypes } from 'electron';
+import { PlatformPreferences } from './preferences';
+
+/**
+ * MacOS specific implementation
+ *
+ * The preferences are stored in defauls, which is MacOS specific.
```
kevinushey, line 7:
```suggestion
 * The preferences are stored in defaults, which is MacOS specific.
```

src/node/desktop/src/main/preferences/preferences.ts
```diff
@@ -0,0 +1,37 @@
...
+    case 'win32':
+      return new FilePreferences(`${homedir()}\\AppData\\Roaming\\RStudio\\desktop.ini`);
+    case 'linux':
+      return new FilePreferences(`${homedir()}/.config/RStudio/desktop.ini`);
+    default:
+      throw Error('unsupported platform');
+  }
+};
+
+export default getPreferenceManager;
```
kevinushey, line 37: I think you could also just export an object called `preferenceManager` and initialize it when this module is loaded. IMHO that's marginally cleaner; do you have a strong opinion?

kevinushey (approved):
LGTM!

src/node/desktop/src/main/preferences/preferences.ts
timtmok: Yes, I think that would be a cleaner option! 👍

Issue #10401, opened 2022-01-20
Provide way to display "Choose R Installation" dialog at startup of Electron app on Windows


gtritchie:
The "Choose R Installation" dialog on Windows is already mostly implemented (#8815) and currently displays every time you launch the Windows desktop app.

In the Qt app we only display this dialog if you hold down Ctrl key while launching. Initial investigations didn't find an obvious way to do this in an electron app. We might need some native code for it (#10376) but be nice to avoid that.

PR #10749, opened 2022-03-08
R Version Selection Window on Windows being shown only with CTRL key being hold

matheustavaresdev:
Addresses #10401

### Intent

Show R Version Selection Window only when holding the CTRL key on Windows.

### Approach

Added a native node module ( @kevinushey ). Handled the ctrl key status through it. The path for the R version is being saved locally for next time the user opens the app. 

Also fixed a bug that would make the selection window not to close when the user finishes selecting a R version through the `Browse` button.

### Automated Tests

Added tests to make sure that this window will not open in other OSs that are not Windows. Need quite some env configs to test for further functionalities of this feature.

### QA Notes

None

### Checklist


src/node/desktop/src/main/detect-r.ts
```diff
@@ -55,34 +58,56 @@ function executeCommand(command: string): Expected<string> {
...
+export async function promptUserForR(platform = process.platform): Promise<Expected<string | null>> {
+  if (platform === 'win32') {
+    const desktop = await import('../native/desktop.node');
 
-  // discover available R installations
-  const rInstalls = findRInstallationsWin32();
-  if (rInstalls.length === 0) {
-    return err();
-  }
+    const storeConfig = new Store({ defaults: kDesktopOptionDefaults });
```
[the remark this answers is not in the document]
timtmok: @matheustavaresdev I just merged changes with how preferences work. I ended up renaming `DesktopOptions` to `ElectronDesktopOptions` so it is differentiated from the `legacyPreferenceManager` which required us supporting the platform-specific code for the preference store.

So I think all that needs to be done is updating `ElectronDesktopOptions` to set and retrieve this new option. Since it didn't exist in Qt, `legacyPreferenceManager` has no need to implement it.

src/node/desktop/src/main/detect-r.ts
```diff
@@ -55,34 +58,56 @@ function executeCommand(command: string): Expected<string> {
...
 
-export async function promptUserForR(): Promise<Expected<string | null>> {
-  // nothing to do if RSTUDIO_WHICH_R is set
-  const rstudioWhichR = getenv('RSTUDIO_WHICH_R');
-  if (rstudioWhichR) {
-    return ok(rstudioWhichR);
-  }
+export async function promptUserForR(platform = process.platform): Promise<Expected<string | null>> {
+  if (platform === 'win32') {
+    const desktop = await import('../native/desktop.node');
```
kevinushey, line 63: Any reason to put the import here rather than at the top level?

src/node/desktop/src/main/detect-r.ts
```diff
@@ -55,34 +58,56 @@ function executeCommand(command: string): Expected<string> {
...
+    const storeConfig = new Store({ defaults: kDesktopOptionDefaults });
+    const rstudioPathKey = 'rstudioPath';
 
-  // ask the user what version of R they'd like to use
-  const dialog = new ChooseRModalWindow(rInstalls);
-  const [path, error] = await dialog.showModal();
-  if (error) {
-    return err(error);
-  }
+    const isCtrlKeyDown = desktop.isCtrlKeyDown() as boolean;
```
kevinushey, line 68: Is the `as boolean` here necessary? Since we already declare the return value in the `.d.ts` file.

src/node/desktop/src/ui/widgets/choose-r/preload.ts
```diff
@@ -97,8 +97,9 @@ const callbacks: Callbacks = {
     ipcRenderer.send('cancel');
   },
 
-  browse: () => {
-    ipcRenderer.send('browse');
+  browse: async () => {
+    const shouldCloseDialog = await ipcRenderer.invoke('browse-r-exe');
```
kevinushey, line 101: Is there a reason for changing the name here? (Do we need something like namespacing for the names used in IPC requests / responses?)

kevinushey (commented):
Overall LGTM, but I agree with @timtmok that we can make use of the existing DesktopOptions tools.

src/node/desktop/src/main/detect-r.ts
matheustavaresdev: Fixed! This was mostly required from my old code.

src/node/desktop/src/ui/widgets/choose-r/preload.ts
matheustavaresdev: Well... `browse` is a very common name. My fear was that this would be used somewhere else and mess up with our code. So I decided to change it already not to lose any time or have headaches in the future.

src/node/desktop/src/main/detect-r.ts
matheustavaresdev: Well, when I was using an external node package for this feature, the package would require to only be installed on Windows OS. So I had to do it the import that way. So due to that, I think that doing it like this would probably save us from some headache that could or could not happen in the future as it won't hurt to import it like that, by the contrary: that's lazy import for this module which will only really import it when it's used...

src/node/desktop/src/main/detect-r.ts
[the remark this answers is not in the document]
matheustavaresdev: Fixed on https://github.com/rstudio/rstudio/pull/10749/commits/67a626fc3a8fcd1a1817d44d846e4a1b8fa81dbc

PR #10755, opened 2022-03-08
Electron preferences part 2: Writing to electron-store, reading from electron-store

timtmok:
### Intent
Address saving font selection #10438

Saves the editor font selection to the electron-store. Reading will check electron-store first and fallback to the legacy option store. Cleaned up the way GWT gets the fixed font. It falls back to the default font list if both option stores return nothing.

### Approach
Refactored the `preferenceManager` to be an interface. Rename the preference manager to `legacyPreferenceManager` since that should only be used for reading and handles platform specific storage. `DesktopOptions` has been renamed to `ElectronDesktopOptions` as it is specific to how Electron will be handling preferences (falling back to legacy options). `PlatformPreferences` is now the interface `DesktopOptions` so that it defines the options available. For now, it's just `fixedWidthFont` but others will be added in a follow-up.

Added an npm launch for debugging. I couldn't debug from VS Code in Windows so I've added this launch so it's possible to connect using Chrome's `chrome://inspect`.

### Automated Tests
I've added tests to check:
* if the font isn't set in electron-store then it's retrieved from the legacy options
* if the font is set in electron-store then it's returned

### QA Notes
All platforms are able to select the font in the Appearance Options. The prompt to restart is required to see the font change. 

### Checklist

### DIFF
--- a/src/node/desktop/src/main/preferences/file-preferences.ts
+++ b/src/node/desktop/src/main/preferences/file-preferences.ts
@@ -14,32 +14,32 @@
  *
  */
 
-import { homedir } from 'os';
 import PropertiesReader from 'properties-reader';
+import { FilePath } from '../../core/file-path';
+import { Xdg } from '../../core/xdg';
 import DesktopOptions from './desktop-options';
 import { preferenceKeys } from './preferences';
 
 const INI_FILE = 'desktop.ini';
 
 class FilePreferences extends DesktopOptions {
-  private iniFile: string; // existing location for some preferences
-  private properties: PropertiesReader.Reader;
+  private properties?: PropertiesReader.Reader;
 
   constructor() {
     super();
-    if (process.platform === 'win32') {
-      this.iniFile = `${homedir()}\\AppData\\Roaming\\RStudio\\${INI_FILE}`;
-    } else if (process.platform === 'linux') {
-      this.iniFile = `${homedir()}/.config/RStudio/${INI_FILE}`;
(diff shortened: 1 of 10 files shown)


src/node/desktop/src/main/preferences/file-preferences.ts
```diff
@@ -1,20 +1,45 @@
...
+class FilePreferences extends DesktopOptions {
+  private iniFile: string; // existing location for some preferences
   private properties: PropertiesReader.Reader;
 
-  constructor(file: string) {
-    this.properties = PropertiesReader(file);
-  }
+  constructor() {
+    super();
+    if (process.platform === 'win32') {
```
gtritchie, line 30: We should use the Xdg methods for determining these paths. These live in https://github.com/rstudio/rstudio/blob/main/src/node/desktop/src/core/xdg.ts, e.g. see `Xdg.userConfigDir()`, `Xdg.userDataDir()`.

gtritchie (changes requested):

src/node/desktop/src/main/preferences/file-preferences.ts
timtmok: Good catch. I've changed it to use `userConfigDir()`. I've also added a check in case the file does not exist. It will fallback to the default font if no font selection is found.

timtmok:
Merging so that zoom level preferences can use this to read the legacy values

PR #10764, opened 2022-03-10
suppress a number of compiler warnings

kevinushey:
Small changes to suppress some compiler warnings on macOS, e.g.

```
[125/535] Building CXX object core/CMakeFiles/rstudio-core.dir/system/PosixSystem.cpp.o
/Users/kevin/rstudio/src/cpp/core/system/PosixSystem.cpp:2576:10: warning: unused variable 's_privUid' [-Wunused-variable]
   uid_t s_privUid;
         ^
1 warning generated.
[144/535] Building CXX object core/CMakeFiles/rstudio-core-tests.dir/system/PosixSystemTests.cpp.o
/Users/kevin/rstudio/src/cpp/core/system/PosixSystemTests.cpp:31:20: warning: unused function 'getNoGroupName' [-Wunused-function]
static std::string getNoGroupName()
                   ^
1 warning generated.
[171/535] Building CXX object server_core/CMakeFiles/rstudio-server-core.dir/ServerDatabase.cpp.o
/Users/kevin/rstudio/src/cpp/server_core/ServerDatabase.cpp:190:21: warning: expression result unused [-Wunused-value]
      OBFUSCATE_KEY(secureKey);
                    ^~~~~~~~~
/Users/kevin/rstudio/src/cpp/server_core/include/server_core/ServerKeyObfuscation.hpp:19:29: note: expanded from macro 'OBFUSCATE_KEY'
#define OBFUSCATE_KEY(key) (key)
                            ^~~
1 warning generated.
[278/535] Building CXX object session/CMakeFiles/rsession.dir/modules/SessionFiles.cpp.o
/Users/kevin/rstudio/src/cpp/session/modules/SessionFiles.cpp:1722:12: warning: unused variable 'ignoreCase' [-Wunused-variable]
      bool ignoreCase   = validateLogical(ignoreCaseSEXP,  "ignore.case");
           ^
1 warning generated.
```

### DIFF
--- a/src/cpp/core/system/PosixSystem.cpp
+++ b/src/cpp/core/system/PosixSystem.cpp
@@ -2572,10 +2572,6 @@ Error permanentlyDropPriv(const std::string& newUsername, const std::string& new
 // privilege manipulation for systems that don't support setresuid/getresuid
 #else
 
-namespace {
-   uid_t s_privUid;
-}
-
 Error permanentlyDropPriv(const std::string& newUsername)
 {
    return permanentlyDropPriv(newUsername, std::string());
--- a/src/cpp/core/system/PosixSystemTests.cpp
+++ b/src/cpp/core/system/PosixSystemTests.cpp
@@ -45,6 +45,7 @@ static std::string getNoGroupName()
    else if (getgrnam("nogroup"))
       group = "nogroup"; // Debian/Ubuntu
 
+   expect_false(group.empty());
    return group;
 }
 
--- a/src/cpp/core/system/PosixSystemTests.cpp
+++ b/src/cpp/core/system/PosixSystemTests.cpp
@@ -28,6 +28,8 @@ namespace core {
 namespace system {
 namespace tests {
 
+#ifdef __linux__
+
 static std::string getNoGroupName()
 {
    std::string group;
@@ -43,9 +45,10 @@ static std::string getNoGroupName()
    else if (getgrnam("nogroup"))
       group = "nogroup"; // Debian/Ubuntu
 
-   expect_false(group.empty());
    return group;
 }
+
+#endif
 
 test_context("PosixSystemTests")
 {
--- a/src/cpp/server_core/include/server_core/ServerKeyObfuscation.hpp
+++ b/src/cpp/server_core/include/server_core/ServerKeyObfuscation.hpp
@@ -16,6 +16,7 @@
 #ifndef SERVER_CORE_SERVER_KEY_OBFUSCATION_HPP
 #define SERVER_CORE_SERVER_KEY_OBFUSCATION_HPP
 
-#define OBFUSCATE_KEY(key) (key)
+// stub for open source
+#define OBFUSCATE_KEY(key) (key[0] = key[0])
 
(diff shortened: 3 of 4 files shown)


src/cpp/core/system/PosixSystem.cpp
```diff
@@ -2572,10 +2572,6 @@ Error permanentlyDropPriv(const std::string& newUsername, const std::string& new
 // privilege manipulation for systems that don't support setresuid/getresuid
 #else
 
-namespace {
-   uid_t s_privUid;
```
kevinushey, line 2576: This appears to be used in Pro, so we need to be careful with the merge here.

src/cpp/core/system/PosixSystemTests.cpp
```diff
@@ -43,10 +45,11 @@ static std::string getNoGroupName()
    else if (getgrnam("nogroup"))
       group = "nogroup"; // Debian/Ubuntu
 
-   expect_false(group.empty());
```
kevinushey, line 46: Not sure if this expectation should be part of the helper function, or part of the test case itself?

gtritchie: Was it causing a build warning?

I put this here to make it easier to spot any future quirks where a container doesn't have one of the expected groups.

gtritchie: Ah I see on Mac it flags getNoGroupName as unused; go ahead and wrap that function in `#ifdef __linux__`.

src/cpp/core/system/PosixSystem.cpp
gtritchie: If that's true I'd try and keep the file the same between open-source and pro. But at a quick glance, I don't see where this gets used in Pro, either. Maybe it was accidentally left behind when things got split into shared_core? The one there does seem to be used:

https://github.com/rstudio/rstudio/blob/2715863751b3553cc5472db2f4532470c60300e4/src/cpp/shared_core/system/PosixSystem.cpp#L247

gtritchie (changes requested):

src/cpp/core/system/PosixSystemTests.cpp
```diff
@@ -28,6 +28,8 @@ namespace core {
 namespace system {
 namespace tests {
 
+#ifdef __linux__
```
gtritchie, line 31: Oops, thanks for fixing that one!

kevinushey:
Aha, there are _two_ PosixSystem.cpp's on Pro:

```
$ rg s_privUid
src/cpp/shared_core/system/PosixSystem.cpp
247:   uid_t s_privUid;
252:   return restorePrivilegesImpl(s_privUid);
307:   s_privUid = oldEUID;

src/cpp/core/system/PosixSystem.cpp
2576:   uid_t s_privUid;
```

So we can indeed remove the version not in shared core.

src/cpp/core/system/PosixSystemTests.cpp
kevinushey: Sounds good; I'll leave it in.

PR #10772, opened 2022-03-11
Fix logging regression for certain RSW binaries

gtritchie:
### Intent

A change to support logging when RStudio Server open source is run as non-root caused logging to fail for some RSW binaries due to permission issues. That change was made here: https://github.com/rstudio/rstudio/pull/10126

Addresses https://github.com/rstudio/rstudio-pro/issues/3307
Addresses https://github.com/rstudio/rstudio-pro/issues/3318

### Approach

When compiling for Pro (RSW), behave the way it was before the regression, but leave the new behaviour for open-source.

Note that binaries such as rsession and rworkspace, which run as non-root, do not rely on the default set in this code; they set their logging locations explicitely when initializing the logging subsystem.

Also, `RSTUDIO_PRO_BUILD` was already defined for pro-builds, just needed to configure that cmake variable into a header file.

https://github.com/rstudio/rstudio/blob/ab97ae298e1258ce54122900b07d9eb9a6027ec5/CMakeGlobals.txt#L60-L65

### Automated Tests

None

### QA Notes

Make sure log files for rserver-session-reaper and rserver-acls are created as expected in RSW. You'll want to up the logging level to `debug` and leave the server running for > 30 minutes for the reaper log to show anything, and enable project sharing for the acls issue.
 
### Checklist

### DIFF
--- a/src/cpp/core/LogOptions.cpp
+++ b/src/cpp/core/LogOptions.cpp
@@ -73,6 +73,11 @@ namespace {
 FilePath defaultLogPathImpl()
 {
 #ifdef RSTUDIO_SERVER
+#ifdef RSTUDIO_PRO_BUILD
+   return FilePath("/var/log/rstudio/rstudio-server");
+#else
+   // For open-source Server, support logging even if RStudio Server is run
+   // as a non-root user.
    if (core::system::effectiveUserIsRoot())
    {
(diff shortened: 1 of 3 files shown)


kevinushey (approved):
LGTM! Thanks for running this down.

PR #10775, opened 2022-03-11
support unicode characters + pipebind operator

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/10757.

### Approach

Add rules to tokenize `_` as an identifier. Note that the diagnostics engine treats it as a regular R identifier, but we should later ensure that the use of `_` outside of a pipe-bind expression is an error.

### Automated Tests

Not included.

### QA Notes

Test that code of the following form:

```
mtcars |> summary(object = _)
```

can execute without issue in R 4.2.0, and also that this code snippet (when saved in an R file) does not produce any diagnostic errors.

### Checklist


kevinushey:
> what does merge: false mean?

Unfortunately, Ace's documentation is sparse -- often, you'll need to delve into the Ace source code to find answers. In this case, the answer lives here:

https://github.com/ajaxorg/ace/blob/68b5029ed4c99f15ab40c59c66a4c8ecb19be5ba/lib/ace/tokenizer.js#L308-L309

Basically, if Ace sees two tokens of the same type next to each other, it tries to merge them together into a single token. So, if you theoretically had the code:

```
_abc
```

which is technically two separate identifiers (`_` and `abc`), Ace would by default try to merge them into a single `_abc` token. However, we don't want that -- we want them to be parsed as separate tokens, so the diagnostics system can flag the fact that `_` and `abc` are separate things.

> Why are some \\ escapes no longer needed?

This is because unicode and non-unicode regular expressions have different escaping rules in JavaScript. For example:

https://stackoverflow.com/questions/63006601/why-does-u-throw-an-invalid-escape-error/63007777#63007777

Our existing regular expression patterns include a number of "identity escapes"; e.g. things like `\'` where the escaping slash is not actually unnecessary. Unfortunately, such usages are considered an error for unicode regular expressions in JavaScript.

---

For the diagnostics issue, my goal was to handle that in a separate PR.

PR #10781, opened 2022-03-14
PT: Fix session restore when pandoc location does not exist

timtmok:
### Intent
Handle PT resuming a session created with GO. Since Pandoc moved locations, the restored `RSTUDIO_PANDOC` location does not exist.

### Approach
When restoring the `RSTUDIO_PANDOC` env var, use the value if the directory exists. Otherwise, use the system set location.

### Automated Tests
None

### QA Notes
I have not been able to reproduce this in a dev environment so I cannot test if this actually works. I'm still working on reproducing so I can test this.

### Checklist


jeffvroom:
Turns out we set the default RSTUDIO_PANDOC after the session resume. So in this case, we need the invalid var to be ignored even if it's not already set in the environment. I updated the branch with that logic and tested that the RMD knits fine after the resume.

jeffvroom (approved):
LGTM

PR #10783, opened 2022-03-15
Add refresh button and functionality to Find in Files pane

jgutman:
### Intent

Addresses #3240 . Adds a refresh button to Find in Files

### Approach

Refresh button and stop search button should not display simultaneously.
Refresh button should clear previous results and begin a new find

### Screenshots

Initial view after Shift + Cmd + F:
Replace toolbar not visible, Stop search button visible, Refresh button not visible, Search in progress
Screen Shot 2022-03-16 at 12 44 05 PM

Search has completed, Refresh button visible, Stop search button not visible
Screen Shot 2022-03-16 at 12 43 37 PM

Toggle on replace and enter replacement, Refresh button still visible, Replace toolbar visible, Replace preview mode active
Screen Shot 2022-03-16 at 12 46 28 PM

Hover over Refresh button to see alt text
Screen Shot 2022-03-16 at 12 47 50 PM

After clicking on Refresh button, returns to initial state, exits replace mode, stop search button visible, refresh button not visible
Screen Shot 2022-03-16 at 12 48 28 PM

Tab back to replace mode, and replacement text has persisted from previous entry
Screen Shot 2022-03-16 at 12 49 33 PM

### Automated Tests

Will plan to add some selenium tests to ensure when replace toolbar is/isn't visible, when refresh button is/isn't visible, when stop search button is/isn't visible.

### QA Notes

1. Open Find in Files pane with Cmd + Shift + F and perform a search.
2. Confirm that when search is in progress, the Stop button is visible, and Refresh button is not. When search has finished, the Stop button is now hidden, and the Refresh button is visible.
3. Toggle over to Replace and enter a potential replacement in the "Replace with" text box.
4. Edit and save a file with a search result to remove the matching line. Hit the refresh button.
5. The search should refresh, the pane should toggle back to the Find mode, "Replace with" toolbar disappears until you click on "Replace" again. The search result you edited away in previous step should no longer show in the search results.

### Checklist

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/FindOutputPresenter.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/FindOutputPresenter.java
@@ -520,11 +520,10 @@ public class FindOutputPresenter extends BasePresenter
                   dialogState_.getPath(),
                   dialogState_.isWholeWord(),
                   dialogState_.isRegex());
-               view_.setStopSearchButtonVisible(true);
 
                super.onResponseReceived(handle);
-
                view_.ensureVisible(true);
+               view_.setStopSearchButtonVisible(true);
             }
          });
    }
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/FindOutputPane.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/FindOutputPane.java
@@ -71,13 +71,12 @@ public class FindOutputPane extends WorkbenchPane
             ToolbarButton.NoText,
             constants_.stopFindInFilesTitle(),
             commands_.interruptR().getImageResource());
-      stopSearch_.setVisible(false);
       toolbar.addRightWidget(stopSearch_);
 
       refreshButton_ = commands_.refreshFindInFiles().createToolbarButton();
       refreshButton_.addStyleName(ThemeStyles.INSTANCE.refreshToolbarButton());
-      refreshButton_.setVisible(true);
       toolbar.addRightWidget(refreshButton_);
+      setStopSearchButtonVisible(false);
 
       showFindButton_ = new LeftRightToggleButton(constants_.findLabel(), constants_.replaceLabel(), true);
       showFindButton_.addClickHandler(new ClickHandler() {
@@ -127,8 +126,8 @@ public class FindOutputPane extends WorkbenchPane
             ToolbarButton.NoText,
             constants_.stopReplaceTitle(),
             commands_.interruptR().getImageResource());
-      stopReplace_.setVisible(false);
       replaceToolbar.addRightWidget(stopReplace_);
+      setStopReplaceButtonVisible(false);
 
       replaceAllButton_ = new ToolbarButton(constants_.replaceAllText(), constants_.replaceAllText(), null);
       replaceToolbar.addRightWidget(replaceAllButton_);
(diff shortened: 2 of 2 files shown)


src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/FindOutputPresenter.java
```diff
@@ -64,6 +64,7 @@
 import org.rstudio.studio.client.workbench.views.output.find.model.FindInFilesServerOperations;
 import org.rstudio.studio.client.workbench.views.output.find.model.FindInFilesState;
 import org.rstudio.studio.client.workbench.views.output.find.model.FindResult;
+import org.rstudio.studio.client.workbench.views.plots.Plots;
```
jgutman, line 67:
```suggestion
```

jgutman:
```suggestion
```

src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/FindOutputPresenter.java
```diff
@@ -497,6 +499,7 @@ public void findInFilesBeginFind()
 
       // find result always starts with !replaceMode
       view_.setReplaceMode(false);
+      view_.turnOffReplaceMode();
```
jgutman, line 501: This fixes a pre-existing bug in the Find in Files. Previously, if I did a find in files search, tabbed over to the replace mode, entered a replacement in the text box, and then did Cmd + Shift + F to start a new find in files search, a new find in files search would open with the replace toolbar visible but disabled, and the previously entered replacement visible in the text box but not actually showing in the preview. This led to a weird UX where the user would have to edit the replacement text box, even if they wanted to keep the replacement the same, in order to actually activate replace mode. 

The fix is to make sure, when we begin a new Find in Files search, we completely turn off replace mode, not just turn off select parts of it and allow other parts to persist. This way when a new search begins, the replacement toolbar will not be visible, but if the user toggles over to it, the replacement they previously entered in the replacement text box will still persist, in case they want to reuse it.

melissa-barca (commented):
Overall this looks great! Could you had a sceen shot of the new button to the PR? It could be beneficial to add a selenium automated test for this new functionality as well.

src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/FindOutputPresenter.java
jgutman: Good idea -- I haven't written any yet, so I'll have to look into how to do that, but seems like a good idea.

Also, I believe `turnOffReplaceMode();` sets replace mode to false so `view_.setReplaceMode(false);` is now redundant/unnecessary -- will remove.

src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/FindOutputPresenter.java
jgutman: I thought I already removed this stray import but seems like I may have missed it

src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/FindOutputPresenter.java
melissa-barca: Looks like we only have some very basic ones for Find in Files so far: https://github.com/rstudio/rstudio-ide-automation/blob/main/rstudio_server_pro/tests/Panes/test_FindInFilesPane.py
It would be great if we could add some more, let me know if you need help getting started.

jgutman: Thanks, even so that's a very helpful starting point! I'll work on adding to that file

Issue #10443, opened 2022-01-26
Pass R code execution error messages to the RPC instead of always replacing with "R code execution error"


jgutman:
When an executing R function hits an error, an error gets passed to the RPC, and can be accessed on the frontend to display the error to the user. Unfortunately, the message that actually gets passed, no matter what the R error message should happen to be, is often just [R code execution error](https://github.com/rstudio/rstudio/blob/8f85b6b8ed4e7db1ac167b2007d17e26c947e4fd/src/cpp/r/RErrorCategory.cpp#L57). The actual error message and location get lost. It would be helpful if the actual error info in errc::CodeExecutionError could get passed more easily into the RPC for access on the frontend, in cases where it would be helpful to display some or all of the error to the user.

gtritchie:
I verified this new behaviour on build 2022.06.0-daily+239 on MacOS, using the steps described in #10793. Got the detailed error message. Closing.

PR #10793, opened 2022-03-16
Propagate R execution error messages in full

jgutman:
### Intent

Addresses #10443. Follow up from #10435, but allows any R code execution error messages to be accessible in the frontend.

### Approach

Instead of using the 2-argument Error constructor `Error::Error(const boost::system::error_code& in_ec, const ErrorLocation& in_location)`, we should use the 3-argument Error constructor `Error::Error(const boost::system::error_code& in_ec, std::string in_message, const ErrorLocation& in_location)`, which already exists, when calling the `rCodeExecutionError` function. This means that instead of using `"R code execution error"` as the error message for all R code execution errors, and not passing any further detailed error messages to the frontend, the frontend will now have access to the actual content of the R error message, and developers can choose to display that error to the user, or handle it further in the frontend if desired.

### Automated Tests

None

### QA Notes

This change is mostly to enable developers, so mostly we'd want to ensure that existing tests don't break. To really kick the tires on it, and test the motivating example:

**OLD BEHAVIOR**
1. Navigate to `PYTHON_PREFIX/lib/pythonX.Y/site-packages/conda/_vendor/toolz/` and rename `itertoolz.py` to `XXXitertoolz.py` (On my computer this is `/Users/jacquelinegutman/Library/miniconda3/lib/python3.9/site-packages/conda/_vendor/toolz/itertoolz.py`. You should then see an error when running `conda env list --json --quiet` in the terminal. (Don't forget to rename it back when you're done testing!)
2. In RStudio, navigate to Tools > Global Options > Python, and under Python interpreters, click Select...
3. If you're running Spotted Wakerobin, let's say version 2022.06.0-daily+208 (prior to the current fix) you will get an error message in the UI like
image
4. When you click OK, you will get a more detailed error message in the console. 
image

**NEW BEHAVIOR**
In the current version, introduced by this change:
2. In RStudio, navigate to Tools > Global Options > Python, and under Python interpreters, click Select...
3. You will see a much more detailed error message, such as 
Screen Shot 2022-03-16 at 2 08 49 PM
Screen Shot 2022-03-16 at 2 08 57 PM
4. Once again when you click OK, you will get the error message in the console as well

### Checklist

### DIFF
--- a/src/cpp/r/RErrorCategory.cpp
+++ b/src/cpp/r/RErrorCategory.cpp
@@ -84,7 +84,7 @@ std::string RErrorCategory::message( int ev ) const
 core::Error rCodeExecutionError(const std::string& errMsg, 
                                 const core::ErrorLocation& location)
 {
-   core::Error error(errc::CodeExecutionError, location);
+   core::Error error(errc::CodeExecutionError, errMsg, location);
    error.addProperty("errormsg", errMsg);
    return error;
 }


melissa-barca (approved):
LGTM! A small change that should be extremely helpful 👍

PR #10769, opened 2022-03-11
Update minimum supported R version to 3.3.0

jgutman:
### Intent
Addresses [rstudio-pro#2887](https://github.com/rstudio/rstudio-pro/issues/2887).

Update minimum supported R version to 3.3.0 in all documentation. Fatal error if using < 3.3.0

Accompanying PRs in [rstudio-pro##3322](https://github.com/rstudio/rstudio-pro/pull/3322), [docs.rstudio.com#820](https://github.com/rstudio/docs.rstudio.com/pull/820), and [rstudio.com#504](https://github.com/rstudio/rstudio.com/pull/504)

### Checklist

- [X] If this PR adds a new feature, or fixes a bug in a previously released version, it includes an entry in `NEWS.md` 
- [X] If this PR adds or changes UI, the updated UI meets [accessibility standards](https://github.com/rstudio/rstudio/wiki/Accessibility)
- [X] A reviewer is assigned to this PR (if unsure who to assign, check Area Owners list)
- [X] This PR passes all local unit tests

PR #10795, opened 2022-03-16
fix path to Windows install

gtritchie:
### Intent

The Windows dockerfile was trying to download `https://cran.rstudio.com/bin/windows/base/old/3.3.0/R-3.3.0-win.exe`. It does not exist. Change was made here: https://github.com/rstudio/rstudio/pull/10769

### Approach

Use `https://cran-archive.r-project.org/bin/windows/base/old/3.3.0/R-3.3.0-win.exe` instead.

### Automated Tests

Build-stuff

### QA Notes

Build-stuff

### Checklist


kevinushey:
Should we instead copy the R binary to our own build infrastructure (rather than downloading from `r-project.org` directly)? Note that the link no longer points to the RStudio CRAN mirror.

gtritchie:
> Should we instead copy the R binary to our own build infrastructure (rather than downloading from `r-project.org` directly)? Note that the link no longer points to the RStudio CRAN mirror.

Yes. Yes we should. I'll do that.

PR #10813, opened 2022-03-21
build fix, accomodate choco's removal of cpack alias

gtritchie:
The choco package manager used to create an alias named "cpack" which clashes with cmake's cpack, so we would delete the alias when setting up the machine.

Windows docker image creation failing because recent version of choco no longer create that alias.

Fix by appending `-ea ig` (shorthand for `-ErrorAction Ignore`) to the command. We could eventually remove this command altogether but let's leave in for the moment in case some machines still pull an older choco.


gtritchie:
The chocolately change is: https://github.com/chocolatey/choco/issues/89

Issue #9409, opened 2021-05-25
Make native pipe toggle a project-level setting


eliocamp:
It will be some time before most of the R community is updated to the new 4.1.0 version and so even if an user has R 4.1.0 and wants to use the native pipe for many applications, it's very likely that in some projects they might need to revert back to magrittr, such as code that needs to be compatible with older versions of R or collaborations with people running old versions. 

Right now the toggle to use magrittr or native pipe with Ctr + Shift + M is a global option, but for the reasons above, it might make sense being able to change it from the default on a per-project basis.

ronblum:
@eliocamp Thank you for the suggestion! We'll review it as we continue to develop RStudio.

colearendt:
Agree

ShixiangWang:
I also encounter such issue. Currently, the RStudio support setting native pipe in "Global Options", how about adding an option to "Project Options" to enable/disable native pipe operator?

jgutman:
One concern is that having both a project level option and a global level option for the pipe operator could lead to confusion, as we've with the tab width setting which is configurable in both settings. 

We could 1) remove the global option entirely and have it only be a project level option -- however that would leave no way to configure the operator outside of the context of a project

Or 2) change the global preference to have 3 options: always use the native pipe, never use the native pipe (always use magrittr) and add a third choice, use the native pipe only when running R 4.1+. This would seem to meet the use case of supporting projects that use older code -- if I switch the R version used by RStudio in order to work on a project that uses R 3.6, for example, my pipe operator would automatically switch to a magrittr pipe.

Does that meet the need?

eliocamp:
I have R 4.1+ installed, but I might be collaborating with some who doesn't, or developing to deploy code somewhere with an old version, or writing a package and trying to keep the R version dependency as low as possible. 

Why would there be confusion? Right now there are several options that can be set up both at the global and project level. Dictionary and saving / restoring .RData, for example.

ShixiangWang:
> One concern is that having both a project level option and a global level option for the pipe operator could lead to confusion, as we've with the tab width setting which is configurable in both settings.
> 
> We could 1) remove the global option entirely and have it only be a project level option -- however that would leave no way to configure the operator outside of the context of a project
> 
[...]

PR #10814, opened 2022-03-21
Make native pipe a project-specific option

jgutman:
### Intent

Address #9409

### Approach

Use of the native pipe operator should be an option on the Project Options level as well as the Global level, making it easier to disable the native pipe in projects that use older versions of R

To help user confusion around a project option that overrides a global option, we've added some text and a button to the global options editing pane, to remind the user that project options may conflict, and to take them over to project options when they hit the button.

### Automated Tests

None

### QA Notes

Set Global Option for "Use native pipe operator" to True via Options pane, set project option to False via Project options pane. Confirm user preferences are respected by entering 
Cmd + Shift + M in the console or editor (will insert magrittr or R native pipe depending on user preference)
Also run `.rs.readUiPref("insert_native_pipe_operator")` and `.rs.readUserPref("insert_native_pipe_operator")` to confirm

Be sure to play around with changing the Global option while project option is set to default, and changing the project option to override the project option. Option should only be written to Rproj file when project option is set to Yes or No, but not when set to default

Make sure changing the project option doesn't change the global option.

### Checklist


src/gwt/src/org/rstudio/studio/client/projects/StudioClientProjectConstants_fr.properties
```diff
@@ -139,6 +139,7 @@ chooseFileCaption=Choisir un fichier
 enableCodeIndexingLabel=Indexer les fichiers sources (pour la recherche/navigation dans le code)
 chkSpacesForTabLabel=Insérer des espaces pour les tabulations
 tabWidthLabel=Largeur de la tabulation
+insertNativePipeOperatorLabel=Utiliser l''opérateur pipe natif de R, |> PLEASE UPDATE
```
jgutman, line 142: This translation is missing the parenthetical (Requires R 4.1+) 
Could someone please add?

src/gwt/src/org/rstudio/studio/client/workbench/prefs/PrefsConstants_fr.properties
```diff
@@ -487,7 +487,7 @@ editingTabWidthLabel=Largeur de la tabulation
 editingAutoDetectIndentationLabel=Détection automatique de l''indentation dans les fichiers
 editingAutoDetectIndentationDesc=Lorsque cette option est activée, l''indentation des documents ne faisant pas partie d''un projet RStudio sera automatiquement détectée.
 editingInsertMatchingLabel=Insertion automatique des parenthèses et des crochets correspondants
-editingInsertNativePipeOperatorLabel=Utiliser l''opérateur pipe natif de R, |>
+editingInsertNativePipeOperatorLabel=Utiliser l''opérateur pipe natif de R, |> PLEASE UPDATE
```
jgutman, line 490: This translation is missing the parenthetical (Requires R 4.1+) 
Could someone please add?

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/AceEditor.java
```diff
@@ -737,6 +737,7 @@ void initialize(CodeToolsServerOperations server,
    {
       server_ = server;
       userPrefs_ = uiPrefs;
+
```
jgutman, line 740:
```suggestion
```

jmcphers:
It can be problematic to create a project pref that shadows a global pref. The issue this will create is that the global preference will still be visible (in the command palette and options dialog), but it will look like it isn't working because the project level preference is taking precedence. 

It is okay to make this a project specific option, but I think that if we do so we either need to remove the global one or have some mechanism that manages the relationship more explicitly.

jgutman:
> It is okay to make this a project specific option, but I think that if we do so we either need to remove the global one or have some mechanism that manages the relationship more explicitly.

Ah okay, is this not what we do with "Insert spaces for Tab" and "Tab width", which exist in both Global and Project options? I tried to follow the model we use there for those options, but maybe there is something managing the relationship for that preference? 

The downside with removing the global preference is that there's no way to set what the pipe operator should be when editing files outside of the context of a project -- also, I imagine this is may be a somewhat transitional project preference, becoming less useful once users don't have a mix of pre and post R 4.1 projects.

jmcphers:
> Ah okay, is this not what we do with "Insert spaces for Tab" and "Tab width", which exist in both Global and Project options? 

Yes! ... and that's exactly why I'm suggesting we may not want to do the same here. These options are a considerable source of confusion. It is not obvious at all where your tab width setting is coming from -- the project settings? global settings? auto detection? -- so users get confused (search the issues list for tab width...). 

If the goal is to handle both pre and post 4.1 projects, what about changing the option to be R version specific instead of project specific? In other words:

1. Remove the existing option
2. Add a new option with 3 values -- `always`, `R 4.1+`, and `never`
3. Use the computed pref layer to check the R version at startup and deliver the appropriate value

jgutman:
Thanks for the idea, that seems reasonable to me, but see discussion on #9409 (where the project setting was originally suggested) about whether or not that meets the need

rstudio.Rproj
```diff
@@ -7,6 +7,7 @@ AlwaysSaveHistory: No
 EnableCodeIndexing: Yes
 UseSpacesForTab: Yes
 NumSpacesForTab: 3
+InsertNativePipeOperator: No
```
kevinushey, line 10: Nit: Can we call this "use native pipe operator"?

kevinushey: Also, does this need to live within the `.Rproj` file? We also have the notion of project-specific preferences that _don't_ live in the project file; those are for project-local preferences that don't necessarily need to be "shared" by anyone else working in that project.

jgutman: Nope it doesn't, I just didn't think of that

jgutman: Actually, could you give me an example of a setting like this? I'm seeing several settings where (Default) inheriting the global option is a choice, and we don't write it to the r project file if it isn't set/different from the default, but it looked to me like we always write the option to the file when it differs from the default

jmcphers: See also: https://github.com/rstudio/rstudio/issues/7135

kevinushey: Some of the options in the Build Tools are an example here; for example, in Roxygen Options, for "Automatically roxygenize when running:":

Screen Shot 2022-03-22 at 10 54 08 AM

For example, the "Install and Restart" preference, which gets set here:

https://github.com/rstudio/rstudio/blob/56597a1893fa5dabcf9aae4c899f38eb39413e83/src/gwt/src/org/rstudio/studio/client/projects/ui/prefs/buildtools/BuildToolsRoxygenOptionsDialog.java#L53

Which is eventually set as part of this, in `auto_roxygenize_options`:

https://github.com/rstudio/rstudio/blob/56597a1893fa5dabcf9aae4c899f38eb39413e83/src/cpp/session/projects/SessionProjects.cpp#L422-L441

And so those are read from a separate project-specific options file:

https://github.com/rstudio/rstudio/blob/56597a1893fa5dabcf9aae4c899f38eb39413e83/src/cpp/session/projects/SessionProjectContext.cpp#L875-L878

And an example where that exists in my local `renv` project:

```
kevin@MBP-P2MQ:~/r/pkg/renv/.Rproj.user [main]
$ find . -name "build_options"
./142DE28A/build_options
./2FA46A95/build_options
```

See also:

https://github.com/rstudio/rstudio/blob/56597a1893fa5dabcf9aae4c899f38eb39413e83/src/cpp/session/projects/SessionProjects.cpp#L517-L541

So basically, project options which might want to be "shared" by any user who is working in that project will become part of the project file, but "user-specific project options" might use their own separate project-local storage.

jgutman: Ah I see. It looks like the VCS options are another example that have their own separate storage, rather than being written to the RProj file. I started down this route, but because there's not an existing appropriate project-local place to write this to (neither build options nor vcs options would be appropriate) I'd have to create my own, and having started doing so it seems like a lot more work than just writing it to the RProj file. 

Is there a significant disadvantage to writing it to the RProj file? While there's not a strict need to share among collaborators on a project, having an inconsistent mix of the different pipe types throughout the code doesn't necessarily seem ideal either (just as you'd prefer not to have a mix of different tab widths between files for collaborators on a project)

kevinushey: In this case, I think it might be fine. It's unlikely that multiple collaborators on the same project would want to use different default pipe styles in one project.

jgutman:
> 3\. Use the computed pref layer to check the R version at startup and deliver the appropriate value

@jmcphers @kevinushey One issue this runs into again is difficulty handling when the R version changes after RStudio startup (user changes R version and restarts the R session, without re-opening RStudio). I'm not sure exactly if the computed pref layer gives us a way around this, but my initial stab at implementing this doesn't handle these cases where the r Version changes

jmcphers:
> I'm not sure exactly if the computed pref layer gives us a way around this

I _think_ it should. Prefs are computed at session start and the new prefs (incl. computed) delivered to the client as part of session initialization.

src/gwt/src/org/rstudio/studio/client/workbench/prefs/PrefsConstants_en.properties
```diff
@@ -236,7 +236,7 @@ buttonDownLabel=Down
 developingPkgHelpLink=Developing Packages
 secondaryReposDialog=Retrieving list of secondary repositories...
 validateSyncLabel=Please select or input a CRAN repository
-showErrorRepoMessage=The repository 
+showErrorRepoMessage=The repository
```
jgutman, line 239:
```suggestion
showErrorRepoMessage=The repository 
```

src/gwt/src/org/rstudio/studio/client/workbench/prefs/PrefsConstants_en.properties
```diff
@@ -317,20 +317,20 @@ helpLinkCaption=Learn more about Quarto
 rMarkdownHeaderLabel=R Markdown
 rMarkdownShowLabel=Show document outline by default
 rMarkdownSoftWrapLabel=Soft-wrap R Markdown files
-docOutlineDisplayLabel=Show in document outline: 
+docOutlineDisplayLabel=Show in document outline:
 docOutlineSectionsOption=Sections Only
 docOutlineSectionsNamedChunksOption=Sections and Named Chunks
 docOutlineSectionsAllChunksOption=Sections and All Chunks
-rmdViewerModeLabel=Show output preview in: 
+rmdViewerModeLabel=Show output preview in:
```
jgutman, line 324:
```suggestion
rmdViewerModeLabel=Show output preview in: 
```

src/gwt/src/org/rstudio/studio/client/workbench/prefs/PrefsConstants_en.properties
```diff
@@ -317,20 +317,20 @@ helpLinkCaption=Learn more about Quarto
...
 rmdViewerModeWindowOption=Window
 rmdViewerModeViewerPaneOption=Viewer Pane
 rmdViewerModeNoneOption=(None)
 rmdInlineOutputLabel=Show output inline for all R Markdown documents
-latexPreviewWidgetLabel=Show equation and image previews: 
+latexPreviewWidgetLabel=Show equation and image previews:
```
jgutman, line 329:
```suggestion
latexPreviewWidgetLabel=Show equation and image previews: 
```

src/gwt/src/org/rstudio/studio/client/workbench/prefs/PrefsConstants_en.properties
```diff
@@ -317,20 +317,20 @@ helpLinkCaption=Learn more about Quarto
...
 latexPreviewWidgetNeverOption=Never
 latexPreviewWidgetPopupOption=In a popup
 latexPreviewWidgetInlineOption=Inline
-knitWorkingDirLabel=Evaluate chunks in directory: 
+knitWorkingDirLabel=Evaluate chunks in directory:
```
jgutman, line 333:
```suggestion
knitWorkingDirLabel=Evaluate chunks in directory: 
```

src/gwt/src/org/rstudio/studio/client/workbench/prefs/PrefsConstants_en.properties
```diff
@@ -351,11 +351,11 @@ visualEditorShowMarginLabel=Show margin column indicator in code blocks
...
 visualModeFontSizeLabel=Editor font size:
 visualModeOptionsMarkdownCaption=Markdown
-visualModeListSpacingLabel=Default spacing between list items: 
-visualModeWrapLabel=Automatic text wrapping (line breaks): 
+visualModeListSpacingLabel=Default spacing between list items:
+visualModeWrapLabel=Automatic text wrapping (line breaks):
 visualModeWrapHelpLabel=Learn more about automatic line wrapping
 visualModeOptionsLabel=Wrap at column:
-visualModeReferencesLabel=Write references at end of current: 
+visualModeReferencesLabel=Write references at end of current:
```
jgutman, line 358:
```suggestion
visualModeReferencesLabel=Write references at end of current: 
```

src/gwt/src/org/rstudio/studio/client/workbench/prefs/PrefsConstants_en.properties
```diff
@@ -391,7 +391,7 @@ ignoreWordsUppercaseLabel=Ignore words in UPPERCASE
 ignoreWordsNumbersLabel=Ignore words with numbers
 checkingHeader=Checking
 realTimeSpellcheckingCheckboxLabel=Use real time spell-checking
-kUserDictionaryLabel=User dictionary: 
+kUserDictionaryLabel=User dictionary:
```
jgutman, line 394:
```suggestion
kUserDictionaryLabel=User dictionary: 
```

src/gwt/src/org/rstudio/studio/client/projects/StudioClientProjectConstants_en.properties
```diff
@@ -103,7 +103,7 @@ projectTypeBook=Book
 engineLabel=Engine:
 engineSelectNone=(None)
 kernelLabel=Kernel:
-chkUseVenvLabel=Use venv with packages: 
+chkUseVenvLabel=Use venv with packages:
```
jgutman, line 106:
```suggestion
chkUseVenvLabel=Use venv with packages: 
```

src/gwt/src/org/rstudio/studio/client/projects/StudioClientProjectConstants_en.properties
```diff
@@ -189,7 +193,7 @@ visualModeZoteroCaption=Visual Mode: Zotero
 rMarkdownText=R Markdown
 sharingText=Sharing
 vcsSelectLabel=Version control system:
-originLabel=Origin: 
+originLabel=Origin:
```
jgutman, line 196:
```suggestion
originLabel=Origin: 
```

src/gwt/src/org/rstudio/studio/client/projects/StudioClientProjectConstants_en.properties
```diff
@@ -145,6 +145,10 @@ chooseFileCaption=Choose File
 enableCodeIndexingLabel=Index source files (for code search/navigation)
 chkSpacesForTabLabel=Insert spaces for tab
 tabWidthLabel=Tab width
+useNativePipeOperatorLabel=Use native pipe operator, |> (requires R 4.1+)
+useNativePipeOperatorNeverText=Never (use magrittr pipe)
+useNativePipeOperatorR41Text=Only when using R 4.1+
+useNativePipeOperatorAlwaysText=Always
```
romainfrancois, line 151: in 🇫🇷: 

```
useNativePipeOperatorLabel=Utiliser le pipe natif, |> (nécessite R 4.1+)
useNativePipeOperatorNeverText=Jamais (utiliser le pipe de magrittr)
useNativePipeOperatorR41Text=Seulement avec R 4.1+
useNativePipeOperatorAlwaysText=Toujours
```

romainfrancois: I don't think we translate "pipe" in 🇫🇷, but we just pronounce it in English. cc @cderv

src/gwt/src/org/rstudio/studio/client/workbench/prefs/PrefsConstants_en.properties
```diff
@@ -487,7 +487,7 @@ editingTabWidthLabel=Tab width
 editingAutoDetectIndentationLabel=Auto-detect code indentation
 editingAutoDetectIndentationDesc=When enabled, the indentation for documents not part of an RStudio project will be automatically detected.
 editingInsertMatchingLabel=Insert matching parens/quotes
-editingInsertNativePipeOperatorLabel=Use native pipe operator, |> (requires R 4.1+)
+editingUseNativePipeOperatorLabel=Use native pipe operator, |> (requires R 4.1+)
```
romainfrancois, line 490: 🇫🇷 

```
editingUseNativePipeOperatorLabel=Utiliser le pipe natif, |> (nécessite R 4.1+)
```

src/gwt/src/org/rstudio/studio/client/projects/StudioClientProjectConstants_en.properties
romainfrancois: There is this in `CmdConstants_fr.properties` : 

```
# insertPipeOperator
insertPipeOperatorLabel = Insérer un opérateur de tube
```

but I don't think I've heard people use `tube`.

src/gwt/src/org/rstudio/studio/client/workbench/prefs/model/UserPrefsAccessorConstants_en.properties
```diff
@@ -934,9 +934,12 @@ pythonPathDescription = The path to the default Python interpreter.
...
 
-# Whether the Insert Pipe Operator command should insert the native R pipe operator, |>
-insertNativePipeOperatorTitle = Use R's native pipe operator, |>
-insertNativePipeOperatorDescription = Whether the Insert Pipe Operator command should insert the native R pipe operator, |>
+# Whether the Insert Pipe Operator command should use the native R pipe operator, |>
+useNativePipeOperatorTitle = Use R's native pipe operator, |>
+useNativePipeOperatorDescription = Whether the Insert Pipe Operator command should use the native R pipe operator, |>
+useNativePipeOperatorEnum_never=Never (use magrittr pipe)
+useNativePipeOperatorEnum_only_R41=Only when using R 4.1+
+useNativePipeOperatorEnum_always=Always
```
romainfrancois, line 942: 🇫🇷 

```
useNativePipeOperatorTitle = Utiliser le pipe natif, |>
 useNativePipeOperatorDescription = Indique si la commande 'Insérer un opérateur de tube' utilise le pipe natif, |>
 useNativePipeOperatorEnum_never=Jamais (utilise le pipe de magrittr)
 useNativePipeOperatorEnum_only_R41=Seulement avec R 4.1+
 useNativePipeOperatorEnum_always=Toujours
```

Same comment as before, I don't think "tube" is being used

rstudio.Rproj
```diff
@@ -7,6 +7,7 @@ AlwaysSaveHistory: No
 EnableCodeIndexing: Yes
 UseSpacesForTab: Yes
 NumSpacesForTab: 3
+UseNativePipeOperator: Never
```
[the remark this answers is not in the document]
romainfrancois: btw, I've been wondering the same thing about `PackageCleanBeforeInstall: Yes`

src/gwt/src/org/rstudio/studio/client/workbench/prefs/model/UserPrefsAccessorConstants_en.properties
romainfrancois: But maybe this is out of context for this pr.

jgutman: Thanks! Apparently we can leave the French translations unedited, and leave them for the Stats Canada folks to translate, but since you've already provided the translation I will update the file to use it. Merci!

(Also thanks for the note about `tube` vs. `pipe` -- I'm assuming someone provided the literal translation previously but I will trust your experience of what is more commonly used

src/gwt/src/org/rstudio/studio/client/projects/StudioClientProjectConstants_en.properties
cderv: I agree @romainfrancois 

I don't think this would be `tube` - which is literally translation for pipe when doing plumbing. 
I would stick with `pipe` or use longer form  `l'opérateur pipe` to add context as done in `CmdConstants_fr.properties`.

jgutman: Thanks for confirming @cderv, updating the translations now

src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/EditingPreferencesPane.java
```diff
@@ -121,6 +125,26 @@ public void onClick(ClickEvent event)
       lessSpaced(keyboardPanel);
       editingPanel.add(keyboardPanel);
 
+      HorizontalPanel projectPrefsPanel = new HorizontalPanel();
```
jgutman, line 128: Screenshot:

image

Please note that Edit Project Preferences button does nothing when no project is activated. We should probably make this panel (text + button) or at least the button invisible when no project is activated

src/gwt/src/org/rstudio/studio/client/projects/ui/prefs/ProjectEditingPreferencesPane.java
```diff
@@ -15,13 +15,12 @@
...
 import com.google.gwt.core.client.GWT;
+import com.google.gwt.user.client.ui.Label;
 import org.rstudio.core.client.ElementIds;
 import org.rstudio.core.client.prefs.PreferencesDialogBaseResources;
 import org.rstudio.core.client.prefs.RestartRequirement;
 import org.rstudio.core.client.resources.ImageResource2x;
-import org.rstudio.core.client.widget.NumericValueWidget;
-import org.rstudio.core.client.widget.OperationWithInput;
-import org.rstudio.core.client.widget.TextBoxWithButton;
+import org.rstudio.core.client.widget.*;
```
jgutman, line 23: Oops -- this is CLion being a little overzealous with the automatic imports, I can narrow this down

src/gwt/src/org/rstudio/studio/client/workbench/prefs/PrefsConstants.java
```diff
@@ -4519,16 +4521,34 @@ public interface PrefsConstants extends com.google.gwt.i18n.client.Messages {
...
+    String editingProjectOverrideInfoText();
+
+    /**
+     * Translated "Edit Project Preferences..."
+     *
+     * @return translated "Edit Project Preferences..."
+     */
+    @DefaultMessage("Edit Project Preferences...")
+    @Key("editProjectPreferencesButtonLabel")
+    String editProjectPreferencesButtonLabel();
```
jgutman, line 4550: These two are missing French translations, but I think that's fine for now

src/gwt/src/org/rstudio/studio/client/projects/ui/prefs/ProjectEditingPreferencesPane.java
```diff
@@ -15,13 +15,17 @@
...
 import org.rstudio.core.client.prefs.PreferencesDialogBaseResources;
 import org.rstudio.core.client.prefs.RestartRequirement;
 import org.rstudio.core.client.resources.ImageResource2x;
+
+import org.rstudio.core.client.widget.FormLabel;
+import org.rstudio.core.client.widget.LayoutGrid;
 import org.rstudio.core.client.widget.NumericValueWidget;
-import org.rstudio.core.client.widget.OperationWithInput;
 import org.rstudio.core.client.widget.TextBoxWithButton;
+import org.rstudio.core.client.widget.OperationWithInput;
```
jgutman, line 28: I'm assuming we prefer listing all these out explicitly, instead of `widget.*`

jgutman:
Sorry for the really bloated PR -- however I think this is finally ready for review!

rstudio.Rproj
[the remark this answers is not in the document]
jgutman: Yes, thanks for the suggestion! You absolutely can have implicit options that are only written to the RProj file if other than the default. I've now made the default "(Default)" as in, inherit the global setting, and if so it will not be written to the RProj file

src/cpp/session/resources/schema/user-prefs-schema.json
```diff
@@ -1535,18 +1535,18 @@
             "title": "Save Retry Timeout",
             "description": "The maximum amount of seconds of retry for save operations."
         },
-        "insert_native_pipe_operator": {
-            "description": "Whether the Insert Pipe Operator command should insert the native R pipe operator, |>",
+        "use_native_pipe_operator": {
```
timtmok, line 1538: @kevinushey @jgutman Since this preference key changed, do we care about a migration if it has been set?

src/gwt/src/org/rstudio/studio/client/projects/ui/prefs/ProjectEditingPreferencesPane.java
kevinushey: Indeed, that is correct.

src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/EditingPreferencesPane.java
kevinushey: > We should probably make this panel (text + button) or at least the button invisible when no project is activated

I agree this should only be shown if the user is currently working within a project.

src/gwt/src/org/rstudio/studio/client/projects/ui/prefs/ProjectEditingPreferencesPane.java
```diff
@@ -58,6 +60,13 @@ public ProjectEditingPreferencesPane(final SourceServerOperations server)
       numSpacesForTab_.setWidth("36px");
       add(numSpacesForTab_);
 
+      LayoutGrid useNativePipeLabeled = new LayoutGrid(1, 2);
```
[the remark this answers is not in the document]
kevinushey: Can we tweak the text a bit?

- Use italics
- Add a bit of extra padding at the bottom
- Include a period at the end of the sentence

src/cpp/session/resources/schema/user-prefs-schema.json
jgutman: I didn't even think of that -- this name didn't really need to be changed, but since we wanted to use "UseNativePipeOperator" for the project setting I thought it would be easier to just be consistent across the code. I can change it back if it's problematic. I don't know how we do migrations though

src/gwt/src/org/rstudio/studio/client/projects/ui/prefs/ProjectEditingPreferencesPane.java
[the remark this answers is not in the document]
jgutman: @kevinushey Do you mean like just put "(requires R 4.1+)" in italics?

Because I can do that for the global preferences which uses a Checkbox

image

but I don't know how to do that for the Project Preferences, as I don't think the FormLabel widget has an `asHtml` param or anything like that
image

If you meant to set the whole line in italics, I could do that in the CSS, but that seems strange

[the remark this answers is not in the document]
kevinushey: Sorry, I was referring to the "Use (default) to inherit the global default setting" text; I don't think we need to tweak the other text.

timtmok (approved):
Looks good with the preference part!

src/gwt/src/org/rstudio/studio/client/projects/ui/prefs/ProjectEditingPreferencesPane.java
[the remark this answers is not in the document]
jgutman: Ah, that makes more sense and you're right -- we already have a CSS style for that bit, I just forgot to apply it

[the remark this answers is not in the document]
jgutman: Better?

image

Issue #10731, opened 2022-03-03
Run Electron Unit Tests with Windows Build


gtritchie:
Work item: Windows variant of #10182

gtritchie:
These are now running and passing.

![Screen Shot 2022-03-22 at 12 41 38](https://user-images.githubusercontent.com/10569626/159552959-2b14b7af-ab43-4202-a08e-b43bee82bf45.png)

PR #10818, opened 2022-03-21
Run electron unit tests as part of the Windows build

gtritchie:
### Intent

Run Electron unit tests on the Windows builder.

Addresses: #10731 

### Approach

Electron (and thus `electron-mocha`) crashes in a Windows Docker container if you try to create a BrowserWindow (even a hidden one).

Until we find a workaround (if that's even possible; I suspect it isn't), I've disabled the broken tests when running inside Docker on Windows.

Testing the Windows classes via unit tests feels wrong, anyways. Probably better covered by integration tests in the long run. Though we likely won't be able to run integration tests on Windows via docker, either.

Also tweaked tests that assumed creating `C:\Program Files\a_test_folder` would fail; that's not the case in a Windows docker container.

### Automated Tests

That's what this is.

### QA Notes

Nothing to test.

### Checklist


src/node/desktop/scripts/run-unit-tests.cmd
```diff
@@ -0,0 +1,22 @@
+::
```
kevinushey, line 1: TIL you can use `::` instead of `REM` for comments.

kevinushey (approved):
LGTM

PR #10830, opened 2022-03-23
Fix group member verification when group member lookup is disabled

### LINKED ISSUE #10829: Group membership verification may fail when group membership lookups are disabled
Calls to `core::system::userBelongsToGroup` may fail (return false even when the user is a member of the target group) in certain environments where `ignore_group_members = True` in `sssd`. This setup is sometimes used when the identity provider has a large number of groups and membership assignments.

We should modify our group membership verification to first use [getgrouplist(3)](https://man7.org/linux/man-pages/man3/getgrouplist.3.html) and then fall back to the current implementation which uses [getgrnam_r(3)](https://linux.die.net/man/3/getgrnam_r)

dbkegley:
### Intent

fixes #10829  

### Approach

First we try to use getgrouplist(3) to determine if one of the
user's groups is the target group. If not, then we fall back
to the legacy implementation, using getgrnam_r(3) to attempt
a group member lookup.

### Automated Tests

Covered by existing tests

### QA Notes

### Checklist

### DIFF
--- a/src/cpp/core/system/PosixSystem.cpp
+++ b/src/cpp/core/system/PosixSystem.cpp
@@ -2398,14 +2398,13 @@ bool isUserNotFoundError(const Error& error)
    return error == systemError(boost::system::errc::permission_denied, ErrorLocation());
 }
 
-Error userBelongsToGroup(const User& user,
-                         const std::string& groupName,
-                         bool* pBelongs)
+Error userBelongsToGroupViaGroupList(const User& user,
+                                     const std::string& groupName,
+                                     bool* pBelongs)
 {
    *pBelongs = false; // default to not found
 
-   // first we try to use getgrouplist(3) to determine if the user is a
-   // member of the target groupName
+   // get a list of the user's groups
    std::vector<group::Group> groups;
    Error error = group::userGroups(user.getUsername(), &groups);
    if (error)
@@ -2417,14 +2416,18 @@ Error userBelongsToGroup(const User& user,
                             [&](const group::Group& group)
                             { return group.name == groupName; }) != groups.end();
 
-   // if the user belongs, then we can return early
-   // otherwise, we fallback to the legacy implementation
-   if (pBelongs)
-      return Success();
+   return Success();
+}
 
-   // next we try to lookup the target group using getgrnam_r(3)
+Error userBelongsToGroupViaGroupName(const User& user,
+                                     const std::string& groupName,
+                                     bool* pBelongs)
+{
+   *pBelongs = false; // default to not found
+
+   // fetch the group by its name
    group::Group group;
-   error = group::groupFromName(groupName, &group);
+   Error error = group::groupFromName(groupName, &group);
    if (error)
       return error;
 
@@ -2445,6 +2448,42 @@ Error userBelongsToGroup(const User& user,
 }
 
 
+// first try to get the user's groups via getgrouplist. we prefer this to
+// fetching information about the group to test membership because
+// this can fail in the case where SSSD group member data is disabled
+// via ignore_group_members (see https://jhrozek.fedorapeople.org/sssd/1.9.91/man/sssd.conf.5.html)
+//
+// we fallback to the old implementation if this method did not yield a positive result
+// to ensure backwards compatibility in existing customer environments. however,
+// group member enumeration of large groups is expensive and this has been known
+// to be slow in some environments.
+//
+// if the first try fails via getgrouplist, we hold the error and attempt
+// the old implementation before returning. if both calls fail, we return the first error
+Error userBelongsToGroup(const User& user,
+                         const std::string& groupName,
+                         bool* pBelongs)
+{
+   *pBelongs = false; // default to not found
+
+   // first we try to use the getgrouplist(3) implementation to determine
+   // if the user is a member of the target groupName.
+   Error groupListError = userBelongsToGroupViaGroupList(user, groupName, pBelongs);
+
+   // if the user belongs, then we can return early.
+   // otherwise, we fallback to the legacy implementation
+   if (!groupListError)
+      return Success();
+
+   // next we try to lookup the target group using the getgrnam_r(3) implementation
+   Error groupNameError = userBelongsToGroupViaGroupName(user, groupName, pBelongs);
+
+   // if both attempts fail, return the first error we encountered
+   if (groupNameError)
+      return groupListError;
+
+   return Success();
+}
 
 /////////////////////////////////////////////////////////////////////
 //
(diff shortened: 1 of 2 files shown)


kfeinauer (commented):
I am personally happy with the implementation, I think everything looks good as written. I would just indicate in the comments why we are using this new behavior and falling back on old behavior, something like:

```
// first try to get the user's groups via getgrouplist. we prefer this to
// fetching information about the group to test membership because
// this can fail in the case where SSSD group member data is disabled
// via ignore_group_members (see https://jhrozek.fedorapeople.org/sssd/1.9.91/man/sssd.conf.5.html)

// we fallback to the old implementation if this method did not yield a positive result
// to ensure backwards compatibility in existing customer environments. however,
// group member enumeration of large groups is expensive and this has been known
// to be slow in some environments.

```

Long term we should consider if we need to fall back to the legacy implementation at all. @jjallaire wrote the initial implementation so he may have more insight, though it's been 10 years at this point.

src/cpp/core/system/PosixSystem.cpp
```diff
@@ -2403,8 +2403,28 @@ Error userBelongsToGroup(const User& user,
...
+
+   // scan the user's groups to see if they are a member of the target group
+   *pBelongs = std::find_if(groups.begin(),
+                            groups.end(),
+                            [&](const group::Group& group)
+                            { return group.name == groupName; }) != groups.end();
+
+   // if the user belongs, then we can return early
+   // otherwise, we fallback to the legacy implementation
+   if (pBelongs)
```
kevinushey, line 2422:
```suggestion
   if (*pBelongs)
```

Presumedly we want to check the pointer value rather than the pointer itself?

src/cpp/core/system/PosixSystem.cpp
```diff
@@ -2403,8 +2403,28 @@ Error userBelongsToGroup(const User& user,
                          bool* pBelongs)
 {
    *pBelongs = false; // default to not found
+
+   // first we try to use getgrouplist(3) to determine if the user is a
+   // member of the target groupName
+   std::vector<group::Group> groups;
+   Error error = group::userGroups(user.getUsername(), &groups);
+   if (error)
+      return error;
```
kevinushey, line 2412: Is it possible that getgrouplist(3) could fail, but the following getgrnam_r(3) call could succeed? If so, it seems to me like we should save this error and return it later only if the following attempt to retrieve group names fails.

It might also be worthwhile to have two separate functions which are called + used here, e.g. `userBelongsToGroupViaGroupList()` and `userBelongsToGroupViaGroupName()`.

jeffvroom:
This may or may not be helpful for this case, but we are in the midst of helping a Workbench customer that also has `ignore_group_members: true` in the ldap configuration. For this customer, the rserver gets a "Permission denied" error when running: `group::userGroups(username, ..)`  The call to `getgrouplist` successfully returns the list of GIDs, but on at least one id `getgrgid_r` returns the permission denied error trying to convert it to a name.  Hopefully you don't see that problem but thought I'd mention it. I think group::userGroups could be more resilient, like maybe it returns a group without a name or creates a dummy name.  It's the GID that's most important for the NFS permissions.

kfeinauer:
@jeffvroom Great point. I was worried that was a potential issue, but had never heard of that being an issue so interesting that you are working on a support issue right now that's hitting that.

We may want to create a separate issue to make the check lenient outside of the scope of this PR. @dbkegley has definitely already gone further than we thought he'd need to to implement this feature :laughing:

dbkegley:
I worked with Cole to get a repro of this environment with ldap/sssd and we were able to verify that this fix works when `ignore_group_members = true`

```
root@ldap-rsp:/test# TARGET_USER=bobo TARGET_GROUP=engineering_group ./debug-sandbox-old.sh id
+ set -o errexit
+ set -o pipefail
+ echo '{
  "context": {
    "username": "bobo",
    "project": "",
    "id": ""
  },
  "password": "",
  "executablePath": "",
  "config": {
    "args": [],
    "environment": {},
    "stdInput": "",
    "stdStreamBehavior": 2,
    "priority": 1,
    "memoryLimitBytes": 0.0,
    "stackLimitBytes": 0.0,
    "userProcessesLimit": 0.0,
    "cpuLimit": 0.0,
    "niceLimit": 0.0,
    "filesLimit": 0.0,
    "cpuAffinity": []
  }
}'
+ /test/rsandbox-old -W / -P su -U bobo -G engineering_group id
2022-03-25T18:10:26.080875Z [rsandbox] ERROR system error 13 (Permission denied); OCCURRED AT rstudio::core::Error rstudio::core::system::temporarilyDropPriv(const string&, const string&, bool) src/cpp/core/system/PosixSystem.cpp:2469; LOGGED FROM: int main(int, char* const*) src/cpp/server/sandbox/SandboxMain.cpp:206


root@ldap-rsp:/test# TARGET_USER=bobo TARGET_GROUP=engineering_group ./debug-sandbox.sh id
+ set -o errexit
+ set -o pipefail
+ echo '{
  "context": {
    "username": "bobo",
    "project": "",
    "id": ""
  },
  "password": "",
  "executablePath": "",
  "config": {
    "args": [],
    "environment": {},
    "stdInput": "",
    "stdStreamBehavior": 2,
    "priority": 1,
    "memoryLimitBytes": 0.0,
    "stackLimitBytes": 0.0,
    "userProcessesLimit": 0.0,
    "cpuLimit": 0.0,
    "niceLimit": 0.0,
    "filesLimit": 0.0,
    "cpuAffinity": []
  }
}'
+ /test/rsandbox -W / -P su -U bobo -G engineering_group id
uid=1002(bobo) gid=501(engineering_group) groups=501(engineering_group),500(rstudio-connect),502(finance_group)
```

dbkegley:
@kfeinauer @MariaSemple Just giving this PR a bump. Anything I can do to help get this merged?

src/cpp/core/system/PosixSystem.cpp
```diff
@@ -2425,6 +2448,42 @@ Error userBelongsToGroup(const User& user,
...
+                         bool* pBelongs)
+{
+   *pBelongs = false; // default to not found
+
+   // first we try to use the getgrouplist(3) implementation to determine
+   // if the user is a member of the target groupName.
+   Error groupListError = userBelongsToGroupViaGroupList(user, groupName, pBelongs);
+
+   // if the user belongs, then we can return early.
+   // otherwise, we fallback to the legacy implementation
```
kfeinauer, line 2474: Minor nit here on the comment: we are returning only if there was no error, not if necessarily if the user belongs to the group. And I think that's the right approach, just need to update the comment.

kfeinauer (approved):
LGTM aside from minor comment nit. An IDE team member (either @kevinushey or @MariaSemple) should give you the permission to merge here.

kevinushey (approved):
LGTM!

MariaSemple (approved):
LGTM!

PR #10838, opened 2022-03-24
Updated icon with transparent background, rounded square edges

jgutman:
Addresses #8395 , attempt #2 (previous PR was merged and reverted)

Got an updated icon from @gregswinehart, this one matches the style and size of the new Mac 11 icons much more closely.

Here we've gone for the trademarked R blue circle, but against a rounded square background with slight drop shadow around the edges

See: 
image
image (1)

Compared to the old icon, which had no background and was overly large:
image (2)


GCRev:
I'm sorry I can't re-approve of this since it has already been merged, but LGTM!

Thanks for including screenies! We even did a better job fitting in with the other apple logos than Zoom.

cbrnr:
Should we see the new icon in the current release? I tried 2022.02.3 Build 492, but it still has the old icon. There seem to have been four releases after this got merged, but I only get the new icon in daily builds.

jgutman:
No, the new icon was only added for the upcoming release which is expected in July. We did not add it for our patch releases, such as 2022.02.3

Issue #10844, opened 2022-03-25
argument help popup does not display with R 4.2.0


kevinushey:
The following error is printed in the console.

> 2022-03-25T18:33:11.596020Z [rserver] ERROR CLIENT EXCEPTION (rsession-kevin): Unexpected help format, no argblock table found;|||rstudio-0.js@8#559::Throwable_4_g$|||rstudio-0.js@18#3070::Error_3_g$|||rstudio-0.js@14#3133::new AssertionError_7_g$|||rstudio-0.js@10#2174::makeAssertionError_Object_0_g$|||rstudio-0.js@23#470267::parseArguments_0_g$|||rstudio-0.js@9#470239::parse_0_g$|||rstudio-0.js@17#429948::onResponseReceived_345_g$|||rstudio-0.js@8#429934::onResponseReceived_344_g$|||rstudio-0.js@36#328440::onResponseReceived_222_g$|||rstudio-0.js@36#328498::onResponseReceived_223_g$|||rstudio-0.js@38#135415::onResponseReceived_6_g$|||rstudio-0.js@17#43353::fireOnResponseReceived_0_g$|||rstudio-0.js@28#43702::onReadyStateChange_0_g$|||rstudio-0.js@18#94072::<anonymous>|||rstudio-0.js@28#24557::apply_0_g$|||rstudio-0.js@16#24616::entry0_0_g$|||rstudio-0.js@14#24595::<anonymous>|||Client-ID: a61434ec-5d99-4360-925d-307080067334|||User-Agent: Mozilla/5.0 (Macintosh  Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/99.0.4844.83 Safari/537.36

jgutman:
We had some difficulty reproducing the originally reported bug, but tested the current behavior on:

RStudio Desktop 2022.02.2-485, Mac OS 12.3.1
RStudio Desktop 2022.06.0-322, Mac OS 12.3.1
RStudio Desktop 2022.06.0-322, Windows 11

and all verified to be working as expected.

PR #10845, opened 2022-03-25
be more permissive when finding the R help argument table

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/10844.

### Approach

With the new enhanced help system for R 4.2.0, the argument table is no longer given the attribute `summary="R argblock"`. However, because we know that the argument table must come after the "Arguments:" header, it's safe to just iterate forward over siblings and use the first table we find.

### Automated Tests

Should already be covered by existing Help pane tests.

### QA Notes

Check that the Help popup tooltip displays R help appropriately.

Screen Shot 2022-03-25 at 11 37 15 AM

### Checklist

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/help/model/HelpInfo.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/help/model/HelpInfo.java
@@ -18,6 +18,7 @@ import com.google.gwt.core.client.JavaScriptObject;
 import com.google.gwt.core.client.JsArrayString;
 import com.google.gwt.dom.client.*;
 
+import org.rstudio.core.client.StringUtil;
 import org.rstudio.core.client.dom.DomUtils;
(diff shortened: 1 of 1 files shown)


kevinushey:
It looks like this change will be necessary to support R 4.2.0 in PT, so I've rebased to target there.

jgutman:
I'm taking a look at testing this, and I've seen the help error come up in my session logs once or twice, but I'm not seeing an error when looking at the tooltip for `stats::rn` in R 4.2+ on older versions of PT or SW. Any advice on reproducing the original error so I can verify this is fixed?

kevinushey:
Weird. I can't reproduce anymore with the latest R 4.3.0 development builds either. :-/

- Maybe this only reproduces in development builds of RStudio? I'm not sure if our Java asserts still fire in a package build.
- Maybe something changed in R 4.3.0 to fix this from when I last tested?

Issue #10484, opened 2022-02-01
Cleanup UI Zooming in Electron Desktop


gtritchie:
Cleanup and finish implementing the zoom feature. It seems to be working in general, but review and resolve the following TODOs to ensure it is fully implemented.

Be sure to test via keyboard shortcuts, command-palette, and via invocation of the View / Actual Size / Zoom In / Zoom Out menu.

Also look at persistence of the zoom level. Currently it is being saved (somewhere...) but in a different place than Qt desktop. Decide if we care about zoom level staying the same if you are going between Qt-based and Electron-based RStudio.

Note, currently relies on Electron's menuItemRoles versus the explicit command-handling we do in the Qt desktop. That might be ok (don't quite recall the gory details 7 months later).

- https://github.com/rstudio/rstudio/blob/e869a097415c5e8ded835bdd70a6eae465d4f226/src/node/desktop/src/main/desktop-browser-window.ts#L194-L197
- https://github.com/rstudio/rstudio/blob/e869a097415c5e8ded835bdd70a6eae465d4f226/src/node/desktop/src/main/main-window.ts#L111-L114
- https://github.com/rstudio/rstudio/blob/e869a097415c5e8ded835bdd70a6eae465d4f226/src/node/desktop/src/main/menu-callback.ts#L302-L323

Also including these TODOs related to dpi as part of this:

- https://github.com/rstudio/rstudio/blob/e869a097415c5e8ded835bdd70a6eae465d4f226/src/node/desktop/src/main/utils.ts#L48-L51
- https://github.com/rstudio/rstudio/blob/e869a097415c5e8ded835bdd70a6eae465d4f226/src/node/desktop/src/main/utils.ts#L344-L349
- https://github.com/rstudio/rstudio/blob/b9cfb0cead45532141eb11bb772ff501a047bd51/src/node/desktop/src/main/gwt-callback.ts#L286-L289

matheustavaresdev:
After many testing and code walkthrough, here's my outcomes on this:

## 1) Image comparison
- **Default ( 0 ) Zoom**
  - Electron 
![electron_0](https://user-images.githubusercontent.com/36674530/154419211-15a2e820-66fa-4c62-8511-ead9d7e00d78.png)
  - QT
  ![QT](https://user-images.githubusercontent.com/36674530/154419236-0e963b73-e6e6-4ccd-a96e-170dcc6eaf6c.png)
  
[...]

PR #10846, opened 2022-03-25
Electron preference defaults and legacy priority

timtmok:
### Intent
Address #10484

This sets the preference priority to Electron-store > legacy Qt > default value for zoom level, fixed font, and R path. All other preferences can follow these as examples for preference priority. There is now a generated type for the user state based on `user-state-schema.json`. The file is also included during the webpack so that default values are taken from there. Hard-coded defaults have now been removed.

There is a new step in `postinstall` to generate the type. So running `npm ci` or `npm install` will generate the necessary type for the user state schema. The default values are pulled in at runtime and directly read from `user-state-schema.json`.

### Approach
The default values are taken directly from `user-state-schema.json` now and there is type validation when setting values. An error is returned if the value is invalid.

The `rBinDir` looks to have stored the selected R version for Windows. So this is now used for `rstudiopath` and will read the legacy value if available. There is a difference where the legacy value does not include the executable so that is checked before using the file path. There's also a mixture of how the path separator is defined (`\\` vs. `/`) but that still works. It's caused by the legacy value using `\\` and Electron using `/`.

I tried to set a webpack alias for importing `user-state-schema.json` but `mocha` couldn't understand the alias. There are some packages available to handle translating the file path but have since been deprecated and/or don't work with Mocha 9.x.

Any issues with VS Code intellisense with the json values might require restarting the TS server using the command palette (or restarting VS Code).

### Automated Tests
There are unit tests for being able to set the value and get a legacy value if that's all that is available.

### QA Notes
MacOS is the only vastly different platform that stored legacy values in Mac defaults. For the most part, legacy values can be set using a PT build. Clearing these places of Electron user state can trigger reading the legacy values:

Windows: `C:\Users\<user>\AppData\Roaming\RStudio\config.json`
MacOS: `/Users/<user>/Library/Application Support/RStudio/config.json`
Linux: `/home/<user>/.config/RStudio/config.json`

### Checklist

### DIFF
--- a/src/cpp/session/resources/schema/user-state-schema.json
+++ b/src/cpp/session/resources/schema/user-state-schema.json
@@ -13,10 +13,6 @@
                     "items": {
                         "type": "string"
                     }
-                },
-                "clipboardMonitoring": {
-                    "type": "boolean",
-                    "description": "Linux clipboard monitoring for middle-click paste"
                 }
             },
             "default": {
@@ -32,16 +28,12 @@
                 },
                 "fixedWidthFont": {
                     "type": "string"
-                },
-                "useFontConfigDb": {
-                    "type": "boolean"
                 }
             },
             "description": "Font options from the Appearance category",
             "default": {
                 "proportionalFont": "",
-                "fixedWidthFont": "",
-                "useFontConfigDb": true
+                "fixedWidthFont": ""
             }
         },
         "view": {
--- a/src/cpp/session/resources/schema/user-state-schema.json
+++ b/src/cpp/session/resources/schema/user-state-schema.json
@@ -4,6 +4,139 @@
     "title": "RStudio User State",
     "type": "object",
     "properties": {
+        "general": {
+            "type": "object",
+            "properties": {
+                "ignoredUpdateVersions": {
+                    "type": "array",
+                    "description": "A list of ignored updates. Ignored updates will not trigger an available update.",
+                    "items": {
+                        "type": "string"
+                    }
+                },
+                "clipboardMonitoring": {
+                    "type": "boolean",
+                    "description": "Linux clipboard monitoring for middle-click paste"
+                }
+            },
+            "default": {
+                "ignoredUpdateVersions": [],
+                "clipboardMonitoring": true
+            }
+        },
+        "font": {
+            "type": "object",
+            "properties": {
+                "proportionalFont": {
+                    "type": "string"
+                },
+                "fixedWidthFont": {
+                    "type": "string"
+                },
+                "useFontConfigDb": {
+                    "type": "boolean"
+                }
+            },
+            "description": "Font options from the Appearance category",
+            "default": {
+                "proportionalFont": "",
+                "fixedWidthFont": "",
+                "useFontConfigDb": true
+            }
+        },
+        "view": {
+            "type": "object",
+            "properties": {
+                "zoomLevel": {
+                    "type": "number",
+                    "description": "UI zoom level, 1.0 == 100%",
+                    "enum": [0.25, 0.5, 0.75, 0.8, 0.9, 1.0, 1.1, 1.25, 1.5, 1.75, 2.0, 2.5, 3.0, 4.0, 5.0]
+                },
+                "windowBounds": {
+                    "type": "object",
+                    "description": "Window location and size",
+                    "properties": {
+                        "width": { "type": "number" },
+                        "height": { "type": "number" },
+                        "x": { "type": "number" },
+                        "y": { "type": "number" }
+                    }
+                },
+                "accessibility": {
+                    "type": "boolean",
+                    "description": "Screen reader support" 
+                }
+            },
+            "default": {
+                "zoomLevel": 1.5,
+                "windowBounds": { "width": 1200, "height": 900, "x": 0, "y": 0 },
+                "accessibility": false
+            }
+        },
+        "session": {
+            "type": "object",
+            "properties": {
+                "lastRemoteSessionUrl": {
+                    "type": "string"
+                },
+                "authCookies": {
+                    "type": "array",
+                    "items": {
+                        "type": "string"
+                    }
+                },
+                "tempAuthCookies": {
+                    "type": "array",
+                    "items": {
+                        "type": "string"
+                    }
+                }
+            },
+            "default": {
+                "lastRemoteSessionUrl": "",
+                "authCookies": [],
+                "tempAuthCookies": []
+            }
+        },
+        "renderer": {
+            "type": "object",
+            "properties": {
+                "engine": {
+                    "type": "string",
+                    "enum": ["auto", "desktop", "gles"]
+                },
+                "useGpuExclusionList": {
+                    "type": "boolean"
+                },
+                "useGpuDriverBugWorkarounds": {
+                    "type": "boolean"
+                }
+            },
+            "default": {
+                "engine": "auto",
+                "useGpuExclusionList": true,
+                "useGpuDriverBugWorkarounds": true
+            }
+        },
+        "platform": {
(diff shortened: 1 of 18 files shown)


src/cpp/session/resources/schema/user-state-schema.json
```diff
@@ -4,6 +4,139 @@
...
+            "type": "object",
+            "properties": {
+                "ignoredUpdateVersions": {
+                    "type": "array",
+                    "description": "A list of ignored updates. Ignored updates will not trigger an available update.",
+                    "items": {
+                        "type": "string"
+                    }
+                },
+                "clipboardMonitoring": {
```
kevinushey, line 17: FWIW this preference might not be necessary anymore with Electron.

src/cpp/session/resources/schema/user-state-schema.json
```diff
@@ -4,6 +4,139 @@
...
+        "font": {
+            "type": "object",
+            "properties": {
+                "proportionalFont": {
+                    "type": "string"
+                },
+                "fixedWidthFont": {
+                    "type": "string"
+                },
+                "useFontConfigDb": {
```
kevinushey, line 36: Do you know if we still need this?

src/node/desktop/src/main/detect-r.ts
```diff
@@ -295,7 +295,7 @@ function isValidInstallationWin32(installPath: string): boolean {
 
 function findDefaultInstallPathWin32(version: string): string {
   // query registry for R install path
-  const keyName = `HKEY_LOCAL_MACHINE\\SOFTWARE\\R-Core\\${version}`;
+  const keyName = `HKEY_LOCAL_MACHINE\\SOFTWARE\\R-Core\\R\\${version}`;
```
kevinushey, line 298: This change looks incorrect to me, given how this function is used:

```
$ rg findDefaultInstallPath
src/node/desktop/src/main/detect-r.ts
296:function findDefaultInstallPathWin32(version: string): string {
329:    const x64InstallPath = findDefaultInstallPathWin32('R64');
338:  const i386InstallPath = findDefaultInstallPathWin32('R');
351:  return findDefaultInstallPathWin32('R');
355:  return findDefaultInstallPathWin32('R64');
```

Note that the caller is responsible for passing `R` or `R64`, since it might depend on whether we're enumerating 32-bit versions of R versus 64-bit versions of R. (The "version" parameter here probably makes this unclear.)

src/cpp/session/resources/schema/user-state-schema.json
```diff
@@ -4,6 +4,139 @@
...
+                "useFontConfigDb": true
+            }
+        },
+        "view": {
+            "type": "object",
+            "properties": {
+                "zoomLevel": {
+                    "type": "number",
+                    "description": "UI zoom level, 1.0 == 100%",
+                    "enum": [0.25, 0.5, 0.75, 0.8, 0.9, 1.0, 1.1, 1.25, 1.5, 1.75, 2.0, 2.5, 3.0, 4.0, 5.0]
```
kevinushey, line 53: Is this a good opportunity to allow arbitrary zoom levels? (That is, we could just enforce that this is a number.)

kevinushey: Or does it need to be this way to support the "legacy" options?

src/node/desktop/src/main/detect-r.ts
kevinushey: Also note that the registry keys `R-core/R` and `R-core/R/3.6.3` both contain InstallPath strings; the first one is just the "default" value associated with the most recently installed copy of R.

src/cpp/session/resources/schema/user-state-schema.json
timtmok: Good point, I'll take another look at how it's used and see about cleaning this up.

src/cpp/session/resources/schema/user-state-schema.json
timtmok: I'll take it out since the font DB isn't necessary now but I also wasn't sure what performance would be like when you mentioned that some users have a large number of fonts installed. I suppose the font DB wouldn't be the first choice when dealing with that performance. If there were a performance problem with a lot of fonts, I would probably refactor the call to retrieve the fonts asynchronously and caching it instead of building the cache at startup.

src/cpp/session/resources/schema/user-state-schema.json
timtmok: Arbitrary zoom levels is possible and would be neat to have a zoom slider instead of the presets. It only has to be enforced for legacy options so I think it would be a requirement for Qt. Only validating for type number should be fine since the legacy value is not overwritten so a user could downgrade RStudio and go back to whatever settings they had.

Electron only needs a value greater than 0. Looks like Qt can only be 0.25 - 5.0. So I think keeping the minimum and maximum but allowing arbitrary values in between makes a lot of sense.

src/node/desktop/src/main/detect-r.ts
kevinushey: Interesting... I see some registry keys in both locations as well. I tried re-installing some versions of R, and I'm now always seeing the registry keys updated in the WOW632Node folder, whereas previously some were updated in the parent directory. Did something change in the latest builds of Windows? This seems a little surprising to me.

kevinushey: I did a brief comparison of where the registry entries seem to be getting updated:

---

### HKEY_LOCAL_MACHINE\SOFTWARE\R-Core

- R 3.3.3

### HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\R-Core

- R 3.6.3
- R 4.1.3
- R 4.2.0 (Pre-release)

---

Even so, RStudio is still able to find all of these installations...

Screen Shot 2022-03-28 at 10 06 17 AM

... however, if I'm running from a command prompt, I need to use `/reg:32` as you said.

```
C:\Users\kevin>reg query HKEY_LOCAL_MACHINE\SOFTWARE\R-Core\R /reg:32

HKEY_LOCAL_MACHINE\SOFTWARE\R-Core\R
    InstallPath    REG_SZ    C:\R\R-3.3.3
    Current Version    REG_SZ    3.3.3

HKEY_LOCAL_MACHINE\SOFTWARE\R-Core\R\3.3.3
HKEY_LOCAL_MACHINE\SOFTWARE\R-Core\R\3.6.3
HKEY_LOCAL_MACHINE\SOFTWARE\R-Core\R\4.1.3
HKEY_LOCAL_MACHINE\SOFTWARE\R-Core\R\4.2.0 Pre-release

C:\Users\kevin>reg query HKEY_LOCAL_MACHINE\SOFTWARE\R-Core\R

HKEY_LOCAL_MACHINE\SOFTWARE\R-Core\R
    InstallPath    REG_SZ    C:\R\R-3.3.3
    Current Version    REG_SZ    3.3.3

HKEY_LOCAL_MACHINE\SOFTWARE\R-Core\R\3.3.3
```

For posterity, this is with Windows 11 on ARM, running through Parallels on my M1 macOS machine. This seems worth testing on native hardware...

timtmok: Yes, it would be great to test natively. I can only get it to work (Win 11 ARM, Parallels) if the `/reg:32` switch is set and that seems wrong. I also had issues with detecting the architecture. Despite this being an arm64 install, `process.architecture` returns x64.

timtmok:
I'll merge given that Windows ARM is not a supported platform. Maybe if one of us can figure out how to detect Windows ARM properly, we could do some sort of easy workaround to query the R installations. That would be helpful for those of us with Windows ARM VM's.

Issue #7992, opened 2020-10-07
Exporting from viewer onto clipboard provides a nonsensical screenshot


kiernang:
### System details

    RStudio Edition : Desktop
    RStudio Version : Version 1.3.1093
    OS Version      : Windows 10
    R Version       :  4.02

### Steps to reproduce the problem
Using a data.frame and the gt package (as well as dplyr):
```
example_df<- data.frame(a = c('This', 'Does', 'Not', 'Work'), b = c(0,0,0,0))
example_table <- example_df %>% gt () %>%
  tab_header(title = "Broken Export") %>% opt_row_striping(row_striping = TRUE) %>%
  tab_source_note(source_note = "This does not export properly")

example_table
```

From there, click 'Export' and then 'Copy to clipboard'. When you paste the picture, the picture will include your code.

### Describe the problem in detail
Trying to export a table from the viewer to an external report. When the table is copied to the clipboard, it is not the table, but instead a screenshot of the table as well as the code in the background. See attached picture.
![image](https://user-images.githubusercontent.com/66456229/95353807-fd861580-08c3-11eb-93a9-a0cd540c0119.png)

### Describe the behavior you expected
Expected just the table. Similar to below image.
![image](https://user-images.githubusercontent.com/66456229/95354251-7b4a2100-08c4-11eb-9f1d-37b1d563cab6.png)

- [ x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ x] If I am reporting a RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [ x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
@kiernang Thank you for raising the issue! We'll review it as a possible fix in the upcoming 1.4 Wax Begonia release.

mikebessuille:
Consider for 1.5.  Need QA to reproduce this; does it happen all the time for copy to clipboard?

kiernang:
Happens all the time.

ronblum:
I can reproduce this on MacOS 11 Big Sur beta 10, as well. Note that this was the behavior in 1.2.5042, as well.

iago-pssjd:
[...]

PR #10858, opened 2022-03-29
Copy viewer plot to clipboard

timtmok:
### Intent
Address #7992

Implements the feature for Electron. This fixes the export to clipboard issue present in Qt builds when the zoom level has been changed.

### Approach
From GWT, it can still do the same thing and the values can be used with Electron's API to call `capturePage`. The preview can be resized and the zoom level can be adjusted without affecting what is copied to the clipboard.

### Automated Tests
None

### QA Notes
I discovered that in #7992 that the issue occurs when changing the zoom level before exporting the plot. This does not fix the issue in Qt but the Electron implementation fixes the issue.

Sample code to generate something in the Viewer pane:
```R
library(gt)
library(tidyverse)
library(glue)

# Define the start and end dates for the data range
start_date <- "2010-03-07"
end_date <- "2010-06-14"

# Create a gt table based on preprocessed
# `sp500` table data
sp500 %>%
  filter(date >= start_date & date <= end_date) %>%
  select(-adj_close) %>%
  gt() %>%
  tab_header(
    title = "S&P 500",
    subtitle = glue("{start_date} to {end_date}")
  ) %>%
  fmt_date(
    columns = date,
    date_style = 3
  ) %>%
  fmt_currency(
    columns = c(open, high, low, close),
    currency = "USD"
  ) %>%
  fmt_number(
    columns = volume,
    suffixing = TRUE
  )
```

Generates an image like:
image

### Checklist

### DIFF
--- a/src/node/desktop/src/main/gwt-callback.ts
+++ b/src/node/desktop/src/main/gwt-callback.ts
@@ -15,7 +15,18 @@
 
 // TODO clean this up
 
-import { app, BrowserWindow, clipboard, dialog, ipcMain, screen, shell, webContents, webFrameMain } from 'electron';
+import {
+  app,
+  BrowserWindow,
+  clipboard,
+  dialog,
+  ipcMain,
+  Rectangle,
+  screen,
+  shell,
+  webContents,
+  webFrameMain,
+} from 'electron';
 import { IpcMainEvent, MessageBoxOptions, OpenDialogOptions, SaveDialogOptions } from 'electron/main';
 import EventEmitter from 'events';
 import { mkdtempSync, writeFileSync } from 'fs';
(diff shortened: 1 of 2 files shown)


kevinushey (approved):
LGTM! Very nice to see that we can make use of existing Electron APIs here :)

src/node/desktop/src/main/gwt-callback.ts
```diff
@@ -411,10 +422,18 @@ export class GwtCallback extends EventEmitter {
       GwtCallback.unimpl('desktop_close_named_window');
     });
 
-    ipcMain.on(
+    ipcMain.handle(
```
kevinushey, line 425: As an aside, should we prefer using `handle` to `on` in general, or was there another motivation for this change?

timtmok: I believe `handle` returns a promise. Using `on` requires passing a callback or firing the message with `sendSync`. Either can work to get the main process to respond to the event. I think `handle` is appropriate here to separate the main process doing the actual work from the renderer that is passing along the event from GWT.

PR #10864, opened 2022-03-29
Require rmarkdown for rmd and qmd if any R code chunks are found

jgutman:
### Intent

Addresses [pro#3310](https://github.com/rstudio/rstudio-pro/issues/3310) -- only scan quarto docs for R package dependencies if any R code chunks are found, and if so, always add rmarkdown to the list of required packages so users can get the package install prompt. Follow up to #10767.

### Approach

We have an extractRcode function, but it was a little broken. We use this to extract any R code chunks from a document (specifically Rmd and qmd, for .R files we extract the entire file) and since we're already checking the file contents for mentions of any packages we need to be installed, we add rmarkdown to the list -- but only if the R content is non empty.
That way users get the package install prompt if they have at least 1 R chunk, but if they have no code chunks, or only python chunks, we won't check their R package dependency situation.

### Automated Tests

None

### QA Notes

`remove.packages("rmarkdown")` for testing
Check a quarto doc with at least 1 R code chunk, and check a quarto doc with no R code chunks (may have python chunks for example)

Python only
image

R only
image

### Checklist

### DIFF
--- a/src/cpp/session/modules/SessionCodeTools.R
+++ b/src/cpp/session/modules/SessionCodeTools.R
@@ -2229,7 +2229,7 @@
       contents
    else if (identical(extension, ".Rmd"))
       .rs.extractRCode(contents,
-                       "^[\t >]*```+\\s*\\{([Rr]+.*)\\}\\s*$",
+                       "^[\t >]*```+\\s*\\{[Rr]\\b.*\\}\\s*$",
                        "^[\t >]*```+\\s*$")
    else if (identical(extension, ".Rnw"))
       .rs.extractRCode(contents,
--- a/src/cpp/session/modules/SessionCodeTools.R
+++ b/src/cpp/session/modules/SessionCodeTools.R
@@ -2237,7 +2237,7 @@
                        "^\\s*@\\s*(%+.*|)$")
    else if (identical(extension, ".qmd"))
      .rs.extractRCode(contents,
-                      "^[\t >]*```+\\s*\\{([Rr]+.*)\\}\\s*$",
+                      "^[\t >]*```+\\s*\\{[Rr]\\b.*\\}\\s*$",
                       "^[\t >]*```+\\s*$")
    if (is.null(code) || identical(code, .rs.scalar("")))
       return(character())
--- a/src/cpp/session/modules/SessionCodeTools.R
+++ b/src/cpp/session/modules/SessionCodeTools.R
@@ -56,7 +56,7 @@
 {
    if (!length(string))
       string <- ""
-   
+
    n <- nchar(string)
    (nchar(strings) >= n) & (substring(strings, 1, n) == string)
 })
@@ -70,10 +70,10 @@
 {
    if (!length(string))
       string <- ""
-   
+
    nstrings <- nchar(strings)
    nstring <- nchar(string)
-   (nstrings >= nstring) & 
+   (nstrings >= nstring) &
       (substring(strings, nstrings - nstring + 1, nstrings) == string)
 })
 
@@ -159,7 +159,7 @@
 {
    envir <- .rs.resolveEnvironment(envir)
    parsed <- parse(text = name)[[1L]]
-   
+
    # for plain symbols, we can attempt lookup of a function object
    # directly; for more complex calls, we try to evaluate it in
    # the requested environment and hope we got a function
@@ -167,10 +167,10 @@
       get(name, envir = envir, mode = "function")
    else
       eval(parsed, envir = envir)
-   
+
    if (!is.function(fn))
       return(NULL)
-   
+
    fn
 })
 
@@ -208,7 +208,7 @@
      control <- append(control, "useSource")
 
    code <- deparse(func, width.cutoff = 59, control = control)
-   
+
    # if we were asked not to use source refs, or we were but there wasn't a
    # source ref to use, then format the code according to user pref
    if (!useSource || is.null(attr(func, "srcref", exact = TRUE)))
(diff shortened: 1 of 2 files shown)


src/cpp/session/modules/SessionCodeTools.R
```diff
@@ -2222,29 +2222,30 @@
...
    else if (identical(extension, ".qmd"))
      .rs.extractRCode(contents,
-                      "^[\t >]*```+\\s*\\{([a-zA-Z0-9_]+.*)\\}\\s*$",
+                      "^[\t >]*```+\\s*\\{([Rr]+.*)\\}\\s*$",
                       "^[\t >]*```+\\s*$")
-   if (is.null(code))
+   if (is.null(code) || identical(code, .rs.scalar("")))
       return(character())
-   
+
```
jgutman, line 2244: Sorry for the IDE whitespace changes -- this is the relevant section of this file.

src/cpp/session/modules/SessionCodeTools.R
```diff
@@ -2441,27 +2442,33 @@
...
    # parse to find packages
    packages <- .rs.parsePackageDependencies(contents, extension)
-   
+   # return early if no R code is parsed
+   if (length(packages) == 0)
+      return(character())
+
+   # If file is qmd or rmd, and there is R code, rmarkdown is REQUIRED
+   if (identical(extension, ".Rmd") || identical(extension, ".qmd"))
+      packages <- c("rmarkdown", packages)
```
jgutman, line 2459: Here is the other relevant set of changes

src/cpp/session/modules/SessionCodeTools.R
```diff
@@ -2222,29 +2222,30 @@
...
-   # package without the user's consent
+   # They were modified to search only R code chunks and skip other chunks
+   # such as python, bash, sql, etc that do not contain R dependencies
+   # Inline R code in markdown docs is also skipped
    code <- if (identical(extension, ".R"))
       contents
    else if (identical(extension, ".Rmd"))
       .rs.extractRCode(contents,
-                       "^[\t >]*```+\\s*\\{([a-zA-Z0-9_]+.*)\\}\\s*$",
+                       "^[\t >]*```+\\s*\\{([Rr]+.*)\\}\\s*$",
```
kevinushey, line 2232: Nit: this would capture an engine called Rcpp, so I think we need something a bit more restrictive, e.g.

```suggestion
                       "^[\t >]*```+\\s*\\{[Rr]\\b.*\\}\\s*$",
```

src/cpp/session/modules/SessionCodeTools.R
```diff
@@ -2222,29 +2222,30 @@
...
                        "^[\t >]*```+\\s*$")
    else if (identical(extension, ".Rnw"))
       .rs.extractRCode(contents,
                        "^\\s*<<(.*)>>=.*$",
                        "^\\s*@\\s*(%+.*|)$")
    else if (identical(extension, ".qmd"))
      .rs.extractRCode(contents,
-                      "^[\t >]*```+\\s*\\{([a-zA-Z0-9_]+.*)\\}\\s*$",
+                      "^[\t >]*```+\\s*\\{([Rr]+.*)\\}\\s*$",
```
kevinushey, line 2240: Same as above?

src/cpp/session/modules/SessionCodeTools.R
kevinushey, line 2459: Nit: only add rmarkdown if it's not already a dependency? (For chunks that might be using `library(rmarkdown)` already)

src/cpp/session/modules/SessionCodeTools.R
jgutman: Great point!

src/cpp/session/modules/SessionCodeTools.R
jgutman: I think this is not going to make a difference, because ultimately we're just adding rmarkdown to the list of known dependencies before doing `missing <- setdiff(packages, installed)` at which point it doesn't matter if rmarkdown shows up once or more than once in packages, it will show up 0 or 1 times in missing depending on whether it's installed or not. I think filtering it out before the set different is just a redundant extra step

src/cpp/session/modules/SessionCodeTools.R
kevinushey:
```suggestion
                      "^[\t >]*```+\\s*\\{[Rr]\\b+.*\\}\\s*$",
```

jgutman:
Thanks @kevinushey !

src/cpp/session/modules/SessionDiagnostics.R
```diff
@@ -110,46 +110,46 @@
...
+
       # If the chunk end lies immediately after the chunk start, bail
       if (start + 1 == end)
          next
-      
+
       new[(start + 1):(end - 1)] <- splat[(start + 1):(end - 1)]
    }
-   
+   new <- new[new != ""]
```
kevinushey, line 152: This line might be the culprit in https://github.com/rstudio/rstudio/issues/10984, since we rely on the blank lines being preserved so that markers are rendered in the right location.

jgutman: Yes, that was the intent of that PR -- I just tagged for review before I got a chance to actually mention the issue number. Just removing this line on its own won't work, but that paired with a small tweak to SessionCodeTools.R fixes both the original issue and the regression

PR #10880, opened 2022-03-31
Electron: Include font styles for Mac

timtmok:
### Intent
Address #10438

Uses the postscript name on Mac to select a specific font style.

image

### Approach
This seems to work well for MacOS. Linux and Windows only seems to take the font family name as a valid way to specify the font. Actually, Linux allows with certain fonts to append the style to the end of the family name. So `Fira Code Light` would work. However, `Fira Code Bold` is invalid (not sure why or what is different about that font file). There are likely further improvements that could be made in this area. It's probably better to port the old Qt code to get that specific querying logic as `node-system-fonts` sets up different parameters for `font-config`.

### Automated Tests
None

### QA Notes
There should be a _much_ longer font list for Mac now. Linux is still missing font styles and Windows already has a full list of fonts with styles.

### Checklist


gtritchie (approved):
LGTM!

kevinushey:
Belated LGTM!

Issue #10488, opened 2022-02-02
Windows Server user was given access to the RStudio Desktop session of another user


WetRobot:
### System details

    RStudio Edition : Desktop
    RStudio Version : RStudio 2021.09.0+351 "Ghost Orchid" Release (077589bcad3467ae79f318afe8641a1899a51606, 2021-09-20) for Windows Mozilla/5.0 (Windows NT 6.3; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) QtWebEngine/5.12.8 Chrome/69.0.3497.128 Safari/537.36
    OS Version      : Microsoft Windows Server 2012 R2 Standard
    R Version       : 4.1.1 (2021-08-10) -- "Kick Things"

### Steps to reproduce the problem

I am unable to give steps to reproduce this because of its random nature. In many years of using RStudio Desktop on a Windows Server with the projects on a network drive, the issue has only occurred once. The issue may potentially be reproduced by forcing two RStudio Desktop sessions to use the same port, which I at least was unable to do.

#### What we know

Users A and B logged onto Windows Server 2012. User A had a running RStudio Desktop session from the previous day. User B is uncertain whether they had a previous session open or whether they opened a new RStudio Desktop session upon log-in. User A had their project on a network drive. User A saw upon log-in that Rstudio Desktop session had complained of failed file monitoring (related: https://github.com/rstudio/rstudio/issues/2414):

```
Warning message:
File monitoring failed for project at "X:/path/to/network/drive/location"
Error 2 (The system cannot find the file specified)
Features disabled: R source file indexing, Diagnostics
```

RStudio Desktop worked normally for user A. However, the RStudio Desktop session for user B was that of user A: User B was able to see the session of user A and even to run code in it. Our IT has confirmed that Windows processes themselves were not shared. We have not encountered an issue like this with any other software.

#### Speculation 

[...]

PR #10885, opened 2022-04-01
reduce chance of collisions when generating port for desktop <-> rsession communication

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/10488. Candidate for backport.

### Approach

- Adjust the random seed used by our RNG, so that the process PID is included. This should help reduce the chance of collisions when multiple RStudio sessions are launched at the same time.

- Use a UUID for the desktop shared secret, rather than just the system RNG.

- Validate that the port we've selected is actually open via TCP, and use it only if it appears open.

### Automated Tests

Not included; this fix is hard to test via automation.

### QA Notes

On Windows, in a command prompt, try launching RStudio in the following way:

```
REM Change path below with wherever you have RStudio installed.
cd "C:\Program Files\RStudio\bin"  
for %i in (1 2 3) do start /b rstudio.exe
```

You should see 3 separate instances of RStudio launched. Try to execute some code in one of those sessions, and confirm that output is seen only in that 1 session, and not in the other 2 sessions.

### Checklist

### DIFF
--- a/src/cpp/core/include/core/Random.hpp
+++ b/src/cpp/core/include/core/Random.hpp
@@ -16,22 +16,11 @@
 #ifndef CORE_RANDOM_HPP
 #define CORE_RANDOM_HPP
 
-#ifndef _WIN32
-# include <unistd.h>
-#else
-# include <process.h>
-#endif
-
-#include <ctime>
-#include <functional>
 #include <limits>
-#include <random>
 
 #include <boost/chrono.hpp>
 #include <boost/random.hpp>
 #include <boost/thread.hpp>
-
-#include <core/system/System.hpp>
 
 namespace rstudio {
 namespace core {
--- a/src/cpp/core/include/core/Random.hpp
+++ b/src/cpp/core/include/core/Random.hpp
@@ -38,9 +38,9 @@ T uniformRandomInteger(
       T minValue = std::numeric_limits<T>::min(),
       T maxValue = std::numeric_limits<T>::max())
 {
-   boost::random::mt19937 rng(getpid() + std::time(NULL));
-   boost::random::uniform_int_distribution<> generator(minValue, maxValue);
-   return generator(rng);
+   boost::random::mt19937 gen(getpid() + std::time(NULL));
+   boost::random::uniform_int_distribution<> dist(minValue, maxValue);
+   return dist(gen);
 }
 } // namespace random
 } // namespace core
--- a/src/cpp/core/include/core/Random.hpp
+++ b/src/cpp/core/include/core/Random.hpp
@@ -22,10 +22,14 @@
 # include <process.h>
 #endif
 
-#include <limits>
 #include <ctime>
+#include <functional>
+#include <limits>
+#include <random>
 
+#include <boost/chrono.hpp>
 #include <boost/random.hpp>
+#include <boost/thread.hpp>
 
 #include <core/system/System.hpp>
 
@@ -38,10 +42,17 @@ T uniformRandomInteger(
       T minValue = std::numeric_limits<T>::min(),
       T maxValue = std::numeric_limits<T>::max())
 {
-   boost::random::mt19937 gen(getpid() + std::time(NULL));
+   // choose random seed
+   long long seed = boost::chrono::high_resolution_clock::now().time_since_epoch().count();
+
+   // construct our generator
+   boost::random::mt19937 gen(seed);
    boost::random::uniform_int_distribution<> dist(minValue, maxValue);
+
+   // generate a value
    return dist(gen);
 }
+
 } // namespace random
 } // namespace core
 } // namespace rstudio
--- a/src/cpp/core/include/core/Random.hpp
+++ b/src/cpp/core/include/core/Random.hpp
@@ -23,6 +23,7 @@
 #endif
 
 #include <limits>
+#include <ctime>
 
 #include <boost/random.hpp>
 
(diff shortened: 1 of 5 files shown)


src/cpp/core/include/core/http/TcpIpSocketUtils.hpp
```diff
@@ -89,7 +89,10 @@ inline Error initTcpIpAcceptor(
    acceptor.open(endpoint.protocol(), ec);
    if (ec)
       return Error(ec, ERROR_LOCATION);
-   
+
+   // TODO: is this the desired behavior on Windows and desktop?
+   // this could allow multiple clients to try connecting to the
+   // same session
    acceptor.set_option(tcp::acceptor::reuse_address(true), ec);
```
kevinushey, line 96: @jeffvroom I'd be curious to know your opinion here. Windows has [SO_EXCLUSIVEADDRUSE](https://docs.microsoft.com/en-us/windows/win32/winsock/so-exclusiveaddruse) to disallow multiple applications to bind to the same port, which I _think_ we want here... 

My understanding is that the `reuse_address()` option maps to the `SO_REUSEADDR` socket option, which (for Linux) has documentation, emphasis mine:

> SO_REUSEADDR
>               Indicates that the rules used in validating addresses
>               supplied in a [bind(2)](https://man7.org/linux/man-pages/man2/bind.2.html) call should allow reuse of local
>               addresses.  For AF_INET sockets this means that a socket
>               may bind, **except when there is an active listening socket
>               bound to the address**.  When the listening socket is bound
>               to INADDR_ANY with a specific port then it is not possible
>               to bind to this port for any local address.  Argument is
>               an integer boolean flag.

But on Windows, it means something different (https://docs.microsoft.com/en-us/windows/win32/winsock/using-so-reuseaddr-and-so-exclusiveaddruse):

> The SO_REUSEADDR socket option **allows a socket to forcibly bind to a port in use by another socket**. The second socket calls [setsockopt](https://docs.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-setsockopt) with the optname parameter set to SO_REUSEADDR and the optval parameter set to a boolean value of TRUE before calling [bind](https://docs.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-bind) on the same port as the original socket. Once the second socket has successfully bound, the behavior for all sockets bound to that port is indeterminate. For example, if all of the sockets on the same port provide TCP service, any incoming TCP connection requests over the port cannot be guaranteed to be handled by the correct socket — the behavior is non-deterministic. A malicious program can use SO_REUSEADDR to forcibly bind sockets already in use for standard network protocol services in order to deny access to those service. No special privileges are required to use this option.

It seems like we'd want the behavior to be uniform here, and so on Windows we either don't want to set this option, or we want to use `SO_EXCLUSIVEADDRUSE` to avoid multiple applications binding on the same port.

src/cpp/core/include/core/Random.hpp
```diff
@@ -16,33 +16,32 @@
...
-  DistributionType distribution(std::numeric_limits<T>::min(),
-                                std::numeric_limits<T>::max());
-
-  // create variate generator
-  boost::variate_generator<GeneratorType, DistributionType> vg(generator,
-                                                               distribution);
-
-  // return random number
-  return vg();
+   boost::random::mt19937 gen(getpid() + std::time(NULL));
```
kevinushey, line 41: We include `::getpid()` here just to provide a little bit of extra certainty that the starting seed should be different. This should be sufficient for desktop, at least -- and shouldn't hurt for server configurations (although this API doesn't appear to be used outside of desktop)

src/cpp/desktop/DesktopMain.cpp
```diff
@@ -117,9 +117,7 @@ Error removeStaleOptionsLockfile()
 
 void initializeSharedSecret()
 {
-   sharedSecret = QString::number(rand())
-                  + QString::number(rand())
-                  + QString::number(rand());
+   sharedSecret = QString::fromStdString(core::system::generateUuid());
```
kevinushey, line 120: On Windows, I was seeing `rand()` producing identical values in different machines when launched at the same time; this should be altogether safer.

src/cpp/core/include/core/http/TcpIpSocketUtils.hpp
jeffvroom: Nice catch on the meaning of those flags. I totally agree... we should be using SO_EXCLUSIVEADDRUSE on windows and not reuse_address() in boost (that maps to SO_REUSEADDR even on windows).  I found this code snippet:

typedef boost::asio::detail::socket_option::boolean<BOOST_ASIO_OS_DEF(SOL_SOCKET), SO_EXCLUSIVEADDRUSE> excluse_address;  acceptor.set_option(excluse_address(true));

in a comment on this page:

https://stackoverflow.com/questions/7164879/boost-asio-why-dont-i-get-bind-address-already-in-use-in-windows-but-do-ge

src/cpp/core/include/core/Random.hpp
jeffvroom: I recognize this particular use case doesn't have to be "secure" in the sense that port is not guessable, we are just trying to avoid collissions. But since this is in our core utils, how about if we use: std::chrono::high_resolution_clock::now().time_since_epoch().count() (which is "clock ticks") in place of std::time(NULL), that I think is in seconds.  I don't see any other uses of this util that require "secure random" - if so, we'd have to do a lot more work!

jeffvroom (approved):
This looks fantastic!

kevinushey:
@atocharnaud, let's continue discussion in https://github.com/rstudio/rstudio/issues/10488.

Issue #10871, opened 2022-03-30
"State not recoverable" error when creating new Quarto doc (Electron)


gtritchie:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2022.06.0-daily+247
    OS Version      : macOS 12.3
    R Version       : 4.1.3

### Steps to reproduce the problem

- Start RStudio Desktop (Electron)
- Enable Quarto preview in Global Options / R Markdown / Quarto, restart if necessary
- File / New File / Quarto Document...

### Describe the problem in detail

- after a delay, get error reading "State not recoverable"
- no document is created

![Screen Shot showing error dialog](https://user-images.githubusercontent.com/10569626/160919910-b56c396d-eee8-4591-b209-cc475e4b2387.png)

### Describe the behavior you expected

- new Quarto document; works in the Qt-based desktop build of same version

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

timtmok:
Seems to only occur in a packaged build. The error:
`/Applications/RStudio Electron.app/Contents/Resources/app/bin/quarto/bin/quarto: line 77: cd: /Applications/RStudio Electron.app/Contents/Resources/app/bin/quarto/bin/../share: No such file or directory
/Applications/RStudio Electron.app/Contents/Resources/app/bin/quarto/bin/quarto: line 101: /Applications/RStudio: No such file or directory`

I also get the error in Qt. It's the wrong path to the Quarto executable.

timtmok:
I added some logging to debug and realized I was getting an error due to my rename of the app to include a space.

Here's the actual error:
`/Applications/RStudioElectron.app/Contents/Resources/app/bin/quarto/bin/quarto: line 77: cd: /Applications/RStudioElectron.app/Contents/Resources/app/bin/quarto/bin/../share: No such file or directory
[91mERROR: NotFound: No such file or directory (os error 2), readdir '/Users/tmok/source/test/formats/html/bootstrap/themes'[39m`

timtmok:
[...]

PR #10891, opened 2022-04-04
Fix File > New File > Quarto Document

timtmok:
### Intent
Addresses #10871 

Quarto only searches for the `share` directory in specific locations. With the difference in Electron packaging, Quarto can't find `share` in `RStudio.app/Contents/Resources/app`. Calling Quarto normally is fine until the new file dialog calls `quarto capabilities` which can't find the `share` directory.

### Approach
Move the Quarto `share` directory to root of `Resources` where Quarto can find it using the relative path check. This option was chosen since it keeps the Quarto `bin` with the other `bin` files while keeping the `share` out of it. I believe this is intended with how Mac apps, in general, are packaged. The alternative is placing both Quarto `bin` and `share` directories together.

### Automated Tests
None. Must be built to reproduce the error as dev environments use Quarto from your installed dependencies location.

### QA Notes
Ran `package/osx/make-package --electron` to check that it what it builds has resolved the issue.

Note: The issue is only reproduced when using `File > New File > Quarto Document`. The new file action in the Files pane bypasses the dialog that calls `quarto capabilities`.

### Checklist

### DIFF
--- a/CMakeGlobals.txt
+++ b/CMakeGlobals.txt
@@ -297,6 +297,8 @@ if (APPLE)
    if (RSTUDIO_ELECTRON)
       set(RSTUDIO_INSTALL_BIN        RStudio.app/Contents/Resources/app/bin)
       set(RSTUDIO_INSTALL_SUPPORTING RStudio.app/Contents/Resources/app)
+      # handles Quarto share when not stored alongside bin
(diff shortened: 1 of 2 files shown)


gtritchie (approved):
LGTM!

PR #10901, opened 2022-04-04
command to show accessibility diagnostics in Electron

gtritchie:
### Intent

Addresses #10899 and #10826

### Approach

Added Help / Diagnostics / "Show Accessibility Diagnostics" command. This command is hidden in RStudio Server and Qt Desktop.

Invoking shows this built-in Chromium window:

Screen Shot 2022-04-04 at 15 07 36

Also fixed so Cmd/Ctrl+W closes this window (and the GPU diagnostics window) instead of closing the current editor buffer.

### Automated Tests

None

### QA Notes

Verify that the command only shows up in Electron desktop, and displays the window. It is expected (useless, but expected) that you can display the window multiple times (the same thing happens with GPU Diagnostics on both Electron and Qt).

### Checklist

- [x] If this PR adds a new feature, or fixes a bug in a previously released version, it includes an entry in `NEWS.md` 
- [x] If this PR adds or changes UI, the updated UI meets [accessibility standards](https://github.com/rstudio/rstudio/wiki/Accessibility)
- [x] A reviewer is assigned to this PR (if unsure who to assign, check Area Owners list)
- [x] This PR passes all local unit tests

Issue #10895, opened 2022-04-04
Finish implementing screen-reader accessibility setting in Electron Desktop


gtritchie:
In Qt desktop, we use the RStudio "enable screenreader" setting to decide if we should tell QtWebEngine to avoid auto-enabling accessibility support due to issues such as https://github.com/rstudio/rstudio/issues/1990.

This happens here in the Qt code. https://github.com/rstudio/rstudio/blob/4daa0bc3e34e80cf296a55b4e979264dcc381cfa/src/cpp/desktop/DesktopMain.cpp#L553-L562

For parity, we should support the same behaviour with Electron.

## Dev Notes below

There are multiple settings for enabling screen reader support: (1) session preference, (2) desktop option (Qt), (3) user state (Electron).

### Session preference

- Defined here: https://github.com/rstudio/rstudio/blob/4daa0bc3e34e80cf296a55b4e979264dcc381cfa/src/cpp/session/resources/schema/user-prefs-schema.json#L1275-L1279
- This setting is used by GWT to make certain UI elements more screen-reader friendly (though potentially less efficient, for example, the xterm.js library will create a DOM-based representation of the terminal when this is enabled, in addition to its canvas-based representation)
- The user can toggle this via the Accessibility preferences pane, or via a dedicated command (toggleScreenReaderSupport)
- For Server, this is the only setting for this feature (the web browser decides on its own if it should enable screen-reader support via operating-system APIs)

### Qt Desktop Option:

- This gets stored using the desktopOptions mechanism (actual storage mechnism varies by platform): https://github.com/rstudio/rstudio/blob/4f5659a7b0371387a129604669de00cc3954517d/src/cpp/desktop/DesktopOptions.cpp#L328-L337
- When user changes the Session preference via UI or command, a desktop callback is invoked to synchronize the desktop setting with the session setting
  - invoked here from GWT: https://github.com/rstudio/rstudio/blob/4f5659a7b0371387a129604669de00cc3954517d/src/gwt/src/org/rstudio/studio/client/workbench/prefs/model/UserPrefs.java#L264-L268
  - received by the Qt desktop here: https://github.com/rstudio/rstudio/blob/4f5659a7b0371387a129604669de00cc3954517d/src/cpp/desktop/DesktopGwtCallback.cpp#L1350-L1353

### Electron Desktop User State

[...]

Issue #10748, opened 2022-03-07
changing the rendering engine should force a full restart of the IDE


kevinushey:
Currently, it seems to just reload the main window.

gtritchie:
The same problem happens when enabling/disabling screen reader setting.

gtritchie:
Research notes: for scenarios where we need to pass new flags to Chromium, I believe we need to fully restart (terminate the process, re-launch), we can't just reload in-place.

Electron supports this via `app.relaunch()`: https://www.electronjs.org/docs/latest/api/app#apprelaunchoptions

We don't want to use such a heavy hammer in all the scenarios that flow through this path, so need to differentiate this scenario in this code branch:

https://github.com/rstudio/rstudio/blob/90e4d797141fcc6338e253969c387ae4a0097547/src/node/desktop/src/main/session-launcher.ts#L377-L378

gtritchie:
Not going to tackel this at the moment, putting back on backlog. :-)

kevinushey:
I believe this was addressed by https://github.com/rstudio/rstudio/pull/10964.

kevinushey:
https://github.com/rstudio/rstudio/issues/10962 is likely also the same issue.

PR #10903, opened 2022-04-04
disable chromium renderer a11y based on user setting

gtritchie:
### Intent

Addresses #10895

### Approach

Port the behaviour from Qt, namely, ensure the desktop setting is synchronized with the session setting, and use that value to decide if we should disable renderer accessibility at startup.

After changing the "screen reader support" toggle it is necessary to restart Electron for it to take effect, but the restart performed is insufficient. Same problem as described here for gpu settings: #10748

### Automated Tests

None

### QA Notes

Verify that when screen reader support is off, Electron doesn't expose the accessibility tree. You can use a screen reader to tell, or more easily, look at the accessibility diagnostics once #10901 is merged.

### Checklist


kevinushey (approved):
LGTM!

Issue #10551, opened 2022-02-09
RStudio uses different path to application (e.g., `rsync`, `ffmpeg`, ...)


kalaschnik:
I was working with FFmpeg in R, and I realized that RStudio points to a different instance than R running in the Terminal.

Is that a bug, or can I change that somewhere within RStudio?

I am in favor of the actual terminal location: `/usr/local/bin/`

<img width="865" alt="image" src="https://user-images.githubusercontent.com/9831678/153290061-3e4251c5-ca94-4260-903f-b62463061d1b.png">

kalaschnik:
I just noticed that some other commands are affected; e.g., `rsync`
- bash: `which rsync` → `/usr/local/bin/rsync`
- R shell: `system("which rsync")` → `/usr/local/bin/rsync`
- RStudio: `system("which rsync")` → `/usr/bin/rsync`

That is very strange. Seems RStudio is not respecting that actual $PATH order in `/etc/paths` ?

kevinushey:
What is the output of `Sys.getenv("PATH")` versus `echo $PATH` in your terminal?

RStudio does explicitly read and load paths from `/etc/paths` and `/etc/paths.d/`, and there should be a fix in the preview release of RStudio regarding the ordering of those path entries.

kalaschnik:
**Terminal**
```
❯ cat /etc/paths
/usr/local/bin
/usr/bin
/bin
/usr/sbin
/sbin

❯ echo $PATH
/Users/steven_kalinke/opt/anaconda3/condabin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/munki:/opt/X11/bin:/Library/Apple/usr/bin:/Users/steven_kalinke/.fig/bin:/Applications/Visual Studio Code.app/Contents/Resources/app/bin
```

**RStudio**
```
> Sys.getenv("PATH")
[1] "/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin:/usr/local/munki:/opt/X11/bin:/Library/Apple/usr/bin:/Applications/RStudio 4.app/Contents/MacOS/postback"
```

Indeed for some reason `/usr/bin` happens first in RStudio, which may cause the issue over shadowing my actual commands in `/usr/local/bin`

kevinushey:
That does indeed sound like something that may be resolved in the preview release of RStudio: https://www.rstudio.com/products/rstudio/download/preview/

I see, with RStudio 1.4.1717:

```
[...]

PR #10907, opened 2022-04-05
read PATH from default shell on startup

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/10551.

### Approach

Rather than attempt to read the PATH via `/etc/paths` and `/etc/paths.d`, invoke the user's default shell as a login shell and ask that shell what the PATH is. This approach has the added bonus in that we can respect the user's configured PATH on Linux as well.

### Automated Tests

TODO

### QA Notes

Test that the PATH entries within a new RStudio session match what you see in a new terminal session. That is,

```
Sys.getenv("PATH")
```

and

```
echo $PATH
```

should print the same things in an RStudio session and a terminal session (assuming you aren't modifying your PATH within `.Renviron` or `.Rprofile`).

### Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,7 +10,7 @@
 - The source button uses `cpp11::source_cpp()` on C++ files that have `[[cpp11::register]]` decorations (#10387)
 - New *Relative Line Numbers* preference for showing line numbers relative to the current line, rather than the first line (#1774)
 - Upgraded SOCI library dependency from version 4.0.0 to 4.0.3 (#10792)
-- RStudio now reads the PATH from the user's default shell on startup (#10551)
+- (macOS only) RStudio now reads the PATH from the user's default shell on startup (#10551)
 
 #### Find in Files
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,7 @@
 - The source button uses `cpp11::source_cpp()` on C++ files that have `[[cpp11::register]]` decorations (#10387)
 - New *Relative Line Numbers* preference for showing line numbers relative to the current line, rather than the first line (#1774)
 - Upgraded SOCI library dependency from version 4.0.0 to 4.0.3 (#10792)
+- RStudio now reads the PATH from the user's default shell on startup (#10551)
 
 #### Find in Files
 
--- a/src/cpp/session/modules/SessionPath.cpp
+++ b/src/cpp/session/modules/SessionPath.cpp
@@ -19,6 +19,7 @@
 #include <vector>
 
 #include <boost/bind/bind.hpp>
+#include <boost/regex.hpp>
 #include <boost/system/errc.hpp>
 
 #include <shared_core/Error.hpp>
@@ -105,7 +106,13 @@ Error initializePathViaShell(const std::string& shellPath,
 // might.
 std::string initializePath()
 {
-   std::string path;
+   // if the user's path already contains '/usr/local/bin', assume that
+   // they're running RStudio through a shell / terminal and so we don't
+   // need to re-read the shell PATH
+   std::string defaultPath = core::system::getenv("PATH");
+   boost::regex reUsrLocalbin("(^|:)/usr/local/bin(:|$)");
+   if (boost::regex_search(defaultPath, reUsrLocalbin))
+      return defaultPath;
    
    // first, try to initialize with user's default shell
    // (RSTUDIO_SESSION_SHELL is primarily for internal use)
--- a/src/cpp/session/modules/SessionPath.cpp
+++ b/src/cpp/session/modules/SessionPath.cpp
@@ -95,7 +95,8 @@ Error initializePathViaShell(const std::string& shellPath,
    {
       // only include last line of output, in case the shell printed
       // something extra during login / processing of scripts
-      std::vector<std::string> lines = core::algorithm::split(result.stdOut, "\n");
+      std::string output = core::string_utils::trimWhitespace(result.stdOut);
+      std::vector<std::string> lines = core::algorithm::split(output, "\n");
       auto n = lines.size();
       if (n == 0)
       {
--- a/src/cpp/session/modules/SessionPath.cpp
+++ b/src/cpp/session/modules/SessionPath.cpp
@@ -122,6 +122,7 @@ std::string initializePath()
    
    if (!shell.empty())
    {
+      std::string path;
       Error error = initializePathViaShell(shell, &path);
       if (!error)
          return path;
@@ -130,6 +131,7 @@ std::string initializePath()
    // next, try to initialize with default shell
    if (shell != "/bin/sh")
    {
+      std::string path;
       Error error = initializePathViaShell("/bin/sh", &path);
       if (!error)
          return path;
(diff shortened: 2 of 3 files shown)


kevinushey:
All that said, I feel a little worried about this PR because:

1. Users might configure their shell in arbitrary ways;
2. Users might run arbitrary code on startup;
3. There could be other side-effects from us trying to run their login shell in this way.

I sort of wonder if we might take this only for SW, and consider a simpler fix for backporting.

NEWS.md
```diff
@@ -10,6 +10,7 @@
 - The source button uses `cpp11::source_cpp()` on C++ files that have `[[cpp11::register]]` decorations (#10387)
 - New *Relative Line Numbers* preference for showing line numbers relative to the current line, rather than the first line (#1774)
 - Upgraded SOCI library dependency from version 4.0.0 to 4.0.3 (#10792)
+- (macOS only) RStudio now reads the PATH from the user's default shell on startup (#10551)
```
jgutman, line 13: I know we mentioned potentially considering this for Linux users; however my instinct is, we've not seen any questions/concerns/issues from Linux users about the path not being as expected, so I think there's no strong justification for extending it now -- it's easy enough to change that later if it did turn out to be a need for Linux users as well, but no reason to pre-emptively do so

src/cpp/session/modules/SessionPath.cpp
```diff
@@ -39,129 +43,165 @@ namespace session {
...
-                           std::vector<std::string>* pPaths)
+Error initializePathViaShell(const std::string& shellPath,
+                             std::string* pPath)
 {
-   Error error = readPathsFromFile(filePath, pPaths);
+   // double-check that the requested shell exists
+   if (!FilePath(shellPath).exists())
+      return fileNotFoundError(shellPath, ERROR_LOCATION);
+   
+   std::vector<std::string> args = { "-l", "-c", "printf \"%s\" \"$PATH\"" };
```
jgutman, line 68: what's -l and -c do here

src/cpp/session/modules/SessionPath.cpp
```diff
@@ -39,129 +43,165 @@ namespace session {
...
+// might.
+std::string initializePath()
 {
-   // tolerate paths with trailing slashes
-   for (std::string item : {entry, entry + "/"})
+   // if the user's path already contains '/usr/local/bin', assume that
+   // they're running RStudio through a shell / terminal and so we don't
+   // need to re-read the shell PATH
+   std::string defaultPath = core::system::getenv("PATH");
+   boost::regex reUsrLocalbin("(^|:)/usr/local/bin(:|$)");
```
jgutman, line 128:
```suggestion
   boost::regex reUsrLocalbin("(^|:)/usr/local/bin/?(:|$)");
```

src/cpp/session/modules/SessionPath.cpp
```diff
@@ -39,129 +43,165 @@ namespace session {
...
+   if (!shell.empty())
+   {
+      std::string path;
+      Error error = initializePathViaShell(shell, &path);
+      if (!error)
+         return path;
+   }
+   
+   // next, try to initialize with default shell
+   if (shell != "/bin/sh")
```
jgutman, line 147: wouldn't shell always be empty at this point? is this `if` necessary?

src/cpp/session/modules/SessionPath.cpp
```diff
@@ -39,129 +43,165 @@ namespace session {
...
-      if (it != pPathComponents->end())
-         return;
+      std::string path;
+      Error error = initializePathViaShell("/bin/sh", &path);
+      if (!error)
+         return path;
    }
+   
+   // all else fails, use the current application path
+   return core::system::getenv("PATH");
```
jgutman, line 156: if this fails, don't we need to pull in paths from /etc? Thought the fallback would essentially be like your conservative PR 2

jgutman (changes requested):
I prefer this overall -- some small nits

NEWS.md
kevinushey: That sounds good to me as well.

src/cpp/session/modules/SessionPath.cpp
kevinushey: `-l` is used to ask the shell to behave like a login shell (so that it will source a user's shell profile, e.g. `.bash_profile` for bash).

`-c` is used to pass a command to the shell to be executed.

These flags are supported by the most commonly used macOS shells (bash, zsh, fish) so I think this invocation should be safe.

src/cpp/session/modules/SessionPath.cpp
kevinushey: This should handle the case where `SHELL` points to an invalid shell, or if we fail to initialize it. In that case, we would still be able to fall-through to initializing with the default shell, so we'd still have a chance of getting the default path correct.

src/cpp/session/modules/SessionPath.cpp
kevinushey: That's true -- I should add that back in.

jgutman (approved):

stefanocoretta:
Hi! Does this solution also works if the path is set in `.zshrc`?

jgutman:
Yes if zsh is your default shell than it should work from there

stefanocoretta:
I tried now with the latest RStudio version and doesn't seem to work.

I get:

```r
Sys.getenv("PATH")
[1] "/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Library/TeX/texbin:/opt/X11/bin:/Users/ste/Applications/quarto/bin:/usr/texbin:/Applications/RStudio.app/Contents/MacOS"
```

but:

```bash
echo $PATH
# /opt/homebrew/Caskroom/miniforge/base/bin:/opt/homebrew/Caskroom/miniforge/base/condabin:/opt/homebrew/bin:/opt/homebrew/sbin:/Users/ste/repos/backhome/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Library/TeX/texbin:/opt/X11/bin:/Users/ste/Applications/quarto/bin:/usr/texbin:/Applications/RStudio.app/Contents/MacOS:/opt/R/arm64/gfortran/bin
```

My session

```
─ Session info ─────────────────────────────────────────────────────────
 setting  value
 version  R version 4.2.1 (2022-06-23)
 os       macOS Monterey 12.4
 system   aarch64, darwin20
 ui       RStudio
 language (EN)
 collate  en_US.UTF-8
 ctype    en_US.UTF-8
 tz       Europe/London
 date     2022-07-28
 rstudio  2022.07.1+554 Spotted Wakerobin (desktop)
 pandoc   NA

─ Packages ─────────────────────────────────────────────────────────────
 package     * version date (UTC) lib source
 cli           3.3.0   2022-04-25 [1] CRAN (R 4.2.0)
 sessioninfo   1.2.2   2021-12-06 [1] CRAN (R 4.2.0)

 [1] /Library/Frameworks/R.framework/Versions/4.2-arm64/Resources/library

────────────────────────────────────────────────────────────────────────
```

PR #10908, opened 2022-04-05
tweak PATH initialization on macOS

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/10551.
Narrower alternative of https://github.com/rstudio/rstudio/pull/10907.

### Approach

Detect if `/usr/local/bin` is on the PATH, and if so, assume the PATH is already initialized. Otherwise, ignore the current state of the PATH and initialize it manually by reading `/etc/paths` and `/etc/paths.d`.

### Automated Tests

Not included.

### QA Notes

Try launching RStudio with something like:

```
PATH=/usr/bin:/bin:/usr/sbin:/sbin /Applications/RStudio.app/Contents/MacOS/RStudio
```

and check that `Sys.getenv("PATH")` reports that `/usr/local/bin` is on the PATH in the right location.

### Checklist

### DIFF
--- a/src/cpp/session/modules/SessionPath.cpp
+++ b/src/cpp/session/modules/SessionPath.cpp
@@ -89,11 +89,16 @@ void addToPathIfNecessary(
    pPathComponents->push_back(entry);
 }
 
-} // anonymous namespace
-
-
-Error initialize()
+std::vector<std::string> initializePath()
 {
+   // if '/usr/local/bin' is already on the PATH, then assume
+   // we're running RStudio through a terminal and so the PATH
+   // has already been initialized
+   std::string path = core::system::getenv("PATH");
+   boost::regex reUsrLocalBin("(^|:)/usr/local/bin(:|$)");
+   if (boost::regex_search(path, reUsrLocalBin))
+      return core::algorithm::split(path, ":");
+   
    // read /etc/paths
    std::vector<std::string> paths;
    safeReadPathsFromFile(FilePath("/etc/paths"), &paths);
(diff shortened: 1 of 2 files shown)


src/cpp/session/modules/SessionPath.cpp
```diff
@@ -116,39 +121,46 @@ Error initialize()
...
 
-   // build the PATH
-   std::vector<std::string> parts =
-         core::algorithm::split(core::system::getenv("PATH"), ":");
-
    // add in components from paths.d etc.
    std::for_each(paths.begin(),
                  paths.end(),
-                 boost::bind(addToPathIfNecessary, _1, &parts));
+                 boost::bind(addToPathIfNecessary, _1, &paths));
```
jgutman, line 127: Wait, I'm confused about these lines. We're looping through paths, and then adding paths to paths? Isn't this completely unnecessary, now that we're not starting with parts anymore?

jgutman (changes requested):
This is okay (but see my comment about the lines that I think should be removed?), although if we go with this route we'd 1) have to add some handling for `/opt/homebrew/bin` and 2) have to add some kind of support article to explain to users what to do if they want their path to be same as system path. 

However, I really don't love the solution -- I just have a strong preference for using the actual system path (slightly modified if necessary, e.g. adding the quarto and Tex stuff) with whatever intentional modifications a user might have made to it. 

Going to review the other PR now as well!

src/cpp/session/modules/SessionPath.cpp
kevinushey: You are correct -- let me take another look.

kevinushey: Should be fixed up now; thanks for catching that!

kevinushey:
Superceded by https://github.com/rstudio/rstudio/pull/10907.

Issue #10929, opened 2022-04-07
respect logging environment variables in desktop


kevinushey:
For example, here:

https://github.com/rstudio/rstudio/blob/49ed99ad10f7af4f2f589c5ca22b82605fb818b3/src/cpp/core/LogOptions.cpp#L47-L51

The desktop client should read and parse those when determining the right way to log.

kevinushey:
Note: RS_LOG_LEVEL is handled in https://github.com/rstudio/rstudio/pull/10930.

ronblum:
Should we document this in the Desktop Guide, as we do in the RSW Admin Guide?

kevinushey:
That's a good idea; yes we should.

jonvanausdeln:
Verified on 2022.12.0-daily+294 Linux Server and Desktop

Environment variables are respected for logging.

Issue #10695, opened 2022-02-28
Implement Electron rendering engine settings


gtritchie:
In Qt desktop we provide several ways to control the rendering engine used by Chromium. This is usually to work around problems on specific platforms/graphics cards/resolutions, etc.

- command-line switches
- UI in the "Choose R Installation" dialog shown at startup on Windows
- Global Options / General / Graphics

We need to ensure these are properly plumbed to work with Electron.

There's a related unimplemented callback, as well: `desktop_set_desktop_rendering_engine`.

kevinushey:
For testing:

- Confirm that if you hold down Ctrl while launching RStudio, you get the "Choose R" dialog, with an option to select the rendering engine at the bottom. (Windows only)

- Confirm that you are prompted to restart RStudio after changing the rendering engine through the aforementioned dialog. (Windows only)

- Use Help -> Diagnostics -> Show GPU Diagnostics to confirm whether the requested rendering engine is being honored. (If you select software rendering, you should see most "features" using software rendering)

- Confirm that the same rendering engine options can be changed via Tools -> Global Options -> Advanced, and that the options displayed + used there also reflect the same options selected in the Choose R dialog.

ronblum:
Checked in the following, using the above guidance, on:
- MacOS 12.4
- Red Hat 8.5
- Windows

All pass, with one exception: On Red Hat, when using "OpenGL for Embedded Systems," Electron appears to be using software acceleration, while Qt is using hardware acceleration; and the `OpenGL` property is set to `Disabled`. Here are the values under "Graphics Feature Status" as a comparison:

*Electron*
![openglembedded_en.png](https://codetreeusercontent.s3.amazonaws.com/uploads/d1c3b22a-0659-4f80-b80e-c7af91a1332b/openglembedded_en.png)

*Qt*
![openglembedded_qt.png](https://codetreeusercontent.s3.amazonaws.com/uploads/d1c3b22a-0659-4f80-b80e-c7af91a1332b/openglembedded_qt.png)

[Thank you @kevinushey for explaining to me that this is primarily about the settings, etc., and not a thorough evaluation of the rendering engines, per se.]

ronblum:
Note that that's on Red Hat 8.5—on Ubuntu 20.04.3, OpenGL is enabled and hardware acceleration is being used:
[...]

PR #10930, opened 2022-04-07
allow user to modify rendering engine in startup window

kevinushey:
### Intent

Addresses part of https://github.com/rstudio/rstudio/issues/10695.
Also part of https://github.com/rstudio/rstudio/issues/10929, for `RS_LOG_LEVEL`.

### Approach

Ports the rendering engine portion of the "Choose R" dialog, and uses that to set the rendering engine option. The plumbing is mostly straightforward; the main difference is around what data gets marshaled around in the Choose R dialog (now, we pass along the selected rendering engine, and that gets handled after the user requests a version of R).

### Automated Tests

Not included; it isn't trivial to test restart of the application.

### QA Notes

To test, from a console, trying running:

```
set RSTUDIO_DESKTOP_PROMPT_FOR_R=TRUE
C:\path\to\rstudio.exe
```

And confirm that you see a select box for changing the rendering engine.

Screen Shot 2022-04-07 at 1 01 23 PM

And check that Help -> Diagnostics -> Show GPU Diagnostics matches the selection. For example, for Software rendering:

Screen Shot 2022-04-07 at 1 07 10 PM

### Checklist


src/node/desktop/.eslintrc.js
```diff
@@ -14,7 +14,7 @@ module.exports = {
     sourceType: 'module',
   },
 
-  plugins: ['@typescript-eslint', 'prettier'],
```
kevinushey, line 17: I can back this out, but honestly, I find the 'prettier' warnings far too annoying / restrictive. The number of warnings spit out by VSCode when this is active is far too distracting and (compounded with the existing pain of Windows development) makes development on Windows just hugely unpleasant.

gtritchie: No objections

src/node/desktop/src/ui/widgets/choose-r/ui.html
```diff
@@ -25,6 +25,17 @@
...
     <select id="select" class="select" name="select" size="7"></select>
 
+    <p id="i18n-customizeRenderingEngine"></p>
+
+    <div>
+      <label id="i18n-renderingEngineLabel" for="rendering-engine"></label>
+      <select name="rendering-engine" id="rendering-engine">
+        <option label="Auto-detect (recommended)" value="auto"></option>
+        <option label="Desktop" value="desktop"></option>
+        <option label="Software" value="software"></option>
```
gtritchie, line 35: Can we put these UI strings into the string file instead of hardcoding them? Ideally, new UI should be localizable from the start.

src/node/desktop/src/main/detect-r.ts
```diff
@@ -85,18 +92,41 @@ export async function promptUserForR(platform = process.platform): Promise<Expec
...
     ElectronDesktopOptions().setRBinDir(path);
+
+    // if the user has changed the default rendering engine,
+    // then we'll need to ask them to restart RStudio now
+    const enginePref = ElectronDesktopOptions().renderingEngine() || 'auto';
+    const engineValue = data.renderingEngine || 'auto';
+    if (enginePref !== engineValue) {
+      ElectronDesktopOptions().setRenderingEngine(engineValue);
+      dialog.showMessageBoxSync({
+        title: 'Rendering Engine Changed',
```
gtritchie, line 117: We should make these UI strings localizable instead of hardcoded.

gtritchie (changes requested):

kevinushey:
I ended up streamlining the implementation for localization of independent UI elements (that is, windows that have their own UI HTML). Roughly speaking:

- Any localizable element should have an appropriate `data-18n-id` attribute set,
- The associated load script should call `localize(document, scope)` to automatically apply translations from the associated appropriate scope.

Let me know if this seems reasonable.

kevinushey:
I also recently learned that the approach I'm taking is pretty similar in spirit to jquery-i18next:

https://github.com/i18next/jquery-i18next

but I think keeping our own implementation here should be fine.

gtritchie (approved):
LGTM!

Issue #10866, opened 2022-03-29
Checkable menu items not checked in Electron


gtritchie:
### System details

    RStudio Edition : Desktop (Electron)
    RStudio Version : 2022.06.0-daily+244
    OS Version      : All
    R Version       : All

### Steps to reproduce the problem

- look at main main items that should have checkboxes such as View / Panes, or Debug / On Error

### Describe the problem in detail

- On Electron builds (all platforms) there aren't any checked menu items (screenshot from Electron):

![Screen Shot 2022-03-29 at 16 00 46](https://user-images.githubusercontent.com/10569626/160714394-6495d23e-d2cc-4619-aaef-749687e38535.png)

### Describe the behavior you expected

- Checked menu items (screenshot from Qt version):

![Screen Shot 2022-03-29 at 16 02 32](https://user-images.githubusercontent.com/10569626/160714590-eda8a030-0887-4a84-86e6-07348d40b634.png)

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

timtmok:
`Electron.MenuItem` does have a checked property and I've confirmed that the menu template is updated with the latest state coming from GWT. The menu is rebuilt so it will have the correct state.

I think the fix is setting `MenuItem.type` to `checkbox`. https://github.com/rstudio/rstudio/blob/d39b759d59140787ce5972a04a58156f6b4d4788/src/node/desktop/src/main/menu-callback.ts#L336

gtritchie:
Reactivating. The check marks are showing up, but the state isn't correct. For example:

- Look at "Debug / On Error" submenu
- Only one item should have a checkbox
- Click on a different one
- Open the menu again

On Mac, at least, I'm seeing both items with checkboxes. Can repeat, and get three items with checkboxes.

![Screen Shot 2022-03-31 at 11 12 14](https://user-images.githubusercontent.com/10569626/161112271-0a232907-bb3f-459f-aecb-977c0a882068.png)

Same with Console on Left/Right, able to get both selected:
[...]

Issue #10876, opened 2022-03-31
Mark mutually-exclusive menu items as type:radio in Electron Desktop


gtritchie:
Main menu items such Debug / On Error contain multiple items where only one at a time can be checked.

![Screen Shot 2022-03-31 at 11 40 16](https://user-images.githubusercontent.com/10569626/161116815-1aef50ed-808c-4a31-9040-cf8fcfc8260f.png)

For accessibility we should mark these menu items as type:radio, not type:checkbox as we currently do. We decided to only do this if the accessibility switch is on (the appearance of radio menus is quite jarring; not seen very commonly, and really only beneficial to screen-reader users).

These are annotated as `radio` in command.cmd.xml: see https://github.com/rstudio/rstudio/commit/47fc0dd5 for when I hooked that up.

timtmok:
I think the `radio` just might work. The real test will be the View > Panes > Console on left/right items.

matheustavaresdev:
It works only when we click on it twice which is kind of weird...

ronblum:
Tested with build RStudio Desktop 2022.06.0-daily-316. Marking as failed for the following reasons: 
- MacOS 12.4: Works as before, and no change expected—success
- Red Hat 8.4: Shows checks as in MacOS, instead of radio buttons—fail
- Ubuntu 20.04: Looks like a checkbox, instead of radio buttons—fail (unless this is the only option on Ubuntu)

Windows TBD due to #10162.

Red Hat Example: 
![radio_redhat.png](https://codetreeusercontent.s3.amazonaws.com/uploads/d10f0c30-dfea-433a-8825-88517054a697/radio_redhat.png)

Ubuntu Example:
![radio_ubuntu.png](https://codetreeusercontent.s3.amazonaws.com/uploads/d10f0c30-dfea-433a-8825-88517054a697/radio_ubuntu.png)

gtritchie:
It is possible none of the Linux desktop managers support displaying these as checkboxes; we have no direct control over that.

So, if this works on Mac and Windows I'd consider it sufficient.

gtritchie:
Build 2022.06.0-daily+333 for RHEL8 on Rocky Linux (with Gnome) is showing me the radio menus as expected. Did you have accessibility enabled when you tried? Or maybe using a different window manager like KDE Plasma (which I haven't tried)?

gtritchie:
[...]

PR #10935, opened 2022-04-08
Electron Menu Items Improvements

matheustavaresdev:
### Intent

Fix some issues that would make Electron Menu Items not to be checked when clicked.

### Approach

Menus would not be checked due to the menu not being rebuilt when changes would occur. This was fixed.

When running tests, I found out that:
- Sometimes the menu would be rebuilt many times in a row. 
- When the app would start, menu would also be rebuilt many times in a row
- Other functions that would update a menu Label, Visibility and to set whether or not a menu was Enabled, would not be applied
All of the above scenarios have been fixed.

### Automated Tests

None

### QA Notes

#10866

### Checklist

Addresses #10866


src/node/desktop/src/main/menu-callback.ts
```diff
@@ -224,31 +234,67 @@ export class MenuCallback extends EventEmitter {
     const accelerator = this.convertShortcut(shortcut);
     template.accelerator = accelerator;
 
-    if (this.setShortcutDebounceId) clearTimeout(this.setShortcutDebounceId);
-
-    this.setShortcutDebounceId = setTimeout(() => {
-      this.updateMenus();
-    }, 5000);
+    this.debounceUpdateMenuLong();
```
matheustavaresdev, line 237: @timtmok I have removed your code in favor of this simpler one. I kept your 5000 ms timeout, but I was wondering: Does it need to be that long? If a menu update is triggered, we will wait 5 seconds for it to be applied. This function is quite fast when running a single time. Wouldn't a smaller timeout be enough? For app initialization, I found that even 250 would be enough... Unless this is triggered hundreds of times in a 5 seconds time frame... 

Haven't tested this bit to see how many times is this triggered... Any thoughts?

timtmok: This was originally there when I last worked on the menu and I kept the 5000 ms timeout because that's just what was there. 250 ms sounds like it would be enough to have all the shortcut changes updated in the model before updating the Electron menu. 👍

timtmok:
LGTM

I did a quick check of Qt and how it handles mutually exclusive menu items. This seems to replicate the behaviour in the same way with the View menu (only one that seemed to have a complicated set of checkable menu items). The one odd behaviour that existed in Qt and still exists here is zooming a primary (is that what it's called?) pane (source, console, help) when a secondary pane is zoomed (history, files, plots, environment, etc.) resets the zoom. You have to activate the option again to get the desired zoom. Since this existed in Qt, I wouldn't consider it a regression.

src/node/desktop/src/main/menu-callback.ts
```diff
@@ -333,7 +381,7 @@ export class MenuCallback extends EventEmitter {
     };
 
     if (checkable) {
-      menuItemOpts.type = 'checkbox';
+      menuItemOpts.type = 'radio';
```
gtritchie, line 384: We don't want all checkable menu items to be `radio`, only those marked as such in the Commands.cmd.xml. See comment in the issue: #10876

https://github.com/rstudio/rstudio/blob/09be8fa4417fce8b42aea96d9ed9e1f76470864a/src/gwt/src/org/rstudio/studio/client/workbench/commands/Commands.cmd.xml#L31-L33

gtritchie (changes requested):
I haven't fully reviewed yet, just want to get you the comment about not using `radio` unless the command was annotated as such.

src/node/desktop/package.json
```diff
@@ -74,6 +75,7 @@
     "electron-store": "8.0.1",
     "i18next": "21.6.10",
     "line-reader": "0.4.0",
+    "lodash.debounce": "^4.0.8",
```
gtritchie, line 78: We need to add this to the `NOTICE.electron` file.

src/node/desktop/src/main/menu-callback.ts
timtmok: It is odd to have `radio` set for everything when that's not what comes from the `Commands.cmd.xml`. It seemed to work though and even exactly to the odd behaviour that Qt has with the Zoom Source/Help/Console panes when zoomed on another pane.

It probably is worth using the `radio` only when specified. GWT will need an update to specify whether the menu item is checkable or a radio. I just can't find an existing menu item that requires checkable, which is why this seems to work for now using `radio`.

timtmok: Gary pointed me to the checkable menu items. The ones that I think should be tested are the Soft Wraps and Rainbow Parentheses in the Code menu. They are set as checkable and can both be checked at the same time.

matheustavaresdev: Yea, that's why I kept radio for everything. As the menu is entirely rebuilt, the items can be unchecked if needed, as it seems that is GWT is handling the state for all menu items and passing over to electron. I thought of keeping everything as radio not to spend more time going around to see that part as I'm not used to GWT but I think that could be good to maintain a better code.

gtritchie: There are a bunch of menu items in Commands.cmd.xml annotated as checkable vs. radio. Some might make more sense as `radio`, but there are some that are definitely not radios. Even if it seems to be behaving correctly, it is an accesibility problem for screen reader users to have a single checkable menu item marked as radio (and may also have the wrong visual representation).

I can do the bit of plumbing needed to get the metadata on radio vs. checkable where we need it.

gtritchie: For this PR, let's keep the menutype as "checkbox". We will track using "radio" (if we do it at all) via #10876.

gtritchie, line 384: Leave this as "checkbox". This matches what we do in Qt.

gtritchie (changes requested):

src/node/desktop/src/main/menu-callback.ts
gtritchie: Also, remove mention of #10876 from the PR notes. That's a separate work item and isn't implemented by this PR.

gtritchie (approved):
LGTM! Thanks!

PR #10948, opened 2022-04-11
Log message fixes

kfeinauer:
### Intent

Currently, log message properties that contain new lines will forward those new lines into the log. This is undesired because log lines themselves should not contain new lines, as that is the delimiter between log lines.

Also, the log location of a log statement is not currently being logged for JSON and should be added.

### Approach

Filter out newlines for log message properties formatted via the pretty formatter. No changes need to be made to the JSON formatting because JSON already encodes newlines as literal `\n`.

If there is a log location associated with a log statement and we are using the JSON log format, simply format it for JSON and put it on the `location` property of the log line.

### Automated Tests

Added two new tests to ensure that newlines in log message properties are not logged.

### DIFF
--- a/src/cpp/core/LoggingTests.cpp
+++ b/src/cpp/core/LoggingTests.cpp
@@ -970,6 +970,72 @@ test_context("Logging")
       REQUIRE_FALSE(logFile.remove());
       REQUIRE_FALSE(defaultLogFile.remove());
    }
+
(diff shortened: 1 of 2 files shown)


jeffvroom (approved):
LGTM!

MariaSemple (approved):
LGTM!

PR #10950, opened 2022-04-11
Improve accessibility of mutually-exclusive menu items in Electron

gtritchie:
### Intent

Addresses #10876

### Approach

The "radio" type wasn't being passed through the desktop callback; now it is.

On Windows and Linux, the "radio" menus will now actually display as radio menu items. Mac doesn't have this concept so there won't be any difference.

To avoid the visual clutter for most users, I only show these menu items as radio buttons if screen-reader support is enabled. It has become fairly rare for desktop software to use radio menus so they are a bit visually jarring.

There are two radio menu groups. "View / Panes / Console on Left|Right" and "Debug / On Error / ...":

console-left-right

error-radio

### Automated Tests

None

### QA Notes

Verify on Windows and Linux that these two menu groups now display as radio buttons.

### Checklist


kevinushey:
Radio buttons in a global menu look really weird to me. Are we sure we want this versus checkboxes?

Is there some compromise where we can display checked vs. unchecked menu items, but signal to screen readers that they behave like radio buttons?

gtritchie:
Yeah these aren't as commonly seen as they used to be.

I can make annotating them as radios conditional on screen-reader being enabled.

timtmok:
Looks good! So glad that the radio style is used when accessibility is on.

timtmok (approved):

PR #10986, opened 2022-04-14
Move Electron path test to after Qt path test for Mac

timtmok:
### Intent
Fix Qt accidentally using Resources/quarto as the quarto binary path. This caused the IDE to not show File > New File > Quarto Document/Presentation

### Approach
Move the Electron path check to after Qt. It turns out that the Electron portion of the check would look in `app` for Quarto. In a Qt package, it looks in `Resources` which holds the Quarto `share` directory. The session ends up using that path and the Quarto checks for the menu items would see that Quarto is not available at that location so the items aren't shown.

The Electron path check was first but I don't think it has to be the first location to check for the binary.

### Automated Tests
None

### QA Notes
I've locally patched a daily Qt build and local Electron build and it works now.

### Checklist

### DIFF
--- a/src/cpp/session/SessionOptions.cpp
+++ b/src/cpp/session/SessionOptions.cpp
@@ -503,7 +503,8 @@ FilePath macBinaryPath(const FilePath& resourcePath,
 {
    // otherwise, look in default Qt location
    FilePath qtPath = resourcePath.getParent().completePath("MacOS").completePath(stem);
-   return qtPath;
+   if (qtPath.exists())
+      return qtPath;
 
    FilePath electronPath =
          resourcePath.completePath("bin").completePath(stem);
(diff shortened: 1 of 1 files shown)


src/cpp/session/SessionOptions.cpp
```diff
@@ -501,7 +501,10 @@ namespace {
 FilePath macBinaryPath(const FilePath& resourcePath,
                        const std::string& stem)
 {
-   // first check for Electron binary path
+   // otherwise, look in default Qt location
+   FilePath qtPath = resourcePath.getParent().completePath("MacOS").completePath(stem);
+   return qtPath;
```
kevinushey, line 506: Does this code need a Qt-only (or existence) check? It seems like the following Electron code would now be dead code?

PR #10991, opened 2022-04-14
Use cookie for selecting UI language in GWT instead of local storage

gtritchie:
### Intent

Fixes another aspect of #10455

The original fix (https://github.com/rstudio/rstudio/pull/10923) used browser local storage for tracking the selected UI language. However, because browser local storage is tied to the origin, and Desktop uses a random port on localhost, info put in local storage in one RStudio Desktop instance is unlikely to show up in the next instance. This caused the UI to load twice when loading non-English UI (the first time it sees that the value isn't set to match user preferences, so sets it and reloads the page).

### Approach

Switch to using a cookie instead of local storage. The cookie is of the form `LOCALE=fr`.

This approach is cleaner in some ways because it doesn't require the bit of JavaScript in the page to insert a meta-tag based on the local storage; GWT can key directly off the cookie.

This works for Server and Electron Desktop, but Qt-desktop doesn't persist the cookie so still has the extra reload when running with French UI. I opened #10993 to track that, but we probably won't fix it since multilingual UI is experimental in the next release, does work (just a flash at startup of Qt desktop), and the release after that will deprecate the Qt-based desktop.

Fixing likely involves using QtWebEngine APIs for cookie management, possibly including a new desktop callback for that purpose.

### Automated Tests

No automated tests for non-English UI scenarios.

### QA Notes

### Checklist


kevinushey (approved):
LGTM!

PR #10998, opened 2022-04-15
Add the ability to create ED25519 SSH keys, and set ED25519 as the default

jgutman:
### Intent

Addresses #8255.

- Adds support for ED25519 encrypted keys to be created (by default)
- Increase size for RSA keys so if they are created, they follow Github recommendation (4096 bytes)
- Update all dialogs/labels that refer to RSA SSH keys to refer more generally to SSH keys

### Approach

RPC call already had an argument for SSH key type to be passed along, it was just set to "rsa" as a constant
Updated so user can choose their preferred key type when they create a new key. As before, when searching for existing keys, we check for ED25519 keys first, and then look for RSA only if not found -- change is only to enable the creation of new ED25519 keys (and better RSA keys) from within the IDE

### Automated Tests

None

### QA Notes

- Create a new ED25519 key
- Create a new RSA key
- Check which key is detected if both key types exist
- Check which key is detected if only one key type exists (eg should detect RSA if user has only RSA)

(You may have to restart RStudio to re-detect deleted keys)

### Checklist


src/cpp/session/modules/SessionWorkbench.cpp
```diff
@@ -298,6 +298,9 @@ Error createSshKey(const json::JsonRpcRequest& request,
    // type
    cmd << "-t" << type;
 
+   if (type == "rsa")
```
gtritchie, line 301: Minor, but maybe compare to `kSshKeyTypeRsa` instead of hardcoded string?

src/cpp/session/resources/schema/user-prefs-schema.json
```diff
@@ -1144,10 +1144,17 @@
             "default": "",
             "description": "The path to the terminal executable to use."
         },
-        "rsa_key_path": {
+        "ssh_key_path": {
```
gtritchie, line 1147: I wouldn't change an existing preference's name. Yes, it would be more accurate and self-documenting, but if you do this you should also implement migration of the setting from the old value to the new, and I don't think that's worth the effort to code and test. Plus nothing you can do if user goes backwards (downgrades to an older version).

gtritchie (changes requested):

src/cpp/session/resources/schema/user-prefs-schema.json
jgutman: ah right, yes good point!

src/cpp/session/prefs/UserPrefsComputedLayer.cpp
```diff
@@ -65,6 +65,7 @@ Error UserPrefsComputedLayer::readPrefs()
    FilePath sshKeyDir = modules::source_control::defaultSshKeyDir();
 
    // Github recommends using ed25519, so look for that first
+
```
jgutman, line 68:
```suggestion
```

src/gwt/src/org/rstudio/studio/client/workbench/prefs/model/UserPrefsAccessorConstants_fr.properties
```diff
@@ -153,7 +153,7 @@ insertMatchingTitle= Insertion automatique des parenthèses et des crochets corr
 insertMatchingDescription= Insérer ou non, les paires correspondantes, telles que () et [], lorsque la première est tapée.
 
 # Whether to insert spaces around the equals sign in R code.
-insertSpacesAroundEqualsTitle= Insérer des espaces autour 
+insertSpacesAroundEqualsTitle= Insérer des espaces autour
```
jgutman, line 156:
```suggestion
insertSpacesAroundEqualsTitle= Insérer des espaces autour 
```

gtritchie (approved):
LGTM!

PR #11008, opened 2022-04-19
Removed redundant call to invalidateAuthCookie

jeffvroom:
This is now done in invalidateUserSession for all cookies.

For postgres, the extra call causes an error in the logs on signout.

Another part of the fix for: https://github.com/rstudio/rstudio-pro/issues/3287

(This is another small change part of that previous fix so I'm going to merge before the review again...)


kevinushey:
Belated LGTM!

Issue #10893, opened 2022-04-04
Mac - Open associated files when clicked


timtmok:
After some light testing, dragging files onto the app will launch and attempt to open the file. R projects don't open though. Double-clicking the associated files in Finder should open RStudio as well but it doesn't appear to work.

The existing behaviour should be documented as well.

Split from #10437

ronblum:
Verified that that both double-clicking on the file and dropping it on the application will open RStudio Desktop 2022.06.0-daily-333. Checked the following file types:
- .R
- .qmd
- .Rmd
- .Rproj

PR #11011, opened 2022-04-19
Electron Mac file associations

timtmok:
### Intent
Addresses #10893

Allows opening associated files, dragging files on to the app icon, and command-line arguments

### Approach
Feeds in the `Info.plist.in` template when building the app. This associates the files with MacOS.

Listens to the `open-file` event which triggers for associated file open events and dragging files on to the app icon. For project files, this sets the env var before the session is created so that the project is opened.

For command-line arguments, they are searched for unswitched options, checks if it's a valid file, and opens it.

An `initialized` property was added to the GWT callback so it can track when GWT has initialized the workbench. Responding to the app open event needs to know if the workbench is ready to call the open file action.

### Automated Tests
None

### QA Notes
Main cases to test:
* dragging files on to the app icon (in Finder or the dock)
* double-clicking files in Finder
* opening from the command-line (`open RStudio.app --args /path/to/file`)
* opening a second project should open a new instance of RStudio

### Checklist

### DIFF
--- a/src/node/desktop/forge.config.js
+++ b/src/node/desktop/forge.config.js
@@ -46,7 +46,7 @@ const config = {
   packagerConfig: {
     icon: './resources/icons/RStudio',
     appBundleId: 'com.rstudio.desktop',
-    extendInfo: '../../cpp/desktop/Info.plist.in',
+    extendInfo: './Info.plist',
   },
 };
 
--- a/src/node/desktop/forge.config.js
+++ b/src/node/desktop/forge.config.js
@@ -46,6 +46,7 @@ const config = {
   packagerConfig: {
     icon: './resources/icons/RStudio',
     appBundleId: 'com.rstudio.desktop',
+    extendInfo: '../../cpp/desktop/Info.plist.in',
   },
 };
 
--- a/src/node/desktop/.gitignore
+++ b/src/node/desktop/.gitignore
@@ -139,6 +139,9 @@ package/
 # Electron-Forge
 out/
 
+# Generated plist during CMake
+Info.plist
+
 .prettier_and_git_ignores
 
 # End of https://www.toptal.com/developers/gitignore/api/node
--- a/src/node/desktop/CMakeLists.txt
+++ b/src/node/desktop/CMakeLists.txt
@@ -117,6 +117,10 @@ endif()
 
 set(ELECTRON_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}" CACHE INTERNAL "")
 if (APPLE)
+   # configure Info.plist
+   configure_file (${CMAKE_CURRENT_SOURCE_DIR}/Info.plist.in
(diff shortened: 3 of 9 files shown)


src/node/desktop/forge.config.js
```diff
@@ -46,6 +46,7 @@ const config = {
   packagerConfig: {
     icon: './resources/icons/RStudio',
     appBundleId: 'com.rstudio.desktop',
+    extendInfo: '../../cpp/desktop/Info.plist.in',
```
gtritchie, line 49: We can't directly use this file, it needs to be consumed after cmake has generated the output file (Info.plist) via [configure_file](https://cmake.org/cmake/help/latest/command/configure_file.html#command:configure_file). Currently this doesn't happen for the Electron build (it's in the Qt-desktop branch of the cmakefile): https://github.com/rstudio/rstudio/blob/88d5f9b1359d261f56a8721ca1bc03d13fbc9701/src/cpp/desktop/CMakeLists.txt#L423-L424

There are some substitutions that happen in this file, so we do need to let cmake process it before we consume it. One example: https://github.com/rstudio/rstudio/blob/88d5f9b1359d261f56a8721ca1bc03d13fbc9701/src/cpp/desktop/Info.plist.in#L270

gtritchie: We could either create another copy of this file that lives in the electron project, and have the electron cmake file process this, or do it from the qt desktop location. Eventually we'll be removing the Qt folder altogether, but so we'll need to resolve such overlaps, but this wouldn't be the first one.

gtritchie (changes requested):

src/node/desktop/forge.config.js
gtritchie: Note that we already do a bit of Electron stuff from the Qt desktop folder, see the end of the CMakelists.txt file: https://github.com/rstudio/rstudio/blob/88d5f9b1359d261f56a8721ca1bc03d13fbc9701/src/cpp/desktop/CMakeLists.txt#L681

timtmok: Ah I missed that. Makes sense to have it go through cmake before feeding it to the Electron packaging.

timtmok: It's running through CMake first now and I see that the package now has the `Info.plist` with the variables filled in.

kevinushey:
> Listens to the open-file event which triggers for associated file open events and dragging files on to the app icon. For project files, this sets the env var before the session is created so that the project is opened.

Do you know if this will also work for Windows?

timtmok:
> > Listens to the open-file event which triggers for associated file open events and dragging files on to the app icon. For project files, this sets the env var before the session is created so that the project is opened.
> 
> Do you know if this will also work for Windows?

No, this event is for MacOS only. However, the changes to handle the command-line arguments should also apply to Windows but I haven't tested it.

gtritchie (approved):
LGTM!

PR #11014, opened 2022-04-19
Ignore duplicate key errors for postgres and sqlite for revoked cookies

jeffvroom:
In a load balancing scenario, the UserSession from more than one server
may end up storing the session same session cookie. This is particularly
true when one rserver routes the request to another for load balancing...
in that case, both rservers will get the same auth cookie to represent
that session.

Now that we invalidate all auth cookies when a user signs out, on all
servers in the cluster, that means some of them will try to insert a
record into the DB that's already there. We just need to ignore those
specific errors.

Part of the fix for: https://github.com/rstudio/rstudio-pro/issues/3287


melissa-barca (approved):
LGTM!

PR #11025, opened 2022-04-20
Sync cpp lib launcher

kfeinauer:
### Intent

Sync the C++ libraries between the rstudio and Launcher repos.

### Approach

On our side, we will use some merge strategies to rewrite the histories of the rstudio-pro repo to extract out just the library code changes, and then pull them into the Launcher code base regularly (see https://github.com/rstudio/launcher/blob/6e2227de2e1e0939317740aee1ef503ecf17c230/sync-repo.sh). The changes made here sync up the two codebases to ensure that when a sync is done, nothing is overwritten in the Launcher repo.

Changes had to be made to unify things, such as the log location (which differed between the codebases). This is unified via a CMake variable which is set in files that are not shared within the libraries themselves.

Similarly, we do not support any database functionality as of yet, so we conditionally skip compilation in the Launcher case by not setting the required CMake variable.

The change to options parsing was made to make it obvious when a user has specified a problematic option on the command line, instead of in a config file. This was introduced here, which I failed to upstream: https://github.com/rstudio/launcher/commit/98505fdf8202b489f04695664aa57da0ec2cac76

### Automated Tests

N/A

### QA Notes

N/A - development/build only.

### DIFF
--- a/CMakeGlobals.txt
+++ b/CMakeGlobals.txt
@@ -64,6 +64,8 @@ else()
    set(RSTUDIO_PRO_BUILD 0)
 endif()
 
+set(RSTUDIO_DEFAULT_LOG_PATH "/var/log/rstudio/rstudio-server")
+
 # detect architecture
 if(UNIX)
 
--- a/CMakeGlobals.txt
+++ b/CMakeGlobals.txt
@@ -64,7 +64,9 @@ else()
    set(RSTUDIO_PRO_BUILD 0)
 endif()
 
-set(RSTUDIO_DEFAULT_LOG_PATH "/var/log/rstudio/rstudio-server")
+if (NOT DEFINED RSTUDIO_DEFAULT_LOG_PATH)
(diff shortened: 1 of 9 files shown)


CMakeGlobals.txt
```diff
@@ -64,6 +64,8 @@ else()
    set(RSTUDIO_PRO_BUILD 0)
 endif()
 
+set(RSTUDIO_DEFAULT_LOG_PATH "/var/log/rstudio/rstudio-server")
```
kevinushey, line 67: Should we guard this by something like `if(NOT DEFINED <...>)`? (in case someone needed to override this from the command line invocation of CMake)

kevinushey (approved):
LGTM!

CMakeGlobals.txt
kfeinauer: Good idea. I'll make that change.

kfeinauer:
@MariaSemple Will wait for your feedback before merging.

Issue #10922, opened 2022-04-06
Can't create a Quarto document on Windows with bundled binaries


ronblum:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2022.06.0-daily+272
    OS Version      : Windows 11
    R Version       : 4.1.3 & 4.2.0

### Steps to reproduce the problem

This is on Windows only.

1. Don't install the Quarto CLI from https://quarto.org. If it's installed, remove it.
2. In RStudio Desktop, try to open a new Quarto document.

### Describe the problem in detail

The document fails to open. Instead, the following pop up appears:

![demo](https://user-images.githubusercontent.com/31009952/162063355-fdbd93e7-3e51-4119-9127-dca7fe1d51f3.png)

Going to the executables: `quarto.cmd` and `demo.exe` both result in the pop up and fail to run.

### Describe the behavior you expected

The Quarto document should be created.

Note that this works fine when Quarto is installed from https://quarto.org.

Triage note: Is this something we need to consider for a backport to Prairie Trillium?

- [ ] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ ] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [ ] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
RStudio installs various runtime libraries it requires on Windows here:

https://github.com/rstudio/rstudio/blob/7a6d81f19b465f5c74efc75e8ad0635f12e18980/src/cpp/session/CMakeLists.txt#L797-L801

https://github.com/rstudio/rstudio/blob/7a6d81f19b465f5c74efc75e8ad0635f12e18980/src/cpp/desktop/CMakeLists.txt#L519-L521

We might need to add vcruntime140_1.dll to the list. (I'm a little surprised this isn't distributed by default with Windows, though.)

timtmok:
I tried installing build 309 and it looks like it installs `vcruntime140_1.dll` already. I also see it packaged in the archive version of the build as well.

![image](https://user-images.githubusercontent.com/9591545/163819825-6a1d68b0-20f9-4c35-943a-2ca16ee464d6.png)

timtmok:
[...]

PR #11034, opened 2022-04-21
Include missing DLLs in Electron Windows package

timtmok:
### Intent
Address #10922 where `vcruntime140_1.dll` is not included in the package.

### Approach
I'm not exactly sure why `CMAKE_INSTALL_SYSTEM_RUNTIME_LIBS_SKIP` produces the correct output. I tried without it and it did the correct thing but also placed a duplicate `bin` with the same files in package root. It appears to have the correct DLLs in `resources/app/bin/x86`:

image

### Automated Tests
None

### QA Notes
The Windows Electron package should have the `vcruntime140_1.dll` in `resources/app/bin/x86` of the install directory.

Creating a Quarto document still works (there appears to be a check that there must be a valid Quarto install _and_ a Quarto document in the existing project in order to enable File > New File > Quarto Document).

The new file button in the Files pane always allows creating a Quarto document.

### Checklist

### DIFF
--- a/src/node/desktop/CMakeLists.txt
+++ b/src/node/desktop/CMakeLists.txt
@@ -191,6 +191,14 @@ if(WIN32)
       DESTINATION "${RSTUDIO_INSTALL_ELECTRON}"
    )
 
+   set(CMAKE_INSTALL_SYSTEM_RUNTIME_LIBS_SKIP TRUE)
+
+   include(InstallRequiredSystemLibraries)
+   install(
+      PROGRAMS ${CMAKE_INSTALL_SYSTEM_RUNTIME_LIBS}
+      DESTINATION "${RSTUDIO_INSTALL_BIN}/x86"
+   )
+
 elseif(LINUX)
 
    if(UNAME_M STREQUAL aarch64)


src/node/desktop/CMakeLists.txt
```diff
@@ -191,6 +191,14 @@ if(WIN32)
       DESTINATION "${RSTUDIO_INSTALL_ELECTRON}"
    )
 
+   set(CMAKE_INSTALL_SYSTEM_RUNTIME_LIBS_SKIP TRUE)
```
kevinushey, line 194: Can we add a comment why we're doing this here?

kevinushey: I also wonder if we need this for https://github.com/rstudio/rstudio/blob/64ee43493bae9754916b9f18f91fe79a6ba9b8a9/src/cpp/desktop/CMakeLists.txt#L519-L521.

kevinushey: Ah, the CMake documentation at https://cmake.org/cmake/help/latest/module/InstallRequiredSystemLibraries.html makes it more clear:

> CMAKE_INSTALL_SYSTEM_RUNTIME_LIBS_SKIP
> Set to TRUE to skip calling the [install(PROGRAMS)](https://cmake.org/cmake/help/latest/command/install.html#command:install) command to allow the includer to specify its own install rule, using the value of CMAKE_INSTALL_SYSTEM_RUNTIME_LIBS to get the list of libraries.

In other words, we set that if we want to provide our own `install()` command.

kevinushey:
https://github.com/rstudio/rstudio/commit/3eeba2e4c0507c332c0f835c0f79a6d6aee4c2d9

Issue #10208, opened 2021-12-08
Make tab/pane names localizable


gtritchie:
The tab/pane names are currently hardcoded. Investigate how to make these localizable (there may be some non-obvious assumptions in the code that the names are fixed that I want to check on).

Specific things to check include:

PaneManager -> "Console"
BuildPane -> "Build"
BuildTab -> "Build"
ConnectionsTab -> "Connections"
ConnectionsPane -> "Connections"
ConsolePane -> "Console"
FilesTab -> "Files"
Help -> "Help"
HelpPane -> "Help"
HelpTab -> "Help"
HistoryTab -> "History"
HistoryPane -> "History"
ViewerTab -> "Viewer"
ViewerPane -> "Viewer"
TutorialTab -> "Tutorial"
TutorialPane -> "Tutorial"
TerminalTabPresenter -> ...
TerminalTab -> "Terminal"
TerminalPane -> "Terminal"
Presentation2Tab -> "Presentation"
Presentation2Pane -> "Presentation"
PresentationPane -> "Presentation"
PresentationFrame -> "Presentation Frame"
PlotsTab -> "Plots"
PlotsPane -> "Plots"
PackagesTab -> "Packages"
PackagesPane -> "Packages"
TestsOutputTab -> "Tests"
TestsOutputPresenter -> "Tests"
SourceCppOutputTab -> "Source Cpp"
SourceCppOutputPane -> "Source Cpp"
RSConnectDeployOutputTab -> "Deploy"
RSConnectDeployOutputPresenter -> "Deploy"
RenderRmdOutputTab -> "Render"
MarkersOutputTab -> "Markers"
MarkersOutputPane -> "Markers"
FindOutputTab -> "Find in Files"
FindOutputPane -> "Find Results"
DataOutputTab -> "SQL Results"
DataOutputPane -> "Data Output"
CompilePdfOutputTab -> "Compile PDF"
CompilePdfOutputPresenter -> "Compile PDF"
LauncherJobsPane -> "Launcher"
JobsPane -> "Jobs"
JobsTab -> "Jobs"
LauncherJobsTab -> "Launcher"
EnvironmentPane -> "Environment"
EnvironmentTab -> "Environment"

rohank07:
@gtritchie wanted to follow up on this.

gtritchie:
@rohank07 missed your ping on this, sorry. Have added to the Spotted Wakerobin milestone.

PR #11060, opened 2022-04-25
Make workbench tab labels localizable

gtritchie:
### Intent

Addresses #10208, making the IDE tab labels localizable (e.g. "Environment", "Console", and so forth).

This does not tackle making the tab names in the Pane Layout section of Global Options localizable; those remain hardcoded in English (using the internal tab identifiers, which in English are the same as the labels, but not true in non-English). This PR is already huge, so will open a separate tracking issue on that (#11061).

### Approach

Use localizable strings for the labels when tabs are being created, and created and used constants for the internal tab name identifiers to make it clearer what's intended for the UI versus internal.

### Automated Tests

We don't have tests for non-English UI, but the existing tests will help confirm nothing has been regressed in English.

### QA Notes

Nothing to test specificially; you can switch into French UI and confirm that tab names are now localized, but mainly we are concerned with ensuring we haven't broken anything in English, and that is adequately covered by existing automation and regular usage.

### Checklist


src/gwt/src/org/rstudio/studio/client/workbench/ui/PaneManager.java
```diff
@@ -1871,37 +1873,37 @@ private String getTabLabel(Tab tab)
...
+      if (name.equalsIgnoreCase(ENVIRONMENT_PANE))
          return Tab.Environment;
-      if (name.equalsIgnoreCase("viewer"))
+      if (name.equalsIgnoreCase(VIEWER_PANE))
          return Tab.Viewer;
-      if (name.equalsIgnoreCase("connections"))
+      if (name.equalsIgnoreCase(CONNECTIONS_PANE))
          return Tab.Connections;
-      if (name.equalsIgnoreCase("source"))
+      if (name.equalsIgnoreCase(UserPrefsAccessor.Panes.QUADRANTS_SOURCE))
```
kevinushey, line 1902: Small question: why do we need the extra package names prefixes here, compared to the other entries? (Why not have a separate `SOURCE_PANE = "Source"` entry below?)

src/gwt/src/org/rstudio/studio/client/workbench/views/jobs/JobsConstants_fr.properties
```diff
@@ -32,4 +33,5 @@ jobListLabel=Les {0} travaux suivants sont toujours en cours d''exécution.
 terminateRunningJobsCaption=Terminer les travaux en cours d''exécution
 terminateJobsCaption=Terminer les travaux
 jobsTabLabel=Onglet des travaux
-viewAllJobsTitle=Voir tous les travaux
\ No newline at end of file
+viewAllJobsTitle=Voir tous les travaux
+launcherTitle=Launcher
```
kevinushey, line 37: Is this the right French translation?

kevinushey (approved):
LGTM! A couple of questions just to double-check I understand what's going on.

src/gwt/src/org/rstudio/studio/client/workbench/views/jobs/JobsConstants_fr.properties
gtritchie: No, but that's ok. These will all get reviewed and (re)-localized when we hand it off. I've made an effort to provide the French strings in some cases, but otherwise it's fine for us to just put in English as a placeholder (in general).

src/gwt/src/org/rstudio/studio/client/workbench/ui/PaneManager.java
gtritchie: I was just using existing constants we already had for the quadrants (tabset1, source, etc.), but agree it's not pretty (or consistent).

PR #11062, opened 2022-04-26
Locale Cookie Handled by Electron

matheustavaresdev:
### Intent

Detect language changes from GWT and update Electron's texts based on it's set cookie.

### Approach

Fetching the cookies and setting them as a `Window LocalStorage Item` when a `BrowserWindow` is created or if the cookies are updated at any given point.

### Automated Tests

None

### QA Notes

Change the language by going to `Tools > Global Options > Advanced`. The screen should refresh. To test some other pages like Error, just give the `--session-exit` flag when running the app.

### Checklist


src/node/desktop/src/assets/locales/fr.json
```diff
@@ -0,0 +1,113 @@
...
+    "paste": "Paste",
+    "selectAll": "Select All",
+    "inspectElement": "Inspect Element",
+    "errorDownloadingImage": "Error downloading image",
+    "downloadCancelledMessage": "The download was cancelled.",
+    "downloadInterruptedMessage": "The download was interrupted.",
+    "save": "Save"
+  },
+  "chooseRDialog": {
+    "documentTitle": "Choose R Installation",
```
gtritchie, line 25: We will get this file localized before the release. In the meantime, let's change a few of these strings so we can tell if the "French" ones are being used (for testing of real builds). Especially for the "other" windows like "Choose R" and the Error page. Something like "FR Choose R Installation FR".

gtritchie (changes requested):

src/node/desktop/src/assets/locales/fr.json
matheustavaresdev: Done on [b155366](https://github.com/rstudio/rstudio/pull/11062/commits/b155366f62316119db252025d40652f3c3db7e26)

gtritchie (approved):
LGTM!

PR #11077, opened 2022-04-27
install same cmake for debian10 used by other Linux builds

gtritchie:
### Intent

When I made the debian10 dockerfile I decided to install cmake using `apt` instead of building a specific version from source like we do on other platforms.

For consistency (and especially now that we're building a relatively new version of cmake) switch debian10 to same technique.

I will do the same for the monitor dockerfile (Pro) via separate PR.

### Approach

Same approach as other dockerfiles.

### Automated Tests

None

### QA Notes

Build mechanics, nothing to test.

### Checklist


kevinushey (commented):
Belated LGTM!

PR #11083, opened 2022-04-28
Fix backend yaml chunk evaluation of options

jgutman:
### Intent

Addresses #10645 so that YAML style chunk options are evaluating when executing a chunk

### Approach

This PR includes only backend changes so the YAML chunk options are respected and combined with the standard chunk options when executing a chunk inline. This does not affect the UI of the ChunkOptionsPopupPanel -- however, upon investigation that would require a lot more lift for seemingly minimal benefit, and there is no clear way forward, as the LineWidget really only ever parses the first line of a chunk on the frontend. Given this, and given that only a very small set of chunk options are even considered in the PopupPanel (see below), I'm not sure the frontend changes are worth doing at this time. We can always revisit if this becomes important to users

image

### Automated Tests
None

### QA Notes

Run chunks with both YAML and standard style chunk options, and compare behavior.

```
```{r}
#| eval: false
  
# doesn't runs with "Run all chunks above", correct behaviour
1 + 1
```

```{r eval = FALSE}
  
# doesn't run with "Run all chunks above", correct behaviour
2 + 2
```

```{r}
#| warning: false
#| message: false
# doesn't output warnings or messages, correct behaviour

library(ggplot2)
ggplot(airquality, aes(Temp, Ozone)) + 
        geom_point() + 
        geom_smooth(method = "loess", se = FALSE)
```

### Checklist

### DIFF
--- a/src/cpp/session/modules/SessionRmdNotebook.R
+++ b/src/cpp/session/modules/SessionRmdNotebook.R
@@ -627,16 +627,16 @@ assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
         opts <- .rs.fromYAML(opt)
         # parse truthy and falsy style yaml options
         opts <- lapply(opts, function(value)
-                             {
-                               if (!is.character(value))
-                                 return(value)
-                               if (tolower(value) %in% c("y", "yes", "on"))
-                                 return(TRUE)
-                               if (tolower(value) %in% c("n", "no", "off"))
-                                 return(FALSE)
-                               else
-                                 return(value)
-                             })
+           {
+             if (!is.character(value))
+               return(value)
+             if (tolower(value) %in% c("y", "yes", "on"))
+               return(TRUE)
+             if (tolower(value) %in% c("n", "no", "off"))
+               return(FALSE)
+             else
+               return(value)
+           })
     }
     return(opts)
 })
@@ -659,7 +659,15 @@ assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
         break
     yamlChunkOpts <- c(yamlChunkOpts, match)
   }
-  opts <- .rs.parseYamlOpt(yamlChunkOpts)
+  
+  if (length(yamlChunkOpts) > 0)
+  {
+     opts <- tryCatch(
+     .rs.parseYamlOpt(yamlChunkOpts),
+     error = function(e) {
+        warning("Failed to parse chunk options in body:\n", e)
+     })
+  }
 
   # strip chunk indicators if present
   matches <- unlist(regmatches(rmdChunkOpts, regexec(.rs.reRmdChunkBegin(), rmdChunkOpts)))
--- a/src/cpp/session/modules/SessionRmdNotebook.R
+++ b/src/cpp/session/modules/SessionRmdNotebook.R
@@ -38,6 +38,12 @@ assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
    "^[\t >]*```+\\s*$"
 })
 
+.rs.addFunction("reYamlOptChunkBegin", function()
+{
+   "^#\\| [a-zA-Z_-]+:\\s*.+$"
+})
+
+
 .rs.addFunction("rnb.cachePathFromRmdPath", function(rmdPath)
 {
    .Call("rs_chunkCacheFolder", rmdPath)
@@ -608,41 +614,67 @@ assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
    .rs.fromJSON(.rs.base64decode(encoded))
 })
 
+.rs.addFunction("parseYamlOpt", function(opt)
+{
+    opt <- sub("^#\\|\\s*", "", opt)
+    opt <-  .rs.trimWhitespace(unlist(strsplit(opt, ":\\s*")))
+    key <- opt[[1]]
+    val <- opt[[2]]
+    opts <- list()
+    opts[key] <- val
+    return(opts)
+})
+
 .rs.addFunction("evaluateChunkOptions", function(code)
 {
   opts <- list()
 
   Encoding(code) <- "UTF-8"
-  # if several lines of code are passed, operate only on the first 
-  code <- unlist(strsplit(code, "\n", fixed = TRUE))[[1]]
+  # if several lines of code are passed, we expect the first line to contain the standard Rmd chunk options
+  # the next several lines may or may not contain YAML-style chunk options
+  code <- unlist(strsplit(code, "\n", fixed = TRUE))
+  rmdChunkOpts <- code[[1]]
+  yamlChunkOpts <- code[-1]
+
+  for (line in yamlChunkOpts) {
+    matches <- unlist(regmatches(line, regexec(.rs.reYamlOptChunkBegin(), line)))
+    # there may be variable number of yaml chunk opts, but they should all be at the top of the chunk
+    if (length(matches) == 0)
+        break
+    opts <- append(opts, .rs.parseYamlOpt(matches))
+  }
 
   # strip chunk indicators if present
-  matches <- unlist(regmatches(code, regexec(.rs.reRmdChunkBegin(), code)))
+  matches <- unlist(regmatches(rmdChunkOpts, regexec(.rs.reRmdChunkBegin(), rmdChunkOpts)))
   if (length(matches) > 1)
-    code <- matches[[2]]
+    rmdChunkOpts <- matches[[2]]
 
   tryCatch({
     # if this is the setup chunk, it's not included by default
     setupIndicator <- "r setup"
-    if (identical(substring(code, 1, nchar(setupIndicator)), 
+    if (identical(substring(rmdChunkOpts, 1, nchar(setupIndicator)),
                   setupIndicator)) {
       opts$include <- FALSE
     }
 
     # extract and remove engine name (can be overridden below with engine=)
-    opts$engine <- unlist(strsplit(code, split = "(\\s|,)+"))[[1]]
-    code <- substring(code, nchar(opts$engine) + 1)
+    opts$engine <- unlist(strsplit(rmdChunkOpts, split = "(\\s|,)+"))[[1]]
+    rmdChunkOpts <- substring(rmdChunkOpts, nchar(opts$engine) + 1)
 
     # parse them, then merge with the defaults (evaluate in global environment)
     opts <- .rs.mergeLists(opts,
-                           eval(substitute(knitr:::parse_params(code)),
+                           eval(substitute(knitr:::parse_params(rmdChunkOpts)),
                                 envir = .GlobalEnv))
     
-    # convert T, F to TRUE, FALSE as appropriate
+    # convert T, F (knitr chunk style) or "true", "false" (yaml chunk style) to TRUE, FALSE as appropriate
     opts <- lapply(opts, function(opt) {
       if (identical(opt, as.name("T")))
         TRUE
+      else if (identical(opt, "true"))
+        TRUE
       else if (identical(opt, as.name("F")))
+        FALSE
+      else if (identical(opt, "false"))
         FALSE
       else
         opt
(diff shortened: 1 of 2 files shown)


src/cpp/session/modules/SessionRmdNotebook.R
```diff
@@ -38,6 +38,12 @@ assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
    "^[\t >]*```+\\s*$"
 })
 
+.rs.addFunction("reYamlOptChunkBegin", function()
+{
+   "^#\\| [a-zA-Z_-]+:\\s*.+$"
```
kevinushey, line 43: One thing we'll have to be careful of here is that the inline chunk options syntax is _very_ liberal. From https://yihui.org/knitr/options/#chunk-options, it looks like you could even just include the chunk header inline:

`````
```{r}
#| my-chunk, echo = FALSE, fig.width = 10,
#| fig.cap = "This is a long long
#|   long long caption."

plot(cars)
```
`````

Hopefully this will not be common!

Comment (kevinushey) [reply src/cpp/session/modules/SessionRmdNotebook.R:43]: More concretely, I think we need to adjust the pattern here to tolerate:

- Option names containing a `.`, e.g. `fig.width`
- Comments placed after the option value.

An example test case:

`````
---
title: "Untitled"
output: html_document
date: "2022-04-28"
---

```{r}
#| eval: true     # this is a comment
#| fig.width: 42  # this is also a comment
knitr::opts_current$get("fig.width")
```
`````

File: src/cpp/session/modules/SessionRmdNotebook.R [hunk 2]
```diff
@@ -608,42 +614,68 @@ assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
    .rs.fromJSON(.rs.base64decode(encoded))
 })
 
+.rs.addFunction("parseYamlOpt", function(opt)
```
kevinushey, line 617: FWIW, we have `.Call("rs_fromYAML")` if you need to parse a YAML string into an R object. Would that be helpful here?

I think you could just trim off the `#| ` prefix and then use that to parse the whole thing together.

src/cpp/session/modules/SessionRmdNotebook.R
```diff
@@ -608,42 +614,68 @@ assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
...
-    # convert T, F to TRUE, FALSE as appropriate
+    # convert T, F (knitr chunk style) or "true", "false" (yaml chunk style) to TRUE, FALSE as appropriate
     opts <- lapply(opts, function(opt) {
       if (identical(opt, as.name("T")))
         TRUE
+      else if (identical(opt, "true"))
+        TRUE
       else if (identical(opt, as.name("F")))
         FALSE
+      else if (identical(opt, "false"))
```
kevinushey, line 677: Do we also need to handle other YAML "truthy" and "falsy" values here? https://yaml.org/type/bool.html

src/cpp/session/modules/SessionRmdNotebook.R
jgutman: Oh, I was reviewing those docs but I don't think I even saw that you could have multiple options per line, I'll have to double check as I may not have accounted for that

src/cpp/session/modules/SessionRmdNotebook.R
jgutman: Good question, I have no idea. I'll check if they are recognized when knitting

src/cpp/session/modules/SessionRmdNotebook.R
jgutman: Yeah, so based on the documentation you sent, the string could be in two different format styles, R and YAML



```
#| tag = Rvalue, tag = Rvalue, tag = Rvalue
```
or
```
#| tag: value
#| tag: value
```

kevinushey: Right, but _hopefully_ users will prefer to use YAML rather than R code here. I don't know if it's reasonable to try-catch with a YAML parser first and fall back to an R parser.

src/cpp/session/modules/SessionRmdNotebook.R
jgutman: It appears that truthy and falsy values are recognized when knitting, however they are not recognized by `.Call("rs_fromYAML")` . Maybe we can update that function to handle these values directly

jgutman: Okay, it looks like these truthy / falsy values are supported in YAML standard 1.1 but not YAML standard 1.2. If yaml-cpp is not automatically converting these values, I'm not sure if we should be here; but we can add some handling specifically in SessionRmdNotebook.R so that it matches the knitr behavior

src/cpp/session/modules/SessionRmdNotebook.R
```diff
@@ -608,37 +614,76 @@ assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
    .rs.fromJSON(.rs.base64decode(encoded))
 })
 
+.rs.addFunction("parseYamlOpt", function(opt)
+{
+    opt <- sub("^#\\|\\s*", "", opt)
+    if (grepl(".+=.+", opt)) # R-style chunk options
+        opts <- knitr:::parse_params(opt)
```
jgutman, line 621: Oops, need to update this to check for comments

jgutman: Actually, @kevinushey inline comments don't seem to be accepted by knitr either, so I'm not sure that we need to support comments. Where did you see that inline comments starting with `#` would be supported?

jgutman: However, multiline chunk options are supposed to be supported, so I think I need to add that support here

```
#| warning = F, message = F, fig.width = 8,
#| fig.cap = "This is a long long
#|   long long caption."
```

jgutman: Okay, added support for multiline, but not for comments, which I don't think are actually meant to be supported

kevinushey: I was using this as an example:

`````
---
title: "Use R Markdown"
output: html_document
---

```{r}
#| eval: false  # this is a comment
print("Hello, world!")
```
`````

When I render this R Markdown document, the `eval: false` option is parsed and read, and the associated chunk is not evaluated.

```
label: unnamed-chunk-1 (with options) 
List of 1
 $ eval: logi FALSE
```

Comment (jgutman) [reply src/cpp/session/modules/SessionRmdNotebook.R]: Weird because I tried that and got an error... I'll check again

Comment (jgutman) [reply src/cpp/session/modules/SessionRmdNotebook.R]: Oh I think I tried it with the R style chunk options, `#| eval = FALSE # this is a comment`

Comment (jgutman) [reply src/cpp/session/modules/SessionRmdNotebook.R]: Okay -- so in-line comments work with YAML format but not with R code format. This is true both when rendered and when running code in the console (with run all chunks above). No need for specific further changes to the code to support the inline comments but will make sure to add that to QA notes

File: src/cpp/session/modules/SessionRmdNotebook.R [hunk 5]
```diff
@@ -608,37 +614,80 @@ assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
...
-                           eval(substitute(knitr:::parse_params(code)),
+                           eval(substitute(knitr:::parse_params(rmdChunkOpts)),
                                 envir = .GlobalEnv))
+
+    # if a chunk option is provided in both body (yaml chunk options) and header (rmdChunkOpts)
+    # the body should override the header option
+    opts <- opts[unique(names(opts))]
     
-    # convert T, F to TRUE, FALSE as appropriate
+    # convert T, F (knitr chunk style) or "true", "false" (yaml chunk style) to TRUE, FALSE as appropriate
```
jgutman, line 690:
```suggestion
    # convert T, F for R code chunk options to TRUE, FALSE as appropriate
```

src/cpp/session/modules/SessionRmdNotebook.R
```diff
@@ -608,37 +614,80 @@ assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
...
+  yamlChunkOpts <- c()
+
+  for (line in code[-1]) {
+    match <- unlist(regmatches(line, regexec(.rs.reYamlOptChunkBegin(), line)))
+    # there may be variable number of yaml chunk opts, but they should all be at the top of the chunk
+    if (length(match) == 0)
+        break
+    yamlChunkOpts <- c(yamlChunkOpts, match)
+  }
+  opts <- .rs.parseYamlOpt(yamlChunkOpts)
```
kevinushey, line 662: Should we handle errors here somehow? (I think we at least need a tryCatch() that returns a sensible default on error; not sure if we need to log or report anything to the user)

src/cpp/session/modules/SessionRmdNotebook.R
```diff
@@ -608,37 +614,80 @@ assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
...
     # extract and remove engine name (can be overridden below with engine=)
-    opts$engine <- unlist(strsplit(code, split = "(\\s|,)+"))[[1]]
-    code <- substring(code, nchar(opts$engine) + 1)
+    opts$engine <- unlist(strsplit(rmdChunkOpts, split = "(\\s|,)+"))[[1]]
+    rmdChunkOpts <- substring(rmdChunkOpts, nchar(opts$engine) + 1)
 
     # parse them, then merge with the defaults (evaluate in global environment)
     opts <- .rs.mergeLists(opts,
-                           eval(substitute(knitr:::parse_params(code)),
+                           eval(substitute(knitr:::parse_params(rmdChunkOpts)),
```
kevinushey, line 683: Similar here; we should consider handling errors (if any) that occur when parsing params. (I realize we're repairing a deficiency from the old code, but it's worth being defensive given the new YAML option style)

src/cpp/session/modules/SessionRmdNotebook.R
```diff
@@ -608,37 +614,80 @@ assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
    .rs.fromJSON(.rs.base64decode(encoded))
 })
 
+.rs.addFunction("parseYamlOpt", function(opt)
+{
+    opt <- sub("^#\\|\\s*", "", opt)
+    if (any(grepl(".+=.+", opt))) # R-style chunk options
+    {
+        opts <- paste(opt, collapse=" ")
```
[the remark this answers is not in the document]
kevinushey: FWIW you can enable style diagnostics here if you want RStudio to give you some feedback regarding style lint like this. (see the option called Provide R style diagnostics)

Screen Shot 2022-05-05 at 1 54 05 PM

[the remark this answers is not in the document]
jgutman: Thanks, I have those all on, but was making changes quickly in my C++ IDE. Too many IDEs to choose from!

src/cpp/session/modules/SessionRmdNotebook.R
jgutman: Because we aren't really parsing the options one at a time, I don't think it'll be really possible to parse some options but ignore only the failing ones; however we can just ignore all chunk body options if they fail to parse for any reason

We could do something like 
```
if (length(yamlChunkOpts) > 0)
  {
     opts <- tryCatch(
     .rs.parseYamlOpt(yamlChunkOpts),
     error = function(e) {
        warning("Failed to parse chunk options in body:\n", e)
     })
  }
```

I don't have a good sense of whether those error messages would be helpful, but I think it's probably best to display something to the user so they know the options aren't just being silently ignored (as they are now)

src/cpp/session/modules/SessionRmdNotebook.R
jgutman: I realize I don't actually understand what this bit is doing, and how evaluating the chunk header in the global environment merges in the default chunk options. I could try to add some handling here but 1) I don't really understand what the original code is doing anyway and 2) this is currently only being applied to the chunk header, not the yaml option in-body style and 3) knitr:::parse_params has some error handling built in, so I'm not sure what we'd do here beyond just propagating those errors

Issue #3095, opened 2018-06-29
trim trailing whitespace from edited lines on saving


kevinushey:
This is a potential enhancement. The 'Strip horizontal whitespace when saving' preference can sometimes be too aggressive -- it can introduce whitespace diffs that can make the history (e.g. for `git blame`) more difficult to process on files which already contain lots of trailing whitespace.

However, cleaning up whitespace is still a definite nice-to-have. As a compromise, it would be useful if the IDE could track what lines have actually been touched by the user since the last save, and then trim whitespace only on those lines when forming a save. This behavior would be controlled by a preference, e.g:

```
Strip trailing horizontal whitespace when saving: [Always / Edited Lines / Never] 
```

PR #11090, opened 2022-04-28
Add Session Server RPC to open source

MariaSemple:
### Intent

First part of [Pro #3095](https://github.com/rstudio/rstudio-pro/issues/3095)

### Approach

* Pull the SocketRpc, ServerSessionRpc, and SessionServerRpc code into OS.
* Remove Pro specific components and add overlays.
* Ensure all three modules are correctly initialized.
* Add an RPC endpoint for testing purposes.

### Automated Tests

I will add a selenium test to call the `/server_version` RPC.

### QA Notes

The behaviour here will be different depending on the environment. A simple test would be to see if the Launcher tab still works with these changes.

### Checklist


src/cpp/session/modules/SessionOverlay.R
```diff
@@ -12,3 +12,10 @@
 # AGPL (http://www.gnu.org/licenses/agpl-3.0.txt) for more details.
 #
 #
+
+.rs.addFunction("invokeServerRpc", function(method, ...)
+{
+   # callback to session to invoke server RPC
+   args <- list(...)
+   .Call("rs_invokeServerRpc", method, .rs.scalarListFromList(args))
```
kevinushey, line 20: nit: use `PACKAGE = "(embedding)"` as otherwise `.Call()` will search all loaded DLLs for a routine called `rs_invokeServerRpc`, rather than just directly look for our own embedded routine.

MariaSemple: For sure, I can add that. I just copied this out of Pro into OS. Should we do a pass on the whole code base to make sure we're setting the package everywhere?

kevinushey: I think that's reasonable, although until now I've just tried to make that change piecemeal whenever I encounter it in code that I'm modifying (or stumble upon for other reasons).

AndrewMcClain (approved):
LGTM!

PR #11092, opened 2022-04-28
properly handle R rawstring state in multimode docs

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/11087.

### Approach

Multi-mode documents embed the highlight rules of other languages by prepending the associated state names with a language-specific prefix. For example, R highlight rules are embedded for R Markdown documents here:

https://github.com/rstudio/rstudio/blob/0a4befe8ef2003ea69b554f1ae2280141f6f6aaa/src/gwt/acesupport/acemode/rmarkdown_highlight_rules.js#L58-L66

Importantly, this is done by giving each "state" name from that highlight group a unique prefix, so that the highlight definitions from one language don't collide with those from another. However, this implies that highlight rules in those languages need to be careful, since the "raw" state names might be transformed in such cases.

### Automated Tests

N/A

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/11087.

### Checklist

### DIFF
--- a/src/gwt/acesupport/acemode/rainbow_paren_highlight_rules.js
+++ b/src/gwt/acesupport/acemode/rainbow_paren_highlight_rules.js
@@ -44,19 +44,30 @@ define("mode/rainbow_paren_highlight_rules", ["require", "exports", "module"], f
           return this.token;
         }
 
+        // NOTE: The 'stack' object here (which is really just an array) is
+        // shared by different highlight rules, and is a general way to set and
+        // persist state in the tokenizer. However, because it's visible to
+        // each matching highlight rule, and each individual highlight rule
+        // might want to manipulate that stack, we need to choose the index
+        // carefully to avoid stomping on state declared from other highlight
+        // rules. 15 is chosen as a "large enough" number to avoid stepping
+        // other rules, which normally would be manipulating the first few
+        // slots of the 'stack' array.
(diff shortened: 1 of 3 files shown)


src/gwt/acesupport/acemode/rainbow_paren_highlight_rules.js
```diff
@@ -44,19 +44,30 @@ define("mode/rainbow_paren_highlight_rules", ["require", "exports", "module"], f
...
         stack[0] = state;
-        stack[1] = stack[1] || 0;
+        stack[15] = stack[15] || 0;
 
         switch(val) {
 
         case "[": case "{": case "(":
-          this.token = "paren.paren_color_" + (stack[1] % $numParenColors);
-          stack[1] = stack[1] + 1;
+          this.token = "paren.paren_color_" + (stack[15] % $numParenColors);
```
jgutman, line 65: I truly have very little understanding of why this works, but indeed it does, so LGTM. I guess you're saying because there were collisions in `stack[1] % $numParenColors` so the wrong left and right token were being paired together, but with `stack[15]` we don't encounter these collisions? I had some trouble parsing out the above comment

Is it typical to reference an issue in the code comments

jgutman (approved):

PR #11094, opened 2022-04-29
Small text tweaks based on feedback

jgutman:
Follow up to #8255 -- small tweaks to text based on feedback.

image

Separate question based on other feedback from the demo session last week:

image

There was feedback that it was difficult to visually distinguish the Project Preferences Note in order to realize it was not another setting. Do you think it's okay as is, or would you suggest any other visual changes? Perhaps a tad more padding at the top between Keybindings and the note? We could indent it as well but that would make it look like it was nested under keybindings. We could also put the whole note in parentheses, like the python note is. Or maybe you feel it's still fine as is

### DIFF
--- a/src/cpp/session/resources/schema/user-prefs-schema.json
+++ b/src/cpp/session/resources/schema/user-prefs-schema.json
@@ -1152,7 +1152,7 @@
         "ssh_key_type": {
             "type": "string",
             "enum": ["ed25519", "rsa"],
-            "enumReadable": ["ED25519-encrypted key", "RSA-encrypted key"],
+            "enumReadable": ["ED25519", "RSA"],
             "default": "ed25519",
             "description": "The encryption type to use for the SSH key file."
         },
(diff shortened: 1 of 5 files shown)


kevinushey:
> There was feedback that it was difficult to visually distinguish the Project Preferences Note in order to realize it was not another setting. Do you think it's okay as is, or would you suggest any other visual changes? Perhaps a tad more padding at the top between Keybindings and the note? We could indent it as well but that would make it look like it was nested under keybindings. We could also put the whole note in parentheses, like the python note is. Or maybe you feel it's still fine as is

IMHO it's fine as-is, but we could try and see if some of these tweaks help. Of these options, I think a couple extra pixels of padding could be enough to "separate" it from the existing options.

kevinushey (approved):
LGTM! Up to you whether you want to merge this PR as-is and consider tweaks for the Global Options dialog separately, or add them to this PR.

PR #11095, opened 2022-04-29
Add Edit Project Prefs button

jgutman:
Since I added a Edit Project Preferences button on the Code tab, I thought I'd add one here too, next to the already existing note about pre-configured Python interpreters. Perhaps now that there's a button there we can edit the text slightly (instead of "Use the Project Options dialog...")

image


src/gwt/src/org/rstudio/studio/client/workbench/prefs/PrefsConstants.java
```diff
@@ -2633,11 +2633,11 @@ public interface PrefsConstants extends com.google.gwt.i18n.client.Messages {
     String pythonPreferencesText();
 
     /**
-     * Translated "(NOTE: This project has already been configured with its own Python interpreter. Use the Project Options dialog to change the version of Python used in this project.".
+     * Translated "(NOTE: This project has already been configured with its own Python interpreter. Use the Edit Project Preferences button to change the version of Python used in this project.".
```
gtritchie, line 2636: The Project Options dialog (and the menu item that launches it) is named "Project Options", not "Project Preferences". Would it be more consistent if the button said "Edit Project Options..." instead of "Preferences"?

We call them preferences internally, but the UI generally calls them options.

Not specific to this change, just an observation.

jgutman: Thanks for calling that out! I was getting confused by that, I can update in both places for consistency

src/gwt/src/org/rstudio/studio/client/workbench/prefs/PrefsConstants.java
```diff
@@ -2633,11 +2633,11 @@ public interface PrefsConstants extends com.google.gwt.i18n.client.Messages {
...
      *
-     * @return translated "(NOTE: This project has already been configured with its own Python interpreter. Use the Project Options dialog to change the version of Python used in this project."
+     * @return translated "(NOTE: This project has already been configured with its own Python interpreter. Use the Edit Project Preferences button to change the version of Python used in this project."
      */
-    @DefaultMessage("(NOTE: This project has already been configured with its own Python interpreter. Use the Project Options dialog to change the version of Python used in this project.")
+    @DefaultMessage("(NOTE: This project has already been configured with its own Python interpreter. Use the Edit Project Preferences button to change the version of Python used in this project.")
```
kevinushey, line 2640: Nit: missing a closing `)`

src/gwt/src/org/rstudio/studio/client/workbench/prefs/PrefsConstants.java
gtritchie: Plus "options" is shorter so the button won't be quite as big.

gtritchie (approved):
LGTM!

PR #11143, opened 2022-05-05
Fix SW bugs in Find in Files path and find in files replacements

jgutman:
### Intent

Fix bugs described in #11041 and #11125

### Approach

1. We previously added two new git grep related params to the begin Find RPC calls, but not to the preview Replace and complete Replace calls, leading to parameter mismatch errors when attempting to do replacements with regex selected (the replace methods are actually only called when regex is True)
2. We previously stripped off the leading `./` from non git grep search result paths, when instead we should have stripped only the `.`, which resulted in a missing separator `/` between the search directory and the result path

### Automated Tests

Will check existing tests

### QA Notes

Find in Files search and replace with regex selected


kevinushey (approved):
LGTM, thanks!

PR #11151, opened 2022-05-06
Bugfix/followup native pipe default

jgutman:
Follow up to #9409 (PR #10814). 

During QA it was found that there were some issues around how it handled changing the global default for the `insert_native_pipe_operator` preference when a project was active and the project setting was set to (Default).

Fixed that here by not writing anything to the project prefs layer when the project setting is Default, instead of previous behavior which wrote the global user preference to the project layer. Initial testing shows the issue seems to be resolved, but will rely on initial QA plan and point of failure from #9409 . 

Also removed some unnecessary code and did some minor refactoring when I was trying to figure out what was going on here.

### DIFF
--- a/src/cpp/session/projects/SessionProjectContext.cpp
+++ b/src/cpp/session/projects/SessionProjectContext.cpp
@@ -767,24 +767,6 @@ void ProjectContext::updatePackageInfo()
    }
 }
 
-bool useNativePipeOption(int useNativePipeOperator)
-{
-   // allow project override
-   switch(useNativePipeOperator)
-   {
-      case r_util::YesValue:
-         return true;
-      case r_util::NoValue:
-         return false;
-      default:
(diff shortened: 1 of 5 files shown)


src/cpp/session/projects/SessionProjectContext.cpp
```diff
@@ -767,32 +767,18 @@ void ProjectContext::updatePackageInfo()
...
-         return true;
-      case r_util::NoValue:
-         return false;
-      default:
-         // fall through
-         break;
-   }
-
-   // no project override
-   return prefs::userPrefs().insertNativePipeOperator();
```
jgutman, line 785: This fallback seemed to be the source of the bug

kevinushey (approved):
LGTM; thanks for tracking this down!

PR #11154, opened 2022-05-07
Fix jsonEscapeString() in SessionCodeTools.R

okkn:
### Intent

If `x` is a multibyte character, `charToRaw(x)` returns a vector longer than 1. Since R 4.2.0 or later, calling if() with a condition of length greater than 1 gives an **error** rather than a warning (https://cran.r-project.org/doc/manuals/r-release/NEWS.html), so the L1218 `else if (charToRaw(x) < 20)`  in `jsonEscapeString()` in SessionCodeTools.R should be corrected.

https://github.com/rstudio/rstudio/blob/438f4c08d77b1424a9b3bb193d16060cdae06aaa/src/cpp/session/modules/SessionCodeTools.R#L1211-L1226

In fact, due to this bug, R 4.2.0 sometimes fails to load R Notebooks (which are probably contain multibyte characters) with an error when opening them.

### Approach

Evaluate whether `charToRaw(x)` is less than 20 only when the length of `x` is equal to 1.

### Automated Tests

### QA Notes

### Checklist

### DIFF
--- a/src/cpp/session/modules/SessionCodeTools.R
+++ b/src/cpp/session/modules/SessionCodeTools.R
@@ -1215,7 +1215,7 @@
    chars <- vapply(chars, function(x) {
       if (x %in% c('"', '\\', '/'))
          paste('\\', x, sep = '')
-      else if (charToRaw(x) < 20)
+      else if (length(charToRaw(x)) == 1 && charToRaw(x) < 20)
          paste('\\u', toupper(format(as.hexmode(as.integer(charToRaw(x))),
                                      width = 4)),
                sep = '')
--- a/src/cpp/session/modules/SessionCodeTools.R
+++ b/src/cpp/session/modules/SessionCodeTools.R
@@ -1215,7 +1215,7 @@
    chars <- vapply(chars, function(x) {
       if (x %in% c('"', '\\', '/'))
          paste('\\', x, sep = '')
-      else if (length(charToRaw(x)) == 1 && charToRaw(x) < 20)
+      else if (length(charToRaw(x)) == 1 && charToRaw(x) < 32)
          paste('\\u', toupper(format(as.hexmode(as.integer(charToRaw(x))),
                                      width = 4)),
                sep = '')


jgutman:
Hi, thanks very much for contributing! Have you reviewed and submitted a [contributor agreement](https://www.rstudio.com/wp-content/uploads/2014/06/rstudioindividualcontributoragreement.pdf)? 

In order to contribute code to this repo, we require:
>Ensure that you have signed the [individual](https://www.rstudio.com/wp-content/uploads/2014/06/rstudioindividualcontributoragreement.pdf) or [corporate](https://www.rstudio.com/wp-content/uploads/2014/06/rstudiocorporatecontributoragreement.pdf) contributor agreement as appropriate. You can send the signed copy to [contribute@rstudio.com](mailto:contribute@rstudio.com).

Thanks very much for submitting this PR! It will be reviewed shortly

kevinushey:
It might be simpler to use something like `utf8ToInt(x) < 32`. (I'm not sure why the number 20 was chosen)

kevinushey:
It might be simpler to use something like `utf8ToInt(x) < 32`. (I'm not sure why the number 20 was chosen)

okkn:
I have submitted my signed contributor agreement.

Similar codes are found in [rstudio/markdown](https://github.com/rstudio/markdown) too:
https://github.com/rstudio/markdown/blob/7e49047b63180694b43b35ea53642421e7bab808/R/rpubsUpload.R#L74-L86

By the way, since the `charToRaw()` function returns code in hexadecimal, the literal value of `20` in `else if (charToRaw(x) < 20)` could be a mistake for 32 (0x20 in hexadecimal), which corresponds to the ASCII code for SPACE.

```r
charToRaw(" ")
## [1] 20
charToRaw(" ") == 20
## [1] FALSE
charToRaw(" ") == 32
## [1] TRUE
```

kevinushey:
@okkn that is a good point -- we should probably be testing against 32 rather than 20.

okkn:
@kevinushey Whether the boundary value should be 20 or 32 is another important issue that needs to be carefully investigated, but first we should quickly correct the code that could produce errors in R 4.2.0.

src/cpp/session/modules/SessionCodeTools.R
```diff
@@ -1215,7 +1215,7 @@
    chars <- vapply(chars, function(x) {
       if (x %in% c('"', '\\', '/'))
          paste('\\', x, sep = '')
-      else if (charToRaw(x) < 20)
+      else if (length(charToRaw(x)) == 1 && charToRaw(x) < 20)
```
kevinushey:
```suggestion
      else if (length(charToRaw(x)) == 1 && charToRaw(x) < 32)
```

Issue #11156, opened 2022-05-09
R versions menu on Windows always selects default even when a specific version was chosen


ronblum:
### System details

    RStudio Edition : Desktop (Electron)
    RStudio Version : 2022.06.0-daily+374 (Electron)
    OS Version      : Windows 11
    R Version       : Multiple

### Steps to reproduce the problem

1. Install multiple versions of R, if not already installed.
2. <kbd>Tools -> Global Options -> Basic -> R Version -> Change</kbd>
    - Default 64-bit version is selected.
3. <kbd>Choose a specific version of R</kbd> (pick something that's not the default).
4. Pick a version and click "OK".
5. Apply global options.
6. Restart RStudio Desktop.
7. In Console Pane, the version should be correct, e.g. 4.0.3.
8. <kbd>Tools -> Global Options -> Basic -> R Version -> Change</kbd>
    - Default 64-bit version is selected, not the specific version.

The same occurs if choosing the version by holding down <kbd>Ctrl</kbd> while starting up.

### Describe the problem in detail

When the "Choose R Installation" menu pops up, it always defaults to "Use your machine's default 64-bit version of R" as the chosen version.

### Describe the behavior you expected

The version should be whatever's been picked. Cf. the behavior in Qt Spotted Wakerobin.

- [ ] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ ] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [ ] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
Verified in RStudio Desktop 2022.06.0-daily+437 on Windows 11. Checked in both
- The menu in Global Options.
- The window that appears when holding <kbd>Ctrl-C</kbd> at startup.

PR #11168, opened 2022-05-10
Previously Selected R Version is now being correctly displayed on Choose R Modal

matheustavaresdev:
### Intent

To show the user the correct option for the last R Version that was previously selected when the `Choose R` modal opens again.

### Approach

Saving that path and comparing with the options on the modal to correctly pre-select the current one.

### Automated Tests

None

### QA Notes

Refer to #11156

### Checklist

Addresses #11156


src/node/desktop/src/ui/widgets/choose-r/preload.ts
```diff
@@ -132,3 +169,9 @@ const callbacks: Callbacks = {
 };
 
 contextBridge.exposeInMainWorld('callbacks', callbacks);
+
+function isRVersionSelected(selectedVersion: string, versionToCompare: string) {
+  return (
+    selectedVersion.replace(/\//g, '/').replace(/\\/g, '/') === versionToCompare.replace(/\//g, '/').replace(/\\/g, '/')
```
gtritchie, line 175: Create and use utility function instead of repeating these regexprs over and over.

gtritchie (changes requested):

src/node/desktop/src/main/preferences/electron-desktop-options.ts
```diff
@@ -272,7 +273,7 @@ export class DesktopOptionsImpl implements DesktopOptions {
     if (process.platform !== 'win32') {
       return;
     }
-    this.config.set(kRBinDir, rBinDir);
+    this.config.set(kRBinDir, rBinDir.replace(/\//g, '/').replace(/\\/g, '/'));
```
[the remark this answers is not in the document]
gtritchie: Be sure to include unit tests for the function(s) you create.

src/node/desktop/src/ui/widgets/choose-r/preload.ts
```diff
@@ -73,9 +84,16 @@ ipcRenderer.on('initialize', (event, data) => {
...
   });
 
+  const selectWidget = document.getElementById('select') as HTMLSelectElement;
+  selectWidget.disabled = true;
+
   rInstalls.forEach((rInstall) => {
     // normalize separators, etc
-    rInstall = path.normalize(rInstall).replace(/[/\\]+$/g, '');
+    rInstall = path
+      .normalize(rInstall)
```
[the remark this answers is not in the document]
matheustavaresdev: Done!

src/node/desktop/src/main/preferences/electron-desktop-options.ts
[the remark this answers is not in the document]
matheustavaresdev: Done!

src/node/desktop/src/main/preferences/electron-desktop-options.ts
```diff
@@ -291,6 +292,30 @@ export class DesktopOptionsImpl implements DesktopOptions {
     return rBinDir;
   }
 
+  // Windows-only option
+  public setRstudioPath(rstudioPath: string): void {
+    if (process.platform !== 'win32') {
+      return;
+    }
+    console.log('setting rstudio path to ',rstudioPath.replace(/\//g, '/').replace(/\\/g, '/'));
```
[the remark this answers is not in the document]
matheustavaresdev: Done!

src/node/desktop/src/ui/widgets/choose-r/preload.ts
matheustavaresdev: Done!

src/node/desktop/src/core/file-path.ts
```diff
@@ -43,6 +43,11 @@ function normalizeSeparatorsNative(path: string) {
   return normalizeSeparators(path, separator);
 }
 
+// Makes sure that the path only has `/` instead of `\` or `//`
+export const threatPathString = (path: string) => {
```
kevinushey, line 47: I don't quote understand what this function is doing. It seems identical to https://github.com/rstudio/rstudio/pull/11168/files?w=1#diff-e763e5885fa9463dcc1490ef5c031c7eb737975eace1eecf59f4694e6ca82276R37-R39, except it doesn't let you choose the separator?

I also don't understand the name selected.

src/node/desktop/src/main/detect-r.ts
```diff
@@ -113,6 +112,7 @@ export async function promptUserForR(platform = process.platform): Promise<Expec
     // save the stored version of R
     const path = data.binaryPath as string;
     ElectronDesktopOptions().setRBinDir(dirname(path));
+    ElectronDesktopOptions().setRstudioPath(path);
```
kevinushey, line 115: This is the path to R, not RStudio. Shouldn't this be something like `setRBinaryExecutablePath()`?

In addition, if we're setting this, it seems like `rBinDir` is redundant, or at least the executable path could be computed from the binary path. In other words, I don't see why this is necessary.

src/node/desktop/src/ui/utils.ts
```diff
@@ -40,3 +41,8 @@ export const checkForNewLanguage = () => {
     }, 100);
   });
 };
+
+// Makes sure that the path only has `/` instead of `\` or `//`
+export const threatPathString = (path: string) => {
```
kevinushey, line 46: This looks like a duplicate definition. Is it needed?

src/node/desktop/src/main/preferences/electron-desktop-options.ts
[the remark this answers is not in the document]
matheustavaresdev: Done!

src/node/desktop/src/core/file-path.ts
matheustavaresdev: I was afraid of messing with the `normalizeSeparators` or `normalizeSeparatorsNative`. 

In my tests for solving this particular issue, only removing any kind of `\\`, `\` over to `/` would work as expected. 

So I thought of using this standalone method. I think that we could even remove `normalizeSeparatorsNative` and use only `normalizeSeparators`, with the same Regexp you mentioned below. Altough I was afraid of breaking something and having to spend more time by testing out stuff. What are your thoughts on it?

src/node/desktop/src/main/detect-r.ts
matheustavaresdev: I followed the naming used by `rBinDir` as `rstudioPath` was also declared there and it worked. Not sure how `rBinDir` is being used right now. That's why I created that new one for saving the exact R binary executable path so we could compare on the Choose R Modal to correctly check the current set option. Just let me know if you think I should rename something or do it differently so I can deploy a change!

src/node/desktop/src/core/file-path.ts
```diff
@@ -43,6 +43,11 @@ function normalizeSeparatorsNative(path: string) {
...
+  return path.replace(/\//g, '/').replace(/\\/g, '/');
```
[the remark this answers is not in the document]
kevinushey: But there's no need to replace `/` with `/`, so in your above implementation the only thing really happening is `\` is being transformed to `\`.

src/node/desktop/src/main/detect-r.ts
kevinushey: Perhaps we should replace the `RBinDir` preference and replace it with the `RExecutable` preference, and compute the `RBinDir` from that preference where appropriate. It's easier to compute the binary directory from the executable path anyhow, I think.

src/node/desktop/src/ui/widgets/choose-r/preload.ts
```diff
@@ -132,3 +166,7 @@ const callbacks: Callbacks = {
 };
 
 contextBridge.exposeInMainWorld('callbacks', callbacks);
+
+function isRVersionSelected(selectedVersion: string, versionToCompare: string) {
+  return threatPathString(selectedVersion) === threatPathString(versionToCompare);
```
kevinushey, line 171: If I understand correctly, you want to compare the paths after normalization. If that's the case, I think using `normalizeSeparators()` is appropriate here.

src/node/desktop/src/core/file-path.ts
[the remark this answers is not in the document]
matheustavaresdev: I just reused `.replace(/\//g, '/')` due for this [line](https://github.com/rstudio/rstudio/blob/main/src/node/desktop/src/main/gwt-callback.ts#L407) being used before, so as I didn't really know what was it for I ended up keeping it haha

src/node/desktop/src/ui/widgets/choose-r/preload.ts
matheustavaresdev: I just want to make sure both are equal, with no `\`

src/node/desktop/src/core/file-path.ts
```diff
@@ -35,10 +35,10 @@ const homePathAlias = '~/';
 const homePathLeafAlias = '~';
 
 function normalizeSeparators(path: string, separator = '/') {
-  return path.replace(/[\\/]/g, separator);
+  return path.replace(/[\\/]+/g, separator);
 }
 
-function normalizeSeparatorsNative(path: string) {
+export function normalizeSeparatorsNative(path: string) {
   const separator = process.platform === 'win32' ? '\\' : '/';
```
gtritchie, line 42: Use Node's constant for path separators. `path.sep`

https://nodejs.org/dist/latest-v16.x/docs/api/path.html#pathsep

src/node/desktop/src/ui/utils.ts
```diff
@@ -40,3 +41,13 @@ export const checkForNewLanguage = () => {
     }, 100);
   });
 };
+
+// Makes sure that the path only has `/` instead of `\` or `\\` or `//`
+function normalizeSeparators(path: string, separator = '/') {
```
gtritchie, line 46: Why does this function (and the next one) exist in both utils.ts and file-path.ts?

src/node/desktop/test/unit/core/file-path.test.ts
```diff
@@ -603,6 +603,23 @@ describe('FilePath', () => {
       assert.strictEqual(cPath.completeChildPath('../bar'), cPath);
       assert.strictEqual(cPath.completeChildPath('/path/to/quux'), cPath);
     });
+    it('Paths only contain forward slashes with no duplicates', () => {
+      const paths = ['c:\\www\\app\\my/folder/file.r',
+        'C:\\R\\4.1.2\\bin\\\\R.exe',
+        'c:\\\\www\\\\app\\my/folder/file.r'];
+      paths.forEach((path) => {
+        assert.isTrue(path.includes('\\'), `Path ${path} should contain at least a single backward slash for this test to be valid`);
+        const threatedPath = normalizeSeparatorsNative(path);
```
gtritchie, line 612: Not sure what this variable name means, specificially the use of "threat".

src/node/desktop/src/core/file-path.ts
```diff
@@ -35,10 +35,10 @@ const homePathAlias = '~/';
 const homePathLeafAlias = '~';
 
 function normalizeSeparators(path: string, separator = '/') {
```
gtritchie, line 37: Add a documentation comment to this method describing what it does, /** */.

gtritchie (changes requested):

gtritchie (changes requested):

src/node/desktop/test/unit/core/file-path.test.ts
```diff
@@ -603,6 +603,23 @@ describe('FilePath', () => {
       assert.strictEqual(cPath.completeChildPath('../bar'), cPath);
       assert.strictEqual(cPath.completeChildPath('/path/to/quux'), cPath);
     });
+    it('Paths only contain forward slashes with no duplicates', () => {
+      const paths = ['c:\\www\\app\\my/folder/file.r',
+        'C:\\R\\4.1.2\\bin\\\\R.exe',
+        'c:\\\\www\\\\app\\my/folder/file.r'];
```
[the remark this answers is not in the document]
matheustavaresdev: Thanks for the reminder. That's true. I will also check this up.

src/node/desktop/test/unit/core/file-path.test.ts
```diff
@@ -603,6 +603,23 @@ describe('FilePath', () => {
...
+      paths.forEach((path) => {
+        assert.isTrue(path.includes('\\'), `Path ${path} should contain at least a single backward slash for this test to be valid`);
+        const threatedPath = normalizeSeparatorsNative(path);
+        assert.isFalse(
+          threatedPath.includes('\\'),
+          `Path ${threatedPath} should NOT contain backward slashes for this test to be valid`,
```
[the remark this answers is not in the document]
matheustavaresdev: I'm gonna double check this as well.

src/node/desktop/src/ui/utils.ts
```diff
@@ -40,3 +41,13 @@ export const checkForNewLanguage = () => {
     }, 100);
   });
 };
+
+// Makes sure that the path only has `/` instead of `\` or `\\` or `//`
```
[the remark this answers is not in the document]
matheustavaresdev: Ok

src/node/desktop/src/core/file-path.ts
matheustavaresdev: Alright

src/node/desktop/test/unit/core/file-path.test.ts
matheustavaresdev: I meant treat! Gonna also fix this.

src/node/desktop/src/ui/utils.ts
matheustavaresdev: As it's being used in the `UI` folder, sometimes code from other parts of the code are not available at some point, like in some standalone pages. Kinda like another module. Had some issues with it previously, but also realized that for this part, there are some code being used outside the `ui` folder. So I'm gonna test removing that and leaving in only one place.

src/node/desktop/src/core/file-path.ts
matheustavaresdev: Alright

matheustavaresdev:
# This is still in progress. Will re-request review when it's ready again.

src/node/desktop/src/ui/utils.ts
matheustavaresdev: `renderer` folder cannot access some node packages. As `file-path.ts do contain some, it's best to have a different file for functions that are strictly non-node dependent

matheustavaresdev:
This is now ready for review again. cc @gtritchie @kevinushey

matheustavaresdev:
@gtritchie , after testing network R versions, the app cannot load `R.exe` files from the network. Altough, if we mount the network drive, it will work just as it should ( example: `T:\R-3.6.3\bin\x64\R.exe` )

That's the error if we try to access directly a network file. Not sure if this was working previously on QT app. Will try it out.
image

matheustavaresdev:
After tests, it does work on regular QT app. I think there might be something wrong with the current implementation on Windows. Should an issue be opened for this?

cc @kevinushey @gtritchie 

image

matheustavaresdev:
Also, on electron, we select the `R.exe` directly and after selected it opens the app already. 

On QT we select the `folder`, then the `R.exe` is added to the list of options so we can choose it manually and only then open the app. This behavior is not occuring on Electron.

matheustavaresdev:
Let me take a look.

kevinushey:
I'm referring to your comment in https://github.com/rstudio/rstudio/pull/11168#issuecomment-1128408415.

matheustavaresdev:
Oh Gotcha! Yea. I still haven't taken a look on that, but do you think that this issue should be part of this issue or should we open another issue for this error?

kevinushey:
Let's open a separate issue for that.

src/node/desktop/src/main/preferences/electron-desktop-options.ts
```diff
@@ -268,27 +271,42 @@ export class DesktopOptionsImpl implements DesktopOptions {
   }
 
   // Windows-only option
-  public setRBinDir(rBinDir: string): void {
+  public rBinDir(): string {
+    if (process.platform !== 'win32') {
+      return '';
+    }
+
+    let rBinDir: string = dirname(this.rExecutablePath());
```
kevinushey, line 279: Will this do the right thing if `rExecutablePath()` is not set, or is the empty string?

```
$ node -p 'require("path").dirname("")'
.
```

I think you need to check the value of  `this.rExecutablePath()` before using it.

src/node/desktop/src/ui/utils.ts
```diff
@@ -1,3 +1,5 @@
+import path from "path";
```
kevinushey, line 1: Is this import still needed? The comment below suggests `path` cannot be used here.

matheustavaresdev: Removing! Thanks for noticing. Forgot to remove from there

src/node/desktop/src/main/preferences/electron-desktop-options.ts
matheustavaresdev: Added this check on [fc3a811](https://github.com/rstudio/rstudio/pull/11168/commits/fc3a8110b86fb07c1bf205513b80fa921b242af7)

src/node/desktop/src/ui/utils.ts
```diff
@@ -40,3 +43,27 @@ export const checkForNewLanguage = () => {
...
+ * Removes duplicated separators from a path based on platform.
+ *
+ * @export
+ * @param {string} _path
+ * @return {*} 
+ */
+export function normalizeSeparatorsNative(_path: string) {
+  /* using conditional to set the separator based on platform as `path` is not available here */
+  const separator = process.platform === 'win32' ? '\\' : '/';
+  return normalizeSeparators(_path, separator);
```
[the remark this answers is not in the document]
matheustavaresdev: Fixed

src/node/desktop/src/ui/utils.ts
```diff
@@ -40,3 +43,27 @@ export const checkForNewLanguage = () => {
...
+}
+
+/**
+ * Removes duplicated separators from a path based on platform.
+ *
+ * @export
+ * @param {string} _path
+ * @return {*} 
+ */
+export function normalizeSeparatorsNative(_path: string) {
```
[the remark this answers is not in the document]
matheustavaresdev: Fixed

src/node/desktop/src/ui/utils.ts
```diff
@@ -40,3 +43,27 @@ export const checkForNewLanguage = () => {
...
+ */
+function normalizeSeparators(path: string, separator = '/') {
+  return path.replace(/[\\/]+/g, separator);
+}
+
+/**
+ * Removes duplicated separators from a path based on platform.
+ *
+ * @export
+ * @param {string} _path
```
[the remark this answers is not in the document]
matheustavaresdev: Fixed

src/node/desktop/src/main/preferences/electron-desktop-options.ts
```diff
@@ -268,27 +271,42 @@ export class DesktopOptionsImpl implements DesktopOptions {
...
 
   // Windows-only option
-  public rBinDir(): string {
+  public rExecutablePath(): string {
     if (process.platform !== 'win32') {
       return '';
     }
 
-    let rBinDir: string = this.config.get(kRBinDir, properties.platform.default.windows.rBinDir);
+    const rExecutablePath: string = this.config.get(kRExecutablePath, properties.platform.default.windows.rstudioPath);
```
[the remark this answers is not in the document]
matheustavaresdev: Fixed

src/node/desktop/src/main/preferences/electron-desktop-options.ts
```diff
@@ -39,7 +42,7 @@ const kRendererEngine = 'renderer.engine';
 const kRendererUseGpuExclusionList = 'renderer.useGpuExclusionList';
 const kRendererUseGpuDriverBugWorkarounds = 'renderer.useGpuDriverBugWorkarounds';
 
-const kRBinDir = 'platform.windows.rBinDir';
+const kRExecutablePath = 'platform.windows.rstudioPath';
```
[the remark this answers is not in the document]
matheustavaresdev: Fixed

src/node/desktop/src/core/file-path.ts
```diff
@@ -34,13 +34,27 @@ export class FilePathError extends Error {
...
+/**
+ * Removes duplicated separators from a path based on platform.
+ *
+ * @export
+ * @param {string} _path
+ * @return {*} 
+ */
+export function normalizeSeparatorsNative(_path: string) {
+  const separator = path.sep;
+  return normalizeSeparators(_path, separator);
```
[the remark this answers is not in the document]
matheustavaresdev: Fixed

src/node/desktop/src/core/file-path.ts
```diff
@@ -34,13 +34,27 @@ export class FilePathError extends Error {
...
 }
 
-function normalizeSeparatorsNative(path: string) {
-  const separator = process.platform === 'win32' ? '\\' : '/';
-  return normalizeSeparators(path, separator);
+/**
+ * Removes duplicated separators from a path based on platform.
+ *
+ * @export
+ * @param {string} _path
```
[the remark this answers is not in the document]
matheustavaresdev: Fixed

src/node/desktop/src/core/file-path.ts
```diff
@@ -34,13 +34,27 @@ export class FilePathError extends Error {
...
+ * @return {*} 
+ */
+export function normalizeSeparatorsNative(_path: string) {
```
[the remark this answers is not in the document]
matheustavaresdev: Fixed

kevinushey (approved):
LGTM!

@gtritchie, do you want to handle the issues related to UNC / network paths in this PR, or separately?

gtritchie:
> @gtritchie, do you want to handle the issues related to UNC / network paths in this PR, or separately?

Separately. We do want to fix it but isn't high priority.

gtritchie (approved):

PR #11173, opened 2022-05-10
Check for nulls when testing equality

jgutman:
### Intent

PRs #9922 and #10238 changed the definition of equality used by the debugger to check the context stack. However, this change introduced a new regression, described in #10664 . This small change fixes the original issues (#9918 and #10045) while also resolving the new regression (10664)

### Approach

Add one more constraint to the equality test -- check if other isNull(). Original equality test was too strict, then it became too liberal, hoping it is now juuuuuust right

Needs PT backport

### Automated Tests

No new tests added. Will ensure that [ide-automation#497](https://github.com/rstudio/rstudio-ide-automation/issues/497) still passes.

### QA Notes

Test the reprexes in all three linked issues:
#9918 
#10045
#10664

### DIFF
--- a/src/cpp/r/RCntxt.cpp
+++ b/src/cpp/r/RCntxt.cpp
@@ -192,7 +192,7 @@ bool RCntxt::operator==(const RCntxt& other) const
 
    // Also equivalent if they refer to the same call at the same stack position and have the same
    // source references
-   if (!other.isNull() &&
+   if (other.isNull() == isNull() &&
        other.call() == call() &&
        other.evaldepth() == evaldepth() &&
        other.srcref() == srcref())
--- a/src/cpp/r/RCntxt.cpp
+++ b/src/cpp/r/RCntxt.cpp
@@ -192,7 +192,8 @@ bool RCntxt::operator==(const RCntxt& other) const
 
    // Also equivalent if they refer to the same call at the same stack position and have the same
    // source references
-   if (other.call() == call() && 
+   if (!other.isNull() &&
+       other.call() == call() &&
        other.evaldepth() == evaldepth() &&
        other.srcref() == srcref())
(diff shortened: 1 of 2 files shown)


src/cpp/r/RCntxt.cpp
```diff
@@ -192,7 +192,8 @@ bool RCntxt::operator==(const RCntxt& other) const
 
    // Also equivalent if they refer to the same call at the same stack position and have the same
    // source references
-   if (other.call() == call() && 
+   if (!other.isNull() &&
```
kevinushey, line 195: For symmetry, would it make sense to test `other.isNull() == isNull()` here?

If not, could we move this check earlier just to sort of indicate the check here is "special" compared to the other members being checked?

(If I understand correctly, the issue here is that we had a non-null context that "looked like" a null context, because all of call / evaldepth / srcref had their "default" values... is that right?)

kevinushey (approved):
LGTM! Main comment is more of a nit + a question to make sure I understand what was going wrong.

kevinushey:
Also, if we want to backport to PT, it might be easier to target this for the prairie-trillium branch and then merge PT back to main after.

src/cpp/r/RCntxt.cpp
jgutman: Yeah, that seemed to be what was going on. I think I probably could change it to compare isNull to isNull and it would have the same effect, it didn't seem likely to me that they could both be null, otherwise the function would have returned true on the direct comparison in the first if block, but for the sake of readability I'm fine with spelling it out here

jgutman, line 195:
```suggestion
   if (other.isNull() == isNull() &&
```

PR #11174, opened 2022-05-10
don't use localized key names for shortcuts on desktop

gtritchie:
### Intent

Addresses: Keyboard shortcuts containing Shift or Enter don't display in menu in French (Desktop) #11073

For Desktop IDE, keyboard shortcuts whose key names were localized (including "Shift/Maj" and "Enter/Entrée") were causing errors to be emitted to console, and those shortcuts weren't displaying at all (or only partially) in the menus. Note that "Ctrl", "Alt", and "Cmd" were the same in English and French so weren't causing problems.

This problem would also manifest with shortcuts containing "Left", "Right", "PageUp", "Backspace", "Space", and so forth, but none of our built-in shortcuts use those. You can modify keyboard shortcuts and try them, though.

### Approach

When running via desktop, don't pass back localized key names as the desktop frameworks don't understand them and are expecting the English key names.

### Automated Tests

N/A

### QA Notes

On Qt desktop, the shortcuts will no longer be localized in the menus when running in French (e.g. Shift will be "Shift", not "Maj"). This might be solvable (perhaps we need to include localized Qt resources) but given we are retiring Qt desktop in an upcoming release it isn't worth messing with.

On Electron, the shortcuts should still show up localized (this is handled by Electron).

On Server, behaviour should be the same as before, and the shortcuts will display localized.

### Checklist

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/command/KeyCombination.java
+++ b/src/gwt/src/org/rstudio/core/client/command/KeyCombination.java
@@ -22,6 +22,7 @@ import org.rstudio.core.client.dom.EventProperty;
 
 import com.google.gwt.dom.client.NativeEvent;
 import com.google.gwt.event.dom.client.KeyCodes;
+import org.rstudio.studio.client.application.Desktop;
 
 public class KeyCombination
 {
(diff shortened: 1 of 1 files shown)


romainfrancois (approved):
LGTM. Thanks.

PR #11181, opened 2022-05-11
don't remove empty lines of code -- we need them for the source code markers

jgutman:
### Intent

I made a whoopsie! When I made [this change ](https://github.com/rstudio/rstudio/pull/10864), I filtered out empty lines because they were empty, so we don't need to check whether or not they contain R code. That was a mistake, because we DO need them in order to make sure the diagnostic source code markers in the gutter actually line up with the appropriate code line, which means they need to include the blank spaces. We should filter them out only later, when we are checking for the existence of R code, rather than in SessionDiagnostics.R

### Approach

Remove 1 silly line of code so we don't filter out blanks. Filter out the newline characters later instead

### Automated Tests

### QA Notes

Test that diagnostic code markers show up on the correct line. Also retest the issue from the original PR [pro#3310](https://github.com/rstudio/rstudio-pro/issues/3310) but this is probably unnecessary

### Checklist

### DIFF
--- a/src/cpp/session/modules/SessionCodeTools.R
+++ b/src/cpp/session/modules/SessionCodeTools.R
@@ -2243,7 +2243,7 @@
       return(NULL)
    # If Quarto and no R code chunks, don't parse further
    # Rmd may have yaml front matter to parse even if no R chunks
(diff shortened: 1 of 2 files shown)


kevinushey (approved):
LGTM!

PR #11182, opened 2022-05-11
Make newly created SSH key the preferred key

jgutman:
### Intent

Follow up to #8255. Address issue that @romainfrancois raised during testing. There was previously no connection between the SSH key you just created, and the SSH key that would be chosen (based on the computed user preference) when you re-open the Global options pane. Update so that the key you just created takes precedence, no session restart necessary.

### Approach

On successful response to RPC call creating the SSH key, reset the computed user preference to point to the new key.

### Automated Tests

None

### QA Notes

Can follow Romain's test plan in the #8255 [comments](https://github.com/rstudio/rstudio/issues/8255#issuecomment-1117141856). 
Create new SSH key. Close and re-open Global options pane, check that SSH key path and View public key file point to the newly created key, and not the old one you had before.

### Checklist


romainfrancois (approved):
LGTM, thanks. I'll test the issue in the morning.

Issue #11184, opened 2022-05-11
Provide option to use web dialogs with Electron Desktop


gtritchie:
Electron has an issue on some Linux distros where the native dialogs don't display on top of the main window as they should: #11100. Although you can get to the dialogs, if you know to go looking for them, it's ugly.

We will hopefully get a fix for this from Electron (or make the fix ourselves), but if this doesn't happen in time for our first preview builds, I propose providing a user preference to tell Electron to use the RStudio Server web-based dialogs instead of the native ones.

This would have a secondary benefit in making it possible to use more of our existing Selenium automation for Electron testing (right now we don't have the ability for Selenium tests to query or operate the native dialogs).

ronblum:
@gtritchie Does the recent Electron update fix this, or do we still need to rely on the option?

In any case, I've tested this, but I'm doing everything with that (mostly) to get some implicit testing.

gtritchie:
The Electron issue is still open: https://github.com/electron/electron/issues/32857

ronblum:
Verified in multiple environments:
- MacOS 12.5
- Red Hat 8.5
- Ubuntu 22.04
- Windows 11

Plus passive testing.

PR #11185, opened 2022-05-11
add option to use web dialogs on Electron desktop

gtritchie:
### Intent

Addresses #11184

This is intended both to provide a workaround for an Electron bug with File dialogs on Linux, and as a way to leverage more of our existing Selenium automation for Electron.

### Approach

On Electron desktop, added new user preference to the General / Advanced / OS Integration settings (the default is "on", matching current behaviour):

Screen Shot 2022-05-11 at 4 42 48 PM

When unchecked, a cookie is set and the page is refreshed, and GWT detects this cookie and changes to use the web-based dialogs instead of the native ones.

This isn't actually breaking (much) new ground, as we already do this in Desktop Pro when connected to a server.

For example, here's Mac Electron IDE showing result of File / Open when this option was unchecked.

Screen Shot 2022-05-11 at 4 45 13 PM

Versus the normal default behaviour:

Screen Shot 2022-05-11 at 4 45 39 PM

### Automated Tests

N/A

### QA Notes

Check that with Electron Desktop that operations such as File / Open File, File / Open Project, and any other entrypoints that bring up filesystem browsing dialogs obey this setting.

### Checklist

### DIFF
--- a/src/gwt/src/org/rstudio/studio/RStudio.gwt.xml
+++ b/src/gwt/src/org/rstudio/studio/RStudio.gwt.xml
@@ -47,6 +47,24 @@
       return window.navigator.userAgent.indexOf("Electron") > 0 ? "true" : "false";
    ]]></property-provider>
 
+   <!-- Detect option to stop use of native OS file dialogs on Electron Desktop -->
+   <define-property name="rstudio.nativedialogs" values="true,false"/>
+   <property-provider name="rstudio.nativedialogs"><![CDATA[
+      try {
+         if (window.navigator.userAgent.indexOf("Electron") > 0) {
+            var cookieArr = window.document.cookie.split(";");
+            for( var i = 0; i < cookieArr.length; i++) {
+               var cookiePair = cookieArr[i].split("=");
+               if ('WEBDIALOG' === cookiePair[0].trim().toUpperCase()) {
+                  return "false";
+               }
+            }
+         }
+      } catch (e) {
+         console.error(e);
+      }
+      return "true";
+   ]]></property-provider>
 
    <!-- Set the user agent. Note that we no longer support older IE. -->
    <!-- Note that 'safari' here is effectively an alias for modern browsers. -->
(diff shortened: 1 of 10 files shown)


src/gwt/src/org/rstudio/studio/client/workbench/prefs/model/UserPrefsAccessorConstants_en.properties
```diff
@@ -977,4 +977,8 @@ checkNullExternalPointersDescription = When enabled, RStudio will detect R objec
 uiLanguageTitle = User Interface Language
 uiLanguageDescription = The IDE's user-interface language.
 
+# Whether RStudio Desktop will use operating system File and Message dialog boxes.
+nativeFileDialogsTitle = Use operating system File and Message dialog boxes
```
kevinushey, line 981: Label tweak: Use native file and message dialogs? (I don't think capitalization is necessary here)

kevinushey (approved):
LGTM; a couple comments for naming / label tweaks.

src/gwt/src/org/rstudio/studio/client/workbench/prefs/model/UserPrefsAccessorConstants_en.properties
gtritchie: Agree, not Sure why i Capitalized thOse.

src/gwt/src/org/rstudio/studio/RStudio.gwt.xml
```diff
@@ -47,6 +47,24 @@
       return window.navigator.userAgent.indexOf("Electron") > 0 ? "true" : "false";
    ]]></property-provider>
 
+   <!-- Detect option to stop use of native OS file dialogs on Electron Desktop -->
+   <define-property name="rstudio.nativedialogs" values="true,false"/>
```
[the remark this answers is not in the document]
gtritchie: That would be more descriptive, will change.

PR #11186, opened 2022-05-12
update embedded libclang to 13.0.1

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/7554.

### Approach

Update the embedded version of libclang on Windows to 13.0.1, and also make a number of tweaks to ensure it behaves well with rtools42.

### Automated Tests

N/A

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/7554.

### Checklist


src/cpp/session/session-options.json
```diff
@@ -719,13 +719,6 @@
             "defaultValue": {"code": "kDefaultRsclangPath", "description": "bin/rsclang"},
             "description": "Specifies the path to the libclang shared library"
          },
-         {
-            "name": "external-libclang-headers-path",
```
gtritchie, line 723: Will this cause problems (like session failing to start) if someone had this setting in rsession.conf? Maybe we should leave this in and mark it as `isDeprecated`, such as seen here?

https://github.com/rstudio/rstudio/blob/7781d230ce5abdcd20d185b385f1a0ee01df6da8/src/cpp/session/session-options.json#L575-L581

gtritchie (changes requested):

gtritchie (commented):
We should also mention this in the NEWS, I think.

src/cpp/session/session-options.json
kevinushey: Good point; I've added it back in. Better safe than sorry.

gtritchie (approved):
LGTM!

PR #11207, opened 2022-05-13
use ucrtbase.dll with R 4.2.0 or newer

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/11206.

### Approach

Newer versions of R will set errno in ucrtbase.dll, so read it from there instead of msvcrt.dll.

### Automated Tests

N/A

### QA Notes

Given a file with the contents:

```
AAAAAAAAAAHHHHHHHHHHHHHHHHHHHH
AAAAAAAAAAHHHHHHHHHHHHHHHHHHHH
AAAAAAAAAAHHHHHHHHHHHHHHHHHHHH
AAAAAAAAAAHHHHHHHHHHHHHHHHHHHH
AAAAAAAAAAHHHHHHHHHHHHHHHHHHHH
AAAAAAAAAAHHHHHHHHHHHHHHHHHHHH
AAAAAAAAAAHHHHHHHHHHHHHHHHHHHH
AAAAAAAAAAHHHHHHHHHHHHHHHHHHHH
AAAAAAAAAAHHHHHHHHHHHHHHHHHHHH
AAAAAAAAAAHHHHHHHHHHHHHHHHHHHH
AAAAAAAAAAHHHHHHHHHHHHHHHHHHHH
AAAAAAAAAAHHHHHHHHHHHHHHHHHHHH
AAAAAAAAAAHHHHHHHHHHHHHHHHHHHH
AAAAAAAAAAHHHHHHHHHHHHHHHHHHHH
AAAAAAAAAAHHHHHHHHHHHHHHHHHHHH
```

(Screaming is an important part of the test procedure)

- Open a file test.R, and paste these contents in,
- Try to save the file,
- Try File -> Open with Encoding -> UTF-8, and confirm the file contents did not change,
- Try file -> Open with Encoding -> ISO-8859-1, and confirm the file contents did not change.

Run this test with both R 4.2.0 and R 3.6.3, just to confirm we did indeed fix the issue with R 4.2.0, and didn't break anything for R 3.6.3.

### Checklist

### DIFF
--- a/src/cpp/core/system/Win32RuntimeLibrary.cpp
+++ b/src/cpp/core/system/Win32RuntimeLibrary.cpp
@@ -36,6 +36,12 @@ namespace runtime {
 
 namespace {
 
+// forward-declare
+class Library;
+
+// a handle to the runtime library now loaded
+Library* s_pRuntimeLibrary = nullptr;
+
 namespace defaults {
 int s_zero = 0;
 int* _errno() { return &s_zero; }
@@ -46,11 +52,11 @@ class Library : boost::noncopyable
 {
 public:
 
-   Library()
+   Library(const char* libraryName)
       : pLib_(nullptr),
         _errno(defaults::_errno)
    {
-      Error error = core::system::loadLibrary("msvcrt.dll", &pLib_);
+      Error error = core::system::loadLibrary(libraryName, &pLib_);
       if (error)
       {
          LOG_ERROR(error);
@@ -59,8 +65,6 @@ public:
 
       RS_LOAD_SYMBOL(_errno);
    }
-
-   int* (*_errno)(void);
 
    ~Library()
    {
@@ -76,18 +80,23 @@ public:
       CATCH_UNEXPECTED_EXCEPTION
    }
 
-private:
+public:
    void* pLib_;
+   int* (*_errno)(void);
 };
 
 Library& library()
 {
-   static Library instance;
-   return instance;
+   return *s_pRuntimeLibrary;
 }
 
 } // end anonymous namespace
 
+core::Error initialize(bool isUcrt)
+{
+   s_pRuntimeLibrary = new Library(isUcrt ? "ucrtbase.dll" : "msvcrt.dll");
+   return Success();
+}
 
 int errorNumber()
 {
--- a/src/cpp/session/SessionMain.cpp
+++ b/src/cpp/session/SessionMain.cpp
@@ -501,7 +501,9 @@ Error rInit(const rstudio::r::session::RInitInfo& rInitInfo)
          LOG_ERROR(error);
 
       // initialize runtime library
-      rstudio::core::runtime::initialize(crt == "ucrt");
+      error = rstudio::core::runtime::initialize(crt == "ucrt");
+      if (error)
+         LOG_ERROR(error);
    }
 #endif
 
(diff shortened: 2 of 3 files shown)


src/cpp/session/SessionMain.cpp
```diff
@@ -1879,6 +1889,15 @@ int main(int argc, char * const argv[])
       ::setpgrp();
 #endif
 
+#ifdef _WIN32
+      // initialize runtime library
+      Version rVersion(getDLLVersion());
+      bool isUcrt = rVersion >= Version("4.2.0");
```
jgutman, line 1895: Will r >= 4.2 on windows always be ucrt? Aren't there some exceptions based on windows version? I recall Tomas K saying something to that effecr

kevinushey: I believe so, but we should probably have a better check for this. The note at https://cran.r-project.org/bin/windows/base/ says:

> This build requires UCRT, which is part of Windows since Windows 10 and Windows Server 2016. On older systems, UCRT has to be installed manually from [here](https://support.microsoft.com/en-us/topic/update-for-universal-c-runtime-in-windows-c0514201-7fe6-95a3-b0a5-287930f3560c).

Let me see if there's a better place for me to figure this out.

src/cpp/core/system/Win32RuntimeLibrary.cpp
```diff
@@ -60,8 +66,6 @@ class Library : boost::noncopyable
       RS_LOAD_SYMBOL(_errno);
    }
 
-   int* (*_errno)(void);
-
```
jgutman, line 64: What was this errno doing exactly?

kevinushey: I just moved that down below (to resolve a compiler warning).

It's basically a reference to the `_errno` symbol, which might be coming from either `msvcrt.dll` or `ucrtbase.dll`.

src/cpp/session/SessionMain.cpp
jgutman: > I just wanted to clarify that to use
UTF-8 as the native encoding in R 4.2, if you are using long-term
support Windows servers, you need Windows Server 2022 (Windows Server
2016 is not enough). If you run R 4.2 on Windows Server 2016, it will
still use the Windows default encoding (as R 4.1 and older).

> More precisely, as "R for Windows FAQ" says,


> For UTF-8 to be the native encoding, you need R at least 4.2 and at least
Windows 10 (version 1903) on desktop systems, Windows Server 2022 on
long-term support server systems or Windows Server 1903 from the semi-annual
channel.

src/cpp/session/SessionMain.cpp
```diff
@@ -487,6 +492,21 @@ Error rInit(const rstudio::r::session::RInitInfo& rInitInfo)
    module_context::registerWaitForMethod(kHandleUnsavedChangesCompleted);
    module_context::registerWaitForMethod(kRStudioAPIShowDialogMethod);
 
+#ifdef _WIN32
+   {
+      // on Windows, check if we're using UCRT
+      std::string crt;
+      Error error = rstudio::r::exec::evaluateString("R.version$crt", &crt);
```
jgutman, line 499: Thanks, this seems less error prone than just checking the R version

jgutman (approved):

PR #11209, opened 2022-05-13
Feature/3094 session metadata direct database

AndrewMcClain:
### Intent

Implement a direct to database metadata channel for use by server.

### Approach

Split out the ActiveSessions and ActiveSession, collective and singular. This helps break a circular dependency. Storage mechanisms are broken out from the ActiveSession* classes themselves into the IActiveSessionsStorage and IActiveSessionStorage, File base implementations of both exist. DB based implementations are added.

### Automated Tests

Automated tests exist for the singular case DB implementation, the collection does not have automated tests at this time.

### Checklist


src/cpp/server/DBActiveSessionStorage.cpp
```diff
@@ -0,0 +1,436 @@
...
+                     {
+                        a += ", '" + b.second + "'";
+                     }
+                     return a;
+                });
+      return values;
+
+   }
+
+   std::string getUpdateString(std::map<std::string, std::string> sourceMap)
```
MariaSemple, line 63: I don't think we should need to do a copy here.
```suggestion
   std::string getUpdateString(const std::map<std::string, std::string>& sourceMap)
```

src/cpp/server/DBActiveSessionStorage.cpp
```diff
@@ -0,0 +1,436 @@
...
+   {
+      std::string setValuesString = std::accumulate(++sourceMap.begin(), sourceMap.end(),
+                  sourceMap.begin()->first + " = '" + sourceMap.begin()->second + "'", [](std::string a, std::pair<std::string, std::string> iter){
+                     return a + ", " + iter.first + " = '" + iter.second + "'";
+                  });
+      return setValuesString;
+   }
+
+   std::string getColumnNameList(std::set<std::string> colNames)
```
MariaSemple, line 72: Similar to above
```suggestion
   std::string getColumnNameList(const std::set<std::string>& colNames)
```

src/cpp/server/DBActiveSessionStorage.cpp
```diff
@@ -0,0 +1,436 @@
...
+         {
+            targetMap->insert(
+               std::pair<std::string, std::string>{
+                  key,
+                  std::to_string(iter->get<int>(key))
+               }
+            );
+         }
+      }
+   }
```
MariaSemple, line 105:
```suggestion
   void populateMapWithRow(database::RowsetIterator iter, std::map<std::string, std::string>* pTargetMap)
   {
      for (size_t i=0; i < iter->size(); i++)
      {
         std::string key = iter->get_properties(i).get_name();
         if (key != kUserId)
         {
            pTargetMap->insert(
               std::pair<std::string, std::string>{
                  key,
                  iter->get<std::string>(key, "")
               }
            );
         }
         else
         {
            pTargetMap->insert(
               std::pair<std::string, std::string>{
                  key,
                  std::to_string(iter->get<int>(key))
               }
            );
         }
      }
   }
```

src/cpp/server/DBActiveSessionStorage.cpp
```diff
@@ -0,0 +1,436 @@
...
+{
+namespace storage
+{
+
+using namespace server_core::database;
+namespace {
+
+   // This is the column name of the foreign key between the active_session_metadata
+   // And the licensed user table. The only column that isn't a string
+   std::string kUserId = "user_id";
```
MariaSemple, line 33: Indentation is one too deep here and in the other namespace blocks

src/cpp/server/DBActiveSessionStorage.cpp
```diff
@@ -0,0 +1,436 @@
...
+#include "server_core/ServerDatabase.hpp"
+#include "core/Database.hpp"
+#include <numeric>
+
+namespace rstudio
+{
+namespace server
+{
+namespace storage
+{
```
MariaSemple, line 26: For namespace blocks we prefer to keep the opening brace on the same line

```suggestion
namespace rstudio {
namespace server {
namespace storage {
```

src/cpp/server/DBActiveSessionStorage.cpp
```diff
@@ -0,0 +1,436 @@
...
+ * ANY EXPRESS OR IMPLIED WARRANTY, INCLUDING THOSE OF NON-INFRINGEMENT,
+ * MERCHANTABILITY OR FITNESS FOR A PARTICULAR PURPOSE. Please refer to the
+ * AGPL (http://www.gnu.org/licenses/agpl-3.0.txt) for more details.
+ *
+ */
+
+#include "server/DBActiveSessionStorage.hpp"
+#include "server_core/ServerDatabase.hpp"
+#include "core/Database.hpp"
+#include <numeric>
```
MariaSemple, line 19: A general rule of thumb for organising includes (plus some other include things):

* The `hpp` associated with this `cpp` goes first (as you have it)
* Other includes should be separated from it by a blank line
* Includes can be blocked together in groups based on the include location (e.g. all `<core/*>` includes together and all `<boost/*>` together, with newlines between) although for this case, it's probably not necessary to split them up because there are so few
* We should use `<>` instead of `""` wherever possible. `""` will include the current directory in the search path whereas `<>` will not.
* Includes from the current directory (requiring `""`) should go last, in their own block
* I like to organise things within a block alphabetically, but that's purely personal preference

```suggestion
#include <server/DBActiveSessionStorage.hpp>

#include <core/Database.hpp>
#include <server_core/ServerDatabase.hpp>

#include <numeric>
```

src/cpp/core/r_util/RActiveSessions.cpp
```diff
@@ -14,7 +14,9 @@
  */
 
 #include <core/r_util/RActiveSessions.hpp>
-#include <core/r_util/RActiveSessionStorage.hpp>
+#include <core/r_util/RActiveSessionsStorage.hpp>
+
+#include <core/Log.hpp>
```
MariaSemple, line 19: (See the other comment about organising includes)
```suggestion

#include <core/Log.hpp>
#include <core/r_util/RActiveSessionsStorage.hpp>
```

src/cpp/core/r_util/RActiveSessionsStorage.cpp
```diff
@@ -0,0 +1,149 @@
...
+ * ANY EXPRESS OR IMPLIED WARRANTY, INCLUDING THOSE OF NON-INFRINGEMENT,
+ * MERCHANTABILITY OR FITNESS FOR A PARTICULAR PURPOSE. Please refer to the
+ * AGPL (http://www.gnu.org/licenses/agpl-3.0.txt) for more details.
+ *
+ */
+
+#include <core/r_util/RActiveSessionsStorage.hpp>
+#include <core/r_util/RActiveSessionStorage.hpp>
+#include <core/r_util/RActiveSessions.hpp>
+#include <algorithm>
```
MariaSemple, line 19:
```suggestion
#include <core/r_util/RActiveSessionsStorage.hpp>

#include <core/r_util/RActiveSessionStorage.hpp>
#include <core/r_util/RActiveSessions.hpp>

#include <algorithm>
```

src/cpp/core/r_util/RActiveSessionsStorage.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+#define kSessionDirPrefix "session-"
+
+namespace rstudio {
+namespace core {
+namespace r_util {
+
+namespace
+{
```
MariaSemple, line 28: Prefer constexpr for literals to macros.

```suggestion
namespace rstudio {
namespace core {
namespace r_util {

namespace
{

constexpr const char * kSessionDirPrefix = "session-";

```

src/cpp/server/DBActiveSessionStorage.cpp
```diff
@@ -0,0 +1,436 @@
...
+      else
+      {
+         connection = overrideConnection_;
+      }
+      
+      if(error)
+      {
+         return error;
+      }
+
```
MariaSemple, line 154: Nit: reduce line count a bit
```suggestion
      if(overrideConnection_ == nullptr)
         error = getConn(&connection);
      else
         connection = overrideConnection_;
      
      if(error)
         return error;
```

src/cpp/server/DBActiveSessionStorage.cpp
```diff
@@ -0,0 +1,436 @@
...
+
+      database::Query query = connection->query("SELECT "+name+" FROM active_session_metadata WHERE session_id = :id")
```
MariaSemple, line 156: We should have constants for these, particularly for the table and column names.

src/cpp/core/include/core/r_util/RActiveSession.hpp
```diff
@@ -0,0 +1,492 @@
...
+   const std::string kInitial = "initial";
+   const std::string kLabel = "label";
+   const std::string kLastUsed = "last_used";
+   const std::string kProject = "project";
+   const std::string kSavePromptRequired = "save_prompt_required";
+   const std::string kRunning = "running";
+   const std::string kRVersion = "r_version";
+   const std::string kRVersionHome = "r_version_home";
+   const std::string kRVersionLabel = "r_version_label";
+   const std::string kWorkingDir = "working_dir";
```
jeffvroom, line 63: So this used to be `working_directory`, not `working_dir` but doesn't that mean we won't be able to read old suspended sessions after an upgrade?  Same thing with `launch_params`.  I don't know if the clean up is worth the aggravation that will cause.  Also, we also use `working_directory` in settings so maybe we should just go back to using the full name to be consistent?  Same with `launch_parameters`.

src/cpp/core/include/core/r_util/RActiveSession.hpp
```diff
@@ -0,0 +1,492 @@
...
+      return true;
+   }
+   
+   bool operator>(const ActiveSession& rhs) const
+   {
+      if (sortConditions_.executing_ == rhs.sortConditions_.executing_)
+      {
+         if (sortConditions_.running_ == rhs.sortConditions_.running_)
+         {
+
```
jeffvroom, line 346: Why did you remove this test?  I don't know that it's important but maybe lastUsed is 0 and this will mean the list order is not consistent which can create weird bugs.

```            
        if (sortConditions_.lastUsed_ == rhs.sortConditions_.lastUsed_)
               return id() > rhs.id();
````

File: src/cpp/core/include/core/r_util/RActiveSession.hpp (hunk 12 above)
Comment (jeffvroom) [reply src/cpp/core/include/core/r_util/RActiveSession.hpp:63]: Never mind, I see the file name is actually working-dir that comes from the mapping table.

File: src/cpp/server/DBActiveSessionStorage.cpp [hunk 14]
```diff
@@ -0,0 +1,417 @@
...
+{
+
+}
+
+Error DBActiveSessionStorage::readProperty(const std::string& name, std::string* pValue)
+{
+   *pValue = "";
+   boost::shared_ptr<database::IConnection> connection;
+   Error error = Success();
+   if(overrideConnection_ == nullptr)
```
jeffvroom, line 145: How about if we move this block into the getConn method()?

src/cpp/server/DBActiveSessionStorage.cpp
```diff
@@ -0,0 +1,417 @@
+/*
+ * DBActiveSessionStorage.cpp
```
jeffvroom, line 2: Should we have a sub-directory called "sessions" for server specific session feature code?  It could hold all of the DBActiveSession* features.

jeffvroom (commented):
Great work!  It's looking pretty close to me. I have a few comments I added here, and always happy to jump on a call to chat if that's easier.

src/cpp/server/DBActiveSessionStorage.cpp
AndrewMcClain: That's a good idea

src/cpp/core/include/core/r_util/RActiveSessionsStorage.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+namespace rstudio {
+namespace core {
+namespace r_util {
+
+   class IActiveSessionsStorage
+   {
+      public:
+         virtual core::Error createSession(const std::string& id, std::map<std::string, std::string> initialProperties) = 0;
+         virtual std::vector< boost::shared_ptr<ActiveSession> > listSessions(FilePath userHomePath, bool projectSharingEnabled) const = 0;
```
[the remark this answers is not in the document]
MariaSemple: Just talking about this with Andrew - I think it would make sense to have the `IActiveSessionsStorage` only return the list of session IDs and have the `ActiveSessions` object to the validation it wants to do.

src/cpp/core/include/core/r_util/RActiveSession.hpp
AndrewMcClain: Good catch.

PR #11211, opened 2022-05-13
tweak how compiler definitions are passed along to libclang

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/7554#issuecomment-1126513442.

### Approach

The previous implementation tried to write the compiler definitions header to the R temporary directory. However, this path cannot be used for the base compilation arguments, as those get persisted as part of the compilation database, and so we end up persisting a transient path.

Instead, we use a stable scoped scratch path. We also ensure that the compiler definitions are re-generated if we need to re-generate the rest of the compilation database.

### Automated Tests

N/A

### QA Notes

https://github.com/rstudio/rstudio/issues/7554#issuecomment-1126513442

### Checklist


gtritchie (approved):
LGTM! (As much as I understand of it, anyway)

PR #11217, opened 2022-05-16
Bugfix/filter suspend resets

cm421:
### Intent

Fixes https://github.com/rstudio/rstudio-pro/issues/3362 where the actions of another user were keeping a session from suspending.

When an `rsession` is waiting for a particular connection to come in (`console_input`, for example), sometimes other, unrelated connections can come in and reset the session suspend timeout without indicating that they've done so. This PR adds a way to 
1. Generate DEBUG messages indicating which `HttpConnection`s are resetting the session's suspend timeout, and
2. filter out incoming connections that should **not** reset that session suspend timeout.

### Approach

I made the implementation a little more verbose than simply iterating through a list of incoming connection `URI`s to ignore for 3 main reasons:

1. Because this was reported by a user, there's a high possibility of similar issues being created going forward. I pulled the implementation into its own file attempt to make it friendlier for a user to contribute their own filter. All they'd need to do is define a new filter class, add it to the list of filters, and make a PR
2. Right now only a couple URIs are filtered. If filtering needs to be more selective (ie filter by URI **and** some aspect of the `HttpConnection` or HTTP request body), this makes it trivial to add that
3. Adding unit tests was trivial for this approach 

### Automated Tests

RSession Unit tests have been added for this fix.

### QA Notes

See the ticket for steps to reproduce in the Pro product. The short of it is that a different user entering/leaving our shared project or following our cursor should not affect when our session suspends from inactivity.

Setting `session-timeout-minutes=1` to `/etc/rstudio/rsession.conf` speeds up testing.

### Checklist

### DIFF
--- a/src/cpp/session/SessionSuspend.cpp
+++ b/src/cpp/session/SessionSuspend.cpp
@@ -131,9 +131,6 @@ void resetSuspendTimeout(boost::shared_ptr<HttpConnection> pConnection)
 {
    if (s_suspendFilters.shouldResetSuspendTimer(pConnection))
    {
-      LOG_DEBUG_MESSAGE(pConnection->request().uri() +
-                        " URI resetting suspend timeout. Request: " +
-                        pConnection->request().body());
       resetSuspendTimeout();
    }
 }
--- a/src/cpp/session/SessionSuspend.cpp
+++ b/src/cpp/session/SessionSuspend.cpp
@@ -26,6 +26,7 @@
 #include <session/SessionOptions.hpp>
 #include <session/SessionPersistentState.hpp>
 #include <session/SessionSuspend.hpp>
+#include <session/SessionSuspendFilter.hpp>
 
 #include <shared_core/Error.hpp>
 #include <shared_core/json/Json.hpp>
@@ -63,6 +64,8 @@ bool s_dirtyBlockingOps = false;
 bool s_initialNotificationSent = false;
 bool s_timeoutLogSent = false;
 SuspendTimeoutState s_timeoutState = kWaitingForTimeout;
+
+SessionSuspendFilters s_suspendFilters;
 
 } // anonymous namespace
 
@@ -121,6 +124,14 @@ bool isTimedOut(const boost::posix_time::ptime& timeoutTime)
 void resetSuspendTimeout()
 {
    s_suspendTimeoutTime = timeoutTimeFromNow();
+}
+
+void resetSuspendTimeout(boost::shared_ptr<HttpConnection> pConnection)
+{
+   if (s_suspendFilters.shouldResetSuspendTimer(pConnection))
+   {
+      resetSuspendTimeout();
+   }
 }
 
 void blockingTimestamp()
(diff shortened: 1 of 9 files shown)


src/cpp/session/SessionSuspendFilter.cpp
```diff
@@ -0,0 +1,106 @@
...
+   m_filters.insert(m_filters.end(), {
+                       boost::make_shared<DistEventFilter>(),
+                       boost::make_shared<SetEditingFilter>()
+                    });
+}
+
+SessionSuspendFilters::~SessionSuspendFilters() {}
+
+/**
+ * Determines if recieving a given HttpConnection should reset the suspend timer for the session.
```
jeffvroom, line 81:
```suggestion
 * Determines if the request should reset the suspend timer for the session.
```

src/cpp/session/SessionSuspend.cpp
```diff
@@ -123,6 +127,17 @@ void resetSuspendTimeout()
    s_suspendTimeoutTime = timeoutTimeFromNow();
 }
 
+void resetSuspendTimeout(boost::shared_ptr<HttpConnection> pConnection)
+{
+   if (s_suspendFilters.shouldResetSuspendTimer(pConnection))
+   {
+      LOG_DEBUG_MESSAGE(pConnection->request().uri() +
```
jeffvroom, line 134: I think this might be a little low on the "signal to noise" ratio since it will apply to all but a couple of URIs. We already will see in the logs before/after messages for each URI so I'd say this is probably not necessary.

jeffvroom (approved):
This looks good to me. The code is very clean and easy to follow, and there's a lot of flexibility in how suspend filters could be designed. Since at this point, there are only two uri based filters, I would have been ok with a simpler approach (just a list of the two non-interactive URIs we check)... this might come in handy in the future.

src/cpp/session/SessionSuspend.cpp
AndrewMcClain: I would flip it here, if the suspend filters prevent the request from resetting the suspend timeout, log that. Seems like it would be the rarer of the two events.

src/cpp/session/SessionSuspendFilterTests.cpp
```diff
@@ -0,0 +1,79 @@
...
+   test_that("Connections without /rpc/set_currently_editing URI can reset a session's suspend timeout")
+   {
+      expect_true(filters.shouldResetSuspendTimer(notYourEvent));
+   }
+}
+
+} // namespace tests
+} // namespace suspend
+} // namespace session
+} // namespace rstudio
```
AndrewMcClain, line 79:
```suggestion
} // namespace rstudio

```

src/cpp/session/include/session/SessionSuspendFilter.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+protected:
+   std::vector<boost::shared_ptr<SessionSuspendFilter>> m_filters; //!< Expandable list of filters to check connections with
+};
+
+} // namespace suspend
+} // namespace session
+} // namespace rstudio
+
+#endif // SESSION_SUSPEND_FILTER_HPP
```
AndrewMcClain, line 62:
```suggestion
#endif // SESSION_SUSPEND_FILTER_HPP

```

AndrewMcClain (commented):
I like the a approach! Very clear and easy to follow.

Issue #10949, opened 2022-04-11
Linter error is not appearing in the gutter in visual mode


astayleraz:
### System details

    RStudio Edition : 
    RStudio Version : 2022.06.0-daily+284
    OS Version      : ubuntu 20
    R Version       : 4.1.2

Failing automation test: https://build.rstudioservices.com/job/IDE/job/qa-opensource-automation/1247/testReport/rstudio_server_pro.tests.test_VisualMarkdownEditor/TestVisualMarkdownEditor/test_linting_error_appears_in_gutter/

Note: Due to the number of automation failures we are having I was not able to narrow down the exact build this started to fail but it was working previously in SW.

### Steps to reproduce the problem

1. Create a rmarkdown file via console command `.rs.api.executeCommand('newRMarkdownDoc')` and choose to create a empty document.
2. Install the lintr package via the console. `install.packages("lintr")`
3. Restart r via console command `.rs.api.executeCommand('restartR')`
4. Insert the following in the source file:
~~~
```{r cars}
summary(cars,
         planes
         trains
         )
```
~~~
5. Click to view in visual mode.
6. Save the file.

### Describe the problem in detail
The linter error is not appearing in the gutter in visual mode. It shows as the image below:

![Screen Shot 2022-04-11 at 10 07 27 AM](https://user-images.githubusercontent.com/1482677/162799108-7dd0fb7d-af17-4e16-8933-63089f8de3c1.png)

### Describe the behavior you expected
The test is expecting to see the linter error below:
<img width="277" alt="Screen Shot 2022-04-11 at 10 46 21 AM" src="https://user-images.githubusercontent.com/1482677/162798942-143f28c8-8e40-40bc-9750-428dc9dd5dc5.png">

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
Do you mean the error message tooltip?

In your first screenshot, the cursor is not on the line with the associated error, which would explain its absence.

astayleraz:
[...]

Issue #10483, opened 2022-02-01
Source diagnostic gutter symbols are different in source editor versus visual editor, with some symbols missing


jgutman:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2022.02.0-preview+412
    OS Version      : Mac OS 12.1 M1 
    R Version       : R-devel 4.2.0 (64-bit)

### Steps to reproduce the problem

Create an Rmd file with an R code check with some diagnostic issues in it. All R diagnostics are enabled.
![image](https://user-images.githubusercontent.com/7817881/152016735-66b6968f-b0a2-4bc1-93d7-747f67c01029.png)
![image](https://user-images.githubusercontent.com/7817881/152016799-ea5c6e63-28ea-48ee-9880-2b3bad46f22a.png)

Code chunk:
~~~
```{r}
print(123 digits=3)
x <- yy

purrr::map ( 1 : 5)
```
~~~

In the source editor, we see 3 diagnostic marker icons displayed in the gutter.
ERROR (red X in circle): with the tooltip message `unexpected token 'digits'` and `unexpected token '='`
WARNING (yellow exclamation point in triangle): with the tooltip message `no symbol named 'yy' in scope`
INFO (blue `i` in circle): with the tooltip message `unnecessary whitespace` and `unexpected whitespace around extraction operator`

However, when we switch over to the visual editor 
1) At first, there are no diagnostic markers displayed. However if we edit and save the file, they do appear.
2) For lines which had multiple diagnostic messages associated with them in the source editor, only the last one appears.
3) The INFO (blue `i` in circle) symbol does not appear at all. ERROR (red X in circle) and WARNING (yellow exclamation point in triangle) do appear as before though.

![image](https://user-images.githubusercontent.com/7817881/152018067-c4ba9e4c-f213-4526-8116-c7b1b5a3e3b8.png)

- [X] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [X] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

jgutman:
[...]

PR #11225, opened 2022-05-17
Bugfix/visual mode markers

timtmok:
### Intent
Address #10949 and #10483

### Approach
Various fixes to visual mode markers. Some were issues with the virtual gutter because disabling line numbers means ACE hides the gutter. We have to create our own "gutter" to display the markers.

* lint items without the raw text caused blank tooltips (can't use the lint text because it has HTML tags)
* visual mode chunks didn't process _style_ type items so those markers didn't have the correct CSS classes applied
* a visual mode chunk at the beginning of the file wasn't initialized properly (no scope meant no way to tell if the lint item belonged to the chunk)
* appended lint text to gutter marker tooltips

### Automated Tests
This should fix an existing automated test

### QA Notes
* use the sample Rmd code in the issues
* disable visual mode line numbers (Options > R Markdown > Visual > Show line numbers in code blocks
* viewing tooltips requires the cursor to be in the code chunk (ACE limitation?)

### Checklist


src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/visualmode/VisualModeChunkRowState.java
```diff
@@ -136,7 +136,15 @@ protected void removeClazz()
    @Override
    protected void setTitle(String title)
```
kevinushey, line 137: This change feels a little strange. Should we instead have a function like `appendToTitle()` for the cases where we want to append content?

kevinushey: Or instead, the caller of `setTitle()` should be querying the current title and appending to it instead; I don't think that should be happening here.

timtmok: I like the suggestion of adding `appendToTitle()`

kevinushey (approved):
LGTM!

Issue #10677, opened 2022-02-24
Setup chunk automatic execution should not be controlled by "Show chunk output inline" setting


jgutman:
Based on #10592, it is a source of user confusion that Setup chunks only execute automatically when BOTH `Automatically run Setup chunk when needed` and `Show chunk output inline in all documents` are enabled. We should change the behavior so that setup chunks run automatically when `Automatically run Setup chunk when needed` is ON, regardless of whether or not the user is in "notebook mode" with `Show chunk output inline in all documents` enabled.

jgutman:
See also #10645 -- when `Show chunk output inline in all documents` is disabled, chunk options (particularly the new YAML style chunk options) appear to be ignored

FYI @hadley

PR #11237, opened 2022-05-18
change def of isExecutableChunk to recognize `#| eval: false`

jgutman:
### Intent

Addresses edge case of #10645 found by @hadley where the YAML chunk options only work in inline chunk execution ("notebook mode"), not when chunks are executed in the console.

### Approach

Ultimately, we need to revamp our definition of notebook mode a bit -- it seems too expansive, in that chunk execution is handled differently in many ways depending on whether chunk execution is set to inline or console. However, the immediate concern here is that when `#| eval: false` is provided as a chunk option, the chunk erroneously executes when in console mode. It correctly does not execute when in inline mode. We previously changed behavior on the backend to parse these new format chunk options. Here, we additionally update the frontend method `isExecutableChunk` so that it not only returns `false` when the chunk starts with `{r eval = FALSE}` but also when it contains `#| eval: false`.

It feels a little unsatisfying to me to have separate frontend and backend methods for parsing these new chunk options, and ultimately a long-term solution should address the issue in #10677 as well, but I think this is good as a stopgap for now to address the immediate issue. 

### Automated Tests

None

### QA Notes

Create and save the following Rmd file:

~~~~
---
output: html_document
editor_options: 
  chunk_output_type: console
---

```{r}
print("I should run")
```

```{r}
#| eval: false
stop("I shouldn't run")
```
~~~~

Using the gearbox next to the Knit button,  "Chunk Output in Console" should be selected.
image

Select "Run all" (option + cmd + R)

Ensure that `stop("I shouldn't run")` does not run and does not get output to the console

### Checklist

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTarget.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTarget.java
@@ -6181,6 +6181,28 @@ public class TextEditingTarget implements
       if (reEvalFalse.test(headerText))
          return false;
 
+      // Also check for YAML style chunk option with eval false
+      Pattern reYamlOpt = Pattern.create("^#\\| .*");
+      Pattern reYamlEvalFalse = Pattern.create("eval\\s*:\\s*false");
+      int start = chunk.getBodyStart().getRow();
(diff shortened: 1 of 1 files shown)


kevinushey (approved):
LGTM!

Issue #10361, opened 2022-01-14
update rtools42 URL when CRAN-stable link is available


kevinushey:
https://github.com/rstudio/rstudio/blob/01ef6e05e3a5e02f39741e78cc6519b6b074b20d/src/cpp/core/r_util/RToolsInfo.cpp#L292-L296

ronblum:
QA Note: In testing, include verification that
1) A fresh installation of R 4.2 (no previous installation of R and Rtools) triggers installation of Rtools42
2) Installing R 4.2 on top of earlier versions of R and Rtools triggers installation of Rtools42
3) Installing an earlier version of R (such as R 3.6.3) triggers installation of earlier versions of Rtools (e.g. Rtools35)

kevinushey:
For PT, I wonder if we should just use our own copy of Rtools in an S3 bucket, and later update that to an appropriate CRAN URL in a patch release (or later)?

mikebessuille:
@kevinushey this seems like a very reasonable approach to me, because we can control the version for the period of PT that it's needed.  In a patch we can update it.

kevinushey:
Done with https://github.com/rstudio/rstudio/commit/1bf4233bb0230021cf673d96fc4c896d17332f1f.

kevinushey:
Moving this to SW just to indicate it's not part of the current release work (we can bring it back for a patch release milestone later if necessary)

jgutman:
https://cran.r-project.org/bin/windows/Rtools/rtools42/rtools.html

Looks like the link is available now

jgutman:
Ah actually looks like the exe is still versioned so I guess it's not stable yet?

kevinushey:
FWIW I've contacted Tomas at R Core to ask if they'll plan to publish Rtools with a stable URL that RStudio could use here, in tandem with the R 4.2.0 release.

ronblum:
[Note: just tracking, today the link is https://cran.r-project.org/bin/windows/Rtools/rtools42/files/rtools42-5168-5107.exe]

jgutman:
[...]

PR #11248, opened 2022-05-19
Get Cran URL for Rtools 42 windows installer

jgutman:
Addresses #10361 

Since we are not expecting a stable CRAN url, one option is to extract the installer url from the Rtools 42 landing page. 

We can do this in `SessionInstallRtools` using a regex to parse the HTML

### DIFF
--- a/src/cpp/session/modules/build/SessionInstallRtools.cpp
+++ b/src/cpp/session/modules/build/SessionInstallRtools.cpp
@@ -105,10 +105,14 @@ Error installRtools()
                .call();
 
          std::string rToolsHome = file_utils::readFile(tmpFile);
-         boost::regex reLinkPattern("<a\\shref=\"(.*\\.exe)\">\\s*Rtools42 installer<\\/a>", boost::regex::icase);
-         boost::cmatch matches;
-         if (regex_utils::match(rToolsHome.c_str(), matches, reLinkPattern))
+
+         boost::regex reLinkPattern("<a\\shref=\"(.*rtools.*\\.exe)\">\\s*.+<\\/a>", boost::regex::icase);
+         boost::smatch matches;
+         if (regex_utils::match(rToolsHome, matches, reLinkPattern))
             url = matches[1];
+         else
+            // Use RStudio's hosted copy as a fallback if no match is found
+            url = "https://rstudio.org/links/rtools42";
       }
       catch (std::exception& e)
       {
--- a/src/cpp/session/modules/build/SessionInstallRtools.cpp
+++ b/src/cpp/session/modules/build/SessionInstallRtools.cpp
@@ -16,9 +16,12 @@
 #include "SessionInstallRtools.hpp"
 
 #include <boost/format.hpp>
+#include <boost/regex.hpp>
 
 #include <shared_core/Error.hpp>
 #include <core/StringUtils.hpp>
+#include <core/FileUtils.hpp>
+#include <core/RegexUtils.hpp>
 
 #include <core/r_util/RToolsInfo.hpp>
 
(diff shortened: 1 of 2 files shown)


src/cpp/session/modules/build/SessionInstallRtools.cpp
```diff
@@ -89,6 +92,31 @@ Error installRtools()
    if (error)
       return error;
 
+   if (version == "4.2")
```
jgutman, line 95: I still need to test this out on Windows, but wanted to put up for review to see if this is even a good idea

src/cpp/session/modules/build/SessionInstallRtools.cpp
```diff
@@ -89,6 +92,31 @@ Error installRtools()
...
+         FilePath tmpFile;
+         FilePath::tempFilePath(tmpFile);
+         r::exec::RFunction("utils:::download.file")
+               .addParam("url", url)
+               .addParam("destfile", tmpFile.getAbsolutePath())
+               .addParam("mode", "rb")
+               .call();
+
+         std::string rToolsHome = file_utils::readFile(tmpFile);
+         boost::regex reLinkPattern("<a\\shref=\"(.*\\.exe)\">\\s*Rtools42 installer<\\/a>", boost::regex::icase);
```
kevinushey, line 108: I wonder if we should be more permissive? I worry a little that the `Rtools42 installer` text could change; it'd be easier if we could just iterate through all of the `<a href ="<...>">` entries in the document and pick the one that "looks like" the Rtools installer.

For reference, if you want to find and iterate over multiple regular expression matches, you could use a regex_token_iterator. See:

https://github.com/rstudio/rstudio/blob/8be41624a8de5d4caf60ab6510db57c3a22000c4/src/cpp/session/modules/rmarkdown/SessionBlogdown.cpp#L238-L252

for a somewhat similar example.

Alternatively, you could move all of this logic into R and have a helper function like `.rs.findRtools42Installer()`

src/cpp/session/modules/build/SessionInstallRtools.cpp
```diff
@@ -89,6 +92,31 @@ Error installRtools()
    if (error)
       return error;
 
+   if (version == "4.2")
+   {
+      try
+      {
+         FilePath tmpFile;
+         FilePath::tempFilePath(tmpFile);
+         r::exec::RFunction("utils:::download.file")
```
kevinushey, line 101: Should errors from this function be handled / logged? (Do we still want to attempt parsing / reading files on failure?)

src/cpp/session/modules/build/SessionInstallRtools.cpp
```diff
@@ -89,6 +92,31 @@ Error installRtools()
...
+               .addParam("url", url)
+               .addParam("destfile", tmpFile.getAbsolutePath())
+               .addParam("mode", "rb")
+               .call();
+
+         std::string rToolsHome = file_utils::readFile(tmpFile);
+         boost::regex reLinkPattern("<a\\shref=\"(.*\\.exe)\">\\s*Rtools42 installer<\\/a>", boost::regex::icase);
+         boost::cmatch matches;
```
kevinushey, line 109: You could also use `boost::smatch` and match on the `std::string` directly (rather than the `c_str()`).

kevinushey (commented):
I'm overall a fan of the approach; my only suggestion is to be a bit more permissive (in case the text in the document changes while the relevant link remains).

src/cpp/session/modules/build/SessionInstallRtools.cpp
jgutman: I'm not sure what kind of handling we would do -- if we fail to download the file, what would we be parsing? I still kept the overall fallback as our s3 copy in case of any failure

jgutman:
My windows setup is borked, so I haven't been able to fully test this out. Seems like it would work though?

src/cpp/session/modules/build/SessionInstallRtools.cpp
kevinushey: I just think we should have something like:

```
Error error = r::exec::RFunction(<...>);
if (error) {
  LOG_ERROR(error);
  // use fallback url
}
```

In other words, check for and handle an error explicitly there.

kevinushey: You could also put this snippet of code into a helper function, e.g.

```
std::string resolveRtoolsInstallerUrl(const std::string& url) { ... }
```

and that helper would just do what you have here (with a guard that returns the default URL for R < 4.2.0)

kevinushey (changes requested):
Overall approach LGTM, but I think we should still explicitly handle any errors from `utils::download.file()`.

kevinushey (approved):
LGTM!

PR #11257, opened 2022-05-20
Feature/3094 session metadata direct database clean

AndrewMcClain:
### Intent

Implementation of direct database connection

OS Portion, Addresses rstudio/rstudio-pro#3094

### Approach

A cherry-pick / cleanup of the previous pr

### Automated Tests

Automated tests exist for the db connection

### QA Notes

Alone, this PR does nothing, however following PR's will build on it allowing the DB connection to be exercised by users

### Checklist


jeffvroom (approved):
LGTM!!!

src/cpp/core/r_util/RActiveSessionStorage.cpp
```diff
@@ -26,153 +25,165 @@ namespace r_util {
...
+   {
+      std::map<std::string, std::string> propertyValue{};
+      *pValue = "";
+
+      Error error = readProperties({ name }, &propertyValue);
+
+      if (error)
+         return error;
+
+      std::map<std::string, std::string>::iterator iter = propertyValue.find(name);
```
MariaSemple, line 73: Nit: const correctness
```suggestion
      std::map<std::string, std::string>::const_iterator iter = propertyValue.find(name);
```

src/cpp/core/r_util/RActiveSessionStorage.cpp
```diff
@@ -26,153 +25,165 @@ namespace r_util {
...
+         { "r_version" , "r-version" },
+         { "r_version_label" , "r-version-label" },
+         { "r_version_home" , "r-version-home" },
+         { "working_directory" , "working-dir" },
+         { "launch_parameters" , "launch-parameters" }
+       };
+
+   Error FileActiveSessionStorage::readProperty(const std::string& name, std::string* pValue)
+   {
+      std::map<std::string, std::string> propertyValue{};
```
MariaSemple, line 65: No need to supply an empty initializer list (here and below)
```suggestion
      std::map<std::string, std::string> propertyValue;
```

src/cpp/core/r_util/RActiveSessionStorage.cpp
```diff
@@ -26,153 +25,165 @@ namespace r_util {
...
+         *pValue = iter->second;
+      
+      return Success();
+   }
+
+   Error FileActiveSessionStorage::readProperties(const std::set<std::string>& names, std::map<std::string, std::string>* pValues)
+   {
+      std::vector<FilePath> failedFiles{};
+      pValues->empty();
+      for (const std::string &name : names)
```
MariaSemple, line 85: Nit: style
```suggestion
      for (const std::string& name : names)
```

src/cpp/core/r_util/RActiveSessionStorage.cpp
```diff
@@ -26,153 +25,165 @@ namespace r_util {
...
+
+         std::string propertyName = getFileNameProperty(file.getFilename());
+         pValues->insert(std::pair<std::string, std::string>{propertyName, value});
+      }
+
+      if(failedFiles.empty())
+         return Success();
+      else
+         return createError("UnableToReadFiles", "Failed to read from the following files ",
+            failedFiles, ERROR_LOCATION);
```
MariaSemple, line 130: Nit: prefer an unqualified `return Success();` at the end
```suggestion
      if(!failedFiles.empty())
         return createError("UnableToReadFiles", "Failed to read from the following files ",
            failedFiles, ERROR_LOCATION);
            
      return Success();
```

src/cpp/core/r_util/RActiveSessions.cpp
```diff
@@ -14,29 +14,38 @@
...
 #include <core/r_util/RActiveSessions.hpp>
-#include <core/r_util/RActiveSessionStorage.hpp>
 
 #include <boost/bind/bind.hpp>
 #include <boost/algorithm/string/predicate.hpp>
 
+#include <core/r_util/RActiveSessionsStorage.hpp>
+#include <core/Log.hpp>
 #include <core/StringUtils.hpp>
 #include <core/FileSerializer.hpp>
```
MariaSemple, line 24: Nit: include order
```suggestion
#include <core/Log.hpp>
#include <core/StringUtils.hpp>
#include <core/FileSerializer.hpp>
#include <core/r_util/RActiveSessionsStorage.hpp>
```

src/cpp/core/r_util/RActiveSessions.cpp
```diff
@@ -86,92 +86,42 @@ Error ActiveSessions::create(const std::string& project,
    return Success();
 }
 
-namespace {
+namespace
+{
 
 bool compareActivityLevel(boost::shared_ptr<ActiveSession> a,
-                          boost::shared_ptr<ActiveSession> b)
+                        boost::shared_ptr<ActiveSession> b)
```
MariaSemple, line 93: Could you undo theses style changes?

src/cpp/core/r_util/RActiveSessions.cpp
```diff
@@ -86,92 +86,42 @@ Error ActiveSessions::create(const std::string& project,
...
-   if (error)
-   {
-      LOG_ERROR(error);
-      return sessions;
-   }
-   std::string prefix = kSessionDirPrefix;
-   for (const FilePath& child : children)
+   std::vector<std::string> sessionIds = storage_->listSessionIds();
+   std::vector<boost::shared_ptr<ActiveSession>> sessions{};
+   for(std::string id : sessionIds)
```
MariaSemple, line 104: Nit: avoid a copy unless necessary
```suggestion
   for(const std::string& id : sessionIds)
```

src/cpp/core/r_util/RActiveSessions.cpp
```diff
@@ -86,92 +86,42 @@ Error ActiveSessions::create(const std::string& project,
...
-               // written back into them)
-               Error error = pSession->destroy();
-               if (error)
-                  LOG_ERROR(error);
-            }
-         }
-      }
-
+      boost::shared_ptr<ActiveSession> candidateSession = storage_->getSession(id);
+      if(candidateSession->validate(userHomePath, projectSharingEnabled))
```
MariaSemple, line 107: Nit: style
```suggestion
      if (candidateSession->validate(userHomePath, projectSharingEnabled))
```

src/cpp/core/r_util/RActiveSessionsStorage.cpp
```diff
@@ -0,0 +1,114 @@
...
+} // anonymous namespace
+
+FileActiveSessionsStorage::FileActiveSessionsStorage(const FilePath& rootStoragePath)
+{
+   storagePath_ = ActiveSessions::storagePath(rootStoragePath);
+   Error error = storagePath_.ensureDirectory();
+   if (error)
+   {
+      LOG_ERROR(error);
+   }
```
MariaSemple, line 44: Nit: fewer lines
```suggestion
   if (error)
      LOG_ERROR(error);
```

src/cpp/core/r_util/RActiveSessionsStorage.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+core::Error FileActiveSessionsStorage::createSession(const std::string& id, std::map<std::string, std::string> initialProperties)
+{
+   FilePath sessionScratchPath = storagePath_.completeChildPath(kSessionDirPrefix + id);
+   Error error = sessionScratchPath.ensureDirectory();
+
+   if(error)
+   {
+      return error;
+   }
```
MariaSemple, line 61: Nit: fewer lines
```suggestion
   if(error)
      return error;
```

src/cpp/core/r_util/RActiveSessionsStorage.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+   FileActiveSessionStorage session = FileActiveSessionStorage{sessionScratchPath};
+   error = session.writeProperties(initialProperties);
+   return error;
```
MariaSemple, line 65: Nit: fewer lines, possibly one less copy (although the compiler probably already removed it)
```suggestion
   return session.writeProperties(initialProperties);
```

src/cpp/server/DBActiveSessionStorage.cpp
```diff
@@ -0,0 +1,404 @@
...
+               });
+   return setValuesString;
+}
+
+std::string getColumnNameList(const std::set<std::string>& colNames)
+{
+   std::string cols = std::accumulate(++colNames.begin(), colNames.end(), 
+               *(colNames.begin()), [](std::string a, std::string b) {
+               return a + ", " + b;
+               });
```
MariaSemple, line 81: Spacing is weird here. Possible alterantive:
```suggestion
   std::string cols = std::accumulate(
      ++colNames.begin(),
      colNames.end(), 
      *(colNames.begin()),
      [](std::string a, std::string b) {
         return a + ", " + b;
      });
```

src/cpp/server/DBActiveSessionStorage.cpp
```diff
@@ -0,0 +1,404 @@
...
+   *pValue = "";
+   boost::shared_ptr<database::IConnection> connection;
+   Error error = getConnectionOrOverride(&connection);
+   
+   if(error)
+   {
+      return error;
+   }
+
+
```
MariaSemple, line 157: Nit: remove some lines
```suggestion
   Error error = getConnectionOrOverride(&connection);
   if(error)
      return error;

```

src/cpp/server/DBActiveSessionStorage.cpp
```diff
@@ -0,0 +1,404 @@
...
+      }
+      else
+      {
+         return Error{"Session does not exist", errc::SessionNotFound, ERROR_LOCATION};
+      }
+   }
+   else
+   {
+      return Error{"DatabaseException", errc::DBError, "Database exception during property read [ session:"+sessionId_+" property:"+name+" ]", error, ERROR_LOCATION};
+   }
```
MariaSemple, line 190: Nit: re-org for less indentation
```suggestion
   error = connection->execute(query, results);
   if (error)
      return Error("DatabaseException", errc::DBError, "Database exception during property read [ session:"+sessionId_+" property:"+name+" ]", error, ERROR_LOCATION);

   database::RowsetIterator iter = results.begin();
   if(iter != results.end())
   {
      if(name != kUserId)
         *pValue = iter->get<std::string>(name, "");
      else
         *pValue = std::to_string(iter->get<int>(name));

      if(++iter != results.end()) 
         return Error("Too many sessions returned", errc::TooManySessionsReturned, ERROR_LOCATION);
   }
   else
      return Error("Session does not exist", errc::SessionNotFound, ERROR_LOCATION);
```

MariaSemple: Can you make similar adjustments throughout?

src/cpp/server/DBActiveSessionStorage.cpp
```diff
@@ -0,0 +1,404 @@
...
+   boost::shared_ptr<database::IConnection> connection;
+   Error error = getConnectionOrOverride(&connection);
+
+   if(error)
+   {
+      return error;
+   }
+   
+   std::string namesString = getColumnNameList(names);
+   database::Query query = connection->query("SELECT "+namesString+" FROM "+kTableName+" WHERE "+kSessionIdColumnName+"=:id")
```
MariaSemple, line 207: Nit: should have spaces between the `+`s
```suggestion
   database::Query query = connection->query("SELECT " + namesString + " FROM " + kTableName + " WHERE " + kSessionIdColumnName + "=:id")
```

src/cpp/server/DBActiveSessionStorage.cpp
```diff
@@ -0,0 +1,404 @@
...
+         }
+      }
+      else
+      {
+         return Error{"Session does not exist", errc::SessionNotFound, ERROR_LOCATION};
+      }
+   }
+   else
+   {
+      return Error{"DatabaseException", errc::DBError, "Database exception during proprerties read [ session:"+sessionId_+" properties:"+namesString+" ]", error, ERROR_LOCATION};
```
MariaSemple, line 232: It looks like `Error()` and `Error{}` are equivalent, but since we tend to use `()` throughout the code base, could you change the style to match?

src/cpp/server/DBActiveSessionStorage.cpp
```diff
@@ -0,0 +1,404 @@
...
+   error = connection->execute(query, rowset);
+
+   if(!error)
+   {
+      database::RowsetIterator iter = rowset.begin();
+      if(iter != rowset.end())
+      {
+         populateMapWithRow(iter, pValues);
+         // Sanity check number of returned rows, by using the pk in the where clause we should only get 1 row
+         if(++iter != rowset.end())
```
MariaSemple, line 220: There should be spaces between the `if` and the `(`. Could you add throughout?
```suggestion
         if (++iter != rowset.end())
```

src/cpp/server/DBActiveSessionStorage.cpp
```diff
@@ -0,0 +1,404 @@
...
+      .withInput(sessionId_);
+   database::Rowset rowset{};
+
+   *pValue = false;
+
+   error = connection->execute(query, rowset);
+
+   if(error)
+   {
+      return Error{"DatabaseException", errc::DBError, "Error while deleting session metadata [ session:"+sessionId_+" ]", error, ERROR_LOCATION};
```
MariaSemple, line 369: Is this message a copy-paste error?

src/cpp/server/DBActiveSessionStorage.cpp
```diff
@@ -0,0 +1,404 @@
...
+
+   // Sanity checking, but should always have 1 row containing the count of rows
+   if(iter != rowset.end())
+   {
+      int count = iter->get<int>("COUNT(*)");
+
+      // ensure one and only one
+      if(count > 1)
+      {
+         return Error{"Too Many Sessions Returned", errc::TooManySessionsReturned, ERROR_LOCATION};
```
MariaSemple, line 381: I suspect this message won't give enough information. Could you add the amount we found and at least the ID of the session that we were looking for?

src/cpp/server/DBActiveSessionStorage.cpp
```diff
@@ -0,0 +1,404 @@
...
+         }
+      }
+   }
+   return Success();
+}
+
+Error DBActiveSessionStorage::isEmpty(bool* pValue)
+{
+   bool val;
+   Error error = isValid(&val);
```
MariaSemple, line 397: What if the session isn't valid because there are already two entries for the same ID? I think we need a separate check here. Maybe a shared implementation that gets the count and different validators for the count in each method?

src/cpp/server/DBActiveSessionStorage.cpp
```diff
@@ -0,0 +1,404 @@
...
+   {
+      return Error{"DatabaseException", errc::DBError, "Database exception during proprerties read [ session:"+sessionId_+" properties:"+namesString+" ]", error, ERROR_LOCATION};
+   }
+
+   return error;
+}
+
+Error DBActiveSessionStorage::readProperties(std::map<std::string, std::string>* pValues)
+{
+   std::set<std::string> all{"*"};
```
MariaSemple, line 240: In general, we avoid using `*` in select lists to avoid the risk of unexpected column name, order, or amount changes. In this case I think it's fine, but we should add an explanatory comment about why it's safe here.

src/cpp/server/DBActiveSessionsStorage.cpp
```diff
@@ -0,0 +1,147 @@
...
+      database::Rowset rowset{};
+      error = connection->execute(query, rowset);
+
+      if(!error)
+      {
+         return rowset.begin()->get<size_t>("COUNT(*)");
+      }
+   }
+
+   return (size_t) 0;
```
MariaSemple, line 96: Do you need this cast?

src/cpp/server/include/server/DBActiveSessionStorage.hpp
```diff
@@ -0,0 +1,68 @@
...
+   Error destroy() override;
+   Error isValid(bool* pValue) override;
+   Error isEmpty(bool* pValue) override;
+private:
+   std::string sessionId_;
+   boost::shared_ptr<database::IConnection> overrideConnection_;
+   Error getConnectionOrOverride(boost::shared_ptr<database::IConnection>* connection);
+};
+
+Error getConn(boost::shared_ptr<database::IConnection>* connection);
```
MariaSemple, line 51: Do you need this method externally to the CPP? If not, we can just remove the declaration here and put it in the anonymous namespace.

MariaSemple (changes requested):
Looking close! Mostly stylistic changes.

src/cpp/server/DBActiveSessionStorage.cpp
jeffvroom: How about if we add a unique constraint on the active_session_metadata table for the session id?  It's best to avoid this situation and catch those errors on the insert.

AndrewMcClain: Since I've already got duplicate row guards in place, Why don't we make swapping that responsiblity a new issue if we want to tackle it. I really want to put a bow on this and spin off any lower priority changes into new issues if we can.

MariaSemple (approved):
LGTM!

Issue #9947, opened 2021-10-12
Creating a new project from inside an existing project can result in incomplete repo list in renv.lock file


cm421:
### System details

    RStudio Edition : Server
    RStudio Version : 1.4.1106-5
    OS Version      : RHEL 8.4
    R Version       : 3.6.3

### Steps to reproduce the problem

This issue was brought to light in support ticket 66222, and seems related to work done for GH issue #7136. 

To create the issue:

See the steps listed in the post on Oct 01 at 10:51 of the support ticket. The issue comes to light in test 3.

### Describe the problem in detail

When creating a new project from a project that contains a renv lock file that does not have all the repos listed, the new project won't have all the repos listed either. Creating a new project and then calling `renv::init()` does include all the desired repos.

### Describe the behavior you expected

Creating a new project and ticking "use renv" should function the same as when making a new project and then manually calling `renv::init()`

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
This sounds like an issue that should be resolved in the most recent release of RStudio. See:

https://github.com/rstudio/rstudio/commit/d4eac9a41997739b693778f468c9dd2c211d4b62

From the commit, it appears the fix appeared in 1.4.1717.

kevinushey:
And it's worth noting that `renv::snapshot()` just saves whatever repositories are currently active (as seen via `getOption("repos")`) into the lockfile, so whatever they see in the lockfile should be what they had active in the R session.

mikebessuille:
@cm421 would you mind retesting this scenario per Kevin's comment above, and closing?

cm421:
Apologies for this falling through the cracks. Testing on RStudio 2022.02.0+443.pro2 with `renv` 0.15.4 this still occurs.

### Part 1
Steps I took:

1. With only one repo listed in `getOption("repos")`, run `renv::init()` in a directory
2. Add a repo Tools->Global Options->Packages->Add Secondary Repo
3. `getOption("repos")` still only returns the one repo
[...]

PR #11264, opened 2022-05-22
Bugfix/renv repo manage warning

cm421:
### Intent

Fixes https://github.com/rstudio/rstudio/issues/9947

### Approach

Check if `renv` is active, and if it is, display an info bar indicating Global Option repo settings are overwritten by the project's `renv.lock` file

The original `CRAN repositories modified outside package preferences.` info bar can still appear if repositories are indeed being modified outside package preferences, but only when `renv` is not doing the modifying

### Automated Tests

I've written selenium tests for this and will link that PR to this one

### QA Notes

After calling `renv::init()` inside a directory, the new InfoBar message should appear under Global Options->Packages

### Checklist


src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/PackagesPreferencesPane.java
```diff
@@ -342,6 +345,25 @@ public void onError(ServerError error)
             }
          }
       );
+   
+      // If renv is managing our repos, make the info bar more specific
+      server_.readProjectOptions(
```
kevinushey, line 350: Should we check first whether a project is actually open? (Or does this still do the right thing in that case as well?)

kevinushey (approved):
LGTM!

src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/PackagesPreferencesPane.java
cm421: It actually already does the right thing both 

1. in a project (with or without `renv` active), and 
2. in a non-R project (with or without `renv` active)

PR #11278, opened 2022-05-24
update stack protection

Enchufa2:
### Intent

See [Debian/Hardening](https://wiki.debian.org/Hardening):

> Prior to GCC 4.9, `-fstack-protector --param ssp-buffer-size=4' is used to cover functions that defines a 4 or more byte local character array, which is an okay balance for security and performance. For those who want to protect all the functions then -fstack-protector-all is recommended.
>
> Since GCC 4.9, -fstack-protector-strong, an improved version of -fstack-protector is introduced, which covers all the [more paranoid conditions](https://outflux.net/blog/archives/2014/01/27/fstack-protector-strong/) that might lead to a stack overflow but not trade performance like -fstack-protector-all, thus it becomes default.

This is also the default in Fedora, but this definition overrides our default. More generally, it would be nice if RStudio wouldn't add any flags if they are set already.

### Approach

Simple update of the definition.

### Automated Tests

None.

### QA Notes

Stack protection is improved, no regressions expected.

### Checklist


kevinushey (approved):
LGTM!

Issue #10906, opened 2022-04-05
Closed secondary editor window comes back on relaunch (Electron)


gtritchie:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2022.06.0-daily+270 (Electron and Qt)
    OS Version      : macOS 12.3.1
    R Version       : 4.1.3

### Steps to reproduce the problem

- start IDE with no project, no open files
- File / New File / R Script (create an empty file)
- Click on the "Show in new window" icon in editor toolbar, to move the document into a separate window
- Close that separate window by clicking on the red circle

![screenshot of editor window with arrow pointing at close icon](https://user-images.githubusercontent.com/10569626/161796051-aa02aa8c-268b-401f-8334-8fe05722e697.png)

- Exit RStudio
- Restart RStudio

### Describe the problem in detail

- the secondary editor window with Untitled document comes back

### Describe the behavior you expected

- the closed window should not come back
- it works if you close the secondary window by clicking the [x] on the document tab, or via Cmd+W; it's only when clicking the red circle that this happens
- the same scenario works correctly in the same Qt-based desktop build

- [ ] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ ] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [ ] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

gtritchie:
The problem is we aren't receiving this 'unload' event in Electron.

https://github.com/rstudio/rstudio/blob/9efe6afea8f1217110cfa41b9ef13ddf199368b7/src/gwt/src/org/rstudio/studio/client/common/satellite/Satellite.java#L191-L196

The call to `notifyRStudioSatelliteClosed()` is what takes care of actually closing the underlying documents so they don't come back on relaunch.

Note that `unload` is documented as being unreliable and best-avoided:

https://developer.mozilla.org/en-US/docs/Web/API/Window/unload_event

matheustavaresdev:
I will try to add a `pagehide` event and see if it works

matheustavaresdev:
> No, haven't tried fixing it yet.

Alright I'll try those options

matheustavaresdev:
Sure thing @mikebessuille

PR #11282, opened 2022-05-24
Fix issue that would keep Secondary (Satellite) Windows Open

matheustavaresdev:
### Intent

This aims to fix an issue that would make Secondary ( Satellite ) Windows open after being closed by clicking on the `X` button.

### Approach

Added a custom `onclose` event for Satellite windows, and disabled that regular window event for those windows. 

### Automated Tests

None

### QA Notes

Refer to #10906

### Checklist

Addresses #10906


kevinushey:
This change doesn't feel right to me; the `closeEvent()` implementation specifically checks `window.opener`:

https://github.com/rstudio/rstudio/blob/128552161df9c8463cc92c376d5bc0c2304de261/src/node/desktop/src/main/desktop-browser-window.ts#L272-L291

so it seems like it's intentional that this could would be run for satellite windows as well. (That said, I still don't quite understand the underlying issue here; just that this change doesn't feel quite right)

gtritchie:
I didn't get a chance to review this today, will do so tomorrow. I agree this doesn't look quite right but I only glanced at it.

matheustavaresdev:
> This change doesn't feel right to me; the `closeEvent()` implementation specifically checks `window.opener`:
> 
> https://github.com/rstudio/rstudio/blob/128552161df9c8463cc92c376d5bc0c2304de261/src/node/desktop/src/main/desktop-browser-window.ts#L272-L291
> 
> so it seems like it's intentional that this could would be run for satellite windows as well. (That said, I still don't quite understand the underlying issue here; just that this change doesn't feel quite right)
@kevinushey this code was written before my change

matheustavaresdev:
What I did was to run the same code that was running when the user would press `CMD+W` for when the user would click on the X button as follows:

https://github.com/rstudio/rstudio/blob/bugfix/secondary-editor-keeps-open/src/node/desktop/src/main/satellite-window.ts#L44-L46

Then, I disabled the code that would run simultaneously for when the user would click the `X` button:

https://github.com/rstudio/rstudio/blob/bugfix/secondary-editor-keeps-open/src/node/desktop/src/main/desktop-browser-window.ts#L256-L260

If I didn't make the last change above, something would be messed up and the window would still reopen after app relaunch. The code below would always be triggered, adding a new pending window on app relaunch, and then reopening the satellite window that should be closed.

https://github.com/rstudio/rstudio/blob/bugfix/secondary-editor-keeps-open/src/node/desktop/src/main/gwt-callback.ts#L465-L479

gtritchie:
I'll take a closer look today.

gtritchie:
This is definiltely not the right fix and can cause data loss.

For example, open a file in a satellite window, then modify it but don't save. Then close the window via the [x]. By not calling SatelliteWindow.onClose and letting it cancel the window close operation, we bypass code that checks for modified files and gives the user the chance to save/discard/cancel the close operation.

I have an idea what the problem is; might actually need to be fixed in GWT and not Electron. I'll add more notes once I've dug into that a bit more closely.

gtritchie:
Also, Cmd+W is not supposed to be the same as clicking the [x]. The [x] should close the entire window (even if it has multiple editor tabs in it), whereas Cmd+W should only close the active document. If there is only one document, then it will also close the window.

matheustavaresdev:
> This is definiltely not the right fix and can cause data loss.
> 
> 
> 
> For example, open a file in a satellite window, then modify it but don't save. Then close the window via the [x]. By not calling SatelliteWindow.onClose and letting it cancel the window close operation, we bypass code that checks for modified files and gives the user the chance to save/discard/cancel the close operation.
> 
> 
> 
> I have an idea what the problem is; might actually need to be fixed in GWT and not Electron. I'll add more notes once I've dug into that a bit more closely.

Oh probably that's most likely the case. The code seemed right on the electron side. I did this change to fix the issue for the window reopening on app relaunch. Didn't notice those other case scenarios for testing my fix. Probably will have to undo it and dig it on GWT.

gtritchie:
Closing this PR, will open a new one when we find the right fix.

Issue #11234, opened 2022-05-18
Intermittent unit test failure on bionic-amd64-server


skyeturriff:
Looking at this build:

https://build.rstudioservices.com/blue/organizations/jenkins/IDE%2Fopen-source-pipeline/detail/main/1251/pipeline/1246

The are failures in the `core` tests on `bionic-amd64-server`:

```
[2022-05-18T10:00:54.540Z] ProcessTests
[2022-05-18T10:00:54.540Z]   Normal exit code is reported properly
[2022-05-18T10:00:54.540Z] -------------------------------------------------------------------------------
[2022-05-18T10:00:54.540Z] ../../../src/cpp/core/system/ProcessTests.cpp:522
[2022-05-18T10:00:54.540Z] ...............................................................................
[2022-05-18T10:00:54.540Z] 
[2022-05-18T10:00:54.540Z] ../../../src/cpp/core/system/ProcessTests.cpp:591: FAILED:
[2022-05-18T10:00:54.540Z]   REQUIRE( exitCode == 1 )
[2022-05-18T10:00:54.540Z] with expansion:
[2022-05-18T10:00:54.540Z]   143 == 1
```

This has been noted as happening occasionally, but not often (cc @gtritchie).

Given that we’d like to make the unit test failures flag the build as broken, we may want to look into this as part of the effort to reduce/eliminate these types of intermittent problems.

gtritchie:
This was fixed by https://github.com/rstudio/rstudio/pull/11283.

PR #11283, opened 2022-05-24
allow either 143 or 1 for process exit code for sigterm test

jeffvroom:
This test is intermittently failing so allowing either exit code.


gtritchie (approved):
Great, I think this fixes #11234?

PR #11292, opened 2022-05-25
If Quarto has been disabled, then disable the quarto test

jgutman:
Follow up to changes in 39de73fd61812d and #11201.

If quarto has been disabled (due to OS) we don't want to run this test because it will fail, and doesn't make sense.


gtritchie (approved):
LGTM! Thanks.

PR #11293, opened 2022-05-25
use Qt 5.12.10 on MacOS

gtritchie:
### Intent

We've been using Qt 5.12.10 for our Mac Desktop releases for the past couple of releases. On other platforms we still use 5.12.8.

The dependency-installer script for Mac (only used on dev machines) was still looking for 5.12.8.

### Approach

Bump script to look for 5.12.10.

### Automated Tests

N/A

### QA Notes

Dev machine only thing.

### Checklist

### DIFF
--- a/dependencies/osx/install-qt-macos
+++ b/dependencies/osx/install-qt-macos
@@ -21,7 +21,7 @@
 set -e
 
 if [ -z "$QT_VERSION" ]; then
-    QT_VERSION=5.12.8
+    QT_VERSION=5.12.10
 fi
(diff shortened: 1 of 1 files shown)


gtritchie:
> Should we also update the versions of Qt checked here?
> 
> The use of `QT_CANDIDATES` here is a bit of a lazy way to find a close-enough version of Qt, but we might want to just bite the bullet and list out the specific versions of Qt we want to use by default on each platform / OS.

If we weren't planning to retire Qt in the next release I'd say yes, but don't think it's worth touching now.

Issue #11247, opened 2022-05-19
Choose R Version cannot select network executables


matheustavaresdev:
### System details
```
RStudio Edition : Desktop (Electron)
RStudio Version : 2022.06.0-daily+374 (Electron)
OS Version      : Windows 11
R Version       : Multiple
```
### Steps to reproduce the problem
1) Have a R version installed into a network drive
2) Browse to it on `Choose R Installation` modal
3) Select a `R.exe`
### Describe the problem in detail
When the executable is selected, this error shows up:
![image](https://user-images.githubusercontent.com/36674530/169360093-c94ffe18-7711-42a5-b34d-ffd08a23744e.png)

### Describe the behavior you expected
That executable should be selected and used with no errors.

- [X] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [X] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [X] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
This is no longer an issue if the default version of R is on a network drive, and the default version of R is selected in the dialog.

However, now the network version of R no longer shows up in the list of available R binaries. @matheustavaresdev was this intentional?

matheustavaresdev:
@ronblum Please see https://github.com/rstudio/rstudio/pull/11168#issuecomment-1129131395 . We have agreed not to have the need of the new path being added to the list, only for it to be used. Just tested here and both network and mapped drivers are working here

ronblum:
Verified that this is now expected behavior. Done!

PR #11311, opened 2022-05-27
Fix issue when selecting a R Executable from a Network Path - Electron

matheustavaresdev:
### Intent

Fix an issue that would occur when trying to select a R Executable that was located on an unmounted Network Drive.

### Approach

Previously, we were not enabling paths with duplicated slashes. By removing that and enabling users to have double slashes ( Ex: `\\TOWER\downloads\R-3.6.3\bin\x64\R.exe`), this issue was solved.

### Automated Tests

Added a few other tests case scenarios for normalizing paths.

### QA Notes

Refer to #11247 

### Checklist

Addresses #11247

### DIFF
--- a/src/node/desktop/src/core/file-path.ts
+++ b/src/node/desktop/src/core/file-path.ts
@@ -42,7 +42,7 @@ const homePathLeafAlias = '~';
  * @return {*} 
  */
 function normalizeSeparators(path: string, separator = '/') {
-  return path.replace(/[\\/]+/g, separator);
+  return path.replace(/[\\]/g, separator);
 }
 
 /**
--- a/src/node/desktop/src/core/file-path.ts
+++ b/src/node/desktop/src/core/file-path.ts
@@ -35,7 +35,7 @@ const homePathAlias = '~/';
 const homePathLeafAlias = '~';
 
 /**
- * Removes duplicate separators from a path, given a separator for search.
+ * Normalize separators of a given path.
  *
  * @param {string} path
  * @param {string} [separator='/']
@@ -46,7 +46,7 @@ function normalizeSeparators(path: string, separator = '/') {
 }
 
 /**
- * Removes duplicated separators from a path based on platform.
+ * Normalizes separatos of a given path based on the current platform.
  *
  * @export
  * @param {string} path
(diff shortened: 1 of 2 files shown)


matheustavaresdev:
As stated on the description, now the paths can have duplicated slashes. I don't think this could be an issue at all... I have done quite some tests and it always seem to return the correct path. Unless there are some case scenarios that I forgot to test, I think this should work as intended. Please let me know your thoughts on this cc @kevinushey @gtritchie @timtmok

src/node/desktop/src/core/file-path.ts
```diff
@@ -35,18 +35,18 @@ const homePathAlias = '~/';
...
- * Removes duplicate separators from a path, given a separator for search.
+ * Normalize separators of a given path.
  *
  * @param {string} path
  * @param {string} [separator='/']
  * @return {*} 
  */
 function normalizeSeparators(path: string, separator = '/') {
-  return path.replace(/[\\/]+/g, separator);
+  return path.replace(/[\\]/g, separator);
```
kevinushey, line 45: What happens if you run `normalizeSeparators("a/b/c", "\\")`?

src/node/desktop/src/core/file-path.ts
```diff
@@ -35,18 +35,18 @@ const homePathAlias = '~/';
...
 }
 
 /**
- * Removes duplicated separators from a path based on platform.
+ * Normalizes separatos of a given path based on the current platform.
```
kevinushey, line 49:
```suggestion
 * Normalizes separators of a given path based on the current platform.
```

matheustavaresdev: Fixed on [c643326](https://github.com/rstudio/rstudio/pull/11311/commits/c643326cb43ad849e0eaf784c7897ad26fe0f1d5)

src/node/desktop/src/core/file-path.ts
matheustavaresdev: Fixed on [c61c6d5](https://github.com/rstudio/rstudio/pull/11311/commits/c61c6d56ed0b2be7d9227cbb91f664c6502e4452)

PR #11314, opened 2022-05-27
Followup/rtools42 installer

jgutman:
### Intent

Followup to fix #11248 which didn't work correctly

### Approach

1) increase timeout to actually download the exe, because it's now 460+ MB which is really large (is 300 seconds long enough? the download.file help page says "(It is unrealistic to require download times of less than 1s/MB.)")
2) move the whole download/parse logic into an R function 
3) update the logic so that rtools40 is used for R versions 4.0 to 4.1.99

### Automated Tests

None

### QA Notes

See issue

### Checklist


src/cpp/session/modules/build/SessionBuild.cpp
```diff
@@ -2343,6 +2343,7 @@ Error initialize()
       (bind(registerRpcMethod, "devtools_load_all_path", devtoolsLoadAllPath))
       (bind(registerRpcMethod, "get_bookdown_formats", getBookdownFormats))
       (bind(sourceModuleRFile, "SessionBuild.R"))
+      (bind(sourceModuleRFile, "SessionInstallRtools.R"))
```
kevinushey, line 2346: Does this need to be "build/SessionInstallRtools.R"? (Note that `SessionBuild.R` is actually located in the modules directory, not the modules/build directory -- I'm not sure if that's an oversight, though)

src/cpp/session/modules/build/SessionInstallRtools.R
```diff
@@ -0,0 +1,36 @@
...
+# AGPL (http://www.gnu.org/licenses/agpl-3.0.txt) for more details.
+#
+#
+
+.rs.addFunction("findRtools42Installer", function(url, fallbackUrl)
+{
+   tryCatch({
+       tmp <- tempfile()
+       utils::download.file(url, tmp, mode = "w", quiet = TRUE)
+       homePageHtml <- paste(readLines(tmp), collapse=" ")
```
kevinushey, line 21:
```suggestion
       homePageHtml <- paste(readLines(tmp), collapse = " ")
```

src/cpp/session/modules/build/SessionBuild.cpp
jgutman: I tried it different ways, but this way works. I was confused because I noticed that about SessionBuild.R code path. However, since I can find `.rs.findRtools42Installer()` in the R session it must be sourcing the file correctly

jgutman: (I don't really get how these paths work, but trial and error led me to this solution)

src/cpp/session/modules/build/SessionInstallRtools.cpp
```diff
@@ -119,20 +97,40 @@ Error installRtools()
...
 
    // form path to destination file
    std::string rtoolsBinary = url.substr(url.find_last_of('/') + 1);
    FilePath installerPath = tempPath.completeChildPath(rtoolsBinary);
    std::string destfile = string_utils::utf8ToSystem(installerPath.getAbsolutePath());
 
+   // The Rtools installer can be a large file, so we want to increase the timeout option if less than 5 min
+   // and respect the user's original configured timeout settings
+   int originalTimeoutOption = r::options::getOption<int>("timeout");
+   error = r::options::setOption<int>("timeout", std::max(300, originalTimeoutOption));
```
kevinushey, line 120: I think we should just use a very large timeout here (presumedly the user would cancel the download manually if no progress were being made)

```suggestion
   error = r::options::setOption<int>("timeout", std::max(3600, originalTimeoutOption));
```

src/cpp/session/modules/build/SessionInstallRtools.cpp
```diff
@@ -40,36 +41,13 @@ namespace build {
...
 Error installRtools()
 {
    // populate list of known Rtools installers
    bool gcc49 = module_context::usingMingwGcc49();
    FilePath installPath("C:\\Rtools");
    std::vector<r_util::RToolsInfo> availableRtools;
+   availableRtools.push_back(r_util::RToolsInfo("4.2", installPath, gcc49));
    availableRtools.push_back(r_util::RToolsInfo("4.0", installPath, gcc49));
    availableRtools.push_back(r_util::RToolsInfo("3.5", installPath, gcc49));
    availableRtools.push_back(r_util::RToolsInfo("3.4", installPath, gcc49));
```
jgutman, line 53: why are these all gcc49? AFAIK rtools 40 and 42 uses gcc 10, but I don't think this even matters

kevinushey (commented):
LGTM; I'd just recommend increasing the timeout even more.

src/cpp/session/modules/build/SessionBuild.cpp
kevinushey: It's possible that sub-directories aren't respected when module files are copied to the build tree, so perhaps this is fine. If it works, then we can leave it :-)

src/cpp/session/modules/build/SessionInstallRtools.cpp
kevinushey: IIRC this is mostly a historical artifact; there was a transition period for older version of Rtools to using gcc 4.9.3 that required a bunch of IDE changes (which IIUC are mostly no longer relevant).

src/cpp/session/modules/build/SessionBuild.cpp
jgutman: Yep, in the build directory, all of these files are under `session/modules/R`

src/cpp/session/modules/build/SessionInstallRtools.cpp
jgutman: Agreed, seems reasonable

src/cpp/session/modules/build/SessionInstallRtools.cpp
```diff
@@ -119,20 +97,40 @@ Error installRtools()
...
+        url = kFallbackUrl;
+     }
+   }
 
    // form path to destination file
    std::string rtoolsBinary = url.substr(url.find_last_of('/') + 1);
    FilePath installerPath = tempPath.completeChildPath(rtoolsBinary);
    std::string destfile = string_utils::utf8ToSystem(installerPath.getAbsolutePath());
 
+   // The Rtools installer can be a large file, so we want to increase the timeout option if less than 5 min
```
jgutman, line 117:
```suggestion
   // The Rtools installer can be a large file, so we want to increase the timeout option
```

src/cpp/session/modules/build/SessionInstallRtools.cpp
```diff
@@ -119,20 +97,40 @@ Error installRtools()
...
+   // and respect the user's original configured timeout settings
+   int originalTimeoutOption = r::options::getOption<int>("timeout");
+   error = r::options::setOption<int>("timeout", std::max(300, originalTimeoutOption));
+
+   if (error)
+       module_context::consoleWriteOutput("To avoid timeouts when downloading the installer, set `options(timeout=300).\n`");
```
jgutman, line 123:
```suggestion
       module_context::consoleWriteOutput("To avoid timeouts when downloading the installer, set `options(timeout=3600).\n`");
```

src/cpp/session/modules/build/SessionInstallRtools.cpp
```diff
@@ -119,20 +97,40 @@ Error installRtools()
...
    FilePath installerPath = tempPath.completeChildPath(rtoolsBinary);
    std::string destfile = string_utils::utf8ToSystem(installerPath.getAbsolutePath());
 
+   // The Rtools installer can be a large file, so we want to increase the timeout option
+   // and respect the user's original configured timeout settings
+   int originalTimeoutOption = r::options::getOption<int>("timeout");
+   error = r::options::setOption<int>("timeout", std::max(3600, originalTimeoutOption));
+
+   if (error)
+       module_context::consoleWriteOutput("To avoid timeouts when downloading the installer, set `options(timeout=3600).\n`");
```
jgutman, line 123:
```suggestion
       module_context::consoleWriteOutput("To avoid timeouts when downloading the installer, set `options(timeout=3600)`.\n");
```

kevinushey (approved):

PR #11326, opened 2022-05-31
close editors when satellite window closed (Electron)

gtritchie:
### Intent

Addresses #10906

On Electron, when we close the satellite window the `unload` event isn't received, and we were relying on this to close the editor(s) in a satellite window when it closed. MDN does say the unload event is unreliable and recommends against using it and that appears to be a valid warning for Electron.

### Approach

On Electron, directly invoke `notifyRStudioSatelliteClosed()` instead of doing it in response to the unload event.

### Automated Tests

None

### QA Notes

In addition to the basic scenario outlined in the issue, also spot-check that that if you try to close a window containing modified (unsaved) buffers, you get prompted to save, etc.

### Checklist

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/common/satellite/Satellite.java
+++ b/src/gwt/src/org/rstudio/studio/client/common/satellite/Satellite.java
@@ -199,6 +199,9 @@ public class Satellite implements HasCloseHandlers<Satellite>
       $wnd.opener.registerAsRStudioSatellite(name, $wnd);
    }-*/;
 
+   public native void notifyRStudioSatelliteClosed() /*-{
+      $wnd.opener.notifyRStudioSatelliteClosed($wnd.RStudioSatelliteName);
+   }-*/;
 
    // check whether the current window is a satellite
    public static native boolean isCurrentWindowSatellite() /*-{
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/source/SourceWindow.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/SourceWindow.java
@@ -18,6 +18,7 @@ import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.Set;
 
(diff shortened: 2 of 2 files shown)


kevinushey:
Did you also try an approach using "beforeunload" in addition to (or instead of) the "unload" event?

I mainly ask because if "unload" isn't reliable, we might need to check other locations where it's used, e.g.

https://github.com/rstudio/rstudio/blob/bb265d0dbedece7ced0325526ba14cb13684cb2d/src/gwt/src/org/rstudio/studio/client/shiny/ShinyApplicationPresenter.java#L132-L152

https://github.com/rstudio/rstudio/blob/bb265d0dbedece7ced0325526ba14cb13684cb2d/src/gwt/src/org/rstudio/studio/client/pdfviewer/model/PdfJsWindow.java#L130-L134

https://github.com/rstudio/rstudio/blob/bb265d0dbedece7ced0325526ba14cb13684cb2d/src/gwt/src/org/rstudio/studio/client/workbench/views/help/HelpPane.java#L270-L275

At least, these code paths are worth testing / auditing for Electron; so maybe we can just leave a note to test those code paths later on.

kevinushey (approved):
LGTM but I think it's worth auditing our other usages of "unload" "beforeunload" in other parts of the code base for Electron.

gtritchie:
> Did you also try an approach using "beforeunload" in addition to (or instead of) the "unload" event?

I looked at it a bit, but seemed too large a potential change given we already use the Electron `close` event to (optionally) stop the close (e.g. if you decide to cancel due to unsaved files, etc.), and the docs talk of subtle differences between the flow involving 'close' versus the DOM beforeunload approach.

Agree we should audit the other cases, I'll open an issue to track that.

Also agree there's probably/hopefully a more elegent solution here involving usage of beforeunload instead of unload, but for now didn't want to crack that open and risk regressing non-Electron.

PR #11342, opened 2022-06-02
Fix various typos in `src/cpp/**`

luzpaz:
### Intent

Fixes typos found within the source code which includes comments and documentation. Follow-up to #10868  

### Approach

Found using `codespell -q 3 -S *_fr.properties -L ba,enque,optionel,ot,parm,parms,pres,prevend,pevent,pevents,ptd,texline`
 but then combing through results pruning upstream 3rd party deps etc...

### Automated Tests

This PR fixes typos. No unit tests needed.

### QA Notes

N/A

### Checklist

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -767,7 +767,7 @@ OpenSSL License
  * The implementation was written so as to conform with Netscapes SSL.
  *
  * This library is free for commercial and non-commercial use as long as
- * the following conditions are aheared to.  The following conditions
+ * the following conditions are adheared to.  The following conditions
  * apply to all code found in this distribution, be it the RC4, RSA,
  * lhash, DES, etc., code; not just the SSL code.  The SSL documentation
  * included with this distribution is covered by the same copyright terms
@@ -792,7 +792,7 @@ OpenSSL License
  *    must display the following acknowledgement:
  *    "This product includes cryptographic software written by
  *     Eric Young (eay@cryptsoft.com)"
- *    The word 'cryptographic' can be left out if the rouines from the library
+ *    The word 'cryptographic' can be left out if the routines from the library
  *    being used are not cryptographic related :-).
  * 4. If you include any Windows specific code (or a derivative thereof) from
  *    the apps directory (application code) you must include an acknowledgement:
@@ -810,7 +810,7 @@ OpenSSL License
  * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
  * SUCH DAMAGE.
  *
- * The licence and distribution terms for any publically available version or
+ * The licence and distribution terms for any publicly available version or
  * derivative of this code cannot be changed.  i.e. this code cannot simply be
  * copied and put under another distribution licence
  * [including the GNU Public Licence.]
--- a/NOTICE
+++ b/NOTICE
@@ -767,7 +767,7 @@ OpenSSL License
  * The implementation was written so as to conform with Netscapes SSL.
  *
  * This library is free for commercial and non-commercial use as long as
- * the following conditions are adheared to.  The following conditions
+ * the following conditions are aheared to.  The following conditions
  * apply to all code found in this distribution, be it the RC4, RSA,
  * lhash, DES, etc., code; not just the SSL code.  The SSL documentation
  * included with this distribution is covered by the same copyright terms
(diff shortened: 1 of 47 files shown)


src/cpp/core/include/core/r_util/RActiveSessionStorage.hpp
```diff
@@ -50,7 +50,7 @@ namespace r_util {
 
    private:
       FilePath scratchPath_;
-      const std::string propertiesDirName_ = "properites";
+      const std::string propertiesDirName_ = "properties";
```
kevinushey, line 53: We need to preserve the incorrect name here as the folder is indeed (unfortunately) called "properites".

NOTICE
```diff
@@ -767,7 +767,7 @@ OpenSSL License
  * The implementation was written so as to conform with Netscapes SSL.
  *
  * This library is free for commercial and non-commercial use as long as
- * the following conditions are aheared to.  The following conditions
+ * the following conditions are adheared to.  The following conditions
```
kevinushey, line 770: We should not be changing license text; the misspellings here are part of the original document. See https://docs.oracle.com/en/enterprise-manager/cloud-control/enterprise-manager-cloud-control/13.5/oemli/original-ssleay-license.html for another example.

src/cpp/core/include/core/r_util/RActiveSessions.hpp
```diff
@@ -84,7 +84,7 @@ class ActiveSession : boost::noncopyable
       if (error)
          LOG_ERROR(error);
 
-      propertiesPath_ = scratchPath_.completeChildPath("properites");
+      propertiesPath_ = scratchPath_.completeChildPath("properties");
```
kevinushey, line 87: Same here.

src/cpp/core/include/core/rapidxml/rapidxml.hpp
```diff
@@ -1327,7 +1327,7 @@ namespace rapidxml
     
         // Note that some of the pointers below have UNDEFINED values if certain other pointers are 0.
         // This is required for maximum performance, as it allows the parser to omit initialization of 
-        // unneded/redundant values.
+        // unneeded/redundant values.
```
kevinushey, line 1330: I'd recommend making these changes upstream; this is a 3rd-party component.

src/cpp/core/zlib/CMakeLists.txt
```diff
@@ -26,7 +26,7 @@ include(CheckFunctionExists)
 include(CheckIncludeFile)
 include(CheckCSourceCompiles)
 
-# disable gcc visiblity attribute (not supported by some mingw versinos)
+# disable gcc visibility attribute (not supported by some mingw versinos)
```
kevinushey, line 29:
```suggestion
# disable gcc visibility attribute (not supported by some mingw versions)
```

kevinushey (changes requested):

src/cpp/core/include/core/rapidxml/rapidxml.hpp
luzpaz: Reverted in 52dc3e8

src/cpp/core/include/core/r_util/RActiveSessions.hpp
luzpaz: Reverted in 52dc3e8

NOTICE
luzpaz: Reverted in 52dc3e8

src/cpp/core/include/core/r_util/RActiveSessionStorage.hpp
luzpaz: Reverted in 52dc3e8

luzpaz:
Please review changes

src/cpp/server/ServerPAMAuthOverlay.hpp
```diff
@@ -23,7 +23,7 @@ namespace server {
 namespace pam_auth {
 namespace overlay {
 
-void intialize();
+void initialize();
```
kevinushey, line 26: This change is a little surprising / concerning; the name here is mis-spelled, but also does not match the name here:

https://github.com/rstudio/rstudio/blob/b668ec9ab8d0c2a98c1c1d5c6e85dd541184a7a1/src/cpp/server/ServerPAMAuthOverlay.cpp#L41-L44

This would seem to imply we're not actually calling this method as expected?

kevinushey (approved):
LGTM, but I think we need to file an issue for the misspelled initialize method.

src/cpp/server/ServerPAMAuthOverlay.hpp
luzpaz: @kevinushey indeed. Is the plan to open a separate ticket for this?

kevinushey: I've done so on our Pro repository.

Issue #3094, opened 2018-06-29
Edit -> Replace and Find fails to open when there is an existing multiline text selection


kevinushey:
### System details

    RStudio Edition : Desktop
    RStudio Version : v1.2.753
    OS Version      : Ubuntu 16.04
    R Version       : 3.4.4

### Steps to reproduce the problem

Given a document with contents:

```r
abc
def
```

Try highlighting everything with <kbd>Ctrl + A</kbd>, and then attempt to open the Find and Replace dialog with <kbd>Ctrl + Shift + J</kbd>.

### Describe the problem in detail

The Find and Replace dialog fails to open.

### Describe the behavior you expected

We should do something, at least. I don't think our Find and Replace helper supports replacements across multiple lines, but we should probably still just open the dialog with an empty selection?

kevinushey:
As an aside, the name `Replace and Find` feels a bit strange, given the order of operations is that first you find, then you replace. I wonder if we should flip that around?

ronblum:
Reproduced on MacOS 10.12.6. Looks like it happens if more than one line is selected—it doesn't have to be the whole document.

ronblum:
@KBGrammer Interesting. Can you provide a reproducible example of it working? I've it on Ubuntu 18.04 Desktop and Ubuntu 16.04.4 Server, but no luck.

KBGrammer:
I'm using Version 1.2.747 (using it because the current release version has broken middle click paste, but the preview works) on Ubuntu 16.04. All of this holds true for a previous version of 1.1.442. I believe it also holds true for 1.1.453, but I only had that installed for 10 minutes before switching to the preview version.

Steps that fail outright:
1. Load a generic script:
```
print("this")
print("that")
print("the other thing")
```
2. Highlight the last 2 lines.
3. Press `Ctrl+Shift+J`. (nothing happens)

Steps that bring up the menu:
1.  Load a generic script:
```
print("this")
print("that")
print("the other thing")
```
2. Press `Ctrl+Shift+J`. (a search text box, buttons saying Next, Prev, All, etc. shows up at the top of the screen)

Here's where things diverge. The following steps allow me to search and replace within the selection (Sequence A)
3. Highlight the last two lines.
4. Check "In selection"
5. Begin typing 't' in the "Find" box, and it highlights the 't' in print on line 2 (in blue). From here, I could replace 't's individually or all at once. The selection remains highlighted (in gray).

The following steps are not in the proper sequence. (Sequence B) 
[...]

PR #11344, opened 2022-06-02
Shuffle code from rstudio-pro to rstudio for adding users to the DB

MariaSemple:
### Intent

Part of [Pro #3094](https://github.com/rstudio/rstudio-pro/issues/3094): add users to the database in Open Source to support foreign key relationship.

NOTE: please do not merge, this requires a Pro-side companinion

### Approach

Shuffle code out of Pro and into OS

### Automated Tests

We should consider adding a test to check if the user entry is in the database after login, but it's not done yet.

### QA Notes

Once you've logged in with a user, attach to the database and check via `SELECT * FROM licensed_users;`. If you log out and log back in, you should also see the last sign-in time column get updated.

For open source, users should never be locked and never be admins. If the values in the database are updated externally, that should still have no impact on the behaviour (i.e. they're ignored).

### Checklist


src/cpp/server/auth/ServerAuthHandler.cpp
```diff
@@ -358,25 +412,397 @@ bool isCookieRevoked(const std::string& cookie)
...
+      if (!*pLicensed)
+      {
+         LOG_WARNING_MESSAGE("User '" + username + "' is locked or there is no license available");
+      }
+      // named user licensing is in place, so forward the result
+      // of the licensing check
+      return error;
+   }
+
+   return isUserLicensed(user, false, pLicensed);
```
jeffvroom, line 651: We just made this call above and I don't think these statements can be reached:
```suggestion
```

src/cpp/server/auth/ServerAuthHandler.cpp
```diff
@@ -358,25 +412,397 @@ bool isCookieRevoked(const std::string& cookie)
...
-   return true;
+   return (boost::posix_time::microsec_clock::universal_time() - lastSignin <
+           boost::posix_time::time_duration(kHoursInOneYear, 0, 0));
+}
+
+Error addUser(boost::asio::io_service& ioService,
+              const std::string& username,
+              bool isAdmin)
+{
+   boost::shared_ptr<IConnection> connection;
```
jeffvroom, line 486: If we are using the file system storage, should we maybe continue to not use the database?  I'm concerned that for OS users they are going to need to start administering the sqlite db, or at least needing to make sure it's operating properly even though they are getting no benefit. It might cause us less churn if this feature was tied to the DB for session metadata.

MariaSemple: There's already a requirement in OS that users have access to the database for revoked cookies, so I don't think that's a concern

jeffvroom (approved):
LGTM!

PR #11346, opened 2022-06-02
Don't create user-d.ctx file unless project sharing is enabled

jeffvroom:
### Intent

Currently desktop users are getting files username-d.ctx created due to this file being added for project sharing when there's no sessionScope(). 

Requires a companion commit to implement supportsProjectSharing() that will look like:

```
bool Options::supportsProjectSharing() const
{
   return projectSharingEnabled() && programMode() == kSessionProgramModeServer;
}```

### Approach

Although we have options().projectSharingEnabled() it's used all over
the place and is true even for desktop/server so I'm reluctant to change it
to return false for OS.

Instead I'm adding a new supportProjectSharing() that will only be true
when projectSharingEnabled() is true and the environment supports sharing.

### Automated Tests

None required.  

### QA Notes

Just need to verify that the username-d.ctx file is only created in /var/run/rstudio-server/rstudio-rsession/ if project-sharing-enabled=1 and server-multiple-sessions=0.  It should not be created when using the job launcher, server/OS, or when server-multiple-sessions=1.

### DIFF
--- a/src/cpp/session/SessionOptionsOverlay.cpp
+++ b/src/cpp/session/SessionOptionsOverlay.cpp
@@ -44,5 +44,10 @@ bool Options::supportsDriversLicensing() const
    return !allowOverlay();
 }
 
+bool Options::supportsProjectSharing() const
+{
+   return false;
+}
+
 } // namespace session
 } // namespace rstudio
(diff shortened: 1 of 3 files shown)


gtritchie (approved):
LGTM!

Issue #10920, opened 2022-04-06
Consider routinely updating version of Quarto bundled with RStudio


ronblum:
### System details

    RStudio Edition : Desktop and Server
    RStudio Version : 2022.06.0-daily+270
    OS Version      : MacOS 12.3, Red Hat 8.3, Ubuntu 20.04
    R Version       : 4.1.3

### Steps to reproduce the problem

Run `quarto --version` bundled with RStudio:
- RStudio Desktop on MacOS: `/Applications/RStudio.app/Contents/MacOS/quarto/bin/quarto --version`
- RStudio Desktop on Red Hat: `/usr/lib/rstudio/resources/app/bin/quarto/bin/quarto --version`
- RStudio Server on Ubuntu: `/Applications/RStudio-2022.06.0-daily-272-OG.app/Contents/MacOS/quarto/bin/quarto --version`

### Describe the problem in detail

The bundled version is 0.9.80.

### Describe the behavior you expected

The most recent version at https://quarto.org as of April 6, 2022 is 0.9.194. Version 0.9.80 is from March 10. We should consider routinely updating to the latest version. 

- [ ] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ ] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [ ] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
Triage: Also consider as `backport` to Prairie Trillium.

mikebessuille:
I don't think we need to backport this; PT users of quarto are using a prerelease thing; they should update it themselves.  

JJ is going to work with Tim so that our dailies include the latest build of quarto starting shortly.

timtmok:
I got some direction from JJ. The dependency can be updated to `latest` (https://github.com/quarto-dev/quarto-cli/releases/latest) which will point to the tagged release. GitHub has REST API to query the releases (https://api.github.com/repos/quarto-dev/quarto-cli/releases).

This commit is an example of what a Quarto update would change: https://github.com/rstudio/rstudio/commit/acac28db27d2d172834036d9f47d85792b2e0b04

[...]

PR #11354, opened 2022-06-03
Force install quarto from latest

timtmok:
### Intent
Address #10920  for linux platforms

### Approach
Cache bust the docker layer by checking the Quarto release. The layer never changes because the install script doesn't change. This queries the GH API to get the Quarto release list and adds it as a file to the docker image. If the file contents has changed then there has been a new Quarto release. The file changing invalidates future layers and will run the Quarto install script.

This will likely be temporary until the final Quarto release. Then it will go back to the old way with specifying a specific version. At that time, it's probably best to add an argument to the Dockerfile and specify the Quarto version there.

### Automated Tests
None

### QA Notes
Tested locally with a test Dockerfile. The `ADD` is cached if the release list hasn't changed. When it changes, the subsequent layers are rebuilt. 

### Checklist


kevinushey (approved):
LGTM!

PR #11356, opened 2022-06-03
Don't deref storage_ for empty sessions since it's not set

jeffvroom:
Fixes crash on Desktop and rworkspaces - both of which attempt to
get properties from an empty session, expecting to return nothing back.

I don't think the empty() test in validate() is strictly necessary as the only places we call it we already have called empty() but maybe this will save us from another confusing crash in the future :)


MariaSemple (approved):
LGTM!

Issue #10632, opened 2022-02-17
Security vulnerability placeholder - see internal #3292


byteit101:
On November 21st, 2021, I reported a potential security issue to security@rstudio.com. I sent two follow up emails, but have not received any response. If I do not receive a response by Monday, February 21st, 2022, I will post the details publicly, in this issue below. 

### System details

    RStudio Edition : 
    RStudio Version : 
    OS Version      : 
    R Version       : 

### Describe the behavior you expected

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ ] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [ ] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
@byteit101 Thank you for raising the issue—both the original report, and letting us know about the issue here. I'm following up internally to understand the situation and follow up on it. In the meantime, I and the rest of the team would appreciate if you held off on publishing the problem—I can assure you that we're investigating this!

ronblum:
@byteit101 Note the above reference—I filed your security concerns in a private repository for investigation. Thank your findings and report!

mikebessuille:
Leaving this issue open for now just as a placeholder until we communicate with the original poster privately.  We have a private issue tracking this problem and are reaching out by email regarding the severity and attack vectors.

mikebessuille:
Still looking into this for Spotted Wakerobin (2022.06)...

timtmok:
This was fixed with the Spotted Wakerobin release.

byteit101:
As this has been patched, here is the original report below for future public reference:

------------------------

Security: Non-BMP Unicode Characters cause multiple issues in Rmd editors (Text & Visual)


### System details

    RStudio Edition : Desktop
    RStudio Version : 2021.09.0 Build 351
    OS Version      :  Linux Debian 10 x64
    R Version       :  3.6.3

### Steps to reproduce the problem

Open this RMarkdown file in RStudio:

    ---
    title: "Untitled"
    author: "Me"
    date: "11/21/2021"
    output:
      html_document: default
    ---
   
    ```{r setup, include=FALSE}
   
    sample(iris)
    xy="𝚫"
[...]

PR #11368, opened 2022-06-06
Handle non-BMP characters in RMD R chunks

timtmok:
### Intent
Address #10632 

### Approach
Javascript doesn't have built-in support for characters outside of the BMP. A quick and dirty workaround is converting the string to a char array to get the line length. This counts positions as is expected when executing the selected lines of code. This handles emojis and even other non-BMP characters that interact differently in the ACE editor. 😎 is selectable as one character, counted as two in ACE, and should be counted as one for execution. ❤️ is selectable as two characters (first character is a heart and the second is a modifier making it red) and is also counted as two in ACE.

Converting to a char array handles these cases as it separates 😎 as one element and ❤️ as two elements, which is correct for the position count for the code range.

### Automated Tests
None

### QA Notes
See rstudio-pro ticket 3292

### Checklist

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/rmd/NotebookQueueState.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/rmd/NotebookQueueState.java
@@ -455,6 +455,16 @@ public class NotebookQueueState implements NotebookRangeExecutedEvent.Handler,
       
       return unit;
    }
+
+   /**
+    * Calculate string length based on Unicode code points. This is a low-effort workaround
+    * to calculating length on strings that include non-BMP characters like emojis.
+    * @param text input string
+    * @return the string length based on atomic parts of text
+    */
+   private native int codePointLength(String text) /*-{
+      return Array.from(text).length;
+   }-*/;
    
    private NotebookExecRange getNotebookExecRange(Scope scope, Range range)
    {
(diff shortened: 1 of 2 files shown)


src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/rmd/NotebookQueueState.java
```diff
@@ -470,13 +480,14 @@ private NotebookExecRange getNotebookExecRange(Scope scope, Range range)
...
-         for (int col = 0; col <= line.length(); col++)
+         int lineLength = codePointLength(line);
+         for (int col = 0; col <= lineLength; col++)
          {
             if (startPos.getRow() == row && startPos.getColumn() == col)
             {
                start = pos;
             }
-            else if (endPos.getRow() == row && endPos.getColumn() == col)
+            else if (endPos.getRow() == row && (endPos.getColumn() == col || col == lineLength))
```
kevinushey, line 490: Why is the `col == lineLength` check needed? (Maybe add a comment explaining?)

kevinushey:
Does this also work for characters like 👩‍❤️‍💋‍👩? That is, for strings containing zero-width joiners? (I don't think Ace will render it correctly in the document, but it should still render correctly after printing the string)

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/rmd/NotebookQueueState.java
timtmok: I can add a comment here. It's to correctly count the end position since the `endPos` may be greater than the char array length.

PR #11401, opened 2022-06-09
Fix build time quarto version check

timtmok:
### Intent
Too many builds run concurrently for all the platforms which causes the build agents to exceed the GitHub API limit. This fixes the issue in case too many builds start within an hour.

### Approach
Uses the download metadata from quarto.org instead of GitHub to check which release to use for the build. This changes the Dockerfiles for Linux and the installer scripts for Linux/Mac/Windows. Mac and Windows always run the dependency install.

### Automated Tests
None

### QA Notes
None 

### Checklist

### DIFF
--- a/dependencies/common/install-quarto
+++ b/dependencies/common/install-quarto
@@ -27,7 +27,7 @@ fi
 
 # variables that control download + installation process
 # update to latest Quarto release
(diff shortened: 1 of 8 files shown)


gtritchie (approved):
LGTM!

PR #11405, opened 2022-06-09
Fix bug with single-session mode and project sharing

jeffvroom:
### Intent

Fixes a bug the cloud team found with `server-multiple-sessions=0` and project sharing (a feature just added in this release)

### Approach

* Need to create the project id with the path chosen above as the active session's project (session.project() is not set until later)

* Need to use createAliasedPath() so that ~/ is managed consistently
or you can have two different project ids for the same path


melissa-barca (approved):
LGTM!

PR #11414, opened 2022-06-10
Handle special regex characters as escaped literals in Whole word mode

jgutman:
Addresses #11435

### DIFF
--- a/src/cpp/session/modules/SessionFind.cpp
+++ b/src/cpp/session/modules/SessionFind.cpp
@@ -1571,11 +1571,21 @@ core::Error runGrepOperation(const GrepOptions& grepOptions, const ReplaceOption
    return Success();
 }
 
+std::string escapeCharacters(const std::string& query)
+{
+   std::string specialChars("([.|*?+(){}^$\\[\\]])");
+   // Note that we need four backslashes in order to add a backslash before each special character to turn into a literal
+   // \\1 is a backreference that refers to the original symbol we want to escape
+   boost::regex_constants::syntax_option_type flags = 0;
+   std::string updatedQuery = boost::regex_replace(query, boost::regex(specialChars, flags), "\\\\\\1");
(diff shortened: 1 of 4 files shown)


src/cpp/session/modules/SessionFind.cpp
```diff
@@ -1569,18 +1569,29 @@ core::Error runGrepOperation(const GrepOptions& grepOptions, const ReplaceOption
    return Success();
 }
 
+std::string escapeCharacters(const std::string& query)
```
kevinushey, line 1572: I think it's worth documenting what the escaping is being done for here (either in the function name, or as a comment), since the way escaping is performed depends on what is trying to interpret the characters / escapes. Since (I think) the intention here is to escape a regular expression pattern, you might want to call the function `escapeForRegex()` or something like that?

src/cpp/session/modules/SessionFind.cpp
```diff
@@ -1569,18 +1569,29 @@ core::Error runGrepOperation(const GrepOptions& grepOptions, const ReplaceOption
...
+{
+   std::string specialChars("([.|*?+(){}^$\\[\\]])");
```
kevinushey, line 1574: IIRC we also considered whether we could use the PCRE escapes `\Q` and `\E` here (https://www.pcre.org/original/doc/html/pcresyntax.html#SEC2) but that didn't work out. Is that right?

kevinushey (approved):
LGTM! Only small suggestion is tweaking a function name for clarity.

src/cpp/session/modules/SessionFind.cpp
jgutman: Yeah, I believe that's right -- I can double check but that's my recall as well. I haven't gotten back to look at this since we pulled it out of SW so will test this out again before merging

PR #11417, opened 2022-06-10
Provide connection address/port, uri, stream file in connection error

jeffvroom:
### Intent

Fixes https://github.com/rstudio/rstudio-pro/issues/3454

Also hides "ssl short read" errors by default, that I've seen a lot in customer logs that are using SSL servers. Those happen when the browser closes the connection mid-stream, as happens pretty regularly in our streaming connections. 

### Approach

Our errors get handled in the abstract base class so I added a method in each subclass to modify the error with details. 

### Automated Tests

Not necessary

### QA Notes

Aside from our existing automation, I don't think this needs additional testing but you can easily see these errors by mis-configuring one of the many ports or addresses in our configuration. For local streams, you should see the stream-path that helps identify the server. For Tcp, you see the address/port of the server in the connection. For both, you see the uri that client is using.

### DIFF
--- a/src/cpp/core/include/core/http/AsyncClient.hpp
+++ b/src/cpp/core/include/core/http/AsyncClient.hpp
@@ -297,6 +297,9 @@ protected:
 
    void handleError(const Error& error)
    {
+      Error httpError = error;
+      addErrorProperties(httpError);
+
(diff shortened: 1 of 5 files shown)


MariaSemple (approved):
LGTM!

PR #11419, opened 2022-06-13
Add 'starting' debug log right after log is initialized

jeffvroom:
### Intent

Addresses the remaining aspect of: https://github.com/rstudio/rstudio-pro/issues/2641

Should I log PATH for windows?

### Approach

We already have debug logs beginning the R session, before and after resuming the session, and when
listening. This adds a message right when the session starts (after logging has been initialized), and prints the dyn library path value for linux and mac.

### Automated Tests

None required

### QA Notes

Enable debug logging and hopefully get improved timing info on session start.

### DIFF
--- a/src/cpp/session/SessionMain.cpp
+++ b/src/cpp/session/SessionMain.cpp
@@ -1853,6 +1853,32 @@ void afterResume()
    LOG_DEBUG_MESSAGE("Resume complete");
 }
 
+void logStartingEnv()
+{
+#ifdef __linux__
+   LOG_DEBUG_MESSAGE("Starting R session with LD_LIBRARY_PATH: " + core::system::getenv("LD_LIBRARY_PATH"));
+#endif
+#ifdef __APPLE__
+   std::string envVars[] = {"LD_LIBRARY_PATH", "DYLD_LIBRARY_PATH", "DYLD_FALLBACK_LIBRARY_PATH"};
+   bool any = false;
+   for (std::string varName:envVars)
+   {
+      std::string envVal = core::system::getenv(varName);
+      if (!envVal.empty())
+      {
+         LOG_DEBUG_MESSAGE("Starting R session with " + varName + ": " + envVal);
+         any = true;
+         break;
+      }
+   }
+   if (!any)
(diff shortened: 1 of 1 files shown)


src/cpp/session/SessionMain.cpp
```diff
@@ -1853,6 +1853,32 @@ void afterResume()
    LOG_DEBUG_MESSAGE("Resume complete");
 }
 
+void logStartingEnv()
+{
+#ifdef __linux__
+   LOG_DEBUG_MESSAGE("Starting R session with LD_LIBRARY_PATH: " + core::system::getenv("LD_LIBRARY_PATH"));
+#endif
+#ifdef __APPLE__
+   std::string envVars[] = {"LD_LIBRARY_PATH", "DYLD_LIBRARY_PATH", "DYLD_FALLBACK_LIBRARY_PATH"};
```
kevinushey, line 1862: Some thoughts:

1. Do we need to log `LD_LIBRARY_PATH`? As far as I know, that's ignored on macOS.
2. Should we just log _all_ of these environment variables? They effect the dynamic loader in different ways, and it's possible to have multiple values set.

We might also want to log `DYLD_INSERT_LIBRARIES` as well. On macOS desktop, we use this to ensure the right version of R is loaded by the R session executable; logging this could be useful to check that we're loading the version of R we expect to.

src/cpp/session/SessionMain.cpp
```diff
@@ -1853,6 +1853,32 @@ void afterResume()
...
+         LOG_DEBUG_MESSAGE("Starting R session with " + varName + ": " + envVal);
+         any = true;
+         break;
+      }
+   }
+   if (!any)
+      LOG_DEBUG_MESSAGE("Starting R session with default system library path");
+#endif
+#ifdef _WIN32
+   LOG_DEBUG_MESSAGE("Starting R session");
```
kevinushey, line 1878: I agree logging the path entries could be useful here, since Windows uses the PATH for DLL resolution. (https://docs.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order)

kevinushey (approved):
LGTM!

PR #11426, opened 2022-06-14
Fixed 2 compile errors and one warning on Mac OS build

jeffvroom:
- size_t != uint64_t (it seems?) so add some explicit cases
 - warning for comparing int against uint (remove the "max test" because a sized int can never be larger than umax int)
 - don't use readObject directly with size_t as it leads to an undefined method

### DIFF
--- a/src/cpp/r/RJson.cpp
+++ b/src/cpp/r/RJson.cpp
@@ -162,7 +162,7 @@ Error jsonValueFromVectorElement(SEXP vectorSEXP,
                }
                case UINT32:
                {
-                  if ((value >= 0) && (value <= UINT32_MAX))
+                  if (value >= 0)
                   {
                      *pValue = static_cast<uint32_t>(value);
                      break;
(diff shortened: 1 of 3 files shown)


src/cpp/r/RJson.cpp
```diff
@@ -162,7 +162,7 @@ Error jsonValueFromVectorElement(SEXP vectorSEXP,
                }
                case UINT32:
                {
-                  if ((value >= 0) && (value <= UINT32_MAX))
```
MariaSemple, line 165: Why remove this check?

jeffvroom: It's getting a compiler warning on the mac because int is not of the same precision as the uint32 it's being compared to.  From my reading of the code, a signed int is always less than uint32_max so it's not necessary right?  If R supported unsigned ints, we'd have a different better way to fix it!

MariaSemple (approved):
LGTM!

Issue #11348, opened 2022-06-02
Find in files: crash when searching for ()


kevinushey:
### System details

    RStudio Edition : 2022.6.0.468
    RStudio Version : Desktop [Open Source]
    OS Version      : macOS Monterey 12.4
    R Version       : R version 4.2.0 (2022-04-22)

### Steps to reproduce the problem

Open a project, and then try to perform a Find in Files search for the term `()`, with the Whole word option selected.

<img width="425" alt="Screen Shot 2022-06-02 at 3 42 35 PM" src="https://user-images.githubusercontent.com/1976582/171750148-0e93d3ed-88af-4cb0-9ae7-e441119841bd.png">

### Describe the problem in detail

RStudio crashes.

### Describe the behavior you expected

No crash; the search should be successful.

---

<details>

<summary>Session Info</summary>

``` r
R version 4.2.0 (2022-04-22)
Platform: aarch64-apple-darwin20 (64-bit)
Running under: macOS Monterey 12.4

Matrix products: default
LAPACK: /Library/Frameworks/R.framework/Versions/4.2-arm64/Resources/lib/libRlapack.dylib

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

loaded via a namespace (and not attached):
 [1] compiler_4.2.0  fastmap_1.1.0   cli_3.3.0       htmltools_0.5.2
 [5] tools_4.2.0     yaml_2.3.5      rmarkdown_2.14  knitr_1.39     
 [9] xfun_0.31       digest_0.6.29   rlang_1.0.2     evaluate_0.15  
```

</details>

jgutman:
> For what it's worth, it seems to work with the `-P` flag...
> 
> ```
> C:\rtools42\usr\bin>type test.txt
> hello()
> hello()there
> hello() I am Jackie
> 
> C:\rtools42\usr\bin>grep -E "\b\Q()\E\b" test.txt
> 
> C:\rtools42\usr\bin>grep -P "\b\Q()\E\b" test.txt
> hello()there
> ```
> 
> but I'm not sure if that's preferable to just escaping characters individually (or if we even feel comfortable changing from `-E ` to `-P`)
> 
[...]

PR #11430, opened 2022-06-14
Prevent crashes with malformed regex which match all lines of content

jgutman:
### Intent

Addresses #11348 part 1 -- prevent RStudio crashes

### Approach

So it turns out that system grep doesn't actually crash when you search for `()` or `[]` or other bad regexes, even though it doesn't really match any content. We give our Find in Files an escape hatch, where if the grep command has completed without error and returns actual output, but the matchOn/matchOff array is empty (this doesn't happen with true no results, where grep stdout is empty) we skip ahead to the next match. 

Essentially this gives us (No results found) when searching for regexes like `()`, but appropriate results when searching for `\(\)`. True no matches (ie searching for `shdjshfdsjgfsjkd`) continue to behave as usual, since the grep onStdOut callback is not triggered when grep finds no matches.

This is because grep doesn't error when you search for `()` -- it actually matches every line, every file, even blank lines.

Separately, we can put in a PR to allow for escaping these characters when in WholeWord mode

### Automated Tests

Existing unit tests pass. We can add a new automated test case in rstudio-ide-automation once this is closed

### QA Notes

Find in files  with regex selected:
Create a file
````
print()

hello()

[1-3]
````
1) `\(\)`
2) `()`
3) `nooooooo`
4) `[]`
5) `\[\]`

### Checklist

### DIFF
--- a/src/cpp/session/modules/SessionFind.cpp
+++ b/src/cpp/session/modules/SessionFind.cpp
@@ -1034,6 +1034,8 @@ private:
             processContents(&lineInfo.decodedPreview, &lineInfo.decodedContents,
                &matchOn, &matchOff);
 
+            // If we reach here, grep has found a malformed match (usually due to a bad regex)
+            // and processContents was not able to identify the corresponding match string
             if (matchOn.getSize() == 0)
                continue;
 
--- a/src/cpp/session/modules/SessionFind.cpp
+++ b/src/cpp/session/modules/SessionFind.cpp
@@ -665,6 +665,13 @@ private:
          decodedLine.append(
             Replacer::decode(std::string(inputPos, end), encoding_, firstDecodeError_));
 
+      if (pMatchOn->getSize() == 0)
+      {
+         // If we reach here, grep has found a malformed match
+         pContent = nullptr;
+         pFullLineContent = nullptr;
+         return;
+      }
       *pFullLineContent = decodedLine;
       if (!findResults().replace())
          adjustForPreview(&decodedLine, pMatchOn, pMatchOff);
@@ -1026,6 +1033,9 @@ private:
             json::Array replaceMatchOn, replaceMatchOff;
             processContents(&lineInfo.decodedPreview, &lineInfo.decodedContents,
                &matchOn, &matchOff);
+
+            if (matchOn.getSize() == 0)
+               continue;
 
             if (findResults().replace() &&
                 !(findResults().preview() &&

### CODE REVIEW

**Comment by kevinushey** [comment]
If I understand correctly, using "Whole word" causes us to treat the user text as a regular expression, even though (I think) the intention here is to treat the search string as-is. At least from the UI, the `[ ] Regular expression` checkbox remains unchecked, so that was my expectation.

I think that's happening on the client side here:

https://github.com/rstudio/rstudio/blob/9ab60a7b67570e0b0497b76afb8daab0b291421b/src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/FindOutputPresenter.java#L507-L509

Do we know whether support for literal escapes in our regular expression search is supported? For example, we could instead use:

```
\b\Q<query>\E\b
```

to force the query to be treated as-is ... although technically, this would be a change in behavior, but I do think it better matches the user intention.

**Comment by kevinushey** [comment]
For git grep, this works:

```
git grep -P '\b\Q()\E\b'
```

And on macOS, this seems okay:

```
git grep -E '\b\Q()\E\b'
```

**Comment by jgutman** [comment]
That actually wasn't the full intention of this PR; I had a draft PR up that escapes characters that should be treated as literals in whole word mode; however we also noticed that searching for `()` in regex mode also crashes the session. Even though that's a bad regex, I don't think it should crash the session; this ensures that it won't lead to a crash. It doesn't fully fix the issue of allowing these regex characters to be treated as literals in whole word mode, which is what I was trying to do in the other PR.

File: src/cpp/session/modules/SessionFind.cpp [hunk 1]
```diff
@@ -1027,6 +1034,9 @@ class GrepOperation : public boost::enable_shared_from_this<GrepOperation>
             processContents(&lineInfo.decodedPreview, &lineInfo.decodedContents,
                &matchOn, &matchOff);
 
+            if (matchOn.getSize() == 0)
```
melissa-barca, line 1037: Can we add a comment here about when this is the case?

melissa-barca (approved):
LGTM!

kevinushey:
> That actually wasn't the full intention of this PR; I had a draft PR up that escapes characters that should be treated as literals in whole word mode; however we also noticed that searching for () in regex mode also crashes the session. Even though that's a bad regex, I don't think it should crash the session; this ensures that it won't lead to a crash. It doesn't fully fix the issue of allowing these regex characters to be treated as literals in whole word mode, which is what I was trying to do in the other PR.

Ahh, gotcha. Thank you for clarifying!

kevinushey (approved):

jgutman:
@kevinushey Thanks for the suggestion! Unfortunately the `\Q()\E` syntax doesn't appear to work on the version of grep we bundle with Windows 😭 so I think I'll have to stick to \escaping \every \single \special \character 🙃

Issue #11111, opened 2022-05-03
Pasting fails and clears out the clipboard (Electron)


ronblum:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2022.06.0-daily-322 & 348
    OS Version      : Red Hat 8.5, Ubuntu 20.04.4
    R Version       : 4.0.3, 4.1.3

### Steps to reproduce the problem

Using RStudio Desktop (Electron) on a Linux OS:
1. Copy text from outside of RStudio Desktop (Electron), so that we can rule out copying from RStudio as a factor. For example, copy the word "Congratulations" at the top of https://hmpg.net/
2. In RStudio Desktop, open an R Script or click in the Console.
3. Paste with <kbd>Ctrl-V</kbd>.
    - The word will be pasted, e.g. "Congratulations".
4. Paste again with <kbd>Ctrl-V</kbd>.
    - Nothing is pasted.
5. Switch back to a different application that will accept pasting text, such as in the address bar of a web browser.
7. Paste, presumably with <kbd>Ctrl-V</kbd>.
    - Nothing is pasted.
8. Copy text from somewhere, e.g. again "Congratulations" at the top of https://hmpg.net/
9. In RStudio Desktop, open an R Markdown file.
10. Switch to visual mode.
11. Paste with <kbd>Ctrl-V</kbd> multiple times.
    - The word is pasted multiple times, e.g. `CongratulationsCongratulationsCongratulationsCongratulationsCongratulations`
12. Switch to source mode.
13. Paste with <kbd>Ctrl-V</kbd> multiple times.
    - Nothing is pasted.

### Describe the problem in detail

Except with a markdown file in visual mode, pasting will either only happen the first time (R Script or Console) or not at all (R Markdown in source mode). The clipboard is emptied, so that the former contents can no longer be pasted in other applications, either.

### Describe the behavior you expected

Each paste attempt should result in a paste, and pasting shouldn’t clear the clipboard.

Compare:
* R Markdown in visual mode, in which pasting occurs.
* Qt version of RStudio Desktop.
* RStudio Desktop on MacOS 12.4.
* RStudio Server.

- [ ] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ ] I have installed the latest version of RStudio, and confirmed that the issue still persists.
[...]

PR #11431, opened 2022-06-14
Fixed issue that was breaking clipboard on Electron for Linux

matheustavaresdev:
### Intent

Fixed issue that was breaking clipboard on Electron for Linux

### Approach

Fixed an old change.

### Automated Tests
None

### QA Notes

Refer to #11111

### Checklist

Addresses #11111

### DIFF
--- a/src/node/desktop/src/main/gwt-callback.ts
+++ b/src/node/desktop/src/main/gwt-callback.ts
@@ -310,7 +310,7 @@ export class GwtCallback extends EventEmitter {
     });
 
     ipcMain.on('desktop_set_global_mouse_selection', (event, selection: string) => {
-      clipboard.writeText(selection, 'selection');
+      clipboard.writeText(selection, process.platform === 'linux' ? 'selection' : 'clipboard');
     });
 
     ipcMain.handle('desktop_get_global_mouse_selection', () => {
--- a/src/node/desktop/src/main/gwt-callback.ts
+++ b/src/node/desktop/src/main/gwt-callback.ts
@@ -310,7 +310,7 @@ export class GwtCallback extends EventEmitter {
     });
 
     ipcMain.on('desktop_set_global_mouse_selection', (event, selection: string) => {
-      clipboard.writeText(selection, process.platform === 'linux' ? 'selection' : 'clipboard');
+      clipboard.writeText(selection, 'selection');
     });
 
     ipcMain.handle('desktop_get_global_mouse_selection', () => {
--- a/src/node/desktop/src/main/gwt-callback.ts
+++ b/src/node/desktop/src/main/gwt-callback.ts
@@ -310,7 +310,7 @@ export class GwtCallback extends EventEmitter {
     });
 
     ipcMain.on('desktop_set_global_mouse_selection', (event, selection: string) => {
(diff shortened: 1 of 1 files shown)


src/node/desktop/src/main/gwt-callback.ts
```diff
@@ -310,7 +310,7 @@ export class GwtCallback extends EventEmitter {
     });
 
     ipcMain.on('desktop_set_global_mouse_selection', (event, selection: string) => {
-      clipboard.writeText(selection, 'clipboard');
+      clipboard.writeText(selection, 'selection');
```
timtmok, line 313: I guess this is fine on Windows/Mac? The Electron docs doesn't really explain what these types are and that `selection` is only available on Linux but default is `clipboard`.

timtmok (approved):

src/node/desktop/src/main/gwt-callback.ts
```diff
@@ -310,7 +310,7 @@ export class GwtCallback extends EventEmitter {
     });
 
     ipcMain.on('desktop_set_global_mouse_selection', (event, selection: string) => {
-      clipboard.writeText(selection, 'clipboard');
+      clipboard.writeText(selection, process.platform === 'linux' ? 'selection' : 'clipboard');
```
kevinushey, line 313:
```suggestion
      clipboard.writeText(selection, 'selection');
```

It would be an error to try and call this routine on a non-Linux platform (since "global mouse selection" is a Linux-only concept) so I think we should just always use 'selection' here.

kevinushey: Usages on the GWT side are typically guarded via `isLinuxDesktop()`; e.g.

https://github.com/rstudio/rstudio/blob/9ab60a7b67570e0b0497b76afb8daab0b291421b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/AceEditorMixins.java#L32-L67

https://github.com/rstudio/rstudio/blob/9ab60a7b67570e0b0497b76afb8daab0b291421b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/AceEditorMixins.java#L92-L124

(It's also possible that these Ace mixins are no longer necessary with Electron though?)

kevinushey: Usages on the GWT side are typically guarded via `isLinuxDesktop()`; e.g.

https://github.com/rstudio/rstudio/blob/9ab60a7b67570e0b0497b76afb8daab0b291421b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/AceEditorMixins.java#L32-L67

https://github.com/rstudio/rstudio/blob/9ab60a7b67570e0b0497b76afb8daab0b291421b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/AceEditorMixins.java#L92-L124

(It's also possible that these Ace mixins are no longer necessary with Electron though?)

matheustavaresdev: Thanks for the info! Fixed on [5628c28](https://github.com/rstudio/rstudio/pull/11431/commits/5628c28723d1cda56af907343fdfa05bdce8fbdf)

PR #11437, opened 2022-06-15
Fix bug with counting sessions

MariaSemple:
### Intent

Fix a bug @astayleraz and I just found while over-viewing the session metadata feature.

### Approach

Actually return the count variable.

### Automated Tests

Our tests will run against the sessions that use this call.

### QA Notes

`.rs.invokeServerRpc` with `operation = "count"` should return a non-zero value.

### Checklist


jeffvroom (approved):
LGTM!

PR #11468, opened 2022-06-21
Remove the unused libapparmor1 dependency for Debian-based distros

atheriel:
### Intent

AppArmor functionality was removed in e939585a19e9d98722328a0bac82619b0750aa3b as part of the work to create Launcher (in 2018), but it looks like the Debian package dependency on `libapparmor` was never removed. I noticed this while trying to reconstruct what dependencies the session components have from the main Debian package's metadata.

### Automated Tests

I do not know what existing tests cover packaging-related changes. But since this is a system library dependency, probably "the binary starts on a Debian-based distro" should be sufficient.

### QA Notes

Unknown, please advise.

### Checklist

### DIFF
--- a/package/linux/CMakeLists.txt
+++ b/package/linux/CMakeLists.txt
@@ -87,7 +87,7 @@ if(RSTUDIO_SERVER)
   # automatic dependency resolution use e.g.
   #   sudo apt-get install gdebi-core
   #   sudo gdebi rstudio-server-0.97.151-amd64.deb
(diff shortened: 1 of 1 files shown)


kevinushey:
We do some install-time configuration of AppArmor here:

https://github.com/rstudio/rstudio/blob/f53047636f367d26fd6cf0ac0ed0add01525d6d3/package/linux/debian-control/postinst.in#L58-L72

Is that code still necessary?

atheriel:
I noticed that, too. I think the answer is probably no, if AppArmor was removed four years ago presumably that upgrade would be out of support by now.

kevinushey (approved):
LGTM; I agree with your assessment that if this breaks anything we'll learn about it very quickly (e.g. builds don't launch on some platforms) but that should be unlikely.

PR #11469, opened 2022-06-22
Fix problems with list() method and single-session mode resume

jeffvroom:
** Note - requires a companion commit in pro to add the new validate
parameter to list() in a few places. Let me do the merge so I can coordinate.

### Intent

Addresses: https://github.com/rstudio/rstudio/issues/11446

### Approach

Add a new 'validate' param to list that is passed as 'false' in `rserver` so it doesn't try to remove the session if it can't validate it. Also should be passing homePath, not userDataDir to list in that case so fixed the code but it won't matter since we don't use that parameter when validate=false.  These are the safe parts of the fix.

I also changed the code so that when initializing an rsession without a specific session id (i.e. non-multiple sessions case in the pro or OS), if there is more than one active session, use the most recent one
rather than creating a new one each time. If an extra session gets left over,
that will never resolve itself until all sessions are cleared. Users who switch back and forth between OS and pro will sometimes see this situation, or maybe a session fails to get removed?  I can't see a downside to that change but let me know if I missed an important use case!

### QA Notes

This affects OS and server-multiple-sessions=0 in workbench when you are suspending and resuming. For OS, I'm not sure how you do that without "ant devmode".

### DIFF
--- a/src/cpp/core/include/core/r_util/RActiveSessions.hpp
+++ b/src/cpp/core/include/core/r_util/RActiveSessions.hpp
@@ -707,7 +707,7 @@ public:
                       const std::string& editor,
                       std::string* pId) const;
 
-   std::vector<boost::shared_ptr<ActiveSession> > list(FilePath userHomePath, bool projectSharingEnabled) const;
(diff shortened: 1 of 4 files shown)


kevinushey (approved):
LGTM!

MariaSemple (approved):
LGTM!

Issue #10266, opened 2021-12-19
Custom roxygen tags in autocomlete


mikldk:
Autocomplete is available for writing `roxygen2` code in RStudio. This is (I guess) made possible via [this code](https://github.com/rstudio/rstudio/blob/7ce19c54507e53de8859502605774a825cced146/src/cpp/session/modules/SessionRCompletions.R#L221).

My package [`roxytest`](https://github.com/mikldk/roxytest) introduces two new tags, `@tests` and `@testexamples` (see [here](https://github.com/mikldk/roxytest)).

Is it somehow possible to add custom tags into RStudio's autocomlete system?

(This is related to mikldk/roxytest#23 by @shahronak47.)

mikldk:
I hope indeed that this will be implemented at some point, and thus I would like to keep this issue open.

kevinushey:
Right now, the set of known roxygen tags is hard-coded here:

https://github.com/rstudio/rstudio/blob/f72551b40e7a0b53a4950a652356ce192807536a/src/cpp/session/modules/SessionRCompletions.R#L160-L217

We could consider reading tags from other packages loaded in the session, though. Is there a way that RStudio could ask `roxytest` what Roxygen tags it is defining / exporting?

mikldk:
@kevinushey Yes, indeed: I submitted a PR where I added the two tags: rstudio/rstudio#11470. It is not a beautiful way of solving this, but on the other hand - if only a few packages define custom tags - it may not hurt doing it this way.

Doing what you suggest requires agreeing on the packages providing e.g. a special function (or declaring a tag in the `DESCRIPTION`), but RStudio would have to go through all packages everything. It may not take that long, but it would often (too often IMO) be in vein.

mikldk:
@kevinushey I think we continue discussing it here rather than in rstudio/rstudio#11470. My suggestion there was simply to edit the hard-coded list:
https://github.com/rstudio/rstudio/blob/f72551b40e7a0b53a4950a652356ce192807536a/src/cpp/session/modules/SessionRCompletions.R#L160-L217
I.e., by adding the two tags:
```
      # Two tags from roxytest (https://github.com/mikldk/roxytest/).
      # They must be here as later it is checked whether
      #    if (.rs.acContextTypes$ROXYGEN %in% type)
      # and the roxytest tags are used in roxygen context
      "@tests ",
      "@testexamples "
```
[...]

PR #11470, opened 2022-06-22
Roxytest tags

mikldk:
### Intent

From rstudio/rstudio#10266:

Autocomplete is available for writing `roxygen2` code in RStudio. 
My package [`roxytest`](https://github.com/mikldk/roxytest) introduces two new tags, `@tests` and `@testexamples` (see [here](https://github.com/mikldk/roxytest)) used in `roxygen2` context.

This PR adds the two tags.

### Approach

The tags had to be declared next to the `roxygen2` ones as they are used in a `roxygen2` context. This is also noted in the code.

### Automated Tests

Autocompletion is not well-covered by tests and I did not want to invent my own strategy for that.

### QA Notes

Minimal risk.

### Checklist

Contributor agreement signed and mailed to you.


kevinushey:
Sorry, but I don't think we should add non-roxygen2 tags to the hard-coded list. I would prefer a solution that queried packages loaded in the current session for the tags they support.

mikldk:
Closing (thinking about alternative solutions).

mikldk:
> Sorry, but I don't think we should add non-roxygen2 tags to the hard-coded list. I would prefer a solution that queried packages loaded in the current session for the tags they support.

But neither roxygen2 nor roxytest need to be loaded before using devtools::document(), nor using their tags. So I don’t think loaded packages are the correct way. Thoughts?

Issue #11395, opened 2022-06-09
Error 10048 and 10053 on RStudio Startup on Windows with Proxy


wtimmerman-fitp:
### System details

    RStudio Edition : Desktop
    RStudio Version : RStudio 2022.06.0-daily+497 "Spotted Wakerobin" Daily (78ce5eff140c7fc8e06e2bff739571334d3c55e5, 2022-06-09) for Windows Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) QtWebEngine/5.12.8 Chrome/69.0.3497.128 Safari/537.36  (also others tested below);
    OS Version      : Windows 10 Pro 
    R Version       : 4.0.3 (also others tested below)

### Steps to reproduce the problem

1. Have system proxy running (details below)
2. Open an instance of RStudio normally in Windows 10. I have a few versions of R installed, using 4.0.3 in this case. This session works fine.
3. Open a second instance of RStudio. This results in the "Error Starting R" warning and the set of error messages, specifically:

From:
C:/Users/MyName/AppData/Local/RStudio/log/rsession-MyName.log

`2022-06-07T06:52:01.004631Z [rsession-MyName] ERROR system error 10048 (Only one usage of each socket address (protocol/network address/port) is normally permitted); OCCURRED AT class rstudio::core::Error __cdecl rstudio::core::http::initTcpIpAcceptor(class rstudio::core::http::SocketAcceptorService<class rstudio_boost::asio::ip::tcp> &,const class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > &,const class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > &) src/cpp/session/http/SessionWin32HttpConnectionListener.cpp:122; LOGGED FROM: int __cdecl main(int,char *const []) src/cpp/session/SessionMain.cpp:2088`

After further testing (more below), i also received this error:
[...]

PR #11475, opened 2022-06-22
tweaks to discovery of open ports for desktop

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/11395.

### Approach

Because we don't know the underlying cause of the issue, this PR takes the "let the user decide" approach.

- Users can set the `RSTUDIO_DESKTOP_SESSION_PORT_RANGE` environment variable to tell RStudio Desktop to use a specific port range. This should be of the form `MIN-MAX`, for some integers MIN and MAX.
- Users can disable RStudio's port validation attempts by setting `RSTUDIO_DESKTOP_SESSION_PORT_VALIDATION_ENABLED = FALSE`. This is mainly to help support cases where port validation fails for some reason; e.g. the RStudio front-end is unable to bind to a port where `rsession.exe` might otherwise still be able to.
- If RStudio fails to find an open port, it still chooses a random port in the aforementioned port range, so at least there's still a chance of avoiding port collisions.

If we can confirm from those affected in https://github.com/rstudio/rstudio/issues/11395 what piece of this fixes the issue, then we can remove the optional pieces before release (if desired).

Note: I think we can intentionally leave these environment variables undocumented for now; we can consider documenting them if we find it's really necessary down the line.

### Automated Tests

Unfortunately none.

### QA Notes

We may need to request testing + follow up from users in https://github.com/rstudio/rstudio/issues/11395 as I don't think we can easily test + validate the underlying issue locally.

However, we can test a couple things:

- Try setting `RSTUDIO_DESKTOP_SESSION_PORT_RANGE = 4000-8000`, start RStudio, and confirm that the `RSTUDIO_SESSION_PORT` environment variable is set to a value within that range.

- Try setting an invalid port range, and confirm that RStudio falls back to a port in the range 49152-65536.

### Checklist

### DIFF
--- a/src/cpp/desktop/DesktopOptions.cpp
+++ b/src/cpp/desktop/DesktopOptions.cpp
@@ -19,17 +19,126 @@
 #include <QApplication>
 #include <QDesktopWidget>
 
+#include <boost/algorithm/string/split.hpp>
+
+#include <shared_core/SafeConvert.hpp>
+
 #include <core/Random.hpp>
+#include <core/StringUtils.hpp>
 #include <core/system/System.hpp>
 #include <core/system/Environment.hpp>
 
 #include "DesktopInfo.hpp"
 #include "DesktopUtils.hpp"
 
+#define kRStudioDesktopSessionPortValidationEnabled "RSTUDIO_DESKTOP_SESSION_PORT_VALIDATION_ENABLED"
+#define kRStudioDesktopSessionPortRange "RSTUDIO_DESKTOP_SESSION_PORT_RANGE"
+#define kRStudioDesktopSessionPort "RSTUDIO_DESKTOP_SESSION_PORT"
+
 using namespace rstudio::core;
 
 namespace rstudio {
 namespace desktop {
+
+namespace {
+
+bool portIsOpen(int port)
+{
+   // Disable validation checks if requested; mainly for testing in cases
+   // where QTcpSocket is unable to successfully bind to an open port
+   std::string envvar = core::system::getenv(kRStudioDesktopSessionPortValidationEnabled);
+   bool needsValidation = core::string_utils::isTruthy(envvar, true);
+   if (!needsValidation)
+      return true;
+
+   // Create a socket
+   QTcpSocket socket;
+
+   // Set up an error handler
+   QObject::connect(&socket, QOverload<QAbstractSocket::SocketError>::of(&QAbstractSocket::error), [&]()
+   {
+      DLOGF(
+               "An error occurred while attempting to bind to port {}: {}",
+               port,
+               socket.errorString().toStdString());
+   });
+
+   if (!socket.bind(port))
+   {
+      DLOGF("Failed to bind TCP socket to port {}", port);
+      return false;
+   }
+
+   // we found a good port; close our socket and save it
+   //
+   // TODO: it'd be nice to keep the port locked until we are
+   // officially ready to launch the rsession, or let rsession
+   // communicate to us which port it wants to use via some
+   // separate channel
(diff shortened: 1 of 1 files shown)


src/cpp/desktop/DesktopOptions.cpp
```diff
@@ -19,18 +19,126 @@
...
+   DLOGF("Using port range [{}, {})", portRangeStart, portRangeEnd);
+   for (int i = 0; i < 20; i++)
+   {
+      // generate a port number
+      int port = core::random::uniformRandomInteger<int>(portRangeStart, portRangeEnd);
+      if (portIsOpen(port))
+         return port;
+   }
+
+   // we couldn't find an open port; just choose and random port and hope for the best
```
gtritchie, line 134: nitpick: "a random port", not "and random port"

src/cpp/desktop/DesktopOptions.cpp
```diff
@@ -151,64 +259,29 @@ void Options::saveMainWindowBounds(QMainWindow* win)
...
+   QString localPeer =  QStringLiteral("\\\\.\\pipe\\%1-rsession").arg(portNumber_);
+   localPeer_ = localPeer.toUtf8().constData();
+   core::system::setenv("RS_LOCAL_PEER", localPeer_);
 #endif
-    
-      // return discovered port
-      return portNumber_;
-   }
 
-   // if we get here, we failed to find an open port
```
gtritchie, line 209: Are we sure we want to pick a random port in this fallthrough case versus returning the formerly hardcoded 8789? I have no idea if this is a good idea or not, or if anything relies on this old behaviour, but would hate to find out after the fact that somebody, somewhere, was depending on this.

src/cpp/desktop/DesktopOptions.cpp
```diff
@@ -19,18 +19,126 @@
...
+         portRangeEnd   = safe_convert::stringTo<int>(parts[1], portRangeEnd);
+      }
+      else
+      {
+         DLOGF("Unexpected value '{}' for {}", portRangeOverride, kRStudioDesktopSessionPortRange);
+      }
+   }
+
+   DLOGF("Using port range [{}, {})", portRangeStart, portRangeEnd);
+   for (int i = 0; i < 20; i++)
```
gtritchie, line 126: Why the change from looping up to 100 times?

src/cpp/desktop/DesktopOptions.cpp
```diff
@@ -19,18 +19,126 @@
...
+   // allow override if necessary
+   std::string portRangeOverride = core::system::getenv(kRStudioDesktopSessionPortRange);
+   if (!portRangeOverride.empty())
+   {
+      std::vector<std::string> parts;
+      boost::algorithm::split(parts, portRangeOverride, boost::is_any_of(":-,"));
+
+      if (parts.size() == 2)
+      {
+         portRangeStart = safe_convert::stringTo<int>(parts[0], portRangeStart);
```
gtritchie, line 116: Should we validate that start < end?

src/cpp/desktop/DesktopOptions.cpp
kevinushey: That was a change introduced with https://github.com/rstudio/rstudio/commit/d07d88465423fe224d86e3337f531ebfb90b505a (that is, defaulting to 8789 wasn't legacy behavior) so I think this change actually brings us back closer to the older behavior.

However, it might be worth considering -- we previously considered ports in the range 4000 through 8000. Should we consider that range as a fallback? (assuming some existing environments are expecting RStudio + rsession to communicate on a port in that range?)

gtritchie (approved):
LGTM!

Issue #11386, opened 2022-06-08
Dock icon doesn't show project (Electron/Mac)


gtritchie:
### System details

    RStudio Edition : Electron
    RStudio Version : 2022.06.0-daily+492 (any Electron build to date)
    OS Version      : macOS 
    R Version       : Any

### Steps to reproduce the problem

- open a project in RStudio Electron
- look at the dock icon
- repeat with the Qt desktop

### Describe the problem in detail

- Project name not displayed in the dock icon by Electron

![RStudio icon in Mac dock without project name](https://user-images.githubusercontent.com/10569626/172674407-ce1e18df-a053-476d-925d-0013e9d1120d.png)

### Describe the behavior you expected

- Shows the project name ("aaa" in this example) in the dock icon as seen here with Qt desktop:

![RStudio icon in Mac dock with project name](https://user-images.githubusercontent.com/10569626/172674533-8a7e020a-3899-49b4-b31a-adc0f1306bc2.png)

### Implementation Details

- Code from Qt needs to be ported, e.g.:

https://github.com/rstudio/rstudio/blob/be0d860a4650b185132484a0a38dd041bb993cf5/src/cpp/desktop/DockTileView.mm#L65-L123 

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

matheustavaresdev:
This requires adding `DockTileView.mm` and calling it's `setLabel` function from inside `src/node/desktop/src/native/desktop.cc` so it can be called from within the Electron Code. You can change the places at https://github.com/rstudio/rstudio/pull/11480 that call `app.dock.setBadge` into the new native function.

jonvanausdeln:
Verifided on
2022.11.0-70
MacOS Monterey 12.4

Tested for most project types

Issue #11387, opened 2022-06-08
Dock menu missing New RStudio Window command, showing generic icon (Electron/Mac)


gtritchie:
### System details

    RStudio Edition : Electron
    RStudio Version : 2022.06.0-daily+492 (any Electron build to date)
    OS Version      : macOS
    R Version       : Any

### Steps to reproduce the problem

- right-click on RStudio dock icon (Electron)

### Describe the problem in detail

- missing the "New RStudio Window" command
- there's a generic icon shown between checkmark and window title

![RStudio Electron dock menu screenshot](https://user-images.githubusercontent.com/10569626/172676882-9bdfd851-78b4-4756-b8f9-fd48a19b8d19.png)

### Describe the behavior you expected

- should include the "New RStudio Window" command and no generic icon, as seen here in same Qt build:

![RStudio Qt dock menu](https://user-images.githubusercontent.com/10569626/172676928-9e57d32e-c7e7-4075-ba99-fd9b03b0a8f2.png)

### Implementation Notes

The Electron icon not show the project name is tracked via #11386.

Need to port over the code from Qt, e.g.:

https://github.com/rstudio/rstudio/blob/be0d860a4650b185132484a0a38dd041bb993cf5/src/cpp/desktop/DockMenu.cpp#L23-L42

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
Verified that "New RStudio Window" is present and opens a new window, using RStudio Desktop 2022.10.0-daily+55 Elsbeth Geranium.

PR #11480, opened 2022-06-23
Show Project Name on Dock Icon - Electron

matheustavaresdev:
### Intent

Show the current project name on the dock icon.

### Approach

Achieved by setting the badge. 

Notes: 
- The background for it is red instead of light blue as in QT. Tried to make it blue but would take too long as it would mess with electron code as it does not currently support fine customizations on it's Dock Badge api

Examples: ( In red, Electron version. In blue, the QT version )

Screen Shot 2022-06-26 at 23 20 08
Screen Shot 2022-06-26 at 23 19 30

### Automated Tests

None

### QA Notes

Refer to #11386

### Checklist

Addresses #11386

### DIFF
--- a/src/node/desktop/src/main/application.ts
+++ b/src/node/desktop/src/main/application.ts
@@ -13,7 +13,7 @@
  *
  */
 
-import { app, BrowserWindow, Menu, screen, WebContents } from 'electron';
+import { app, BrowserWindow, screen, WebContents } from 'electron';
 import i18next from 'i18next';
 import path from 'path';
 import { getenv, setenv } from '../core/environment';
@@ -213,8 +213,6 @@ export class Application implements AppState {
       this.argsManager.handleAfterSessionLaunchCommands();
     });
 
-    this.setDockMenu();
-
     return run();
   }
 
@@ -295,19 +293,6 @@ export class Application implements AppState {
         owner,
         baseUrl,
       );
-    }
-  }
-
-  setDockMenu(){
-    if (process.platform === 'darwin'){
-      const  menuDock = Menu.buildFromTemplate([{
-        label: 'New RStudio Window',
-        click: () => {
-          this.appLaunch?.launchRStudio({});
-        }
-      }]);
-
-      app.dock.setMenu(menuDock);
     }
   }
 }
--- a/src/node/desktop/src/main/application.ts
+++ b/src/node/desktop/src/main/application.ts
@@ -13,7 +13,7 @@
  *
  */
 
-import { app, BrowserWindow, screen, WebContents } from 'electron';
+import { app, BrowserWindow, Menu, screen, WebContents } from 'electron';
 import i18next from 'i18next';
 import path from 'path';
 import { getenv, setenv } from '../core/environment';
@@ -213,6 +213,8 @@ export class Application implements AppState {
       this.argsManager.handleAfterSessionLaunchCommands();
     });
 
+    this.setDockMenu();
+
     return run();
   }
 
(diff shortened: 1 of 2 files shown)


src/node/desktop/src/main/application.ts
```diff
@@ -295,4 +297,17 @@ export class Application implements AppState {
       );
     }
   }
+
+  setDockMenu(){
+    if (process.platform === 'darwin'){
+      const  menuDock = Menu.buildFromTemplate([{
+        label: 'New RStudio Window',
```
timtmok, line 304: Now that there's string externalization set up, this should be added.

timtmok (changes requested):

src/node/desktop/src/main/application.ts
matheustavaresdev: That's true my bad

timtmok:
We could leave the badge as is and follow up with another change to use native code to overlay the project name on the icon. It looks like we could include the `DockTileView.mm` code and call it instead of `setBadge`. @matheustavaresdev or is that what you mean by it messes with the Electron code?

gtritchie:
@matheustavaresdev can you add a screenshot showing how the badge looks with this PR?

gtritchie:
Also it sounds like this PR also addresses #11387?

src/node/desktop/src/main/main-window.ts
```diff
@@ -217,8 +217,12 @@ export class MainWindow extends GwtWindow {
 
     this.executeJavaScript('window.desktopHooks.getActiveProjectDir()')
       .then((projectDir) => {
+        if (process.platform == 'darwin') app.dock.setBadge('');
+
         if (projectDir.length > 0) {
           this.window.setTitle(`${projectDir} - ${appState().activation().editionName()}`);
+       
+          if (process.platform == 'darwin') app.dock.setBadge(projectDir);
```
kevinushey, line 225: How do longer project names appear here? Are they truncated sensibly?

matheustavaresdev: No, they are truncated on the middle automatically by Electron in this pattern: `[first 4 letters]...[last 5 letters]` . The QT version shows the first 14 letters . I could show only the first X amount of letters instead of letting Electron clip it automatically... How should this be done? 

Screen Shot 2022-06-26 at 23 20 08

Screen Shot 2022-06-26 at 23 19 30

matheustavaresdev:
> Also it sounds like this PR also addresses #11387?

Yes! @gtritchie

matheustavaresdev:
> @matheustavaresdev can you add a screenshot showing how the badge looks with this PR?

Added on the PR Description! cc @gtritchie

matheustavaresdev:
> We could leave the badge as is and follow up with another change to use native code to overlay the project name on the icon. It looks like we could include the `DockTileView.mm` code and call it instead of `setBadge`. @matheustavaresdev or is that what you mean by it messes with the Electron code?

By "messing with electron code", I meant that we could change their `setBadge` function to change the background and other data so we could customize to our needs.

src/node/desktop/src/main/application.ts
matheustavaresdev: Fixed on [6dafb5f](https://github.com/rstudio/rstudio/pull/11499/commits/6dafb5fa11e7d4763a121ada108df55558a6ed90)

matheustavaresdev:
> I agree with @timtmok. Let's not use this approach and wait until we can use the native code to implement the same behaviour we get with Qt.

Alright. Will work on that then!

Issue #9543, opened 2021-06-27
RStudio 1.4.1717 still won't run with R-4.1.0-arm64.pkg


greglum:
### System details

    RStudio Edition : Desktop
    RStudio Version : 1.4.1717
    OS Version      :  macOS Big Sur 11.4
    R Version       : 4.1 (arm)

### Steps to reproduce the problem

- Install the R-4.1.0-arm64.pkg (and ensure it is your only version of R on the system)
- R will run fine on the system
- Install RStudio (1.4.1717) and ensure Rosetta 2 is properly installed ( /usr/sbin/softwareupdate --install-rosetta )
- RStudio will fail with the output 

```
dyld: could not load inserted library
'/Library/Frameworks/R.framework/Resources/lib/libR.dylib'
because no suitable image found. Did find:
/Library/Frameworks/R.framework/Resources/lib/libR.dylib: mach-o, but wrong architecture
/Library/Frameworks/R.framework/Resources/lib/libR.dylib: stat() failed with errno=1
```

### Describe the problem in detail

Hello, with the release notes of RStudio 1.4.1717 it is supposed to run natively with the R arm packages (with RStudio still running via Rosetta 2,  of course). However, I could not get it to work; in order to still run RStudio, I had to uninstall the R arm pkg and replace it with the Intel version, similar to this [thread here](https://community.rstudio.com/t/rstudio-does-not-start-up-with-r-4-1-0-on-macbook-pro-m1-macos-big-sur-11-3-1/105533/3). I have reproduced this off of a fresh OS install as well as multiple resets of RStudio's Desktop State. After running multiple diagnostics, I have isolated it to an RStudio issue as the  R-4.1.0-arm64.pkg would run successfully on my system. 

### Describe the behavior you expected

Would expect RStudio (1.4.1717) to run via Rosetta 2 with only the R(4.1) arm pkg installed on the system.

- [x ] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [ x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

Edit: formatting

ronblum:
[...]

PR #11489, opened 2022-06-23
only set DYLD_INSERT_LIBRARIES if running rsession directly

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/9543.

### Approach

Don't set `DYLD_INSERT_LIBRARIES` if we're running `rsession` through `/usr/bin/arch`, as doing so tries to force macOS to load the R library into _the /usr/bin/arch process_, rather than the `rsession` process! I still don't quite understand why this only fails sometimes, but I'm pretty confident this is the underlying issue (as I was finally able to reproduce locally).

### Automated Tests

N/A

### QA Notes

Test that RStudio can run with both x86_64 and arm64 R installations on M1 macOS machines.

### Checklist

### DIFF
--- a/src/cpp/desktop/DesktopSessionLauncher.cpp
+++ b/src/cpp/desktop/DesktopSessionLauncher.cpp
@@ -97,19 +97,27 @@ void launchProcess(const std::string& absPath,
 #ifdef Q_OS_DARWIN
    
    QProcessEnvironment environment = QProcessEnvironment::systemEnvironment();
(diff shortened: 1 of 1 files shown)


gtritchie (approved):
LGTM! Very weird, though!

gtritchie:
Also asked in Slack, but do we need to port this to Electron, or does it already handle this?

PR #11493, opened 2022-06-24
Change SchemaVersion Comparison Logic

AndrewMcClain:
Now assumes that any unknown flower version is a later rather than prior
version

Add test coverage to ensure the comparisons are correct.

Addresses rstudio/rstudio-pro#3572

### Intent

When a user downgrades the previous version was squashing the db schema, preventing later upgrades from being successful due to a mismatch between the actual schema and the reported schema of the database.

### Approach

Previously unknown flower versions were assumed to be prior versions, unknown flowers are now presumed to be later versions

### Automated Tests

Automated test coverage added to assert the comparisons between the versions, and empty version, unknown version.

### QA Notes

This specifically addresses a scenario in which an installation is downgraded, and then upgraded again.

Previously downgrading would cause the SchemaVersion of the DB to be updated. After upgrading again a duplicate column error when starting the server would prevent the server from starting.

Now doing a PT -> GO -> PT downgrade/upgrade path should result in a working system each step of the way.

### Checklist

### DIFF
--- a/src/cpp/core/Database.cpp
+++ b/src/cpp/core/Database.cpp
@@ -769,8 +769,8 @@ bool SchemaVersion::operator<(const SchemaVersion& other) const
       return false;
 
    const auto& versions = versionMap();
-   int thisFlowerIndex = (versions.find(Flower) != versions.end()) ? versions.at(Flower) : -1;
-   int otherFlowerIndex = (versions.find(other.Flower) != versions.end()) ? versions.at(other.Flower) : -1;
+   int thisFlowerIndex = (versions.find(Flower) != versions.end()) ? versions.at(Flower) : INT_MAX;
+   int otherFlowerIndex = (versions.find(other.Flower) != versions.end()) ? versions.at(other.Flower) : INT_MAX;
 
    if (thisFlowerIndex < otherFlowerIndex)
       return true;
--- a/src/cpp/core/Database.cpp
+++ b/src/cpp/core/Database.cpp
@@ -769,8 +769,8 @@ bool SchemaVersion::operator<(const SchemaVersion& other) const
       return false;
 
    const auto& versions = versionMap();
-   int thisFlowerIndex = (versions.find(Flower) != versions.end()) ? versions.at(Flower) : INT_MAX;
-   int otherFlowerIndex = (versions.find(other.Flower) != versions.end()) ? versions.at(other.Flower) : INT_MAX;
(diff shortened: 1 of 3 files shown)


src/cpp/core/Database.cpp
```diff
@@ -769,8 +769,8 @@ bool SchemaVersion::operator<(const SchemaVersion& other) const
       return false;
 
    const auto& versions = versionMap();
-   int thisFlowerIndex = (versions.find(Flower) != versions.end()) ? versions.at(Flower) : -1;
-   int otherFlowerIndex = (versions.find(other.Flower) != versions.end()) ? versions.at(other.Flower) : -1;
+   int thisFlowerIndex = (versions.find(Flower) != versions.end()) ? versions.at(Flower) : INT_MAX;
```
MariaSemple, line 772:
```suggestion
   int thisFlowerIndex = (versions.find(Flower) != versions.end()) ? versions.at(Flower) : INT_MAX;
```

src/cpp/core/Database.cpp
```diff
@@ -769,8 +769,8 @@ bool SchemaVersion::operator<(const SchemaVersion& other) const
...
+   int otherFlowerIndex = (versions.find(other.Flower) != versions.end()) ? versions.at(other.Flower) : INT_MAX;
```
MariaSemple, line 773: I think this might be more portable and achieve the same result
```suggestion
   int thisFlowerIndex = (versions.find(Flower) != versions.end()) ? versions.at(Flower) : versions.size();
   int otherFlowerIndex = (versions.find(other.Flower) != versions.end()) ? versions.at(other.Flower) : versions.size();
```

src/cpp/core/DatabaseTests.cpp
```diff
@@ -437,6 +437,38 @@ TEST_CASE("Database", "[.database]")
...
    }
 
+   test_that("Schema Version comparisons are correct")
+   {
+      std::vector<SchemaVersion> versions {
+         {},
+         {"", "20200226141952248123456_AddRevokedCookie"},
+         {"Ghost Orchid", "20210712182145921760944"},
+         {"Prairie Trillium", "20210916132211194382021"},
+         {"Aphid", "21210916132211194382021"}
```
MariaSemple, line 447: What is `Aphid`?

MariaSemple: Oh, this is a test, nevermind!

kevinushey:
@AndrewMcClain @MariaSemple -- is this PR ready for merge?

MariaSemple (approved):
LGTM!

Issue #11515, opened 2022-06-28
File cannot be saved when created through File > New File


timtmok:
### Steps to reproduce:
* Create a file using File > New File (any type will do)
* Attempt to save the file (any name will do)

The file is not saved and does not show up in the Files pane. The editor shows the file name as `Untitled`. Attempting to create another file has the same result except that it may show the file in the Files pane. However, it is _not saved_ and refreshing the pane makes the file disappear.

So far, an issue on Windows. Mac saves the file successfully.

### Expected
The file should be saved.

Build: 2022.07.0+544 Electron
OS: Windows

timtmok:
I'm testing a fix locally for this. It seems that there is a bug with normalizing the save path. I also discovered another issue with the path normalization that assumes the user is saving the file to their home directory (or subdirectory of it). Going to run through some other tests to make sure this works correctly.

It's also only occurring in builds because local dev build behaves slightly differently with the environment variables.

timtmok:
Some things to test with the fix are:
* saving somewhere within the user's home directory
* saving on Linux or Mac since the altered code is shared with these platforms
* saving to a network location or outside of the home directory

If the path is not accessible (i.e. no permission), the save does result in a descriptive enough error message on Windows but not on Mac (says the path doesn't exist). On Windows, it does allow choosing a new path. On Mac, the dialog has closed and attempting to save again results in the same error message without the option to choose a new path (same thing occurs in a Qt build).

dfalty:
@andrie the fix would not be in 544. Looks like the fix is in [547](https://dailies.rstudio.com/version/2022.07.0+547.pro4/).

ronblum:
On MacOS 12.5, RStudio Desktop 2022.07.0+547
* The "Save" options work, saving outside of the home directory works
* The "Save All" options don't work (for new files)

anwhite003:
Why does this appear to be a closed issue ??
I just downloaded RStudio 2022.07.0 Electron for latest build preview version on Windows 10 at work (on an HP Z2) and the GUI absolutely fails to save an RScript file, whether I select a new path or leave the standard directory I have for the RStudio Project.  If I quit & switch back to Non-Electron version of PRIOR RStudio Preview release, the R Script file is saved. This would APPEAR to be an Electron issue and AFAIK should NOT be closed. - Andy White (Friday 1 July 2022)

gtritchie:
[...]

PR #11538, opened 2022-06-30
Fix path normalization when saving file

timtmok:
### Intent
Address #11515 - File cannot be saved on Windows

### Approach
In dev, `process.env.HOME` returns the user's home path. In a build, it returns `undefined`. The error is caused by normalizing the home path.

There was another bug uncovered as well where we are assuming the save path is somewhere in the user's home. So this only performs the home alias replacement if the save path starts with the home path. It is a similar check as the CPP code.

### Automated Tests
None

### QA Notes
This only affected Windows but would be good to test the change on Linux or Mac (tested in dev on Mac and Windows). A network path is likely the use case where a user would save outside of their home directory.

### Checklist

### DIFF
--- a/src/node/desktop/src/renderer/desktop-bridge.ts
+++ b/src/node/desktop/src/renderer/desktop-bridge.ts
@@ -13,7 +13,7 @@
  *
  */
 
-import { ipcRenderer, webContents } from 'electron';
+import { ipcRenderer, SaveDialogReturnValue, webContents } from 'electron';
 import { logger } from '../core/logger';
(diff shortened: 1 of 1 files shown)


gtritchie (approved):
LGTM!

gtritchie:
Do we want to take this fix in Spotted Wakerobin so the Electron preview has it? Or is it too late for that?

andrie:
I commented at #11515, but I still this problem on Windows in build 544.

Issue #11487, opened 2022-06-23
Deno processes stick around if Quarto files not closed before closing RStudio


ronblum:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2022.07.0-preview+540
    OS Version      : Windows 11
    R Version       : 4.2.0

### Steps to reproduce the problem

Using RStudio Desktop on Windows:
1. Open the Windows Task Manager in order to view the running processes.
2. Open a Quarto document. The default is sufficient.
3. Render the document.
    - This creates a deno.exe process.
4. Close the document.
    - The deno.exe process terminates.
5. Open a Quarto document, whether the same or another.
6. Render the document.
    - This creates a deno.exe process.
7. Close RStudio Desktop. Do not close the document before exiting RStudio—it must be left open, or else the bug won't occur.
   - The deno.exe process continues to run. It might be a subprocess under an RStudio process that lingers.

### Describe the problem in detail

The deno.exe process sticks around. One problem that this creates is that when trying to install a new version of RStudio Desktop, the installation fails, warning that the process is still running.

### Describe the behavior you expected

The deno.exe process should terminate. This will allow users to install RStudio Desktop updates.

- [ ] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ ] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [ ] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

jjallaire:
@kevinushey we noticed that the normal process termination code for deno processes wasn't actually working on Windows (esp. when it had a web server running). This extra sauce was added to the **stop** action for quarto preview jobs:

https://github.com/rstudio/rstudio/blob/main/src/cpp/session/modules/quarto/SessionQuartoJob.hpp#L59-L84

I think that to resolve this problem we just need to arrange for the **stop** code to be called on session shutdown. I looked a bit and wasn't confident that I could do this safely. I noticed that there is a cleanup operation that is called for each job but it wasn't clear how to hook into it. I also noticed that the jobs list was cleared on session shutdown so that also creates some potential order dependency.  

Thought it better that someone with a more recent "working set" in rsession make this change.

ronblum:
[...]

PR #11544, opened 2022-07-02
let session create its own job object

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/11487.

### Approach

On Windows, we use job objects to ensure that child processes are cleaned up when a parent process exits. However, node creates processes with `JOB_OBJECT_LIMIT_SILENT_BREAKAWAY_OK`, which effectively prevents new processes from being added to an existing job object.

This PR resolves the issue by allowing the R session to create its own job object, so that the lifetime of child processes created by the session is limited by the lifetime of the session itself.

Some open questions...

- Is there a way for us to pass the desktop's job object handle down to the session, when it is initialized? Is this worth doing?
- Can we convince ourselves that node will appropriately kill the session process if the desktop process dies in this case?

### Automated Tests

N/A

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/11487.

### Checklist

### DIFF
--- a/src/cpp/core/system/Win32System.cpp
+++ b/src/cpp/core/system/Win32System.cpp
@@ -88,8 +88,11 @@ Error initJobObject(bool* detachFromJob)
    }
 
    JOBOBJECT_EXTENDED_LIMIT_INFORMATION jeli = { 0 };
-   jeli.BasicLimitInformation.LimitFlags = JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE |
-                                           JOB_OBJECT_LIMIT_BREAKAWAY_OK;
+   jeli.BasicLimitInformation.LimitFlags =
+         JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE |
+         JOB_OBJECT_LIMIT_BREAKAWAY_OK |
+         JOB_OBJECT_LIMIT_DIE_ON_UNHANDLED_EXCEPTION;
+
    ::SetInformationJobObject(hJob,
                              JobObjectExtendedLimitInformation,
                              &jeli,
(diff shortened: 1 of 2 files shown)


kevinushey:
On desktop, we already have some tools for detecting parent termination, e.g.

https://github.com/rstudio/rstudio/blob/d5a1bc209137589551f02a357ba88d0c74ef3e05/src/cpp/session/SessionMain.cpp#L1475-L1496

But as far as I can see, the win32 implementation is basically just a stub:

https://github.com/rstudio/rstudio/blob/d5a1bc209137589551f02a357ba88d0c74ef3e05/src/cpp/core/system/Win32ParentProcessMonitor.cpp#L28-L38

But even without that, from some light testing, it seems like the R session still shuts itself down when the front-end is closed, even if they aren't part of the same job object?

gtritchie:
@kevinushey 

> from some light testing, it seems like the R session still shuts itself down when the front-end is closed, even if they aren't part of the same job object?

Do we initiate an rsession shutdown directly in a normal user-initialized shutdown of the desktop process? I think we do but haven't gone looking.

What happens if you kill rstudio.exe? Does the session still go away with this fix in place?

kevinushey:
> Do we initiate an rsession shutdown directly in a normal user-initialized shutdown of the desktop process? I think we do but haven't gone looking.

We do, regular shutdowns work as expected (e.g. you get asked to save the R workspace)

> What happens if you kill rstudio.exe? Does the session still go away with this fix in place?

It does, along with all the processes that were launched by the R session. Honestly, I'm not exactly sure why, but it seems like the rsession performs a clean shutdown in the case that the front-end is terminated.

gtritchie (approved):
LGTM, not sure this will be the end of the adventure here, but should move us in the right direction!

Issue #10897, opened 2022-04-04
Electron: Open newly rendered docx files with Word


ronblum:
### System details

    RStudio Edition : Desktop
    RStudio Version : RStudio-2022.06.0-daily-266 (Electron)
    OS Version      : MacOS 12.3
    R Version       : 4.1.3

### Steps to reproduce the problem

Render a document in Word (docx) format. The default documents suffice.
- R Markdown: Set `output: word_document` in the YAML.
- Quarto: Set `format: docx` in the YAML.

### Describe the problem in detail

In an Electron Desktop, the .docx file is created, but Word doesn't automatically open.

### Describe the behavior you expected

Word (or LibreOffice in Linux, but I haven't tried there yet) should launch and open the rendered document in read-only mode. This is the behavior in the Qt version.

- [ ] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ ] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [ ] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
Handling this requires implementing the following callbacks:

https://github.com/rstudio/rstudio/blob/185809a4ca28cecc98ac67cd01d9228223d54796/src/node/desktop/src/main/gwt-callback.ts#L365-L367

https://github.com/rstudio/rstudio/blob/185809a4ca28cecc98ac67cd01d9228223d54796/src/node/desktop/src/main/gwt-callback.ts#L353-L355

We may need to port the associated code here for Windows:

https://github.com/rstudio/rstudio/blob/main/src/cpp/desktop/DesktopWordViewer.cpp

For a short-term solution, it might suffice to just use `shell.openPath()` in each case.

https://www.electronjs.org/docs/latest/api/shell#shellopenpathpath

jonvanausdeln:
Verified on
2022.11.0-71
Windows 11
MacOS 12.4

Verified on both RMarkdown and Quarto documents.

PR #11557, opened 2022-07-05
Open docx files rendered by rstudio -- electron

jgutman:
### Intent

Addresses #10897 

### Approach

Open the file using the default application for .docx files (Word/Pages/LibreOffice whatever)
Does not close the most recently rendered file, so if rendering multiple times, user will get multiple windows and have to manage the windows themselves

### Automated Tests

None

### QA Notes

See issue
### Checklist

### DIFF
--- a/src/node/desktop/src/main/gwt-callback.ts
+++ b/src/node/desktop/src/main/gwt-callback.ts
@@ -358,7 +358,10 @@ export class GwtCallback extends EventEmitter {
     });
 
     ipcMain.on('desktop_show_word_doc', (event, wordDoc: string) => {
-      GwtCallback.unimpl('desktop_show_word_doc');
+      shell.openPath(resolveAliasedPath(normalizeSeparatorsNative(wordDoc))).catch((value) => {
+        console.log('error:', value);
+        logger().logErrorMessage(value);
+      });
     });
 
     ipcMain.on('desktop_show_ppt_presentation', (event, pptDoc: string) => {
@@ -370,7 +373,8 @@ export class GwtCallback extends EventEmitter {
     });
 
     ipcMain.on('desktop_prepare_show_word_doc', () => {
-      GwtCallback.unimpl('desktop_prepare_show_word_doc');
+      // if possible, close most recently rendered docx item
+      return '';
     });
 
     ipcMain.on('desktop_prepare_show_ppt_presentation', () => {


src/node/desktop/src/main/gwt-callback.ts
```diff
@@ -358,7 +358,10 @@ export class GwtCallback extends EventEmitter {
     });
 
     ipcMain.on('desktop_show_word_doc', (event, wordDoc: string) => {
-      GwtCallback.unimpl('desktop_show_word_doc');
+      shell.openPath(resolveAliasedPath(normalizeSeparatorsNative(wordDoc))).catch((value) => {
+        console.log('error:', value);
+        logger().logErrorMessage(value);
```
timtmok, line 363: The Qt code falls back to showing the file in the View pane.

I think we can also add this change to the NEWS since it used to only open in the View pane on non-Windows platforms. I'm not sure why this wasn't done before since it should work on all platforms that have an application that can read docx.

timtmok (changes requested):

src/node/desktop/src/main/gwt-callback.ts
jgutman: where are you seeing in the Qt code that it opened in the view pane on failure? I just see it calling the showFile method which just does 
```
void GwtCallback::showFile(QString path)
{
   if (path.isNull() || path.isEmpty())
      return;

   path = desktop::resolveAliasedPath(path);

   desktop::openUrl(QUrl::fromLocalFile(path));
}

void GwtCallback::showWordDoc(QString path)
{
#ifdef Q_OS_WIN32

   path = desktop::resolveAliasedPath(path);
   Error error = wordViewer_.showItem(path.toStdWString());
   if (error)
   {
      LOG_ERROR(error);
      showFile(path);
   }

#else
   // Invoke default viewer on other platforms
   showFile(path);
#endif
}
```
Not clear to me where it's falling back to opening in the viewer pane (can we even open a .docx file in the view pane?)

timtmok: I think you can use either the secondary or satellite window to do this. You might want to check how popping out an editor works to confirm this.

jgutman: sorry, that's really not clear to me -- I'm not following where this is done in the original Qt code, where there doesn't seem to be any kind of fallback for non-windows platforms, and there doesn't seem to be any way to open these kinds of files in the viewer pane currently. Opening the file through the File -> Open File menu or through the File pane will always open an external application. Maybe I'm missing what you're saying

timtmok (approved):

Issue #11340, opened 2022-06-02
Open PPT presentations in external app (Electron)


gtritchie:
### System details

    RStudio Edition : Electron Desktop
    RStudio Version : 2022.06.0-daily+468
    OS Version      : Any
    R Version       : Any

### Steps to reproduce the problem

- install Microsoft PowerPoint (or another app registered to load PPTX files)
- create a Quarto Presentation, with PowerPoint as the output format
- click on the Render button

![Screen Shot 2022-06-02 at 11 25 42 AM](https://user-images.githubusercontent.com/10569626/171688611-19c77b94-c6a0-40bc-83c8-07ea8c87cbeb.png)

### Describe the problem in detail

- the PowerPoint file is created, but not automatically opened

### Describe the behavior you expected

- PowerPoint (or whatever is registered for the filetype) should open the create file

### Background

Similar to #10897; we need to implement various callbacks (for each platform):

Qt implementation:
https://github.com/rstudio/rstudio/blob/c532ae20fd27a48948c6c7bf05f20c7975d07262/src/cpp/desktop/DesktopGwtCallback.cpp#L620-L629

https://github.com/rstudio/rstudio/blob/c532ae20fd27a48948c6c7bf05f20c7975d07262/src/cpp/desktop/DesktopGwtCallback.cpp#L579-L593

https://github.com/rstudio/rstudio/blob/c532ae20fd27a48948c6c7bf05f20c7975d07262/src/cpp/desktop/DesktopGwtCallbackMac.mm#L379-L389

https://github.com/rstudio/rstudio/blob/c532ae20fd27a48948c6c7bf05f20c7975d07262/src/cpp/desktop/DesktopGwtCallbackMac.mm#L63-L80

Electron NYI placeholders:

https://github.com/rstudio/rstudio/blob/c532ae20fd27a48948c6c7bf05f20c7975d07262/src/node/desktop/src/main/gwt-callback.ts#L359-L361

https://github.com/rstudio/rstudio/blob/c532ae20fd27a48948c6c7bf05f20c7975d07262/src/node/desktop/src/main/gwt-callback.ts#L371-L373

- [ ] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ ] I have installed the latest version of RStudio, and confirmed that the issue still persists.
[...]

PR #11558, opened 2022-07-05
open pptx files rendered by rstudio

jgutman:
### Intent

Addresses #11340 

### Approach

same as #11557 
### Checklist

### DIFF
--- a/src/node/desktop/src/main/gwt-callback.ts
+++ b/src/node/desktop/src/main/gwt-callback.ts
@@ -362,7 +362,10 @@ export class GwtCallback extends EventEmitter {
     });
 
     ipcMain.on('desktop_show_ppt_presentation', (event, pptDoc: string) => {
-      GwtCallback.unimpl('desktop_show_ppt_presentation');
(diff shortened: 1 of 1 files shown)


src/node/desktop/src/main/gwt-callback.ts
```diff
@@ -362,7 +362,10 @@ export class GwtCallback extends EventEmitter {
     });
 
     ipcMain.on('desktop_show_ppt_presentation', (event, pptDoc: string) => {
-      GwtCallback.unimpl('desktop_show_ppt_presentation');
+      shell.openPath(resolveAliasedPath(normalizeSeparatorsNative(pptDoc))).catch((value) => {
```
timtmok, line 365: Same here where we can show it in the View pane if there's an error.

timtmok (changes requested):

src/node/desktop/src/main/gwt-callback.ts
timtmok: Looks like it's fine here after the discussion with opening docx files. There's no need for a fallback.

Issue #3650, opened 2018-10-11
Problem Installing RStudio-Server on Alpine-Linux


MattZhao:
### System details

    RStudio Edition : server
    RStudio Version : v1.1.456
    OS Version      : Alpine-Linux 3.8
    R Version       : 3.5.1

### Steps to reproduce the problem
I am trying to build RStudio-Server on Alpine-Linux on Docker. I am using python:3.7.0-alpine3.8 from docker hub as base image and installed glibc with instruction from https://github.com/sgerrand/alpine-pkg-glibc.
Then I followed the official instruction to build rstudio-server, which fails because it's not able to find PAM, although I have explicitly installed linux-pam, as you can see in the console output.

(4/7) Installing libuv (1.20.2-r0)
(5/7) Installing cmake (3.11.1-r2)
(6/7) Installing cmake-bash-completion (3.11.1-r2)
(7/7) Installing linux-pam (1.3.0-r0)
Executing busybox-1.28.4-r1.trigger
OK: 298 MiB in 94 packages
-- The C compiler identification is GNU 6.4.0
-- The CXX compiler identification is GNU 6.4.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Boost version: 1.63.0
-- Using RStudio-provided Boost 1.63.0
-- Performing Test HAVE_SCANDIR_POSIX
-- Performing Test HAVE_SCANDIR_POSIX - Success
-- Looking for SA_NOCLDWAIT
-- Looking for SA_NOCLDWAIT - found
-- Looking for SO_PEERCRED
-- Looking for SO_PEERCRED - found
-- Looking for inotify_init1
-- Looking for inotify_init1 - found
-- Looking for getpeereid
-- Looking for getpeereid - not found
-- Looking for setresuid
-- Looking for setresuid - found
CMake Error at cmake/modules/FindPAM.cmake:72 (message):
  PAM was not found
Call Stack (most recent call first):
  src/cpp/core/CMakeLists.txt:205 (find_package)

[...]

PR #11572, opened 2022-07-07
Check for group_member() and execinfo.h via CMake

atheriel:
### Intent

Folks trying to build RStudio on FreeBSD and Musl-based distributions like Alpine or Void Linux both encounter two assumptions that we make:

1. `group_member()` is available on non-macOS platforms. But `group_member()` is a Glibc extension, so this is not actually true.

2. `backtrace()` and `backtrace_symbols()` from `execinfo.h` are always provided by the system's libc. But they *happen* to be provided by Glibc and macOS's libc; on FreeBSD these are in a separate library, and Musl doesn't implement them at all.

### Approach

Both of these can be fixed by adding more careful checks for these features in our CMake configuration. I've also found numerous other major projects that have near-identical CMake checks for these two features; it's clearly a common problem.

This commit also adds a fallback implementation of `group_member()` rather than simply skipping the check (as currently happens on macOS) because this should improve correctness. However, this has an important consequence that this check will now happen on macOS, where previously it did not. I'm happy to remove this fallback if it is not regarded as necessary for correct behaviour.

Note that this PR *might* fully address the Alpine build errors discussed in #3650.

Also note that the community-based [FreeBSD](https://cgit.freebsd.org/ports/tree/devel/RStudio/files?id=90225d0c1d8cca043672180d39585462c45558b0) and [Void Linux ports](https://github.com/void-linux/void-packages/tree/master/srcpkgs/rstudio/patches) currently carry patches similar to (though less complete than) this one; hopefully this change will require them to carry fewer patches in the future.

### Automated Tests

All of these changes should be adequately tested by "it still complies on Linux and macOS", in my understanding.

### QA Notes

Please advise.

### Checklist

### DIFF
--- a/src/cpp/core/r_util/RSessionContext.cpp
+++ b/src/cpp/core/r_util/RSessionContext.cpp
@@ -48,6 +48,10 @@
 
 #include "config.h"
 
+#if !defined(HAVE_GROUP_MEMBER) && !defined(_WIN32)
+#include <limits.h> // for NGROUPS_MAX
+#endif
+
 // must be included after config.h for RSTUDIO_SERVER define
 #include <core/system/UserObfuscation.hpp>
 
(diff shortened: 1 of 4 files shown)


src/cpp/core/r_util/RSessionContext.cpp
```diff
@@ -207,12 +211,23 @@ bool isSharedPath(const std::string& projectPath,
...
       // (note that this checks supplementary group IDs only, so the check
       // against the primary group ID above is still required)
+#ifdef HAVE_GROUP_MEMBER
       if (::group_member(st.st_gid))
          return false;
+#else
+      // this is basically what glibc, gnulib, and glibcompat do to implement
+      // group_member()
+      gid_t groups[NGROUPS_MAX];
+      int ngroups = ::getgroups(NGROUPS_MAX, groups);
```
kevinushey, line 224: Should we check for and log errors here (if `::getgroups()` returns `-1`)?

kevinushey: (it seems like that should never happen but better safe than sorry IMHO)

atheriel: Yep, makes sense.

kevinushey (approved):
LGTM!

PR #11589, opened 2022-07-08
Make `libpam` optional in the core libraries

atheriel:
### Intent

Launcher does not use any `libpam` features from `server_core` and [it complicates the build and CI process](https://github.com/rstudio/launcher/issues/380) to require it unconditionally. This PR allows Launcher to opt out of using `libpam` (and in fact that is the default, so no changes should be required on their end once the shared libraries are synced).

### Approach

This conditional include is similar to the one carved out for the database/SOCI libraries in d07ca26d978649db0b5de2312cfd4a81ae2b8091, which Launcher *also* does not use.

### Automated Tests

This change does not introduce any new unit tests (it actually looks like there are no tests for the `libpam` support at all), but I did confirm locally that you can build with and without `-DRSTUDIO_USE_PAM=0`:

```
$ make rstudio-core-tests rstudio-server-core-tests
```

### QA Notes

This change *should* have no effect for this project and errors should appear at build time, so the risk seems low.

### Checklist


kfeinauer (approved):
LGTM!

src/cpp/CMakeLists.txt
```diff
@@ -604,6 +604,12 @@ elseif(RSTUDIO_CONFIG_SESSION_ONLY)
    message(STATUS, "Configured to build SESSION_ONLY")
 endif()
 
+# the core libs can be built without libpam (for Launcher, originally), but we
+# use it by default
+if (NOT DEFINED RSTUDIO_USE_PAM)
+  message(STATUS "Configured to build with libpam features")
```
kevinushey, line 610: Minor nit: this check should happen outside of the `NOT DEFINED` block here just so we get a status report when one explicitly sets `cmake -DRSTUDIO_USE_PAM`.

kevinushey (approved):

Issue #11567, opened 2022-07-06
Electron loads wrong/unexpected version of R with Brew


jgutman:
### System details

    RStudio Edition : Desktop 
    RStudio Version : 2022.10.0 Electron any
    OS Version      :  Mac OS 12.4 M1
    R Version       : any

### Steps to reproduce the problem
1. Install R with `brew`
2. Set path to point to non-brew R
3. Set `RSTUDIO_WHICH_R` environment variable in terminal to point to non-brew R
4. Launch RStudio Electron
5. Run `R.home()` and check which R version is being loaded

### Describe the problem in detail

Electron seems to point to the wrong version of R, even if there is another R higher on the path, even if there is a symlink to another R in `/usr/local/bin/R`, even if RSTUDIO_WHICH_R points to another version of R. The wrong R seems to be chosen

### Describe the behavior you expected

1. Respect the RSTUDIO_WHICH_R env var even when not launched from a terminal
2. Prefer the R that comes first in the user's path if RSTUDIO_WHICH_R is not set

- [ ] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ ] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [ ] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

jgutman:
It would be nice if we could get RSTUDIO_WHICH_R and which R actually returned the same values as user's default shell, which is not currently the case. But this PR at least aligns electron behavior with former Qt behavior on macs

jonvanausdeln:
Verified on:
2022.11.0-75
MacOS 12.4

Verified R version. Also noted that issue with env variable changes in zsh are still not picked as noted in the PR fix.

PR #11607, opened 2022-07-12
Copy behavior from Qt desktop for Mac -- check fallback list of R

jgutman:
### Intent

Addresses #11567 to at least mimic the old Qt desktop behavior for Macs

### Approach

Qt desktop code took [1 approach for Mac](https://github.com/rstudio/rstudio/blob/8df7e8c7c6b7727779786613a9089ba14e0c4a68/src/cpp/core/r_util/REnvironmentPosix.cpp#L106-L147) and a[ different approach for Linux](https://github.com/rstudio/rstudio/blob/8df7e8c7c6b7727779786613a9089ba14e0c4a68/src/cpp/core/r_util/REnvironmentPosix.cpp#L203-L244) in terms of finding the right R executable. Electron code was essentially applying the Linux logic to both platforms, and then falling back to the Mac logic, leading to funny behavior on Mac. 

Overall, it would probably be ideal to launch a terminal process in the user's default shell to pick up any profile modifications to the RSTUDIO_WHICH_R and/or PATH variables, since we check those variables as part of detecting the right R. But for now, we don't pick up any user modifications to those variables, so if they are set in bash/zsh profile, changes will be ignored. 

This change doesn't fix that issue, just brings back in line with the former Qt behavior

### QA Notes

See #11567 for details -- probably need to test on Mac with brew installed R on the system
### Checklist

### DIFF
--- a/src/node/desktop/src/main/detect-r.ts
+++ b/src/node/desktop/src/main/detect-r.ts
@@ -278,27 +278,28 @@ function scanForR(): Expected<string> {
 }
 
 function scanForRPosix(): Expected<string> {
-  // first, look for R on the PATH
+  if (process.platform == 'darwin') {
+    // For Mac, we want to look in a list of hard-coded locations
+    const defaultLocations = ['/opt/local/bin/R', '/usr/local/bin/R', '/usr/bin/R'];
+    // also check framework directory and homebrew ARM locations for macOS
+    defaultLocations.push('/Library/Frameworks/R.framework/Resources/bin/R');
+    defaultLocations.push('/opt/homebrew/bin/R');
+
+    for (const location of defaultLocations) {
+      if (isValidBinary(location)) {
+        logger().logDebug(`Using ${location} (found by searching known locations)`);
+        return ok(location);
+      }
+    }
+  }
+
+  // look for R on the PATH
+  // should we launch the default shell to pick up the user modifications to the path?
   const [rLocation, error] = executeCommand('/usr/bin/which R');
   if (!error && rLocation) {
+    logger().logDebug(`PATH: ${getenv("PATH")}`);
     logger().logDebug(`Using ${rLocation} (found by /usr/bin/which/R)`);
     return ok(rLocation);
-  }
-
-  // otherwise, look in some hard-coded locations
-  const defaultLocations = ['/opt/local/bin/R', '/usr/local/bin/R', '/usr/bin/R'];
-
-  // also check framework directory and homebrew ARM locations for macOS
-  if (process.platform === 'darwin') {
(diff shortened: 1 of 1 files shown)


src/node/desktop/src/main/detect-r.ts
```diff
@@ -278,29 +278,30 @@ function scanForR(): Expected<string> {
 }
 
 function scanForRPosix(): Expected<string> {
-  // first, look for R on the PATH
+  if (process.platform == 'darwin') {
+    // For Mac, we want to look in a list of hard-coded locations
+    const defaultLocations = ['/opt/local/bin/R', '/usr/local/bin/R', '/usr/bin/R'];
```
timtmok, line 283: https://github.com/rstudio/rstudio/blob/8df7e8c7c6b7727779786613a9089ba14e0c4a68/src/cpp/core/r_util/REnvironmentPosix.cpp#L109-L115

Qt seems to use this order:
- /usr/bin/R
- /usr/local/bin/R
- /opt/local/bin/R
- /opt/homebrew/bin/R
- /Library/Frameworks/R.framework/Resources/bin/R

src/node/desktop/src/main/detect-r.ts
```diff
@@ -278,29 +278,30 @@ function scanForR(): Expected<string> {
...
+  // should we launch the default shell to pick up the user modifications to the path?
   const [rLocation, error] = executeCommand('/usr/bin/which R');
   if (!error && rLocation) {
+    logger().logDebug(`PATH: ${getenv("PATH")}`);
     logger().logDebug(`Using ${rLocation} (found by /usr/bin/which/R)`);
     return ok(rLocation);
   }
 
-  // otherwise, look in some hard-coded locations
-  const defaultLocations = ['/opt/local/bin/R', '/usr/local/bin/R', '/usr/bin/R'];
```
timtmok, line 289: https://github.com/rstudio/rstudio/blob/8df7e8c7c6b7727779786613a9089ba14e0c4a68/src/cpp/core/r_util/REnvironmentPosix.cpp#L228-L229

Removing this means that Linux has no fallback locations to check.

The priority order is definitely a bit weird. Linux prefers using `which R` first with a fallback location list and Mac only uses a defined location list.

I think making the location list match the platforms in Qt then adding a platform check for whether to use `which R` will give the same behaviour.

src/node/desktop/src/main/detect-r.ts
jgutman: Yeah, but I think putting homebrew R before Cran R is a big mistake for the majority of our users and is just not a good idea, especially with the M1 macs

timtmok (approved):

PR #11608, opened 2022-07-12
Implement project name in dock tile

timtmok:
### Intent
Address #11386

### Approach
Port the native code that redraws the application icon with the project name overlaid. Add in stubs for the other platforms.

I think we can implement this for Windows and Linux since there's support to set an icon: https://www.electronjs.org/docs/latest/api/browser-window#winseticonicon-windows-linux

It wouldn't need native code either! (still needs it for Mac since that API is supported for Windows and Linux only)

image

I've also reorganized the native code folder in case we add more modules. The type definitions go in the root and the actual code are in their respective folders.

### Automated Tests
None

### QA Notes
This is the same code as Qt so the look should be exactly the same.

### Checklist


src/node/desktop/src/main/gwt-callback.ts
```diff
@@ -544,6 +544,7 @@ export class GwtCallback extends EventEmitter {
 
     ipcMain.on('desktop_bring_main_frame_to_front', () => {
       this.mainWindow.window.focus();
+      console.debug('desktop_bring_main_frame_to_front');
```
gtritchie, line 547: Did you mean to leave this in?

src/node/desktop/src/native/dock/dock_mac.mm
```diff
@@ -0,0 +1,320 @@
...
+#define RS_EXPORT_FUNCTION(__NAME__, __FUNCTION__) \
+  exports.Set(                                     \
+    Napi::String::New(env, __NAME__),              \
+    Napi::Function::New(env, __FUNCTION__)         \
+  )
+
+// Implementation based on https://github.com/jessesquires/JSCustomBadge, and
+// https://cgit.kde.org/kdevplatform.git/commit/?id=c0626300e1c18448dd653a3484fd1ca6298e2d8c
+
+@interface MacDockTileView : NSView {
```
gtritchie, line 32: Ah, I have a soft spot for Objective-C. Was doing NeXTSTEP programming in it long, long ago.

gtritchie (approved):
LGTM! One comment about what looks like debugging code.

src/node/desktop/src/main/gwt-callback.ts
timtmok: I'll clean this up. I got side tracked into trying to reproduce another bug!

Issue #11581, opened 2022-07-08
"Select all" with CMD+A in Find/Replace menu doesn't work with latest Electron Elsbeth build


DavisVaughan:
If you open the find/replace menu and type in something to "find", then CMD+A won't highlight the whole word anymore.

I'm doing it in the video, you just can't see it (because nothing happens):

https://user-images.githubusercontent.com/19150088/178024164-39e4dde3-7f50-4165-87e8-2dc1922c2272.mov

RStudio 2022.10.0-daily+55 "Elsbeth Geranium" Daily (f895778c3e83e20672af1434f040a31b8f2cf906, 2022-07-07) for macOS
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) RStudio/2022.10.0-daily+55 Chrome/102.0.5005.115 Electron/19.0.6 Safari/537.36

DavisVaughan:
It also doesn't work correctly in "Go to File/Function", i.e. the menu you get with `Control + .` on a Mac

- `Control + .`
- Type in a function name
- Try to select all with `Cmd + A`

DavisVaughan:
It also doesn't work in `Edit -> Find in Files`. The `Find:` box doesn't work with `Cmd + A` anymore.

timtmok:
I've only been able to reproduce this so far on Mac. Linux and Windows work.

Something else seems to be responding to the keyboard event and preventing the select all.

timtmok:
I'm still not sure of the root cause but I've got a workaround that replicates Qt behaviour.

gtritchie:
I'll take a look at the PR by EOD (Edmonton time).

jonvanausdeln:
Verified on:
2022.11.0-71
MacOS 12.4
Win11
Ubuntu 22.04

Select all works in the input fields in all the mentioned scenarios. Note, the 'wierd' behavior of hitting cmd+a while the 'next' button has focus still exists as it does in the QT version.

kevinushey:
I reverted the PR in https://github.com/rstudio/rstudio/commit/94b86a9bd5821e293f17c1191b3846bae04d85f2; it caused issues with RStudio stealing Cmd + A even when other applications had focus.

timtmok:
Throwing this to the backlog for now. I don't anticipate a fix for EG unless there's more info on why the editor doesn't get or respond to the hotkey event.

[...]

PR #11621, opened 2022-07-14
Fix select all in text input fields

timtmok:
### Intent
Address #11581

### Approach
Implements a workaround to call Electron's `selectAll()` on the focused web view. I'm not sure what is overriding the Cmd+A but it eats the keyboard event. This seems to get around that and results in the same behaviour as Qt (even down to pressing Tab to focus on a button and it selects UI text). We could refine this behaviour by executing Javascript and querying the `document.activeElement` for the active element but this seems cleaner and easier to understand.

I've refactored the wrapper method on `getFocusedWebContents()` since it could be useful in other places.

### Automated Tests
None

### QA Notes
Any text input field would have been affected by this bug. It should also work on editors that have been popped out to its own window.

### Checklist

### DIFF
--- a/src/node/desktop/src/main/application.ts
+++ b/src/node/desktop/src/main/application.ts
@@ -13,7 +13,7 @@
  *
  */
 
-import { app, BrowserWindow, Menu, screen, WebContents } from 'electron';
+import { app, BrowserWindow, globalShortcut, Menu, screen, WebContents } from 'electron';
 import i18next from 'i18next';
 import path from 'path';
 import { getenv, setenv } from '../core/environment';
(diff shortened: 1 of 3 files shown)


gtritchie (approved):
LGTM!

I took a quick look and also couldn't figure out what's apparently eating Cmd+A only on Mac Electron. There is an Electron menuItem.role = 'selectAll' for this, but we don't do anything with that except when creating a right-click context menu [here](https://github.com/rstudio/rstudio/blob/3a87a1ff21a9b53f8c9d0297249ad2339b0c063f/src/node/desktop/src/main/context-menu.ts#L166-L170), and that seems unrelated. Also, problem doesn't happen when I created a simple test page in Electron-Fiddle (Cmd+A works fine in text input fields).

PR #11691, opened 2022-08-02
Allow to find system SOCI_* libraries

Flamefire:
### Intent

There may be system installed SOCI libraries in non-standard locations (e.g in /opt) which can be found by using CMakes `find_library` but currently are not due to the `NO_DEFAULT_PATH` option.
This makes installing this more complicated than it needs to.

### Approach

Add a second `find_library` call which will find the SOCI libraries in default locations if it hasn't been found already in `SOCI_LIBRARY_DIR`


src/cpp/CMakeLists.txt
```diff
@@ -467,12 +467,16 @@ endif()
...
+   if(NOT APPLE AND NOT RSTUDIO_USE_SYSTEM_SOCI)
+      find_library(SOCI_CORE_LIB NAMES "libsoci_core.a" "soci_core" PATHS "${SOCI_LIBRARY_DIR}" NO_DEFAULT_PATH)
+      find_library(SOCI_SQLITE_LIB NAMES "libsoci_sqlite3.a" "soci_sqlite3" PATHS "${SOCI_LIBRARY_DIR}" NO_DEFAULT_PATH)
+      find_library(SOCI_POSTGRESQL_LIB NAMES "libsoci_postgresql.a" "soci_postgresql" PATHS "${SOCI_LIBRARY_DIR}" NO_DEFAULT_PATH)
    endif()
-   find_library(SOCI_CORE_LIB NAMES "libsoci_core.a" "soci_core" PATHS "${SOCI_LIBRARY_DIR}" NO_DEFAULT_PATH)
-   find_library(SOCI_SQLITE_LIB NAMES "libsoci_sqlite3.a" "soci_sqlite3" PATHS "${SOCI_LIBRARY_DIR}" NO_DEFAULT_PATH)
-   find_library(SOCI_POSTGRESQL_LIB NAMES "libsoci_postgresql.a" "soci_postgresql" PATHS "${SOCI_LIBRARY_DIR}" NO_DEFAULT_PATH)
+   # If not found (or not searched) search now again in default paths (including user set paths)
+   find_library(SOCI_CORE_LIB NAMES "libsoci_core.a" "soci_core")
```
kevinushey, line 477: I don't think we want this fallback when `RSTUDIO_USE_SYSTEM_SOCI` is set; in that case, only the above branch should be running.

Flamefire: It actually is a common CMake pattern to do a 2-stage-find:
```
find_*(FOO ... PATHS ... NO_DEFAULT_PATH)
find_*(FOO ...)
```
This allows to search some special paths first before considering the default paths which to me makes sense.

If you still want an if-else with 1 find call (per lib) in each branch I'll change that.

kevinushey: Trying to think out loud:

1. If `RSTUDIO_USE_SYSTEM_SOCI` is set and true, then we should use the system copy of SOCI (not the bundled version);
2. If `RSTUDIO_USE_SYSTEM_SOCI` is not set, then we should only use the bundled copy of SOCI (not a version that we find on the PATH / other system install)

If I understand correctly, with this PR, in the second case, we might fall back to using the system copy of SOCI if the bundled version wasn't available, and I don't think we want that behavior here. Does that make sense?

Flamefire: > If I understand correctly, with this PR, in the second case, we might fall back to using the system copy of SOCI if the bundled version wasn't available, and I don't think we want that behavior here.

Correct. To me this made sense as you want the install to work. If `RSTUDIO_USE_SYSTEM_SOCI` is set you (only) want the system soci, not the bundled copy. If it is not set you likely don't care. But as this also includes the case where it is set to false I did change it, so it either searches the system or the bundled directory.

So with the recent change it now behaves as you describe.

kevinushey:
LGTM; thanks!

PR #11702, opened 2022-08-04
Revert "avoid getClientLeft() / getClientTop()"

jgutman:
Reverts rstudio/rstudio#11640

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/dom/ElementEx.java
+++ b/src/gwt/src/org/rstudio/core/client/dom/ElementEx.java
@@ -46,38 +46,35 @@ public class ElementEx extends Element
       return (result == null) ? '' : result + '';
    }-*/;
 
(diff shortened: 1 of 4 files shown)


jgutman:
Weirdly, this now causes the error it was attempting to fix. And reverting it seemed to fix the error. But no idea why what worked before doesn't work now and vice versa

kevinushey:
That's weird. Best guess; something about how devmode rebuilds things is causing issues here? That is, the issue doesn't show up in "clean" builds but might show up after using devmode to build / rebuild the IDE over a few sessions?

Do we have an open issue tracking a better fix for this? My guess is that the bug will just re-surface in the future unless we figure out what's going on.

jgutman:
No, I haven't filed an issue yet. I basically was always using `ant desktop` and never `ant devmode` and we tried re-running after doing an `ant clean` and still needed to revert the changes to get it to work. couldn't figure out what was going on exactly

Issue #11251, opened 2022-05-19
Problems opening files from Explorer/Finder/etc. (Electron)


gtritchie:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2022.06.0-daily+421
    OS Version      : Any
    R Version       : Any

### Overview

Various scenarios opening files in RStudio from the operating system: goal is for Electron to behave the same as Qt on Mac, Linux, and Windows.

#### Steps to reproduce the problem

- ensure RStudio is not running
- double-click on an `*.RProj` file

#### Describe the problem in detail

- working correctly with Electron (including for completeness)
- Mac: ✅
- Linux: ✅
- Windows: ✅

#### Describe the behavior you expected

- the IDE starts and loads the project

#### Steps to reproduce the problem

- ensure RStudio is running
- double-click on an `.RProj` file

#### Describe the problem in detail

- working correctly with Electron (including for completeness)
- Mac: ✅
- Linux: ✅
- Windows: ✅

#### Describe the behavior you expected

- a new instance of the RStudio starts and loads the project
- this happens even if the project is already open in another instance
- in Desktop Pro (once implemented for Electron), also test `.rdprsp` files behave the same way and trigger a new instance

#### Steps to reproduce the problem

- start RStudio and make sure it doesn't have the file open that you're going to double-click on in next step
- close RStudio
- double-click on a file associated with RStudio such as a `*.R`, `*.Rmd`, etc.; might have to right-click and use "Open With..." depending on the file type and what other programs are installed

#### Describe the problem in detail

- Mac: ✅
- Linux: ✅
- Windows: ✅ (tracked separately via #11369)
   - RStudio starts, the editor region displays (blank), and an error shows "Error while opening file" : "The system cannot find the file specified."
   - After closing the error message dialog, the blank editor region remains

![windows-file-open-error](https://user-images.githubusercontent.com/10569626/169410649-21ae2ed7-52a8-4d77-8c98-3ca3c23f384f.png)

#### Describe the behavior you expected

- the IDE starts and loads the file, even on an obscure platform like Microsoft Windows

#### Steps to reproduce the problem

- start RStudio and make sure it doesn't have the file open that you're going to double-click on in next step
- keep RStudio running
- double-click on a file associated with RStudio such as a `*.R`, `*.Rmd`, etc.; might have to right-click and use "Open With..." depending on the file type and what other programs are installed

#### Describe the problem in detail

- Mac: ✅
- Linux: ❌ 
[...]

PR #11703, opened 2022-08-04
Electron instance handling with file open events

timtmok:
### Intent
Implement single instance logic when opening a file at startup.
Addresses part of #11251 

### Approach
Before attempting to open a file at startup, request the instance lock. If obtained, the instance is the primary instance and should open the file. If not, it exits and the request would send an event to the primary instance that will open the file.

This brings in `net-ipc` to communicate locally among the instances using a Unix/Windows socket. This notifies other instances when the primary instance exits and allows one of the other instances to request the lock. It will then take over opening files that come from `Open With`.

If we do need more IPC events, this will provide everything that's necessary to communicate with the primary instance.

This may or may not fix it on Linux but I would expect it to work.

### Automated Tests
None

### QA Notes
* `rstudio.exe --log-level=DEBUG <some_path_to_a_file_to_open>` should open the file and activate an existing instance.
* if the file is a project then it will always open a new instance
* debug messages have been added to confirm the primary instance handover so check `~/AppData/Local/RStudio/log` (desktop log files are rotated by day)
   * the `net-ipc` messages are the instance handling messages
   * almost all of the `net-ipc` messages include the process id
   * it should log the process id that is the primary instance
   * secondary instances should log the process id and that it connected to the primary instance
   * closing the primary instance should show logs that the other instances disconnected and one of them took over as primary
   * NOTE: sometimes the logs for secondary instances don't flush to file because the logger we use doesn't really handle that very well

### Checklist

### DIFF
--- a/src/node/desktop/package.json
+++ b/src/node/desktop/package.json
@@ -71,6 +71,7 @@
     "i18next": "21.8.13",
     "line-reader": "0.4.0",
     "lodash.debounce": "4.0.8",
+    "net-ipc": "2.0.1",
     "node-addon-api": "5.0.0",
     "node-system-fonts": "1.0.1",
     "properties-reader": "2.2.0",
--- a/NOTICE
+++ b/NOTICE
@@ -68,6 +68,7 @@ used by these components are included below):
 - properties-reader (npm)
 - json-schema-to-typescript (npm)
 - lodash.debounce (npm)
+- net-ipc (npm)
 
 Windows builds of RStudio also include a binary copy of libclang. This
 component is licensed to you under the Apache License v2.0 with LLVM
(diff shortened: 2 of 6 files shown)


src/node/desktop/package.json
```diff
@@ -71,6 +71,7 @@
     "i18next": "21.8.13",
     "line-reader": "0.4.0",
     "lodash.debounce": "4.0.8",
+    "net-ipc": "2.0.1",
```
gtritchie, line 74: Please add this to the NOTICE file. There should be an entry in the list near the top of the file, and the license details later in the file.

gtritchie (changes requested):

Issue #11392, opened 2022-06-08
Can't save Viewer pane contents as image (Electron)


gtritchie:
### System details

    RStudio Edition : Electron
    RStudio Version : All
    OS Version      : Any
    R Version       : Any

### Steps to reproduce the problem

- display something in the viewer page; for example, run this code:

```R
tempDir <- tempfile()
dir.create(tempDir)
htmlFile <- file.path(tempDir, "index.html")
writeLines(c("<html><body><p>Hello World</p></body></html>"), htmlFile)
viewer <- getOption("viewer")
viewer(htmlFile)
```

- click on Export / Save As Image... in Viewer pane toolbar

![Screen Shot 2022-06-08 at 4 50 47 PM](https://user-images.githubusercontent.com/10569626/172730489-2cc0942a-9999-47d1-bf63-98b73381cc68.png)

- click Save in the dialog

![Screen Shot 2022-06-08 at 4 52 56 PM](https://user-images.githubusercontent.com/10569626/172730660-94b3598f-0798-486c-a0ec-589cda4c27c7.png)

### Describe the problem in detail

- the dialog dismisses but the file isn't saved

### Describe the behavior you expected

- the file is created; if you had also clicked "View plot after saving" checkbox in that dialog the file would then open in the default application for that file format

### Implementation details

Need to implement this callback: https://github.com/rstudio/rstudio/blob/e80e1a57761c4f52eb9aa72b3ca3f60b82037efc/src/node/desktop/src/main/gwt-callback.ts#L516-L518

The Qt implemention is here: https://github.com/rstudio/rstudio/blob/e80e1a57761c4f52eb9aa72b3ca3f60b82037efc/src/cpp/desktop/DesktopGwtCallback.cpp#L766-L789

- [ ] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ ] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [ ] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

jgutman:
Note, in Electron 2022.10.0.69, save as image works correctly with actual plots, but does not appear to work with text/html examples as given above. 

```
library(ggplot2)
[...]

PR #11704, opened 2022-08-04
Viewer pane save as image - Electron

jgutman:
Addresses #11392 

follow up to #11705, can be merged first (this PR contains those changes as well)

Change supported file formats for viewer export to remove TIFF export option on Macs, keep BMP option for non Macs
Implement viewer callback in electron native image API

### DIFF
--- a/src/cpp/session/modules/viewer/SessionViewer.cpp
+++ b/src/cpp/session/modules/viewer/SessionViewer.cpp
@@ -158,10 +158,7 @@ Error getViewerExportContext(const json::JsonRpcRequest& request,
    json::Array formats;
    formats.push_back(plotExportFormat("PNG", kPngFormat));
    formats.push_back(plotExportFormat("JPEG", kJpegFormat));
-   #ifndef __APPLE__
-      formats.push_back(plotExportFormat("BMP", kBmpFormat));
-   #endif // !__APPLE__
-   // TIFF and SVG formats not currently supported in Electron
+   // BMP, TIFF and SVG formats not currently supported in Electron
    contextJson["formats"] = formats;
 
    // get directory path -- if it doesn't exist revert to the current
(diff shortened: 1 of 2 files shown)


src/node/desktop/src/main/gwt-callback.ts
```diff
@@ -338,31 +338,29 @@ export class GwtCallback extends EventEmitter {
       });
     });
 
-    ipcMain.on('desktop_show_file', (event, file: string) => {
-      shell.showItemInFolder(file);
-    });
-
-    ipcMain.on('desktop_show_word_doc', (event, wordDoc: string) => {
+    function showFileInSystemViewer(file: string) {
       shell.openPath(resolveAliasedPath(normalizeSeparatorsNative(wordDoc))).catch((value) => {
```
jgutman, line 342:
```suggestion
      shell.openPath(resolveAliasedPath(normalizeSeparatorsNative(file))).catch((value) => {
```

src/cpp/session/modules/viewer/SessionViewer.cpp
```diff
@@ -158,11 +158,10 @@ Error getViewerExportContext(const json::JsonRpcRequest& request,
    json::Array formats;
    formats.push_back(plotExportFormat("PNG", kPngFormat));
    formats.push_back(plotExportFormat("JPEG", kJpegFormat));
-#ifdef __APPLE__
-   formats.push_back(plotExportFormat("TIFF", kTiffFormat));
-#else
-   formats.push_back(plotExportFormat("BMP", kBmpFormat));
-#endif
+   #ifndef __APPLE__
+      formats.push_back(plotExportFormat("BMP", kBmpFormat));
```
jgutman, line 162: I'm not sure if the electron image.toBitmap() method really supports BMP files or not. They can't be opened with Preview on a Mac, but Mac did not previously support the BMP format from the viewer pane anyway. Can check if it works on Linux/Windows and remove support for BMP if not. PNG/JPEG should be sufficient here for most non-plot image exports anyway

timtmok: Looks like `image.toBitmap()` isn't what we really want. BMP support will likely require adding a package to handle this. Maybe https://www.npmjs.com/package/jimp will be useful here. So BMP is in the same boat as TIFF.

Probably best to leave it out for now and determine how best to proceed in another ticket.

jgutman: Okay, removed

PR #11705, opened 2022-08-04
consolidate callbacks, fix desktop_show_file logic

jgutman:
### Intent

desktop_show_file callback doesn't have same behavior as behavior. Used to open the file in a default system viewer, now it just opens a finder/explorer window with the file in question highlighted. This is generally not helpful. Blocks #11392

### DIFF
--- a/src/node/desktop/src/main/gwt-callback.ts
+++ b/src/node/desktop/src/main/gwt-callback.ts
@@ -338,31 +338,29 @@ export class GwtCallback extends EventEmitter {
       });
     });
 
-    ipcMain.on('desktop_show_file', (event, file: string) => {
-      shell.showItemInFolder(file);
-    });
-
-    ipcMain.on('desktop_show_word_doc', (event, wordDoc: string) => {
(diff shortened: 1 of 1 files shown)


src/node/desktop/src/main/gwt-callback.ts
```diff
@@ -338,31 +338,29 @@ export class GwtCallback extends EventEmitter {
       });
     });
 
-    ipcMain.on('desktop_show_file', (event, file: string) => {
-      shell.showItemInFolder(file);
-    });
-
-    ipcMain.on('desktop_show_word_doc', (event, wordDoc: string) => {
+    function showFileInSystemViewer(file: string) {
       shell.openPath(resolveAliasedPath(normalizeSeparatorsNative(wordDoc))).catch((value) => {
```
jgutman, line 342:
```suggestion
      shell.openPath(resolveAliasedPath(normalizeSeparatorsNative(file))).catch((value) => {
```

timtmok (approved):
LGTM!

Issue #11665, opened 2022-07-25
python chunks with empty lines aren't run 


kevinushey:
### System details

    RStudio Edition : 2022.10.0-daily+68
    RStudio Version : Desktop [Open Source]
    OS Version      : macOS Monterey 12.4
    R Version       : R version 4.2.1 (2022-06-23)

### Steps to reproduce the problem

Create an R Markdown document, with a chunk containing the following:

`````
```{python}
def example():
  print("I'm part of the function definition!")

  print("I'm also part of the function definition!")

example()
```
`````

Then, try to run it using the chunk Run button.

### Describe the problem in detail

The empty line "ends" the function definition, and so the second `print()` statement doesn't actually become part of the code.

```
>>> def example():
...   print("I'm part of the function definition!")
... 
>>>   print("I'm also part of the function definition!")
I'm also part of the function definition!
>>> 
>>> example()
I'm part of the function definition!
```

### Describe the behavior you expected

The code should be parsed and run as a single unit, so that the resulting code is:

```
>>> def example():
...   print("I'm part of the function definition!")
...   
...   print("I'm also part of the function definition!")
... 
>>> example()
I'm part of the function definition!
I'm also part of the function definition!
```

---

<details>

<summary>Session Info</summary>

``` r
R version 4.2.1 (2022-06-23)
Platform: aarch64-apple-darwin20 (64-bit)
Running under: macOS Monterey 12.4

Matrix products: default
LAPACK: /Library/Frameworks/R.framework/Versions/4.2-arm64/Resources/lib/libRlapack.dylib

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] renv_0.15.5-43

loaded via a namespace (and not attached):
 [1] Rcpp_1.0.9          here_1.0.1          lattice_0.20-45    
 [4] png_0.1-7           rprojroot_2.0.3     digest_0.6.29      
 [7] grid_4.2.1          jsonlite_1.8.0      evaluate_0.15      
[10] rlang_1.0.4         cli_3.3.0           rstudioapi_0.13    
[13] Matrix_1.4-1        reticulate_1.25     rmarkdown_2.14     
[16] tools_4.2.1         xfun_0.31           yaml_2.3.5         
[...]

PR #11724, opened 2022-08-08
Fix python chunk execution with blank lines

jgutman:
### Intent

Addresses #11665. 

### Approach

Instead of sending python chunks to the executor one line at a time, as we do with R chunks, send all lines at once. This will ensure the preprocessing that already happens for non markdown python code gets applied to python code in chunks, instead of applying the preprocessing one line at a time

image

Sends full function body to the console, and adds a newline to terminate the function definition

### Automated Tests

None

### QA Notes

See issue 
### Checklist

### DIFF
--- a/src/cpp/session/modules/rmarkdown/NotebookQueue.cpp
+++ b/src/cpp/session/modules/rmarkdown/NotebookQueue.cpp
@@ -283,7 +283,7 @@ private:
          execContext_->onExprComplete();
          
       ExecRange range;
-      std::string code = execUnit_->popExecRange(&range, mode);
+      std::string code = execUnit_->popExecRange(&range, mode, execContext_->engine());
       if (code.empty())
       {
          // no code to evaluate--skip this unit
--- a/src/cpp/session/SessionConsoleInput.cpp
+++ b/src/cpp/session/SessionConsoleInput.cpp
@@ -170,9 +170,9 @@ void fixupPendingConsoleInput()
       return;
    
    // if this has no newlines, then nothing to do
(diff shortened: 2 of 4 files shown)


src/cpp/session/modules/rmarkdown/NotebookQueue.cpp
```diff
@@ -283,7 +283,7 @@ class NotebookQueue : boost::noncopyable
          execContext_->onExprComplete();
          
       ExecRange range;
-      std::string code = execUnit_->popExecRange(&range, mode);
+      std::string code = execUnit_->popExecRange(&range, mode, execContext_->engine());
```
jgutman, line 286: passing the engine in as a param here so we don't have to parse it out from the code itself

jgutman:
The preprocessing function is already in thr code. The problem is the code gets sent to the preprocessing function line by line. Then it gets executed all at one time anyway. This sends the python code to the preprocessing function all at once so the preprocessing actually works. There's no need to change the preprocessing because it already works fine for non markdown chunks

jgutman:
So, this implementation works if you click the "Run chunk" button. The full chunk gets sent, and the blank line and following line get included in the function definition. However, if you put your cursor in the chunk, and try to run each line one by one with Cmd(Ctrl)+Enter, we still get the same behavior as before, where the function definition terminates with the blank line and last line gets left out of function. @kevinushey is this worth fixing/coming up with another solution for, or is it sufficient to have the Run Current Chunk/Run all chunks above buttons working as expected?

Issue #11721, opened 2022-08-08
(R 4.2) path.expand("~") resolves incorrectly for non-ASCII usernames on Windows


kevinushey:
### System details

    RStudio Edition : 2022.07.1+554
    RStudio Version : Desktop [Open Source]
    OS Version      : Windows 10 x64 (build 25174)
    R Version       : R version 4.2.1 (2022-06-23 ucrt)

### Steps to reproduce the problem

Create a user account containing non-ASCII (latin1) characters; e.g. Fürz. Then, try to launch RStudio with R 4.2.x.

### Describe the problem in detail

RStudio prints a bunch of warnings, e.g.

> In normalizePath(path.expand(path), winslash, mustWork) : path[1]="C:/Users/F�rz/Documents": The system cannot find the path specified

In addition, `path.expand("~")` gives:

```
> path.expand("~")
[1] "C:/Users/F\xfcrz/Documents"
```

which is evidently latin1-encoded, but not marked as such.

```
> path <- path.expand("~")
> Encoding(path)
[1] "unknown"
> Encoding(path) <- "latin1"
> path
[1] "C:/Users/Fürz/Documents"
```

### Describe the behavior you expected

`path.expand("~")` should produce a correctly-encoded path.

---

<details>

<summary>Session Info</summary>

``` r
R version 4.2.1 (2022-06-23 ucrt)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 25174)

Matrix products: default

locale:
[1] LC_COLLATE=English_United States.utf8 
[2] LC_CTYPE=English_United States.utf8   
[3] LC_MONETARY=English_United States.utf8
[4] LC_NUMERIC=C                          
[5] LC_TIME=English_United States.utf8    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

loaded via a namespace (and not attached):
[1] compiler_4.2.1 tools_4.2.1   
```

</details>

jgutman:
Assigning @kevinushey because you seem to already have a PR in progress, but if you want someone else to take over that's fine too

jonvanausdeln:
Verified on 2022.11.0-173
Windows 11

Tested with several different usernames containing special characters

PR #11726, opened 2022-08-09
keep user home / rhome as UTF-8 with R 4.2.x

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/11721.
May also address https://github.com/rstudio/rstudio/issues/11717.

### Approach

When using the UCRT runtime, environment variables are expected to be encoded as UTF-8, so don't try to convert from UTF-8 to the system encoding here.

### Automated Tests

N/A

### QA Notes

Test that, when using RStudio + R 4.2.x on Windows, no warnings are printed to the console on startup.

### Checklist

### DIFF
--- a/dependencies/windows/install-soci/install-soci.R
+++ b/dependencies/windows/install-soci/install-soci.R
@@ -41,6 +41,7 @@ postgresql_zip <- file.path(owd, "win-postgresql.zip")
 postgresql_zip_url <- "https://rstudio-buildtools.s3.amazonaws.com/win-postgresql.zip"
 
 downloadAndUnzip <- function(outputFile, extractDir, url) {
+   
    # download zip if we don't already have it
    if (!file.exists(outputFile)) {
       section("Downloading '%s' from '%s'", outputFile, url)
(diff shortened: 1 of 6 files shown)


kevinushey:
Some other notes...

`utf8ToSystem()` uses `wctombs()`, which relies on the current locale. Before R has been initialized, this locale has not yet been set to a UTF-8 locale, so setting to the native encoding implies setting (for example) some native code page, e.g. CP850 (ISO-8859-1; or `latin1` in R). However, by the time R is ready to read that value, it will be assuming that it is UTF-8 encoded, and so it interprets some `latin1` bytes as though they were UTF-8, which is incorrect.

For reference, R sets up the locale on Windows here:

https://github.com/wch/r-source/blob/ab26cd7a4bb2f109fe614fcedd5857d6028394d4/src/main/main.c#L866-L906

However, we don't call that until here:

https://github.com/rstudio/rstudio/blob/e7c8d8d28da7b50623acaadc5c95863fac40a347/src/cpp/r/session/REmbeddedWin32.cpp#L388-L389

jgutman (approved):

PR #11756, opened 2022-08-15
copy c++ implementation to resolve user home path using env vars

### LINKED ISSUE #11635: Default home directory can be incorrectly expanded/understood by windows
### System details

    RStudio Edition : Desktop
    RStudio Version : 2022.10.0-68
    OS Version      : Windows11
    R Version       : 

1. Start IDE with default settings & a default project
2. Create a new R Markdown file
3. Select PDF as default output format
4. Knit the file to PDF
5. Accept default location for file. `C:\Users\<user>\Documents`

File is created, but when attempting to open the preview, it fails due to not finding a path. The path is completely incorrect.

failure-tilde

This also happens if you save the file to subdirectory of the default directory (saved to `C:\Users\Jon\Documents\testdir-newproject`):

failure-tilde-2

To work around the issue,
* Save to a directory that is not a sub-directory of the default "home" directory 

The IDE seems to be incorrectly substituting the path when using the `C:\Users\<user>\Documents` directory.  As the issue shows up even when changing the "Default Working directory" in global settings, to something above "Documents" (e.g.  `C:\Users\<user>`), and then saving the file in the "Documents" directory.

I expect the file preview to be opened.

- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).

jgutman:
### Intent

Followup to fix #11635 

Electron needs to resolve home path alias consistently with the backend.

See implementation in:    https://github.com/rstudio/rstudio/blob/d01feac78480189424d22c029dc97100701463f7/src/cpp/session/SessionOptions.cpp#L223
https://github.com/rstudio/rstudio/blob/8f85b6b8ed4e7db1ac167b2007d17e26c947e4fd/src/cpp/shared_core/system/User.cpp#L159-L176

### DIFF
--- a/src/node/desktop/src/core/user.ts
+++ b/src/node/desktop/src/core/user.ts
@@ -20,10 +20,10 @@ import { FilePath } from './file-path';
 
 export function userHomePath(): FilePath {
   const user = getenv('R_USER');
-  if (user != '') 
+  if (user !== '') 
     return new FilePath(user);
   const home = getenv('HOME');
-  if (home != '')
+  if (home !== '')
     return new FilePath(home);
   return new FilePath(os.homedir());
 }
--- a/src/node/desktop/src/core/user.ts
+++ b/src/node/desktop/src/core/user.ts
@@ -14,10 +14,17 @@
  */
 
 import os from 'os';
+import { getenv } from './environment';
 
 import { FilePath } from './file-path';
 
 export function userHomePath(): FilePath {
+  const user = getenv('R_USER');
+  if (user != '') 
+    return new FilePath(user);
+  const home = getenv('HOME');
+  if (home != '')
+    return new FilePath(home);
   return new FilePath(os.homedir());
 }
 


src/node/desktop/src/core/user.ts
```diff
@@ -14,10 +14,17 @@
  */
 
 import os from 'os';
+import { getenv } from './environment';
 
 import { FilePath } from './file-path';
 
 export function userHomePath(): FilePath {
+  const user = getenv('R_USER');
+  if (user != '')
```
timtmok, line 23: Because of Typescript, both sides are ensured to be strings but for style consistency `!==` should be used to also check type.

timtmok (approved):
Minor nit with the comparison operator but I don't think that will affect the behaviour.

src/node/desktop/src/core/user.ts
jgutman, line 23:
```suggestion
  if (user !== '') 
```

src/node/desktop/src/core/user.ts
```diff
@@ -14,10 +14,17 @@
...
+import { getenv } from './environment';
 
 import { FilePath } from './file-path';
 
 export function userHomePath(): FilePath {
+  const user = getenv('R_USER');
+  if (user != '') 
+    return new FilePath(user);
+  const home = getenv('HOME');
+  if (home != '')
```
jgutman, line 26:
```suggestion
  if (home !== '')
```

Issue #11672, opened 2022-07-27
Unable to start Elsbeth Geranium


sfr:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2022.11.0-daily-78
    OS Version:  Windows 10 x64 (build 19044)
    R Version: 4.2.1

### Describe the problem in detail
Unable to start Elsbeth Geranium 2022.11.0-daily-78, actually none of the Elsbeth Geranium versions so far. Other versions e.g.  Spotted Wakerobin, start fine.
![err msg](https://user-images.githubusercontent.com/2040557/181179266-93b49e3c-c66a-4cd4-a25a-ba378756b30b.png)

#### Diagnostics
    c:\development\RStudio-2022.11.0-daily-78\resources\app\bin>"Run Diagnostics.lnk"
    ERROR: The system was unable to find the specified registry key or value.
    ERROR: The system was unable to find the specified registry key or value.

#### Registry
I do not have administrative privileges to `HKEY_LOCAL_MACHINE,` so everything is in `HKEY_CURRENT_USER`

    [HKEY_CURRENT_USER\SOFTWARE\R-core]
    
    [HKEY_CURRENT_USER\SOFTWARE\R-core\R]
    "Current Version"="4.2.1"
    "InstallPath"="c:\\development\\r-4.2.1"
    
    [HKEY_CURRENT_USER\SOFTWARE\R-core\R\4.2.1]
    "InstallPath"="c:\\development\\r-4.2.1"
    
    [HKEY_CURRENT_USER\SOFTWARE\R-core\R64]
    
    [HKEY_CURRENT_USER\SOFTWARE\R-core\R64\4.2.1]
    "InstallPath"="c:\\development\\r-4.2.1"
    
    [HKEY_CURRENT_USER\SOFTWARE\R-core\Rtools]
    "InstallPath"="C:\\rtools42"
    "Current Version"="4.2"
    
    [HKEY_CURRENT_USER\SOFTWARE\R-core\Rtools\4.2]
    "InstallPath"="C:\\rtools42"
    "FullVersion"="4.2.0.1"

#### Environment variables (amongst others)
    Path=c:\development\r-4.2.1\bin\x64\
    RTOOLS42_HOME=C:\rtools42
    R_HOME=C:\development\r-4.2.1
    R_LIBS=c:\development\r-4.2.1\library\

#### Session info from working Spotted Wakerobin
    > sessioninfo::session_info()
[...]

PR #11767, opened 2022-08-16
add HKEY_CURRENT_USER in findRInstallations Electron

jgutman:
### Intent

Addresses #11672 in line with Qt approach to Windows registry keys

### DIFF
--- a/src/node/desktop/src/main/detect-r.ts
+++ b/src/node/desktop/src/main/detect-r.ts
@@ -307,6 +307,27 @@ function scanForRPosix(): Expected<string> {
   return err();
 }
 
+function regQuery(cmd: string, rInstallations: Set<string>): Set<string> {
+  const [output, error] = executeCommand(cmd);
+    if (error) {
+      logger().logError(error);
+      return rInstallations;
+    }
+
+  // parse the actual path from the output
+  const lines = output.split(EOL);
+  for (const line of lines) {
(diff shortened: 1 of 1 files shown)


kevinushey (approved):
LGTM!

One other suggestion I'd make -- now that user-mode installs of R will be placed in the user's %LOCALAPPDATA%/Programs/R folder, we might also want to search for R installations in that location. I think you could add the folder here:

https://github.com/rstudio/rstudio/blob/94b86a9bd5821e293f17c1191b3846bae04d85f2/src/node/desktop/src/main/detect-r.ts#L339-L344

and the existing machinery should hopefully be able to pick it up. (This would be useful just in case users choose not to update the Windows registry after installing R.)

jgutman:
Okay, just now noticing in the logs, after merging this issue -- we get the following messages:
```2022-09-14T15:15:01.968Z error: Command failed: reg query HKEY_LOCAL_MACHINE\SOFTWARE\R-Core /s /v InstallPath /reg:64
ERROR: The system was unable to find the specified registry key or value.

2022-09-14T15:15:02.033Z error: Command failed: reg query HKEY_CURRENT_USER\SOFTWARE\R-Core /s /v InstallPath /reg:64
ERROR: The system was unable to find the specified registry key or value.
```

kevinushey:
Can we suppress the error message / output here?

Issue #11766, opened 2022-08-16
New project that uses renv creates the project directory and contents but doesn't open the project


ronblum:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2022.11.0-daily+101/105
    OS Version      : MacOS 13 and Windows 11
    R Version       : 4.2.1

### Steps to reproduce the problem

1. Open a new project: File -> New Project
2. New Directory -> New Project
3. Check "Use renv with this project"
4. Choose directory
5. Create Project
    - Error with pop up appears (see below).
    - Error appears in Console (see below).
6. In Files Pane, switch to the new project directory
7. Click on the .proj file
    - The project will be opened.
    - The correct message about `renv` appears in the Console Pane (see below).

### Describe the problem in detail

Errors in step 5:
<img width="245" alt="renv" src="https://user-images.githubusercontent.com/31009952/184917640-302a947d-34b8-4a35-aa5c-0ce0051e96c0.png">

```
[Workspace loaded from ~/.RData]

* Initializing project ...
* Discovering package dependencies ... Done!
* Copying packages into the cache ... Done!
The following package(s) will be updated in the lockfile:

# CRAN ===============================
- renv   [* -> 0.15.5]

* Lockfile written to '~/rstudio/Projects/Testing/empty_project_renv/renv.lock'.
Error in gzfile(file, "rb") : cannot open the connection
In addition: Warning message:
In gzfile(file, "rb") :
  cannot open compressed file 'output.rds', probable reason 'No such file or directory'
> 
```

Correct behavior in step 7:

```
* Project '~/rstudio/Projects/Testing/empty_project_renv' loaded. [renv 0.15.5]
>
```

### Describe the behavior you expected

The project should open correctly, and in the new directory. This is a regression, as it works correctly in Spotted Wakerobin 2022.11.1+554, both Qt and Electron.

- [ ] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ ] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [ ] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
[...]

PR #11770, opened 2022-08-16
ensure 'output.rds' written to expected location

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/11766.

### Approach

When running code within an R sub-process, ensure that any output is written to `output.rds` at the expected location.

### Automated Tests

N/A

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/11766. 

### Checklist

### DIFF
--- a/src/cpp/session/modules/SessionRUtil.R
+++ b/src/cpp/session/modules/SessionRUtil.R
@@ -133,6 +133,9 @@
    # define runner script (will load data and execute user-defined callback)
    script <- quote({
       
+      # save the original working directory
+      originalWorkingDir <- getwd()
+      
       # read side-car data file
       bundle <- readRDS("bundle.rds")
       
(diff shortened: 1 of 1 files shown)


jgutman (approved):

jgutman (commented):
Do we not need this bit then? Is this the wrong dir? 

```

      # move to requested working directory
      workingDir <- bundle[["workingDir"]]
      if (!is.null(workingDir)) {
         dir.create(workingDir, recursive = TRUE, showWarnings = FALSE)
         owd <- setwd(workingDir)
         on.exit(setwd(owd), add = TRUE)
      }
```

kevinushey:
That's the directory in which the provided function itself should be run (in case it's sensitive to the working directory).

The scaffolding for how the function is run (the input bundle, and where the output bundle should be written) are separate, and are set for the process when it's launched. So, when writing the output to file, we need to return to the original working directory.

Does that make sense?

jgutman:
Got it

Issue #11782, opened 2022-08-18
include .pdb debug information in debug package builds


kevinushey:
On occasion, some bugs can only be reproduced in package builds. We should make it easier to:

1. Produce local package builds (add a `debug` flag to the `make-package.bat` script)
2. Include all relevant `.pdb` files, so we can get symbols in debug stack traces.

PR #11785, opened 2022-08-18
Developer/win32 debug package builds

kevinushey:
### Intent

- Adds a `debug` command line argument to `make-package`, to explicitly document the mechanism for creating debug builds.
- Also copy `.pdb` files into the created executable, so that we can more easily get debug information when debugging a package build.

Closes https://github.com/rstudio/rstudio/issues/11782.


jgutman (approved):
Oh, neat! Thank you! This is super helpful

PR #11789, opened 2022-08-18
 don't re-encode to utf8

jgutman:
### Intent

Addresses #11717, followup to #11788 
### Approach

#11788  makes it so an error in the crash handler doesn't bring down the r session. This PR addresses the actual error so it doesn't get thrown in the first place. 

### Automated Tests

None yet

### QA Notes

- Use Windows, R 3.6.3
- Create a user with an accented character (e.g. umlaut) in the user name
- Install daily as usual, ensure RStudio starts without crashing


kevinushey (approved):
LGTM!

Issue #11779, opened 2022-08-18
RStudio / Quarto not rendering PDF or HTML when space in filename


quinten-goens:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2022.02.2 Build 485
    OS Version      :  Windows 10
    R Version       :  R-4.2.1

### Steps to reproduce the problem
Make a Quarto file (.qmd) with a space in the filename. E.g. use the following content for the file: 

```
---
title: "Quarto Demo"
author: "Test"
date: "2022-5-24"
---
```

Try to render the file by clicking the ''render'' button. 

### Describe the problem in detail
The rendering is halted. The job in cancelled and returns: 
```
'C:/Program' is not recognized as an internal or external command, operable program or batch file.
```

### Describe the behavior you expected

That the file would render even with a space in the filename. The easy fix is to remove the space in the filename. But ideally spaces would not break the process or alternatively RStudio would give an appropriate message.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ ] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
This seems like a Quarto bug to me, as RStudio is normally able to run other applications containing spaces in their paths without issue.

As a workaround, I think we can try using a short path name for Quarto on Windows, to avoid the issue with quarto components having spaces in their names.

jgutman:
Closing in favor of #11804

PR #11792, opened 2022-08-18
use short path name when running quarto

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/11779.

### Approach

Use a short path name when trying to run Quarto.

### Automated Tests

N/A (Windows only)

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/11779.

### Checklist

### DIFF
--- a/src/cpp/core/FileUtils.cpp
+++ b/src/cpp/core/FileUtils.cpp
@@ -65,6 +65,29 @@ bool copySourceFile(const FilePath& sourceDir,
 
 } // anonymous namespace
 
+// NOTE: Assumes that the path string is UTF-8 encoded.
+std::string shortPathName(const std::string& path)
+{
+#ifdef _WIN32
+
+   std::wstring widePath = string_utils::utf8ToWide(path);
+   DWORD length = ::GetShortPathNameW(widePath.c_str(), nullptr, 0);
+   if (length == 0)
+      return path;
+
+   std::vector<WCHAR> buffer(length, 0);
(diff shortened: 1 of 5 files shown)


jgutman (approved):
@jjallaire just created a new issue for this which got assigned to me. This seems like a reasonable solution to me, although it's odd that we don't have this issue with other executables, which does seem to suggest this could perhaps be addressed on the Quarto side, but I think addressing it this way is sufficient for now to make sure it gets into this release

jgutman:
Also, we will soon have the ability to run automated tests on Windows, and this is a good candidate. Not sure of the ETA on that but I will put this on @jonvanausdeln's radar

src/cpp/core/FileUtils.cpp
```diff
@@ -65,6 +65,29 @@ bool copySourceFile(const FilePath& sourceDir,
...
+
+   std::vector<WCHAR> buffer(length, 0);
+   DWORD result = ::GetShortPathNameW(widePath.c_str(), &buffer[0], length);
+   if (result == 0)
+      return path;
+
+   return string_utils::wideToUtf8(std::wstring(&buffer[0]));
+
+#else
+   return string;
```
jgutman, line 86: This should have been return `path`; fixed on main post-merge

kevinushey: Sorry about that! Thanks for fixing that up.

Issue #11252, opened 2022-05-19
Implement acceptNavigationRequest in Electron


gtritchie:
In the Qt desktop app we restrict what can load in a window here: https://github.com/rstudio/rstudio/blob/edd144e6c64f765f62716e57a31808597e92ab7e/src/cpp/desktop/DesktopWebPage.cpp#L402-L404

This only has a skeleton implementation in Electron, here: https://github.com/rstudio/rstudio/blob/ecd67f3bf4ed54ca704a8f765683b135a7bda277/src/node/desktop/src/main/desktop-browser-window.ts#L167-L169

Among other things, we need to implement various callbacks that communicate valid URLs, then use this information:

https://github.com/rstudio/rstudio/blob/ecd67f3bf4ed54ca704a8f765683b135a7bda277/src/node/desktop/src/main/gwt-callback.ts#L769-L786

timtmok:
The last automation run on `main` has passed all tests.

For the most part, the changes for this ticket shouldn't be much different. The main part is any of the panes will check what is being loaded and ensure it is approved content coming from the IDE. The other piece is it will launch any external URLs displayed in panes in the web browser.

A related issue is #11611 where clicking an external link crashes the IDE. This still needs a solution but is related to Electron not supporting the `will-frame-navigate` event. We need to detect the click on the external link and open it in the user's default web browser.

jonvanausdeln:
Verified on 2022.11.0 Build 178
windows11, Ubuntu 22.04

Tested several different pages, links and file types.

Issue #11611, opened 2022-07-13
Click on link in viewer pane in IDE electron crash IDE


cderv:
### System details

    RStudio Edition : Desktop [Open Source]
    RStudio Version : 2022.10.0.64
    OS Version      : Windows 10 x64 (build 22000)
    R Version       : R version 4.2.0 (2022-04-22 ucrt)

### Steps to reproduce the problem

1. Open RStudio daily Electron
2. Create a new quarto document (probably happening with other type too) with the aim to create a HTML document in the viewer with a link
3. Add this simple content 
````markdown
---
format: html
---

<https://rstudio.com>
````

4. Render the document to HTML
5. It will open in the viewer
6. Click the link in the viewer

### Describe the problem in detail

When clicking the link, the IDE window reload completely, and then crash 

See below
![ide-link-crash](https://user-images.githubusercontent.com/6791940/178683274-815d45da-d3e1-41bd-ae3c-468a363a8fbf.gif)

Crash message is this one

![image](https://user-images.githubusercontent.com/6791940/178682446-23a3f893-550d-4da4-a908-6e6faf9c6306.png)

[diagnostics-report.txt](https://github.com/rstudio/rstudio/files/9100274/diagnostics-report.txt)

### Describe the behavior you expected

I would expect the link to open in external window like preview Qt version but not crash the whole session

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

**Comment by timtmok** [comment]
This is the error produced in the dev console:
```
Uncaught TypeError: ansi_html_community__WEBPACK_IMPORTED_MODULE_0___default(...).setColors is not a function
    at eval (overlay.js?f53e:29:1)
    at ./node_modules/webpack-dev-server/client/overlay.js (index.js:277:1)
    at __webpack_require__ (index.js:489:33)
    at fn (index.js:677:21)
    at eval (index.js?protocol=ws…o&reconnect=10:8:69)
[...]

PR #11799, opened 2022-08-19
Electron will-navigate event handling

timtmok:
### Intent
Address #11252 to check whether navigation to the url is accepted

### Approach
Implement similar logic to `WebPage::acceptNavigationRequest`. URL is checked for:

* malformed URL logs an error, no navigation
* `about:blank` and `chrome://gpu` are allowed
* protocols other than `http`, `https`, `mailto`, and `data` are rejected
* no base url and local is allowed
* matching base url is allowed
* viewer, presentation, tutorial, or Shiny dialog is allowed
* allow if external navigation is allowed or is a safe host
* last resort, opens in a browser

### Automated Tests
Unit tests included cover the util methods that have been added and the `allowNavigation` method that is called for the `will-navigate` event.

### QA Notes
Check the Tutorial, Viewer, and Presentation panes and run a Shiny app in a window and in the Viewer pane.

Opening a new project should have the IDE re-load with the project.

### Checklist


src/node/desktop/src/main/url-utils.ts
```diff
@@ -0,0 +1,111 @@
+/*
+ * url-utils.ts
+ *
+ * Copyright (C) 2022 by RStudio, PBC
```
jgutman, line 4: Dangit, you're going to make me update my in-progress PR changing all RStudio copyrights to Posit, aren't you?

src/node/desktop/src/main/url-utils.ts
```diff
@@ -0,0 +1,111 @@
...
+/**
+ * Checks the `url` if it is local
+ * 
+ * @param url the URL to check
+ * @returns true if it is local
+ */
+export function isLocalUrl(url: URL) {
+  const host = url.hostname;
+
+  return host === 'localhost' || host === '127.0.0.1' || host === '[::1]';
```
jgutman, line 53: is there a particular reason this changed from `host == "::1"` in Qt to `host === '[::1]'` here?

src/node/desktop/src/main/desktop-browser-window.ts
```diff
@@ -276,6 +253,85 @@ export class DesktopBrowserWindow extends EventEmitter {
...
+      if (targetUrl.protocol === baseUrl.protocol &&
+          targetUrl.host === baseUrl.host) {
+        return true;
+      }
+    }
+
+    const viewer = this.viewerUrl ?? this.mainWindow?.viewerUrl;
+    const tutorial = this.tutorialUrl ?? this.mainWindow?.tutorialUrl;
+    const presentation = this.presentationUrl ?? this.mainWindow?.presentationUrl;
+    const shinyDialog = this.shinyDialogUrl ?? this.mainWindow?.shinyDialogUrl;
```
jgutman, line 304: are there really any differences between these?

src/node/desktop/src/main/gwt-callback.ts
```diff
@@ -761,22 +762,26 @@ export class GwtCallback extends EventEmitter {
...
-      GwtCallback.unimpl('desktop_set_tutorial_url');
+      this.getSender('desktop_set_tutorial_url', event.processId, event.frameId).setTutorialUrl(url);
     });
 
     ipcMain.on('desktop_set_viewer_url', (event, url) => {
       this.getSender('desktop_set_viewer_url', event.processId, event.frameId).setViewerUrl(url);
     });
 
-    ipcMain.on('desktop_reload_viewer_zoom_window', (event, url) => {
+    ipcMain.on('desktop_reload_viewer_zoom_window', (_event, url) => {
```
jgutman, line 772: is this our convention to use _event if it isn't being called? I haven't been doing that

jgutman (approved):

src/node/desktop/src/main/url-utils.ts
timtmok: Within a URL, ipv6 addresses are defined in square brackets. The URL object also returns the hostname as `[::1]`. I had to look this up when writing the unit tests.

src/node/desktop/src/main/desktop-browser-window.ts
timtmok: Not really other than GWT has different callbacks for these special URL's. I think it does need to be tracked individually though since the different panes could all be displaying something. It would have been nice to refactor this so that the iframe (or something that represents it) keeps track of the allowed URL it can display.

Most of this was getting an understanding of what the URL navigation request was supposed to do and the URL's that are allowed.

src/node/desktop/src/main/gwt-callback.ts
timtmok: Yes, `eslint` will complain if there is an unused parameter. The underscore signals that it's there, because of argument order, but unused. It's perfectly fine without it but addresses the linter warning (I think it's configured as a warning).

timtmok:
Hopefully, this addresses a lot of the navigation handling. There's still a missing piece with iframe navigation since Electron hasn't implemented the `will-frame-navigate` event. I'm looking into trying to hook that in through GWT but not having any luck so far. It's pretty important to have this to address #11611. It needs to detect that it's an external URL, cancel the navigation, and open the URL in a browser.

PR #11800, opened 2022-08-19
Moved verifyRequestSignature call into authenticate

jeffvroom:
### Intent

Follow on fix for: http://github.com/rstudio/rstudio-pro/issues/3674 

### Approach

Use the existing authenticate() hook for the verifyRequestSignature. We could remove all of the other calls to verifyRequestSignature as they are now redundant but am thinking to leave them in for this patch.

### QA

See the original PR for testing.


MariaSemple (approved):
LGTM!

Issue #11712, opened 2022-08-06
Can't "Open in browser" after rendering Rmarkdown document


ntluong95:
### System details

    RStudio Edition : Desktop
    RStudio Version :  RStudio 2022.11.0-daily+82 "Elsbeth Geranium" Daily
    OS Version      :  Windows 11
    R Version       :  4.2.1

### Steps to reproduce the problem
Write a Rmarkdown file and render it
### Describe the problem in detail
After rendering, a window pop-up that allow us to "Open in browser". However, this function is not working with this build of Rstudio. For previous build, I use and it is fine
### Describe the behavior you expected
After rending Rmarkdown, the document can open in browser

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
@ntluong95 Thank you for raising the issue! Can you check with the latest build of RStudio Desktop, 2022.11.0-daily+88? I was able to reproduce the issue with a previous version, but it appears to be resolved.

kevinushey:
I can reproduce with a user account containing non-ASCII characters in the name; for example:

![Screenshot 2022-08-18 153551](https://user-images.githubusercontent.com/1976582/185507266-6e0e3065-08b2-404a-94d9-1c3465c9f63b.png)

Perhaps that's related?

kevinushey:
Ah, no, it's not that -- I think it's instead that we're not resolving the home directory correctly. Note that on Windows, the home directory is by default the document directory:

```
> path.expand("~/")
[1] "C:/Users/kevin/Documents/"
```

This can be overridden by setting the `HOME` or `R_USER` environment variables.

However, we're resolving the home directory in Electron using `os.homedir()`, which only looks at `USERPROFILE`.

https://github.com/rstudio/rstudio/blob/main/src/node/desktop/src/core/user.ts#L21-L29

See the associated code on Qt here:

https://github.com/rstudio/rstudio/blob/main/src/cpp/shared_core/system/Win32User.cpp#L170-L173

[...]

PR #11801, opened 2022-08-19
Fix windows home path ~ implementation

jgutman:
### Intent

Address implementation of `~` on Windows in #11635 and #11712

### Approach

Port over `currentCSIDLPersonalHomePath` and `defaultCSIDLPersonalHomePath` as native modules and use those when both R_USER and HOME are unset and platform is Windows

### Automated Tests

None

### DIFF
--- a/src/node/desktop/src/native/desktop/desktop.cc
+++ b/src/node/desktop/src/native/desktop/desktop.cc
@@ -212,8 +212,8 @@ wchar_t* currentCSIDLPersonalHomePathImpl()
                                    nullptr,
                                    SHGFP_TYPE_CURRENT,
                                    homePath);
-   if (SUCCEEDED(hr))
-      return homePath;
+   #else
+   wchar_t homePath[1024] = {};
    #endif
    return homePath;
 }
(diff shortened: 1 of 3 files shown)


src/node/desktop/src/native/desktop/desktop.cc
```diff
@@ -197,13 +198,76 @@ Napi::Value isCtrlKeyDown(const Napi::CallbackInfo& info)
...
+   #endif
+   return homePath;
+}
+} // end anonymous namespace
+
+Napi::Value currentCSIDLPersonalHomePath(const Napi::CallbackInfo& info)
+{
+   wchar_t* value = currentCSIDLPersonalHomePathImpl();
+   // convert wide to UTF-8
+   std::wstring_convert<std::codecvt_utf8<wchar_t>> conv1;
```
timtmok, line 226: I think it might be better to use `codecvt_utf8_utf16` although I'm not sure. `codecvt_utf8` seems to convert between UTF-8 and UTF-32. I think in testing we want to be sure that users with special characters in their username work.

timtmok (approved):

Issue #11778, opened 2022-08-18
Daily build called to open a file opens in the user's home instead of file location


Halberdier:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2022.11.0 Build 113, "Elsbeth Geranium" Daily (cc55ded2, 2022-08-17) for Windows
    OS Version      : Microsoft Windows 10 Enterprise, 10.0.19044
    R Version       : 4.2.1 (2022-06-23 ucrt) -- "Funny-Looking Kid"

### Steps to reproduce the problem
Right-click on a file in Windows File Explorer (e.g., an R script)
Choose "Open With" then RStudio, if RStudio is not already the default app for that file extension

### Describe the problem in detail
RStudio opens the session with the working directory in the user's home.
Not really a bug, but the behavior of this daily build is different from the official release.

### Describe the behavior you expected
In the official release, when RStudio is called to open a file, the working directory is the file location.
In this daily build, the file location has to be set explicitly with setwd(...).

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

jgutman:
This may be the same/closely related to #11088

jgutman:
This behavior exists on Electron, on both Mac and Windows

jonvanausdeln:
Verified on 2022.11.0-daily+215
WIndows 11, MacOS 12.4

Tested against several file types in many paths. Included often problematic paths in windows that contain spaces,etc.

PR #11832, opened 2022-08-24
Electron - open file in working directory when launching RStudio with a file arg

jgutman:
### Intent

Address #[#11778](https://github.com/rstudio/rstudio/issues/11778), at least on Windows and Linux. May need some additional handling for Mac to process double-click on Finder events, but will leave to subsequent PR.

### Approach

Set the RS_INITIAL_WD environment variable, used by the C++ session code, in pre-launch Electron code

### Automated Tests
None

### QA Notes
Will need to be tested on at least Windows and Mac -- see notes in issue.

### DIFF
--- a/src/node/desktop/src/main/application-launch.ts
+++ b/src/node/desktop/src/main/application-launch.ts
@@ -17,6 +17,7 @@ import path from 'path';
 import { spawn } from 'child_process';
 import { existsSync, readdirSync } from 'fs';
 import { setenv, unsetenv } from '../core/environment';
+import { kRStudioInitialProject, kRStudioInitialWorkingDir } from '../core/r-user-data';
 import { MainWindow } from './main-window';
 import { app } from 'electron';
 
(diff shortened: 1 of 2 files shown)


src/node/desktop/src/main/application-launch.ts
```diff
@@ -70,19 +71,18 @@ export class ApplicationLaunch {
...
     if (existsSync(projectFile)) {
-      setenv('RS_INITIAL_PROJECT', projectFile);
+      setenv(kRStudioInitialProject, projectFile);
     }
 
     // run it
     spawn(process.execPath, argv, { detached: true });
 
     // restore environment variables
-    unsetenv('RS_INSTALL_WD');
```
jgutman, line 85: RS_INSTALL_WD doesn't even exist, pretty sure that was a typo and RS_INITIAL_WD was intended. These get unset in the C++ code anyway, so I'm not sure that it matters

src/node/desktop/src/main/args-manager.ts
```diff
@@ -146,8 +146,12 @@ export class ArgsManager {
             setenv(kRStudioInitialProject, arg);
             return false;
           }
+          else {
```
jgutman, line 149: for all files that are not Rproj files

timtmok: You might be able to do the same thing for the `open-file` event. However, the event may be called at any time and not necessarily before the session launches. The electron docs specifies that the `open-file` event at startup needs to be handled before the app is ready. But I'm not sure if that guarantees that opening from Finder will always fire the event before app ready.

I guess try that in the next PR and find out! 🤞

timtmok (approved):

src/node/desktop/src/main/args-manager.ts
jgutman: yep, working on that PR now -- seems like we can just have an if/else clause based on whether or not the app is ready yet. That's what we do for handling the Rproj files so I am guessing the same thing will work for other files

PR #11833, opened 2022-08-24
Add Finder handling for Mac setting initial working directory

jgutman:
### Intent
Second part of [#11778](https://github.com/rstudio/rstudio/issues/11778) for Macs

### Approach

set the initial working directory in the `open-file` event used when double clicking file from the Finder window.

### DIFF
--- a/src/node/desktop/src/main/application.ts
+++ b/src/node/desktop/src/main/application.ts
@@ -208,9 +208,7 @@ export class Application implements AppState {
       }
       else {
         // for non-Rproj files, we want to set the initial working directory, before opening the file
-        if (app.isReady()) {
-          this.appLaunch?.launchRStudio({workingDirectory: path.dirname(filepath)});
-        } else {
+        if (!app.isReady()) {
           setenv(kRStudioInitialWorkingDir, path.dirname(filepath));
         }
       }
--- a/src/node/desktop/src/main/application.ts
+++ b/src/node/desktop/src/main/application.ts
@@ -19,7 +19,7 @@ import path from 'path';
 import { getenv, setenv } from '../core/environment';
 import { FilePath } from '../core/file-path';
 import { logger } from '../core/logger';
-import { kRStudioInitialProject } from '../core/r-user-data';
+import { kRStudioInitialProject, kRStudioInitialWorkingDir } from '../core/r-user-data';
 import { generateRandomPort } from '../core/system';
 import { getDesktopBridge } from '../renderer/desktop-bridge';
 import { DesktopActivation } from './activation-overlay';
@@ -205,6 +205,14 @@ export class Application implements AppState {
           setenv(kRStudioInitialProject, filepath);
         }
         return;
+      }
+      else {
+        // for non-Rproj files, we want to set the initial working directory, before opening the file
+        if (app.isReady()) {
+          this.appLaunch?.launchRStudio({workingDirectory: path.dirname(filepath)});
+        } else {
+          setenv(kRStudioInitialWorkingDir, path.dirname(filepath));
+        }
       }
 
       app.whenReady()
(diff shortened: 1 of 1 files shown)


src/node/desktop/src/main/application.ts
```diff
@@ -206,6 +206,14 @@ export class Application implements AppState {
         }
         return;
       }
+      else {
+        // for non-Rproj files, we want to set the initial working directory, before opening the file
+        if (app.isReady()) {
+          this.appLaunch?.launchRStudio({workingDirectory: path.dirname(filepath)});
+        } else {
+          setenv(kRStudioInitialWorkingDir, path.dirname(filepath));
+        }
```
jgutman, line 215: unlike Rproj, we don't want to return early here because we still need to do the below stuff under app.whenReady

timtmok: Is it necessary to launch a new session here? Remember, the `open-file` event can occur if the IDE is currently running or is starting up.

If the event fires and this is an instance starting up then this event handling is done before the app is ready so this case is probably fine.

If the event fires and there is an existing instance, it would open a new instance for this file. However, the single instance logic will run some checks within the new instance. If the new instance is opening a file, not a project, and detects a previous instance, it shuts down. At the same time, the first instance gets an event (see `second-instance` below) and opens the file. Also at the same time in the first instance's `open-file` event, it continues after launching the second instance and opens the file. I think there would be two calls to `openFile` but it doesn't cause any harm.

At least that's what I think will happen.

jgutman: Okay, so thinking about this again, and examining the current Qt behavior. If there is an existing instance of RStudio, we don't want to open a new instance, and we don't want to change the working directory. We just want to open the second file in a new editor tab in the existing instance. So actually, maybe we just completely remove the if clause behavior -- the open file part happens below, and we don't need to or want to set working directory if an instance already exists. But if there's no instance (app not ready) then we do want to set the WD on launch

src/node/desktop/src/main/application.ts
```diff
@@ -206,6 +206,14 @@ export class Application implements AppState {
         }
         return;
       }
+      else {
+        // for non-Rproj files, we want to set the initial working directory, before opening the file
+        if (app.isReady()) {
+          this.appLaunch?.launchRStudio({workingDirectory: path.dirname(filepath)});
+        } else {
```
jgutman, line 213:
```suggestion
        if (!app.isReady()) {
```

src/node/desktop/src/main/application.ts
```diff
@@ -206,6 +206,12 @@ export class Application implements AppState {
         }
         return;
       }
+      else {
+        // for non-Rproj files, we want to set the initial working directory, before opening the file
```
jgutman, line 210:
```suggestion
        // for non-Rproj files, we want to set the initial working directory, before opening the file
        // if a session does not already exist (otherwise, just open the file)
```

src/node/desktop/src/main/application.ts
timtmok: @jgutman I think that would work. My initial implementation is probably incorrect with the `isReady` check. It would be handled anyway below with the `whenReady` promise.

Do you also mean remove the `.rproj` piece as well? I think that's still needed to open in a new session. Otherwise, I think the call to GWT to open the file works but it reuses the window.

jgutman: Oh, you might be confused because I already removed it, and made the change, so I think this was ready for re-review as is. I don't think any changes to rproj files are needed, they should always open a new session, but non rproj files should not open a new session if one already exists. Basically the only behavior we want to change here is:

Current behavior on main: rproj files start new session, non rproj files open in source editor, but never change the working directory

Desired behavior: non rproj files open in source editor, but if no session currently exists, change the working directory by setting the environment var prior to session launch. Rproj behavior unchanged

timtmok: Look good!

timtmok (approved):

PR #11884, opened 2022-08-31
avoid returning pointers to stack memory

kevinushey:
Resolves this warning:

```
  CXX(target) Release/obj.target/desktop/src/native/desktop/desktop.o
../src/native/desktop/desktop.cc:218:11: warning: address of stack memory associated with local variable 'homePath' returned [-Wreturn-stack-address]
   return homePath;
          ^~~~~~~~
../src/native/desktop/desktop.cc:249:11: warning: address of stack memory associated with local variable 'homePath' returned [-Wreturn-stack-address]
   return homePath;
          ^~~~~~~~
2 warnings generated.
```


jgutman:
Thanks!

PR #11886, opened 2022-08-31
use 'webview' for Viewer pane in Electron

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/11611.

### Approach

Unfortunately, Electron doesn't expose some of the events events we need in order to control navigation within a sub-frame (e.g. an iframe).

Fortunately, it _does_ in so-called [webviews](https://www.electronjs.org/docs/latest/api/webview-tag). These are effectively enhanced versions of iframes, with a fair amount more isolation + customizability (despite the warnings). These also expose the functionality we require in order to control navigation.

This PR flips a switch for Electron builds, wherein we use a Webview for the Viewer pane rather than an iframe.

### Automated Tests

Not included.

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/11611. In particular, check that various outputs that we render within the Viewer pane are shown correctly, and links within are navigated as expected.

### Checklist

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/ScrollUtil.java
+++ b/src/gwt/src/org/rstudio/core/client/ScrollUtil.java
@@ -14,7 +14,7 @@
  */
 package org.rstudio.core.client;
 
-import org.rstudio.core.client.widget.RStudioFrame;
+import org.rstudio.studio.client.workbench.views.viewer.ViewerPane;
 
 import com.google.gwt.core.client.Scheduler;
 import com.google.gwt.core.client.Scheduler.RepeatingCommand;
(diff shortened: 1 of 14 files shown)


kevinushey:
Actually, we might want to just make the RStudioFrame class itself behave differently in Electron vs. not-Electron, as I expect we'll want to be able to control navigation (and other behaviors) in every context where we're using Electron. I'll dive in further tomorrow.

timtmok:
Amazing.

I dismissed webviews since the Electron docs mentioned it is in flux and `event.preventDefault()` does nothing. But being able to listen on the `webContents` event is something I didn't think of trying.

I agree on the change to `RStudioFrame`. All usages need the URL checked before starting navigation.

kevinushey:
I'm hitting an ugly roadblock. It seems like the `<webview>` iframe is very strict about CORS -- even with the same origin for the embedded content, any attempts to interact with the underlying `iframe` window via JavaScript are forbidden. This means that interactions we try to perform in e.g. the Help pane will fail (since Help content is displayed in a webview iframe, which we can't interact with externally).

Thus far, I can't see any way to work around or disable this, but this might make the approach using `webview` a non-starter... which is quite unfortunate. But I'm going to keep poking a bit more.

PR #11927, opened 2022-09-09
add 'discard pending console input' preference

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/10391.

Screen Shot 2022-09-09 at 11 52 24 AM

### Approach

Straightforward addition of preference.

### Automated Tests

N/A

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/10391.

### Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,7 @@
 
 ### R
 
+* Whether pending console input is discarded on error can now be controlled via a preference in the Console pane. (#10391)
 * Improved handling of diagnostics within pipeline expressions. (#11780)
 * Improved handling of diagnostics within glue() expressions. 
 * RStudio now provides autocompletion results for packages used but not loaded within a project.
--- a/src/cpp/session/SessionMain.cpp
+++ b/src/cpp/session/SessionMain.cpp
@@ -980,7 +980,8 @@ void rConsoleHistoryReset()
 
 void rConsoleReset()
 {
-   rsession::console_input::clearConsoleInputBuffer();
+   if (prefs::userPrefs().discardPendingConsoleInputOnError())
+      rsession::console_input::clearConsoleInputBuffer();
 }
 
 bool rLocator(double* x, double* y)
--- a/src/cpp/session/include/session/prefs/UserPrefValues.hpp
+++ b/src/cpp/session/include/session/prefs/UserPrefValues.hpp
@@ -410,6 +410,7 @@ namespace prefs {
 #define kUiLanguageEn "en"
 #define kUiLanguageFr "fr"
 #define kNativeFileDialogs "native_file_dialogs"
+#define kDiscardPendingConsoleInputOnError "discard_pending_console_input_on_error"
 
 class UserPrefValues: public Preferences
 {
@@ -1830,6 +1831,12 @@ public:
     */
    bool nativeFileDialogs();
    core::Error setNativeFileDialogs(bool val);
+
+   /**
+    * When enabled, any pending console input will be discarded when an (uncaught) R error occurs.
+    */
+   bool discardPendingConsoleInputOnError();
+   core::Error setDiscardPendingConsoleInputOnError(bool val);
 
 };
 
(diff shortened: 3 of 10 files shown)


timtmok (approved):

DamienFr:
Thank you for implementing this.
In which version will this be available ? i'm currently running
RStudio 2022.07.2+576 "Spotted Wakerobin" Release (e7373ef832b49b2a9b88162cfe7eac5f22c40b34, 2022-09-06) for Ubuntu Bionic

kevinushey:
It will be available in the next release (currently targeted for the end of the year) but if you want to test the daily builds are available at https://dailies.rstudio.com.

ebuhle:
I'm just now encountering the issue described in #10391 in Windows after upgrading from R 4.1.3 to R 4.2.2. (I'm using RStudio 2022.07.1, but that has not changed, so I gather the new behavior is somehow enabled by R 4.2.x itself.) 

I appreciate this proposed option to opt out of the new halt-execution-on-error behavior. However, I think the default should be the old continue-execution-after-error behavior that we've all been accustomed to, i.e. the "Execution" checkbox should be unchecked by default. 

Like other commenters on #10391, I have numerous scripts that will break because of this new behavior. It is NOT necessarily the case that a line in a script (run interactively) that throws an error for some particular input will cause the rest of the script to fail. As another example in addition to those described in #10391, the one that brought me here just now asks the user to assign a personal access token in order to install a package from a private GitHub repo; if the PAT has not been defined, that line should throw an error and the script should carry on running. Sure, I could wrap the `install_github()` call in `try()`, and arguably that would be better programming etiquette, but the reality is that lots of users have idiosyncratic code that relies on the traditional behavior.

The problem with the opt-out approach is that I have colleagues who will also be attempting to run these scripts, and I can't ensure that they all go and uncheck the proposed "Execution" box. Likewise, I could solve the problem on my end pending the next release by reverting to an earlier version of RStudio, but I can't make colleagues do the same.

In summary, I agree with the other users who feel like this is a major and unexpected change in the default behavior. IMO a good solution would be to enable halt-execution-on-error but require active opt-in, keeping the previous behavior as the default.

kevinushey:
> the one that brought me here just now asks the user to assign a personal access token in order to install a package from a private GitHub repo; if the PAT has not been defined, that line should throw an error and the script should carry on running. 

But how is your script going to work if the required package isn't installed? In addition, what if someone wants to run your script non-interactively?

If you're running the script interactively, you can always just select the next lines of code to be run as well (although I recognize it's annoying if you're already used to the "keep-going-on-error" behavior.)

We did an ad-hoc Twitter poll and the responses here were quite divided, so unfortunately I think no matter what we do here a number of users are going to be unhappy.

On the balance, I think the change will encourage better programming habits, even if it impedes certain kinds of interactive workflows.

jmcphers:
> We did an ad-hoc Twitter poll and the responses here were quite divided

Indeed:

image

The results suggest that, while it's far from unanimous and selection bias exists, the new behavior is what most people expect.

ebuhle:
> But how is your script going to work if the required package isn't installed? In addition, what if someone wants to run your script non-interactively?

It could be that the package is optional, or it could already be installed. I'm typically running scripts interactively; on the rare occasions when I `source()` a script, it's usually function definitions or something else that wouldn't throw any errors. My point wasn't to seek advice about this particular example, just to note (along with the commenters in #10391) that there are plenty of cases where the presumption that everything that follows an error must be invalid is false.

> > We did an ad-hoc Twitter poll and the responses here were quite divided
> 
> Indeed:

This has already been addressed:

> Also, jmcphers, I really appreciate the poll. That said, you didn't actually ask the question at hand. You asked what people 'thought' would happen, rather than what they would prefer to happen when executing the code. Based on the question you asked, it turns out that unless they recently updated most people's expectations are wrong.
> 
> Along another line, I suspect the audience who follows any account related to Rstudio on twitter are not representative of the Rstudio user population as a whole. thus, theres a big polling bias :)

Again, I fail to see what harm could come from making the default for the "Execution" checkbox the traditional behavior. If users want to opt in to this encouragement of better programming habits, great. Meanwhile, it would not be a breaking change for a significant number of users' legacy code.

Issue #11100, opened 2022-04-30
system file dialogs are displayed behind RStudio main window when opened


kevinushey:
Seen on Fedora 35. Also seen on Ubuntu 22 (haven't tried any other Linux distros).

For example, File / New Project / New Directory / New Project, then click on the Browse button next to "Create project as subdirectory of":

![image](https://user-images.githubusercontent.com/1976582/166089338-3ba91e72-c2d4-4870-b0b4-361a48483eb2.png)

Note the file dialog on the left-hand side displayed behind the main window. We should bring the window to the front + into focus after it's displayed.

matheustavaresdev:
This is an Electron issue and is being tracked by https://github.com/electron/electron/issues/32857 . This needs to be fixed before our task can be closed, but this should be automatically fixed as soon as Electron issues a version update on NPM with that fix.

mikebessuille:
Moving to Elsbeth - we are awaiting an electron fix, but that's not urgent for the preview due to Gary's workaround listed above.

matheustavaresdev:
https://github.com/electron/electron/issues/32857#issuecomment-1150935043 - Electron team know why this is happening but dont know how to solve. Might take a while until it's fixed. An alternative seems to be changing every file browse dialog into `<input type="file">` but this could be very time consuming.

timtmok:
There is progress upstream! We may have a fix to consume in the next electron release if all goes well with the PR that's currently open.

timtmok:
Looks like that PR stalled and a different approach is currently under investigation.

mikebessuille:
By the end of Aug 2022, if we don't have an Electron fix, we should switch the IDE to using web dialogs by default on linux (at least, for a specific set of Ubuntu versions?)

mikebessuille:
@timtmok is this in Electron yet? If not, I think we should move ahead now with our plan to use web dialogs by default on Linux - Has that already been done?  I'd like to close this ticket.

timtmok:
I'll take it on

timtmok:
I've merged a change to use web dialogs by default on Linux. Mac and Windows will use native dialogs by default.

It might be a while before there's a build since there's a lack of Windows build agents and it's holding up the entire pipeline. Daily 186 or greater should have the change.

jonvanausdeln:
Verified on 2022.11.0-daily+215
Windows 11, Ubuntu 22.04

Tested several types of projects and other paths to get file selection dialogs. All work as expected.

PR #11929, opened 2022-09-09
Use web dialogs on Linux

timtmok:
### Intent
Address #11100

### Approach
Reverts to web dialogs on Linux. The preference is hidden in the options and will be forced to use web dialogs.

### Automated Tests
None

### QA Notes
Mac and Windows should still show the preference in General -> Advanced -> Use native file and message dialogs

### Checklist

### DIFF
--- a/src/cpp/session/prefs/UserPrefsComputedLayer.cpp
+++ b/src/cpp/session/prefs/UserPrefsComputedLayer.cpp
@@ -86,6 +86,10 @@ Error UserPrefsComputedLayer::readPrefs()
    defaultRVersionJson["label"] = versionSettings.defaultRVersionLabel();
    layer[kDefaultRVersion] = defaultRVersionJson;
 
+   #ifdef __linux__
+   layer[kNativeFileDialogs] = false;
+   #endif
+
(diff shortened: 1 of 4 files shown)


gtritchie (commented):
So this prevents the use of native dialogs altogether on Linux (hides/ignores the preference)?

I suspect some users are still going to want to use the native dialogs; could we instead just change the default for Linux but keep the preference operational?

I'm hitting this all the time when using VSCode on Linux (file open dialog not showing up and I have to switch to it), so they're living with it until a fix arrives, it seems!

Plus I think it works correctly on some desktop managers.

timtmok:
> So this prevents the use of native dialogs altogether on Linux (hides/ignores the preference)?
> 
> I suspect some users are still going to want to use the native dialogs; could we instead just change the default for Linux but keep the preference operational?
> 
> I'm hitting this all the time when using VSCode on Linux (file open dialog not showing up and I have to switch to it), so they're living with it until a fix arrives, it seems!
> 
> Plus I think it works correctly on some desktop managers.

Good point. I'll add the preference back in.

timtmok:
I've switched to using the pref layers to set the default on Linux.

PR #11936, opened 2022-09-12
remove old R 2.11 workaround on macOS

kevinushey:
Some cleanup to remove code associated with a now unsupported version of R.


gtritchie (approved):
My favourite code is no code. LGTM!

PR #11954, opened 2022-09-15
Add Timer class and ITimer interface to core library.

kfeinauer:
This allows code that needs to test timeout functionality the ability to mock the timer to allow you to have control over when the timer is actually invoked, making this very useful for unit testing.

This has already been proven out in the Launcher codebase.


jeffvroom (approved):
LGTM!

PR #11956, opened 2022-09-15
Nit: standardize arg names to lowercase in Socket.hpp

AlexMapley:
### Intent
Updates `src/cpp/include/core/http/socket.hpp` with a small casing fix. Any parameter variables should not be title case by default, and we refer to similar variables as the lowercase `handler` as well.

### Automated Tests
This should be a no-op, nowhere do we ever name this argument when making calls to `asyncWrite()`.
If we introduced any bugs our compiler should catch them, given we're just renaming a couple variables.

### Checklist
- [ x If this PR adds or changes UI, the updated UI meets [accessibility standards](https://github.com/rstudio/rstudio/wiki/Accessibility)

### DIFF
--- a/src/cpp/core/include/core/http/Socket.hpp
+++ b/src/cpp/core/include/core/http/Socket.hpp
@@ -39,11 +39,11 @@ public:
 
    virtual void asyncWrite(
                      const boost::asio::const_buffers_1& buffer,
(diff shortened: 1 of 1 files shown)


kevinushey (approved):
LGTM; thank you for the fixup!

Issue #11604, opened 2022-07-12
Long lag on RStudio startup on opening a new session (Windows and Mac)


ronblum:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2022.10.0-daily+59
    OS Version      : Windows 11
    R Version       : 4.2.1

### Steps to reproduce the problem

1. Close RStudio if it's already open.
2. Start RStudio.
    - Startup takes a significant amount of time.
3. Open a project in a new session.
     - The new window takes a while to open.

### Describe the problem in detail

Starting up RStudio and opening new sessions/windows takes a significant amount of time, long enough to make it appear as if nothing is happening.

### Describe the behavior you expected

Ideally, startup and new sessions would be brisk. If that's not possible, it would be very helpful if at least a window opened with a message like "Starting up" or "Initializing Session" or something like that, or a spinning wheel, or anything that'll let the user know that the application is responding but still starting up.

- [ ] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ ] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [ ] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

DavisVaughan:
FWIW I also see this on Mac, so I don't think it is Windows specific. It takes around 9 seconds to open an RStudio Project for me. Definitely long enough for me to wonder if anything is happening

DavisVaughan:
Another follow up: This seems to only be problematic when opening a 2nd RStudio session. i.e.:

- If RStudio is completely closed down, opening a Project is reasonably fast (4 seconds or so)
- If RStudio is open, then opening a 2nd Project is slow (9 seconds)

gtritchie:
This is easy to repro in a dev environment, run first instance with `npm start` then from inside that do Session / New Session. The icon appears in the dock right away, then there's a lag. I strongly suspect (but haven't confirmed) a race condition related to the loc features, e.g. this PR https://github.com/rstudio/rstudio/pull/11062

bhattmaulik:
[...]

PR #11974, opened 2022-09-19
Workaround to Electron second instance hang

timtmok:
### Intent
Address part of #11604 

### Approach
Request instance lock after app is ready to prevent pause until user focuses window. Electron docs show to request before the app is ready but this doesn't seem to make any difference on Linux and Windows. MacOS does briefly show the dock icon since that shows when the app is ready. These seems to be a good compromise to waiting for focus when requesting the instance lock.

I don't know if this addresses the long start up though. It still seems to be long on Windows but that's using an Amazon Workspace, which is already quite slow.

### Automated Tests
None

### QA Notes
This does not necessarily address the long startup for a second window. It does remove the need to focus the second window so that it finishes start up. 

### Checklist

### DIFF
--- a/src/node/desktop/forge.config.js
+++ b/src/node/desktop/forge.config.js
@@ -42,6 +42,9 @@ const config = {
             },
           ],
         },
+        // uncoment and change these ports to launch multiple debug instances
+        // port: 3000,
+        // loggerPort: 9000,
       },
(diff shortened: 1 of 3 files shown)


kevinushey (approved):
LGTM!

PR #12018, opened 2022-09-22
Do an exact version match before resorting to findClosest

jeffvroom:
The findClosest process will return the next higher patch version, even if an exact match exists. So if you have 4.0.1, 4.0.2, and 4.0.3 and are looking for 4.0.1, it will return 4.0.2 instead. The project stores just the version number, not the home dir so we don't get an exact match earlier on.

### Intent

Might address: https://github.com/rstudio/rstudio-pro/issues/3822

### Approach

This addresses a basic logic error in the selectVersion function, used to find a matching R version. It will only change the code path when the homeDir does not match. This case is used when restoring the R version from a project where we only have the version number, not the home dir. 

### Automated Tests

I think existing automation should test this pretty thoroughly. 

### QA Notes

It would be good to have the cloud team see if this fixes their issue.

### DIFF
--- a/src/cpp/core/r_util/RVersionsPosix.cpp
+++ b/src/cpp/core/r_util/RVersionsPosix.cpp
@@ -331,10 +331,16 @@ std::vector<RVersion> enumerateRVersions(
 namespace {
 
 bool isVersion(const RVersionNumber& number,
-               const std::string& rHomeDir,
                const RVersion& item)
 {
-   return number == RVersionNumber::parse(item.number()) &&
+   return number == RVersionNumber::parse(item.number());
+}
+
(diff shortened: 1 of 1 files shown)


cm421 (approved):
LGTM!

mbaynton:
> It would be good to have the cloud team see if this fixes their issue.

We put a recent daily (`2022.12.0-daily-256`) in rstudio.cloud staging and I used it to test this out. It unfortunately did not make a difference in the behavior described in https://github.com/rstudio/rstudio-pro/issues/3822. Happy to discuss, troubleshoot, etc further. cc: @jeffvroom

PR #12019, opened 2022-09-22
read rest of stdout / stderr on process exit

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/11736.

### Approach

When a process exits, we need to check stdout + stderr once more in case any more output was produced in the interval between our last poll and when the exit happened.

Deserves a sanity check because if this is really the issue it's stunning that this is the first time we're encountering it!

Also, if so, then arguably we should be doing the same thing in PosixChildProcess.cpp:

https://github.com/rstudio/rstudio/blob/main/src/cpp/core/system/PosixChildProcess.cpp#L1126-L1165

### Automated Tests

N/A

### QA Notes

N/A

### Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -24,6 +24,7 @@
 - Support for v2 format of Quarto crossref index
 
 ### Posit Workbench
+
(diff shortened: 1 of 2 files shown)


MariaSemple (approved):
LGTM!

PR #12020, opened 2022-09-22
read stdout, stderr on process exit for posix

kevinushey:
### Intent

Addresses the same concern as https://github.com/rstudio/rstudio/issues/11736 for POSIX.

### Approach

Read stdout / stderr after seeing the process has exited, just in case any output was produced in the interval between us checking for the process exit, and the last time we polled for stdout / stderr.

### Automated Tests

N/A

### QA Notes

N/A

### Checklist


jeffvroom (approved):
LGTM!  The only question in my mind is whether readPipe might return an error in some edge case after the process has exited. In that case, we'd be calling reportError() now when previously we would have ignored that error. I don't see any reason that would happen though and I'd rather know about that edge case than ignore it if starting from scratch. The pipe needs to stay open until its closed from reader's side.

PR #12022, opened 2022-09-23
Avoid error when calling socket.cancel() method if the socket is not open

jeffvroom:
Using a call to is_open() before calling cancel.  This allows us to finish the cleanup in that method and call the error handler for the connection.  This is a speculative fix we'd like to get into a build for a customer to try. I think it's pretty safe because the only place that code is used is the connection timeout error handler and it's a simple change. 

### Intent

Addresses: https://github.com/rstudio/rstudio-pro/issues/3812

### Automated Tests

I don't think we have a way to test connection timeouts currently but we might want to investigate a "network fuzzer" or something so we can run load tests with flaky connectivity in a load balancing scenario.


cm421 (approved):
LGTM!

Issue #4178, opened 2019-01-18
session restarts with tibbles in environment crash version 1.2.1241


hrbrmstr:
This initially came up when writing packages but as I tried to isolate it I was able to reproduce it with  the following code in a fresh RStudio launch:

```r
library(tidyverse)
## ── Attaching packages ────────────────────────────── tidyverse 1.2.1 ──
## ✔ ggplot2 3.1.0     ✔ purrr   0.2.5
## ✔ tibble  2.0.0     ✔ dplyr   0.7.8
## ✔ tidyr   0.8.2     ✔ stringr 1.3.1
## ✔ readr   1.3.1     ✔ forcats 0.3.0
## ── Conflicts ───────────────────────────────── tidyverse_conflicts() ──
## ✖ dplyr::filter() masks stats::filter()
## ✖ dplyr::lag()    masks stats::lag()

ls(all.names = TRUE)
## [1] ".First"       ".Random.seed" "iris2" 

rstudioapi::restartSession()
# BOMBS
```

<img width="1402" alt="image" src="https://user-images.githubusercontent.com/509878/51392838-2ec51d80-1b04-11e9-937b-b018f997c28a.png">

session info:

```
─ Session info ──────────────────────────────────────────────────────────────────────────────────
 setting  value                                 
 version  R version 3.5.2 RC (2018-12-17 r75868)
 os       macOS Mojave 10.14.3                 <====== beta 4
 system   x86_64, darwin15.6.0                  
 ui       RStudio                               
 language (EN)                                  
 collate  en_US.UTF-8                           
 ctype    en_US.UTF-8                           
 tz       America/New_York                      
 date     2019-01-18                            

─ Packages ──────────────────────────────────────────────────────────────────────────────────────
 package     * version    date       lib source                        
 assertthat    0.2.0      2017-04-11 [1] CRAN (R 3.5.0)                
 backports     1.1.2      2017-12-13 [1] CRAN (R 3.5.0)                
 base64enc     0.1-3      2015-07-28 [1] CRAN (R 3.5.0)                
 bindr         0.1.1      2018-03-13 [1] CRAN (R 3.5.0)                
 bindrcpp      0.2.2      2018-03-29 [1] CRAN (R 3.5.0)                
 broom         0.5.1      2018-12-05 [1] CRAN (R 3.5.1)                
 callr         3.0.0      2018-08-24 [1] CRAN (R 3.5.0)                
[...]

PR #12030, opened 2022-09-24
Make sure session has been restored before running requests offline

jeffvroom:
### Intent

Addresses: https://github.com/rstudio/rstudio-pro/issues/3870

### Approach

Added a new method used by the two places we handle requests in the offline thread to be sure there's not a pending deferred session deserialization. Otherwise, we can end up running that R code off the main thread. 

### Automated Tests

See the issue for a test case that I found attached to this issue fixed a couple of years ago: https://github.com/rstudio/rstudio/issues/4178

Given that it's reproduced two bugs, it would be a good one for our regression suite.

### DIFF
--- a/src/cpp/r/include/r/session/RSession.hpp
+++ b/src/cpp/r/include/r/session/RSession.hpp
@@ -218,6 +218,9 @@ core::Error run(const ROptions& options, const RCallbacks& callbacks);
    
 // deferred deserialization of the session
 void ensureDeserialized();
+
+// returns false if there is a pending deferred deserialization of session data
+bool isSessionRestored();
       
 // set client metrics 
 void setClientMetrics(const RClientMetrics& metrics);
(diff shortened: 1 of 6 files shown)


kevinushey:
We also have some code in SessionThemes that tries to track whether deferred initialization is complete:

https://github.com/rstudio/rstudio/blob/60785042921738ec4be056634a808a56f8b55405/src/cpp/session/modules/SessionThemes.cpp#L57

Should we consider consolidating that code as well?

The `s_deferredDeserializationAction` is also not always set; e.g. it's initialized here:

https://github.com/rstudio/rstudio/blob/60785042921738ec4be056634a808a56f8b55405/src/cpp/r/session/RInit.cpp#L198-L203

(reading the code, it seems like we always set a hook but the fact that we have the null check suggests that might not have always been the case?)

Either way, it might be cleaner if we just tracked whether initialization was complete via a flag (similar to the SessionThemes solution) and had a global accessor to that value.

kevinushey:
I'm going to merge this PR as-is for now and we can consider requirements (if necessary) later.

PR #12038, opened 2022-09-26
allow users to control 'text-rendering' preference

kevinushey:
cc: @hadley 

Screen Shot 2022-09-26 at 11 23 31 AM


hadley:
THANK YOU!!!

PR #12040, opened 2022-09-27
Fixed test in code for reactivating users

jeffvroom:
### Intent

Addresses: https://github.com/rstudio/rstudio-pro/issues/3874

### Approach

One liner change of || to an && 

### Automated Tests

Let's add a test for reactivating users. We can do it by updating the postgres db directly to set last_sign_in time to an old date, then just try to login that user.


MariaSemple (approved):
LGTM!

PR #12045, opened 2022-09-27
Do not use "inset" -- too recent of a css property for QT

GCRev:
See: https://github.com/rstudio/rstudio-pro/issues/3846
OS-PR, but when this merges to Pro it will fix Pro / RDP.

### Intent

Fix the login page not being centered in RDP, which is caused by an unsupported CSS shorthand.

### Approach

Instead of `inset`, use `top`-`right`-`bottom`-`left` for absolute positioning. `inset` is a shorthand introduced in 2020 which is _after_ we stopped updating the QT versions.

### Automated Tests

This does not require automated tests. It's a layout-specific issue that only requires a one-time fix.

### QA Notes

I verified that this was working in actual RDP.

### Checklist

### DIFF
--- a/src/gwt/www/rstudio.css
+++ b/src/gwt/www/rstudio.css
@@ -26,7 +26,10 @@ html, body {
   background-color: var(--bg-col);
   color: var(--fg-col);
   position: fixed;
-  inset: 0;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
   height: 100vh;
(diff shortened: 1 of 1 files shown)


gtritchie (approved):
LGTM! Sorry you still have to worry about ancient Chromium via Qt.

GCRev:
@gtritchie `inset` is actually much more recent than I thought.

PR #12051, opened 2022-09-28
PyShiny Run App support

jcheng5:
### Intent

* Add a button for "Run App" for PyShiny apps
* Fix some issues with localhost port proxying that were exposed by PyShiny (specifically: chunked HTTP responses were being corrupted 100% of the time, and proxied WebSockets had a race condition that caused them to close 10% of the time).

To see this feature in action
1. Grab an example from https://shinylive.io/py/examples/, and save it as an `app.py` file.
2. Open that file in RStudio, and hit the "Run App" button to see it launch in a new window.
3. Arrange the app window and the RStudio window so that you can see them both, and make a trivial change to app.py. When you hit Save, you should see the app window reload itself.

### Approach

The intent was to do the "simplest possible" integration, unfortunately that ended up still being a bit complicated.

Hitting the Run App button should run the current PyShiny app at the RStudio Terminal, and launch an external browser. This needs to work on RStudio Server as well, and in order to serve multiple users simultaneously, a random port should be used by PyShiny each time.

* Add `RS_SERVER_URL`, `RS_SESSION_URL`, `RS_PORT_TOKEN`, and `rserver-url` functionality from RSP that is necessary for PyShiny to adapt its URLs to proxied URLs.
* Tell PyShiny to pick its own port (`--port 0`) and invoke the web browser `$BROWSER` when ready.
* In server mode, the `$BROWSER` is now an rpostback-browser script that causes a new browser window to be launched.

A couple of fixes to the proxy logic were required as well: 1) support for proxying chunked responses from upstream, and 2) removing a race condition in proxying WebSockets that would cause the response headers to be corrupted.

### Automated Tests

I don't know how to write automated tests for RStudio.

### QA Notes

I don't know how to build RStudio in desktop mode, so I've only tested this in server mode. It'll be important to test this in desktop mode as well.

### Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -16,6 +16,7 @@
 ### Python
 
 - RStudio attempts to infer the appropriate version of Python when "Automatically activate project-local Python environments" is checked and the user has not requested a specific version of Python. This Python will be stored in the environment variable "RETICULATE_PYTHON_FALLBACK", available from the R console, the Python REPL, and the RStudio Terminal (#9990)
+- Shiny for Python apps now display a "Run App" button on the Source editor toolbar. (Requires `shiny` Python package v0.2.7 or later.)
 
 ### Quarto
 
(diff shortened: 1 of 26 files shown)


jcheng5:
Approved by @jeffvroom, @melissa-barca, and @jgutman.

Issue #12059, opened 2022-09-29
Clicking the "Stop" button waits for queued messages


billdenney:
I occasionally have code that sends many, many messages to the console from a package that I often use.  When I press the "stop" button, it still waits for all of the queued messages to print.  Sometimes, that can take minutes, and Rstudio will ask if I want to terminate R.

It would be preferable if the option were first to send remaining messages to /dev/null and go back to the prompt.  Perhaps the UI for that would be if the "Stop" button were pressed twice or if the prompt to kill the R session were made an option would be to drop messages that are still coming in.  (I don't know if this is reasonable or possible.)

Reprex (which will make you want to restart R):

```r
forever <- function() {
  while (TRUE) {
    message("This is the message that never ends.  It goes on and on my friends.")
  }
}

forever()
```

![image](https://user-images.githubusercontent.com/10359931/193066662-b0a4a4ca-9532-4a1b-9684-2b4d445bf0ef.png)

![image](https://user-images.githubusercontent.com/10359931/193066509-81eb1a22-2788-417c-9406-2272cc069adf.png)

kevinushey:
Thanks for the bug report!

Taking a look internally, it seems like the R session itself actually _is_ getting interrupted, but RStudio is getting stuck processing the queue of accumulated messages. Perhaps we could clear out messages in the queue when an interrupt is handled?

billdenney:
Clearing out the message queue makes sense to me. If I hit the stop button, I probably already have all the information that I need.  Maybe add a message indicating that some messages have been dropped, too.

kevinushey:
With the PR in https://github.com/rstudio/rstudio/pull/12061, the behavior is improved, but not perfect -- it still takes some time for the R console to become responsive again, but it does eventually respond.

jonvanausdeln:
Tested on 2022.12.0-daily+257
Stopping does give the console back and is responsive.

However, I did see if you let the example run for a little bit, the entire IDE screen/window goes blank and the window is needs to be force closed (end process )   This happened on both Windows 11 and MacOS 12. ON windows it happened much sooner, usually within one minute. Mac I saw it after about 4 minutes of running. @kevinushey do you think this is a separate issue and should be filed separately? I'm inclined to file it, close this one as it functions as intended... thoughts?

kevinushey:
I think we can file that as a separate issue. I think we're overloading Chromium / Electron itself so much that the renderer itself is crashing, so the issue is (while related) worth considering separately.

PR #12061, opened 2022-09-29
discard pending console output on interrupt

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/12059.

### Approach

Relatively straightforward wiring of callbacks into our existing interrupt handlers:

- The `rConsoleInterrupted` callback sends a client event notifying that an interrupt was sent,
- The client processes that event eagerly, and uses that as a signal to discard any pending output

### Automated Tests

Not included.

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/12059.

### Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,7 @@
 ### R
 
 * Whether pending console input is discarded on error can now be controlled via a preference in the Console pane. (#10391)
+* RStudio now drops pending console input when the console is interrupted. (#12059)
 * Improved handling of diagnostics within pipeline expressions. (#11780)
 * Improved handling of diagnostics within glue() expressions. 
 * RStudio now provides autocompletion results for packages used but not loaded within a project.
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,7 +4,7 @@
 ### R
 
 * Whether pending console input is discarded on error can now be controlled via a preference in the Console pane. (#10391)
-* RStudio now drops pending console input when the console is interrupted. (#12059)
+* RStudio now drops pending console output when the console is interrupted. (#12059)
 * Improved handling of diagnostics within pipeline expressions. (#11780)
 * Improved handling of diagnostics within glue() expressions. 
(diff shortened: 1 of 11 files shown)


jeffvroom:
A concern I have about this fix is that it's designed for this specific test case, where all pending console output is not important.  If we use transactional semantics, when pressing 'stop', I'd expect to see any pending console output that was generated before I pressed stop and here that all gets discarded.  Perhaps the real problem here is that we can create an arbitrarily long buffer of console output events. There's a fixed cost for each one and we have no way to bound the latency debt that's created when a program is spewing output as fast as it can. 

Do we have code to merge console output events on the server side, and somehow join them together on the client if they are coming faster than we can process?  That would allow us to overall handle more console output events. 

As a second measure, we could have a max pending console output setting and discard the oldest output when more than that comes in. That way, we'd always have the most recent output right up until the 'stop' was pressed but wouldn't let the latency get too long.

kevinushey:
> Do we have code to merge console output events on the server side, and somehow join them together on the client if they are coming faster than we can process? That would allow us to overall handle more console output events.

We don't, but that's a good idea. It also raises the question of "why does it take the client so long to process console output?" and it seems like the majority of time is just spent counting lines:

Screen Shot 2022-09-30 at 9 57 00 AM

It looks like we're counting lines here using a plain regular expression, so we could almost surely do better with a plain loop using `charCodeAt()`.

https://github.com/rstudio/rstudio/blob/21efc9735599542c3e9f5799bb66e18ff6f8fb75/src/gwt/src/org/rstudio/core/client/dom/DomUtils.java#L226-L272

It's also probably slow to count lines in the text nodes themselves rather on the strings we're getting from the server; it's worth exploring if we can alleviate that cost as well.

I'll try to see if we can make progress with those before dropping potential output. Thank you!

kevinushey:
Superceded by https://github.com/rstudio/rstudio/pull/12077.

PR #12073, opened 2022-09-30
Add a startup splash screen for long loads

timtmok:
### Intent
Address #11604

### Approach
Adds a simple splash screen with a configurable delay before showing the splash (default 250ms). There is also a command-line argument (`--startup-delay` adds 5s to startup) to delay bringing up the main window so the splash can be force shown.

`splash.ts` is empty but required for webpack. If we desire, it can be used to update the splash with progress indicators, reticulating splines, etc.

First startup:
https://user-images.githubusercontent.com/9591545/193334823-603f2d88-b58b-44a9-8ef3-54a95426b52b.mov

Another instance:
https://user-images.githubusercontent.com/9591545/193334864-b45d84a2-8da2-450a-997e-842a0b040f6f.mov

There might be further improvement to showing the main window sooner by showing it before configuring the `rsession`.

### Automated Tests
None

### QA Notes
The splash can be shown on first startup or a second instance (like opening a project in a new session). 

### Checklist

https://user-images.githubusercontent.com/9591545/193334864-b45d84a2-8da2-450a-997e-842a0b040f6f.mov


NEWS.md
```diff
@@ -1,6 +1,10 @@
 
 ## RStudio 2022-10.0 "Elsbeth Geranium" Release Notes
 
+### New
+
+* Splash screen on startup (#11604)
```
kevinushey, line 6:
```suggestion
* RStudio now displays a splash screen on startup, while the R session is being initialized. (#11604)
```

src/node/desktop/src/main/application.ts
```diff
@@ -60,6 +61,7 @@ export class Application implements AppState {
   gwtCallback?: GwtCallback;
   sessionStartDelaySeconds = 0;
   sessionEarlyExitCode = 0;
+  startupDelay = 0;
```
kevinushey, line 64: Nit: it can be helpful to include the units in the name, e.g. `startupDelayMs`

src/node/desktop/src/main/args-manager.ts
```diff
@@ -131,6 +132,11 @@ export class ArgsManager {
       application.sessionStartDelaySeconds = 5;
     }
 
+    // switch for forcing the splash to show by adding a startup delay
+    if (app.commandLine.hasSwitch(kSplashForce)) {
```
kevinushey, line 136: Do we want to allow users to customize the startup delay here? (also the constant name doesn't match the flag name)

kevinushey (commented):
Overall LGTM; just minor comments / nits.

src/node/desktop/src/main/args-manager.ts
timtmok: Sounds like a good thing to add!

timtmok:
@kevinushey I've added a default startup delay if no value is given.

I've also moved the delay to before the window shows. The delay wasn't really working because of the single-threaded UI requiring `application.run()` to fully execute before it could process the delay to show the splash.

kevinushey (approved):
LGTM!

PR #12077, opened 2022-09-30
optimize handling of console output / errors

kevinushey:
### Intent

Part of https://github.com/rstudio/rstudio/issues/12059.

### Approach

This PR bundles a number of changes that should improve the responsiveness of both the Console pane and the Build pane, in the cases where we're receiving a very large amount of text.

1. We now bundle output on `stderr` in the same way we bundle `stdout`. This is necessary as R users may produce output on `stderr` via the `message()` function, so it can become overloaded in the same way that `stdout` might.

2. We also now truncate output shown in the Build pane. This is especially important for projects that compile C / C++ code, as the Build pane can become overwhelmed by warning output from the compiler.

Just to drive the second point home, this is what I see if I compile the `rstan` package with `-Weverything`:

```
kevin@MBP-P2MQ:~/r/pkg/rstan/rstan/rstan [develop]
$ R CMD INSTALL --preclean . &> output.log
$ ll output.log
-rw-r--r--  1 kevin  staff    70M Oct  3 11:33 output.log
```

### Automated Tests

N/A

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/12059.

### Checklist

### DIFF
--- a/src/cpp/session/SessionClientEventQueue.cpp
+++ b/src/cpp/session/SessionClientEventQueue.cpp
@@ -62,6 +62,7 @@ bool ClientEventQueue::setActiveConsole(const std::string& console)
       {
          // flush events to the previous console
          flushPendingConsoleOutput();
+         flushPendingConsoleErrors();
          
          // switch to the new one
          activeConsole_ = console;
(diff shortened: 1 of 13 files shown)


kevinushey:
Unit tests aren't happy with these changes, so will need to investigate.

kevinushey:
Unit tests are happier now.

kevinushey:
I realized while putting this PR together that we're already coalescing `stdout` in the console process; I think it makes sense to do the same for `stderr` here as well (since R users might write large amounts of output to `stderr()` via the `message()` function).

With this PR, the following is much more performant:

```
for (i in 1:200000) writeLines(paste(i), con = stdout())
for (i in 1:200000) writeLines(paste(i), con = stderr())
```

However, the following can still lock up RStudio:

```
for (i in 1:200000) {
  writeLines("stdout", con = stdout())
  writeLines("stderr", con = stderr())
}
```

but such patterns should hopefully be far less common.

jeffvroom (approved):
LGTM!  This is going to make a big difference!

PR #12099, opened 2022-10-04
alter systemd configuration to avoid KillMode=none

colearendt:
### Intent

Now that we have #12074 , we can relax our `KillMode=none` restriction and the extra `killall USR2` that we send to `rsession` and `rworkspaces`, trusting that they will get cleaned up properly by the process tree / supervisor.

Aside: `KillMode=none` is expressly discouraged in [the systemd docs](https://www.freedesktop.org/software/systemd/man/systemd.kill.html)

> If set to none, no process is killed (strongly recommended against!). In this case, only the stop command will be executed on unit stop, but no process will be killed otherwise. Processes remaining alive after stop are left in their control group and the control group continues to exist after stop unless empty.

Another note from `launchctl` docs at launchd.info:

> Stopping a job will send the signal SIGTERM to the process.

### Approach

- Remove `USR2` kill commands
- remove `KillMode=none` config in systemd unit files
- disregard other init mechanisms, since they are mostly artifacts of older systems we no longer support at this point (i.e. Xenial)

### Automated Tests

N/A

### QA Notes

Install on various systems, start homepage and session processes, stop/restart/offline the service and ensure:
- sessions / rworkspaces processes are stopped and cleaned up
- no "orphans" (process children of pid 1 that should not be - i.e. `rsession` and `rworkspaces`)
- starting sessions after being suspended in this way do not yell / error about an "unclean shutdown" and potentially lost files / etc.

### Checklist

### DIFF
--- a/src/cpp/server/extras/admin/rstudio-server.in
+++ b/src/cpp/server/extras/admin/rstudio-server.in
@@ -39,15 +39,11 @@ case "$1" in
 
     stop)
         daemonCmd "stop"
-        if (pidof rsession >/dev/null); then killall -s USR2 $2 $3 $4 $5 $6 $7 $8 $9 rsession; fi
-        if (pidof rworkspaces >/dev/null); then killall -s USR2 $2 $3 $4 $5 $6 $7 $8 $9 rworkspaces; fi
         ;;
 
     restart)
         testConfig || exit $?
         daemonCmd "restart"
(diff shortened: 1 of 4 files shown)


jeffvroom (approved):
LGTM!

atheriel (approved):

Issue #10719, opened 2022-03-02
resolve 'insecure content-security-policy' warnings with electron


kevinushey:
The following is printed in the developer console when launching RStudio via Electron:

> Electron Security Warning (Insecure Content-Security-Policy) This renderer process has either no Content Security
>   Policy set or a policy with "unsafe-eval" enabled. This exposes users of
>   this app to unnecessary security risks.
> 
> For more information and help, consult
> https://electronjs.org/docs/tutorial/security.
> This warning will not show up
> once the app is packaged.

timtmok:
We should review those Electron security recommendations. We do use `openExternal` and they're already commented that we should review it. Setting a CSP sounds like a good start according to the recommendation. I don't know of any other areas that might affect us but the URL navigation is the most obvious target for an exploit.

matheustavaresdev:
Yea @timtmok . I think we should review the guidelines from Electron and implement them. Perhaps create a new issue with requirements for it and also setting and testing CSP for more clarity. 

Regarding what's being reported in this issue, should we still remove those warnings from dev builds?

gtritchie:
No, the goal isn't to hide the warning on dev builds, but to understand them and implement reasonable defaults.

gtritchie:
This issue can be used to track that (specify and implement CSP).

timtmok:
From my review of how links are handled, any navigation in the main window or iframes (used for panes) go through a URL check that was ported from Qt. This allows authorized URLs to load. The other Electron security recommendations are handled already by our chosen configuration when creating windows. Windows are to be sandboxed, node integration disabled, and context isolated via preload scripts.

I've got a PR now to address setting the authorized links for the iframes. It will ensure local URLs. I think we will require another change for the version update check when it navigates to the download page. It should allow a whitelisted link to check the version (rstudio.com and posit.co). Navigating to the download page, while opening in the user's browser, should also check the link.

It may also be good to add in a caution dialog when opening external links. Users should be aware that the link is going to open their browser and allow a chance to deny it. That will help if a link was navigated unintentionally, which may be a malicious action.

timtmok:
https://github.com/rstudio/rstudio/issues/12135 created to provide a notice to a user opening an external link.

Aaron-bukasa:
Define a CSP that explicitly lists the authorized sources for scripts, etc. in your main HTML file.

<meta http-equiv="Content-Security-Policy" content="script-src 'self';">

PR #12120, opened 2022-10-07
Ensure setting urls are local

timtmok:
### Intent
Address security for #10719 

### Approach
I found this when reviewing the [Electron security guidelines](https://www.electronjs.org/docs/latest/tutorial/security). Most of it is preventing navigating to unintended URLs. We allow specific local URLs and some specific safe hosts. GWT has to set specific local URLs for some panes like viewer, presentation, and tutorial as well as Shiny.

Since Electron has to make available a way for GWT to set this in the main process, this is exposed in the renderer process through the preload scripts. This PR ensures setting these  URLs are local to prevent an attack through the preload. Without this, the IDE could be set to load an arbitrary URL. Normally, we anything disallowed in the IDE is opened externally.

### Automated Tests
Some unit tests to ensure `allowNavigation` returns false when attempting to set these specific URLs.

### QA Notes
Previewing in the panes should still work. If there is an attempt to load an external URL, it will open in the user's default browser.

### Checklist

### DIFF
--- a/src/node/desktop/src/main/desktop-browser-window.ts
+++ b/src/node/desktop/src/main/desktop-browser-window.ts
@@ -399,18 +399,38 @@ export class DesktopBrowserWindow extends EventEmitter {
   }
 
   setViewerUrl(viewerUrl: string): void {
+    const url = new URL(viewerUrl);
+    if (!isLocalUrl(url)) {
+      return;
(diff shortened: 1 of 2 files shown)


kevinushey (approved):
LGTM!

Issue #11888, opened 2022-09-01
"Show folder in New windows" not working anymore on Windows


cderv:
### System details

    RStudio Edition : Desktop [Open Source]
    RStudio Version : 2022.11.0.158
    OS Version      : Windows 10 x64 (build 22000)
    R Version       : R version 4.2.0 (2022-04-22 ucrt)

### Steps to reproduce the problem

* Open RStudio IDE in no project 
* Go to Files pane
* Click More > Show folder in a new windows

Same if RStudio IDE is in a project

### Describe the problem in detail

This won't work with an error like this (in french here)
![image](https://user-images.githubusercontent.com/6791940/187879678-fa528494-5b26-4531-ac4c-306602c2b89a.png)

![image](https://user-images.githubusercontent.com/6791940/187880198-cb7c13fc-091b-4616-a82d-e50e6cdc9e4e.png)

Basically it says that `~` is not found, so this is probably related to other issue regarding Home expansion on Windows

Notes: I have see https://github.com/rstudio/rstudio/issues/11582 but it does not seem directly related. 

### Describe the behavior you expected

File Explorer of Windows should open in the same folder as the file pane. 

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

jgutman:
English screenshot

![Image](https://user-images.githubusercontent.com/7817881/195180201-9e5f4bb5-c0fa-4836-b318-a40ea8f08602.png)

jonvanausdeln:
Verified on 2022.12.0-daily+263
Windows 11

Tested with various different paths, including usernames with special characters.  File Explorer window opens as expected.

PR #12129, opened 2022-10-11
use resolved path when passing to windows explorer event

jgutman:
### Intent
Addresses #11888 on Windows.

### Approach
Use pre-existing resolveAliasedPath since path gets passed/used directly by Windows and not RStudio

### QA Notes

Follow instructions in issue. Now yields a new Windows Explorer window open to the current directory showing in the Files Pane.

### Checklist

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/files/Files.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/files/Files.java
@@ -914,7 +914,9 @@ public class Files
    @Handler
    void onShowFolder()
    {
-      eventBus_.fireEvent(new ShowFolderEvent(currentPath_));
+      String path = server_.resolveAliasedPath(currentPath_);
+      FileSystemItem resolvedCurrentDir = FileSystemItem.createDir(path);
+      eventBus_.fireEvent(new ShowFolderEvent(resolvedCurrentDir));
    }
 
    public void onFileChange(FileChangeEvent event)


src/gwt/src/org/rstudio/studio/client/workbench/views/files/Files.java
```diff
@@ -914,7 +914,9 @@ void onSetWorkingDirToFilesPane()
    @Handler
    void onShowFolder()
    {
-      eventBus_.fireEvent(new ShowFolderEvent(currentPath_));
+      String path = server_.resolveAliasedPath(currentPath_);
+      FileSystemItem resolvedCurrentDir = FileSystemItem.createDir(path);
```
timtmok, line 918: Does the path need to be created?

kevinushey: I agree the name is confusing. The intention is that it follows the pattern for other JavaScript object wrappers in GWT, where they are created with static functions called `create()`. No actual filesystem action is associated with the create method here.

kevinushey: (if we felt refactoring for clarity was worthwhile, we could rename these like `fromDirectoryPath()` and `fromFilePath()`)

timtmok: Makes sense after reading the implementation. I'm good either way.

timtmok (approved):

Issue #10588, opened 2022-02-11
Windows Electron installer doesn't detect running Electron RStudio


gtritchie:
### System details

    RStudio Edition : Electron Desktop
    RStudio Version : Any Windows Electron package build
    OS Version      : Windows-10
    R Version       : Any

### Steps to reproduce the problem
- Install RStudio Electron on Windows
- Start RStudio
- Run the RStudio Electron installer

### Describe the problem in detail

- The installer should warn that you have RStudio running and must close it before installing/uninstalling RStudio
- It does detect the Qt RStudio, via code here: https://github.com/rstudio/rstudio/blob/b936403eb9a1f3fece55a1b66146acfc2a273a8b/package/win32/cmake/modules/NSIS.template.in.electron#L992-L999

### Describe the behavior you expected

- Should detect if either Qt-based or Electron-based RStudio is running
- We can't use the same techinque as Qt detection; the window class of an Electron app is the same as for Chrome, Edge, and likely many other applications; so, if you had downloaded RStudio via Chrome and still have that tab open, it would be detected as an instance of RStudio
- Ideally we'll detect the rstudio.exe process and key off that instead of looking for window classes/titles

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

jonvanausdeln:
Verified on 2022.12.0-daily-268 Windows Desktop

Correctly blocked when another electron instance was running. Also blocked when a QT version was already running.

PR #12134, opened 2022-10-12
Update Electron Windows installer with running instance check

timtmok:
### Intent
Address #10588 to detect if RStudio is running when installer is run

### Approach
Detects if RStudio is running by process name. This detects Qt and Electron since it is checking for a `rstudio.exe` process. The error message remains the same and the installer exits when it sees RStudio is running.

The NSIS plugin has been uploaded to the `rstudio-buildtools` S3 with read permissions set. It's v1.6 and the unicode DLL was used. The zip was repackaged so only the include and DLL are in the archive.

`install-dependencies.cmd` will be called in the Jenkins build to install the NSIS plugin.

### Automated Tests
None

### QA Notes
Running the Windows installer with either Qt or Electron version running will be detected by the installer.

### Checklist


gtritchie (approved):
LGTM! As good as NSIS scripting can look, that is.

jgutman (approved):

PR #12137, opened 2022-10-13
Fix RS_SERVER_URL having double session paths

jcheng5:
### Intent

Fix RS_SERVER_URL environment variable having two copies of `/s/{session-id}` in the URL path.
https://github.com/rstudio/rstudio-pro/issues/3595#issuecomment-1275058183

### Approach

I'm totally not sure if this is a good way to go about this, but I found `r_util::parseSessionUrl` which breaks the base URL apart into the pieces we need.

### Automated Tests

I don't have one

### QA Notes

The Shiny for Python "Run App" button needs to be tested on both RS and RSW.

### Checklist

### DIFF
--- a/src/cpp/session/SessionClientInit.cpp
+++ b/src/cpp/session/SessionClientInit.cpp
@@ -60,6 +60,7 @@
 #include <core/http/Response.hpp>
 #include <core/http/Cookie.hpp>
 #include <core/http/CSRFToken.hpp>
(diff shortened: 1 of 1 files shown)


jgutman (approved):
I'm not familiar with the `r_util::parseSessionUrl` function either -- more of a Workbench thing, but I took a look at the implementation and seems like this should do what you need. Will rely on the QA testing to further confirm post-merge!

Issue #11912, opened 2022-09-07
Save file modal on Windows does not open inside current project by default


cderv:
### System details

    RStudio Edition : Desktop [Open Source]
    RStudio Version : 2022.11.0.166
    OS Version      : Windows 10 x64 (build 22000)
    R Version       : R version 4.2.0 (2022-04-22 ucrt)

### Steps to reproduce the problem

* Open RStudio in no project
* Create a new file 
* Modal should open probably in the home folder
* Save the file. 
* Open a project 
* Repeat the step: create the file and click on save
* Modal will open and still on the previous directory, not on the project directory by default. 

### Describe the problem in detail

This is painful to try to find back the project folder from the modal

### Describe the behavior you expected

I don't think it was like that before ELECTRON, but I may be wrong. I would expect the modal to open by default in the project root folder. 

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

cderv:
@ronblum Yes, it seems I messed it up. I'll edit above the steps.

jonvanausdeln:
I was able to reproduce. It appears that the file modal in windows stays in the last location used.  As this only reproduced when I open the project (step 5) by selecting it from the recent project list.  When I use "Open Project ..." and select it in the file explorer model, the file save works correctly.

mikebessuille:
Sounds like a regression.  I think we should clear the most recent directory when we change projects (including when we create a new project) and reset it to the root of the project.

jgutman:
It also doesn't seem to open in the home folder necessarily when I first go to save a file in a new session with no project. It tries to save in my most recently used directory, which happens to be inside a project, even though that project is not active.

Issue #11088, opened 2022-04-28
File -> Open File should use current project folder as starting folder


kevinushey:
On Windows Electron, the starting directory when entering the File > Open File dialog should default to the current working directory. If a user has recently opened a file, the starting directory should be the last used directory (from the last time File > Open File was used). However, this should reset back to the default (WD) if working directory is changed, or if a new project is opened/closed (thereby changing the working directory)

ronblum:
@kevinushey Did you by chance open a file in the home directory while in a project? What I'm seeing is that Open File is always using the directory from which a file was last opened. So if I'm in one project, open a file in it, and then switch or close projects, Open File still uses the previous project directory. Which is a bug.

jgutman:
See also #11778

kevinushey:
Hmm... Here's another piece of the puzzle? I think the starting directory ultimately depends on the R session's working directory, which I didn't capture in my initial report.

On macOS, the starting directory used by the system dialog in File -> Open File seems to depend on the R session's working directory. If I'm working in a project, but run `setwd("~")`, then File -> Open File uses the home directory. This is true for both Electron and Qt builds.

However, it seems like this doesn't occur in the Electron builds of RStudio on Windows -- there, the project directory is always used.

kevinushey:
It also seems like (if the working directory is not changed) then the last-used directory in the File -> Open File dialog is preserved on future file open attempts (which seems okay). Note that opening files via e.g. the Files pane in RStudio doesn't affect that behavior / state though.

jgutman:
See revised description. On windows the last used working directory also seems to persist between sessions, which doesn't seem to be the case on Mac.

billdenney:
I am a "me too" on this issue.

One thing that I don't see in the list is when multiple sessions are open, the File > Open directory seems to be shared between all sessions (which I would not prefer).  I like the way that it used to work which I will try to describe below.  I'm going to use the words "working directory" to mean what is shown with "File > Open" or "File > Save As...".

* Starting working directory is the same as the R session directory.  If you open a file by double-clicking from explorer, it is the directory of that file.
* Each session tracks its own working directory.
* `setwd()` in R changes the working directory.
* Manually moving to a new directory (I open a file from "c:\tmp" while I'm doing most of my work in "c:\mywork") changes the working directory without changing R's `setwd()`.

jonvanausdeln:
Verified on 2022.12.0-daily+270
windows 11

[...]

PR #12152, opened 2022-10-14
fix initial save/open directory for windows

jgutman:
### Intent

Addresses #11912 (save dialog) and #11088 (open dialog)

### Approach
use normalizeSeparatorsNative to get platform specific path separators


timtmok (approved):

jgutman:
For some reason the default file extension also doesn't seem to populate here on Windows, but that's not even part of the original issues linked here so I'm going to ignore that for now.

Issue #11824, opened 2022-08-24
Can't create new project from Existing directory on Windows 11


cderv:
### System details

    RStudio Edition : Desktop [Open Source]
    RStudio Version : 2022.11.0.135
    OS Version      : Windows 10 x64 (build 22000)
    R Version       : R version 4.2.0 (2022-04-22 ucrt)

### Steps to reproduce the problem

* Open RStudio IDE
* Create new folder in the users Documents directory. For example `test_project`
* Click on File > New Project > Existing Directory 
* Browser and select the newly created directory
* You should have now something like 
   ![image](https://user-images.githubusercontent.com/6791940/186387464-f471655d-16a9-476f-8aae-8d62902a277b.png)

  _Small notes: I think in previous release version, it was a `~` and not the absolute path shown in this windows. Hint for the issue ?_
* Click Create project. This is the error you'll get
   ![image](https://user-images.githubusercontent.com/6791940/186387610-7c877c80-6fe4-4d65-ad14-63a8020396fd.png)

### Describe the problem in detail

New project can't be created from existing directory. 

In addition, this is what we get created inside the selected directory 

````
> fs::dir_tree("test_project/")
test_project/
└── Users
    └── chris
        └── Documents
````

`test_project` was empty, but now it is filled with some files from the whole path. It is like the `C://` was not correctly handle by the IDE. 

### Describe the behavior you expected

Project should be created correctly, and without folder to be created wrongly.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

**Comment by kevinushey** [comment]
I can reproduce. If you look in the request log you'll see something like:

```
== REQUEST ======
{
  "method": "create_project",
  "params": [
    "C:\\Users\\kevin\\scratch\\example/C:\\Users\\kevin\\scratch\\example.Rproj",
    null,
    null,
    null
  ],
  "clientId": "33e600bb-c1b1-46bf-b562-ab5cba070b0e",
  "clientVersion": ""
}

== RESPONSE ======
{
  "error": {
    "code": 100,
[...]

PR #12165, opened 2022-10-17
Fix new proj from existing directory on Windows

jgutman:
also normalize client paths on Windows automatically

### Intent
Addresses #11824

### Approach

- Extract project name (directory name) from absolute directory path to avoid path duplication
- Normalize path separators on Windows

### Automated Tests
No

### QA Notes
See issue reprex

### Checklist

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/files/FileSystemItem.java
+++ b/src/gwt/src/org/rstudio/core/client/files/FileSystemItem.java
@@ -22,6 +22,8 @@ import org.rstudio.core.client.regex.Pattern;
 import org.rstudio.studio.client.common.filetypes.FileIcon;
 import org.rstudio.studio.client.common.filetypes.FileIconResources;
 import org.rstudio.studio.client.common.vcs.StatusAndPathInfo;
+import org.rstudio.core.client.BrowseCap;
+
 
 import com.google.gwt.core.client.JavaScriptObject;
 
@@ -36,11 +38,13 @@ public class FileSystemItem extends JavaScriptObject
 
    public static FileSystemItem createDir(String path)
    {
+      path = normalizeSeparatorsNative(path);
       return create(path, true, -1, 0);
    }
 
    public static FileSystemItem createFile(String path)
    {
+      path = normalizeSeparatorsNative(path);
       return create(path, false, -1, 0);
    }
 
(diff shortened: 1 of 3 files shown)


timtmok (approved):
So it turns out Electron on Windows requires the escaped backslash and GWT requires slash? Wonderful.

src/gwt/src/org/rstudio/core/client/files/FileSystemItem.java
```diff
@@ -288,6 +303,15 @@ public final int compareTo(FileSystemItem other)
 
    }
 
+   private final static String normalizeSeparatorsNative(String path)
```
kevinushey, line 306: Shouldn't this be called `normalizeSeparators()` or `normalizeSeparatorsPosix()`? The name implies you are using native separators here (ie, using `\` on Windows) but the behavior here is the opposite.

src/gwt/src/org/rstudio/core/client/files/FileSystemItem.java
```diff
@@ -132,6 +136,17 @@ public final FileSystemItem getParentPath()
       }
    }
 
+   public final String getRootPath()
```
kevinushey, line 139: Isn't this the same as `getName()`? (except that it doesn't handle paths like `/foo/bar/`)

I also don't think this is the right name for such a method; the "root" of a path is usually either `/` or `C:/` or `//server/blah`.

jgutman: You're right this is a bad name -- I'm open to suggestions here for better names. I kind of felt the name wasn't right that's why Ieft the comment here. But it's not the same as getName() in many cases. 

The idea is we need to handle paths like `foo/bar/baz`, and need the inverse of `getParentPath()` so `getParentPath('foo/bar/baz')` will return `foo/bar`, we want a method that returns `baz`, basically the last part of the path (either the immediate dir name if a dir or the file name if a file). 

Needed this because basically, when creating the Rproj file, we took the dir name: `foo/bar/baz` and added `.Rproj` to make it `foo/bar/baz.Rproj` and then combined with the directory name which resulted in paths like `foo/bar/baz/foo/bar/baz.Rproj` instead of `foo/bar/baz/baz.Rproj`. This didn't seem right, so just needed a method to pull out the last part of the path if the absolute path is passed in.

src/gwt/src/org/rstudio/core/client/files/FileSystemItem.java
jgutman: Yes, you're right, I can rename this

src/gwt/src/org/rstudio/core/client/files/FileSystemItem.java
jgutman: Actually now realizing that getName which calls getNameFromPath was probably failing and returning the full path only because of Windows using the wrong separators `\` instead of `/` and with fixing the separator issue with `normalizeSeparators` it probably should have worked correctly, but I fixed the path issue before fixing the issue of the separators. I'll clean this up because makes sense now -- getName should probably have worked if the path was normalized, but with Windows style paths it was returning the full absolute path resulting in our yucky duplicated path issues

kevinushey: Thanks! That all makes sense to me.

Issue #12167, opened 2022-10-17
FileLogDestination can deadlock when rotating files


kfeinauer:
### System details

    RStudio Edition : OS/Pro
    RStudio Version : Current

### Steps to reproduce the problem

Seen in Launcher, but occurring in common code on some systems where `getLastWriteTime` fails. Not all systems exhibit this behavior, so deferring to @MariaSemple for better repro steps.

### Describe the problem in detail

Deadlock leading to process freeze wherever logs are written to file.

PR #12168, opened 2022-10-17
Fixed bug where failing to delete a rotated log file could cause a log line to be emitted, causing deadlock.

kfeinauer:
### Intent

Fix https://github.com/rstudio/rstudio/issues/12167

This is because the call to `FilePath::getLastWriteTime` can fail and end up emitting a log line, which causes recursion in the logging code leading to attempting to re-acquire a mutex that was already acquired.

### Approach

The fix is to make sure we aren't logging by adding a new method overload that returns the error instead of logging it.

### Automated Tests

Existing unit tests still pass.

### QA Notes

Check with @MariaSemple for better repro steps. Only repros in certain environments.

### DIFF
--- a/src/cpp/shared_core/FileLogDestination.cpp
+++ b/src/cpp/shared_core/FileLogDestination.cpp
@@ -350,7 +350,9 @@ struct FileLogDestination::Impl
       {
          // Check to see if the logfile is stale
          // If so, we will delete it instead of rotating it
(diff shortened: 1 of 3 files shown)


MariaSemple (approved):
LGTM!

jeffvroom (approved):
LGTM!

PR #12175, opened 2022-10-18
Fix bug where zlib decompression can hang on certain invalid string sequences

kfeinauer:
### Intent

In some cases, zlib can indicate that no progress has been made and no progress will ever be made again as there is no more data in the buffer to process.

### Approach

In this case, we should error out.

Addresses https://github.com/rstudio/launcher/issues/497.

### Automated Tests

Added new zlib unit test to test this specific case.

### QA Notes

N/A - covered by unit tests.


MariaSemple (approved):
LGTM!

jeffvroom (commented):
LGTM!

Issue #10945, opened 2022-04-09
Allow system-wide Rprofile.site file on installDependencies


Enchufa2:
Currently, installation of package dependencies (e.g., when you open a file that uses an R package that is not installed, a prompt says "Package %s required but is not installed" and you have the option to click "Install", which launches a job) uses a vanilla session:

https://github.com/rstudio/rstudio/blob/ce2869b15b1c7d4d6221b5edd9efd6a049a7f139/src/cpp/session/modules/SessionDependencies.cpp#L618-L621

I find this too restrictive, because `--vanilla` implies `--no-save`, `--no-restore`, `--no-init-file` and `--no-environ`, but also `--no-site-file`. It would be nice to use all but the last one for this (could be e.g. `R_PROCESS_VANILLA_USER`), so that system-wide R options apply to these installations too.

kevinushey:
Can you elaborate a bit more?

Because user + site profiles have already been executed, the requisite state (e.g. the library paths) should already be set in the parent R session (or, the user may have changed library paths separately), and so it's RStudio's responsibility to ferry that current state to the launched sub-process without having that state again modified by a user or site profile.

I'm guessing this might not be true for certain R options set in the child sessions that you're hoping would be reflected though; is that the case? If so, it might be sufficient for us to set `R_PROFILE_USER` to an empty (or non-existent) values so the site profile can be run without the user profile.

Enchufa2:
Any other options, apart from the library paths, set in the `R.home("etc")`/`Rprofile.site` file are not carried out to the child session. I see good reasons to avoid the user's `.Rprofile`, but not to avoid the system-wide `Rprofile.site` file. Or maybe an option to enable this.

Enchufa2:
Still relevant.

PR #12176, opened 2022-10-18
allow use of site profile in install dependencies

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/10945.

@Enchufa2, let me know if this looks okay to you.

### DIFF
--- a/src/cpp/session/SessionAsyncRProcess.cpp
+++ b/src/cpp/session/SessionAsyncRProcess.cpp
@@ -80,8 +80,16 @@ void AsyncRProcess::start(const char* rCommand,
    std::vector<std::string> args;
 
    if (rOptions & R_PROCESS_VANILLA)
+   {
       args.push_back("--vanilla");
-
+   }
+   else if (rOptions & R_PROCESS_VANILLA_USER)
+   {
+      args.push_back("--no-save");
+      args.push_back("--no-restore");
+      args.push_back("--no-init-file");
+   }
+   
    if (rOptions & R_PROCESS_NO_RDATA)
    {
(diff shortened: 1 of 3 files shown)


src/cpp/session/SessionAsyncRProcess.cpp
```diff
@@ -80,8 +80,16 @@ void AsyncRProcess::start(const char* rCommand,
    std::vector<std::string> args;
 
    if (rOptions & R_PROCESS_VANILLA)
+   {
       args.push_back("--vanilla");
-
+   }
+   else if (rOptions & R_PROCESS_VANILLA_USER)
+   {
+      args.push_back("--no-save");
```
kevinushey, line 88: What about Renviron files? `R --help` has:

```
$ R --help | grep no-environ
  --no-environ          Don't read the site and user environment files
```

Do we want to:

1. Read all Renviron files;
2. Read only the site Renviron file,
3. Read none of the Renviron files (since we're inheriting environment variables anyway)

Enchufa2: There are two site Renviron files: `R_HOME/etc/Renviron` and `R_HOME/etc/Renviron.site`. `--no-environ` blocks `R_HOME/etc/Renviron.site` and the user Renviron, but `R_HOME/etc/Renviron` is always read and cannot be blocked. So we want `--no-environ` because it blocks the user Renviron, but still the core environment variables are always exported.

On the other hand, there's only one site Rprofile, so we want just `--no-init-file`, as you just did, so LGTM. Many thanks for this! :)

Issue #12205, opened 2022-10-21
Link to add new publishing account always assumes account type is Connect


mbaynton:
### System details

    RStudio Edition : Desktop
    RStudio Version : Latest Dailies
    OS Version      : Ubuntu Focal, Mac
    R Version       : 4.2

### Describe the problem in detail
This relates to the Publish dialog that is shown when publishing an Rmd document. It's similar to #12204 .

If I publish a new Rmd file and select Posit Cloud as the destination where I want to publish to, I get this window:
![image](https://user-images.githubusercontent.com/3026002/197285747-556fb933-38d8-42d9-8c4e-6da02144e823.png)

If I choose Add New Account in the above window, I get the window to create a new Connect account, not a new Cloud account
![image](https://user-images.githubusercontent.com/3026002/197285876-e1838fe8-756c-4357-9bff-31dcf187c965.png)

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
Verified that this works correctly in RStudio Desktop 2022.12.0-daily+299 on MacOS 13.
- Publishing a source document in the IDE
- Publishing a rendered document in a window
- Publishing a rendered document in the Viewer Pane
- Setting up <kbd>Global Options -> Publishing</kbd>

(Granted, these all go through the same flow but still worth verifying.)

PR #12219, opened 2022-10-24
If Posit Cloud compatible content, then add new account wizard should…

jgutman:
… allow users to add a new cloud account of their choice

### Intent

Addresses #12205 

### Approach

Click Publish Document on an Rmd. User is shown the following UI:
image

User selects Posit Cloud. Publish wizard appears showing the list of accounts compatible with this content:
image

User doesn't have a Posit Cloud account set up, or wants to set up a new one, so user clicks on Add New Account. Here's what changed: 
image

The Connect Account now shows Cloud accounts in addition to Connect to which the user may want to connect.

User selects Posit Cloud
image
And gets the instructions/link for setting up a new Posit Cloud account as usual.

This isn't a perfect approach (for example it allows the user to connect a new shinyapps.io account, even though those accounts will be filtered out because they aren't compatible with rmarkdown publishing) but it is better than the current state in which a user would not be able to connect a new Posit Cloud account at all from this screen. This I think accomplishes the main purpose and is probably the best that is feasible in the timeframe for the next release, without a major refactoring of our publishing code.


jgutman:
cc @mbaynton

Issue #11783, opened 2022-08-18
pasted image in visual editor are broken


battmanux:
# to reproduce 
paste an image in qmd visual editor. It will create a broken link with href="data:image/png;base64,iVBOR..."
Image will not be displayed nor rendered as it is handled as a local file

# workaround (part 1)
during pandoc rendering an extention is added to the uri. 
with metadata default-image-extension: '' pandoc will render the image as expected

# workaround (part 2)
the regexp identifying uri (see ImagePreviewer.java#L514) does not capture data: as it should (it only capture xxx://)
replacing data: with data:// fools the regexp and is well understood by the chrome browser
https://github.com/rstudio/rstudio/blob/e07e7bcd5133850c4cff145c343cfe7103e642dc/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/ImagePreviewer.java#L514

# expected behavior
either handle data: uri correctly or convert data: to a local file while pasting

ronblum:
@battmanux Thank you for raising this issue!

I can reproduce this issue in:
- RStudio Server 2022.07.1+554 on RHEL via Safari on MacOS 13
- RStudio Server 2022.11.0-daily+187 on Ubuntu 20.04 via Chrome on MacOS 13
- RStudio Desktop 2022.07.1+554 (Electron) on MacOS 13
- RStudio Desktop 2022.11.0-daily+187 on MacOS 13

However, it works as anticipated in:
- RStudio Desktop 2022.07.1+554 (Qt) on MacOS 13

So in addition to this being an issue in RStudio Server, this is a regression in Electron.

jgutman:
Discussed this was JJ and his opinion was it may not be worth trying to get this to work right now/for this release. This is likely more complicated and something needs be intercepted/transformed on the Electron side rather than merely fixing the function linked to above by JJ.

mikebessuille:
I dunno if I agree - pasting an image into a visual editor is pretty central functionality.  having it broken in a general sense in the electron version is a big regression, IMO.

jgutman:
Note: URIs are missing on both Electron and GWT side

kevinushey:
FWIW, a simple workaround would be to switch `data:` to `data://` in the code the OP linked to in ImagePreviewer.

I'd recommend we at least try the simple fix before punting on this issue.

battmanux:
[...]

PR #12224, opened 2022-10-25
Fix Electron clipboard image paste

timtmok:
### Intent
Address #11783 

### Approach
Add Electron specific clipboard check in image paste event handling. Qt adds images to the clipboard event as `application/qt-image`. Electron seems to add images as `image/png` to data items, which can be checked if it is an `image/png`.

### Automated Tests
The random string code was removed from a unit test util. There's a random string util in the core code that is identical. So some unit test refactoring was done to fix the imports.

### QA Notes
This will work on all platforms. It does not address the server issue.

### Checklist

### DIFF
--- a/src/gwt/panmirror/src/editor/src/nodes/image/image-events.ts
+++ b/src/gwt/panmirror/src/editor/src/nodes/image/image-events.ts
@@ -71,6 +71,19 @@ function imagePaste(ui: EditorUI) {
         event.preventDefault();
         return true;
       }
+
+      // Electron specific clipboard check
+      for (const item of clipboardEvent.clipboardData.items) {
+        if (item.type === 'image/png') {
+          ui.context.clipboardImage().then(image => {
+            if (image) {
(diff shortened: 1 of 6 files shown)


src/gwt/panmirror/src/editor/src/nodes/image/image-events.ts
```diff
@@ -71,6 +71,19 @@ function imagePaste(ui: EditorUI) {
         event.preventDefault();
         return true;
       }
+
+      // Electron specific clipboard check
+      for (const item of clipboardEvent.clipboardData.items) {
```
jgutman, line 76: Is there a specific reason for doing it this way instead of just adding another 
`else if (clipboardEvent.clipboardData.types.includes('image/png')` ?

jgutman (approved):

kevinushey (approved):
LGTM; awesome work chasing that down!

PR #12225, opened 2022-10-25
Reduce severity of sentry upload failures

skyeturriff:
### Intent

Issues with sentry uploads have been causing build failures much more consistently and have prevented a successful pro build for over a week. This change lowers the urgency of sentry upload failures so that builds are marked as unstable rather than failing out completely.

### Approach

Wrap the sentry upload stage in a try-catch and set the build result to unstable on failure.

I replayed a previous build, making [these changes directly in the Jenkinsfile](https://build.rstudioservices.com/job/IDE/job/pro-pipeline/job/main/1084/replay/diff). This resulted in getting the builds past the sentry upload failures (as they are now just marked unstable).

One thing to note - the "upload debug symbols" stage is marked as a "pass" (green checkmark) in the [Blue Ocean view](https://build.rstudioservices.com/blue/organizations/jenkins/IDE%2Fpro-pipeline/detail/main/1084/pipeline/2747), even though the upload failed and so it should be marked as unstable. I don't know if this is because I am setting `currentBuild.result` rather than calling `unstable()` (I went with that as it seemed to be what other code in that section was defaulting to).

### Automated Tests

n/a

### QA Notes

n/a

### Checklist

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -157,7 +157,13 @@ def sentry_upload(type, flavor) {
   withCredentials([string(credentialsId: 'ide-sentry-api-key', variable: 'SENTRY_API_KEY')]){
     // timeout sentry in 15 minutes
     timeout(activity: true, time: 15) {
-      sh "cd package/linux/build-${flavor.capitalize()}-${type}/src/cpp && ../../../../../docker/jenkins/sentry-upload.sh ${SENTRY_API_KEY} ${sentryUploadRetryLimit}"
+      try {
+        // attempt to run sentry uplaod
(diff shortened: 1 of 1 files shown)


jeffvroom (approved):
LGTM!  My only thought is whether it would be helpful to add something like:

echo 'Error running sentry-upload: ' + err.toString()

inside of the catch blocks so we might get a log of why it failed. We could always add that as needed though and I'm not 100% sure on the syntax there so maybe just merge it?

skyeturriff:
Merged and I will look into adding @jeffvroom 's  suggestion!

Issue #12130, opened 2022-10-11
IDE goes blank and hangs in a long-running function writing to console


jonvanausdeln:
### System details

    RStudio Edition : Desktop
    RStudio Version :  2022.12.0-daily+258
    OS Version      : Windows 11 & MacOS 12.4
    R Version       : 4.2.1

### Steps to reproduce the problem
Discovered when testing #12059 see https://github.com/rstudio/rstudio/issues/12059#issuecomment-1275301292. A long running function that is writing to the console will make the ide window go blank and has to be force quite.

1. Execute R script
```
forever <- function() {
  while (TRUE) {
    message("This is the message that never ends.  It goes on and on my friends.")
  }
}

forever()
```
2. Let it run for a few minutes

### Describe the problem in detail

The IDE window will go blank and hang.  Process needs to be killed to proceed.
On windows this took less than 1 minute, on MacOS it was around 3 minutes.  This also might be machine resource dependent.

### Describe the behavior you expected

Code to continue executing and IDE to remain responsive

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

timtmok:
The issue appears to be how the content is displayed in the ACE editor.

With Qt, it creates new HTML elements for each message output. The maximum message output that is created in the HTML is 1000 nodes (ie. 1000 lines). So it will scale with large output.

Electron doesn't behave the same for some reason. It appends the message output to the existing node and uses a line break. This bloats the memory and eventually overwhelms the renderer.

This should be fixable by getting the output to render as a new node. Then it's probably ACE editor that handles the scrollback and limiting the nodes to prevent the memory bloat.

timtmok:
Looks like Electron does cap the nodes to 1000. It actually processes the events much faster than Qt and `VirtualConsole.java` aggregates the output to prevent updating the DOM too quickly. This was done to address: https://github.com/rstudio/rstudio/issues/5518

[...]

PR #12258, opened 2022-10-31
Lower limit for max line console queue

timtmok:
### Intent
Partially address #12130 

### Approach
Lower the max line count for the console output queue. This seems to not crash the render thread.

Electron seems to process the R console output callbacks a lot faster than Qt. I don't know if the rsession is faster at processing or if there's less overhead with Electron allowing GWT to queue the messages faster. In any case, the renderer is overwhelmed by the output and the many updates to the DOM. It keeps updating the one node until it becomes too large for the renderer. This lower limit results in a smaller node and prevents the render thread from crashing.

### Automated Tests
None

### QA Notes
This is the same issue as reported in #5518 but with a lower limit.

### Checklist


timtmok:
I guess the proper fix is addressing the tech debt and implementing virtualization as Kevin mentioned in https://github.com/rstudio/rstudio/pull/5528#issuecomment-540114798

kevinushey (approved):
LGTM; nice work chasing this down!

PR #12259, opened 2022-10-31
Allow session suspend when Quarto job is running

midnightprioriem:
### Intent

Addresses rstudio/rstudio-pro#3889. Adds the ability to mark jobs as transient and child processes as `allowParentSuspend` so that certain jobs and child processes, like the Quarto preview job, won't prevent session suspension when session idle suspension is triggered.

### Approach

- Add the `allowParentSuspend` variable to `ProcessOptions`, defaulting to `false`
- Modify `hasActiveChildren` to `haveDurableChildren` and include `allowParentSuspend` in the logic so that child processes that have `allowParentSuspend` marked `true` don't count towards preventing suspension
- Add the `transient` tag option to the `Jobs` API, which will make a job suspendible even if it's running
- Update the Session Quarto Job to include the new `transient` tag
With the above changes, sessions will be suspendible even if the Quarto preview is running.

### Automated Tests

N/A

### Checklist

### DIFF
--- a/src/cpp/core/include/core/system/ChildProcess.hpp
+++ b/src/cpp/core/include/core/system/ChildProcess.hpp
@@ -82,6 +82,9 @@ public:
    // Has this process generated any recent output?
    virtual bool hasRecentOutput() const;
 
+   // Check if process allows for parent process to be suspended
+   bool allowParentSuspend() const;
+
 protected:
    Error run();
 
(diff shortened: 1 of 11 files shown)


jeffvroom (approved):
LGTM!  Nice work. This is very clean and easy to read and I think gives us a way to differentiate between transient and durable jobs.  I checked the job launcher and it does not use the hasActiveChildren api. It seems likely no one else is using that to me so I think replacing it is ok.

melissa-barca (approved):
LGTM!

PR #12261, opened 2022-11-01
Specify sentry-cli version to install

skyeturriff:
### Intent

The Upload to Sentry build stage fails intermittently. This PR specifies the version of sentry-cli installed by the dependency scripts allowing us to know which version is being used by the build containers and to easily update the version when needed. The latest is 2.8.

### Approach

For Linux, the existing script should be grabbing latest. Given that this layer of the build container image is cached, it's uncertain what version is currently being used. I've set the `SENTRY_CLI_VERSION` env var, which is used by the script retrieved by the curl command.

For Windows, the version was hard coded into the download URL. I've extracted this out into an env var.

I've also added some echos to confirm the version that is installed... this may not may not be generally useful so if it seems too noisy it can be removed!

### Automated Tests

n/a

### QA Notes

n/a... I ran the scripts in my Linux and Windows VM to confirm it executes ok


jeffvroom (approved):
LGTM!

timtmok (approved):
Nice! I'll merge it now since builds are quiet in the morning.

Issue #12256, opened 2022-10-31
Restarting RStudio launches many external windows on the restart of R


jthomasmock:
### System details

    RStudio Edition : Desktop
    RStudio Version : Version 2022.12.0-daily+296 (2022.12.0-daily+296)
    OS Version      :  macOS Monterey
    R Version       : 4.2.0

### Steps to reproduce the problem

- Generate HTML such as a `gt` table in RStudio
- Might throw a Javascript exception (?)
- Restart the R session with Cmd + Shift + 0
- 
https://user-images.githubusercontent.com/29187501/199028857-6fe86fb6-6f2c-44e1-aba9-d230fe29ca87.mp4

### Describe the problem in detail

Generating HTML such as a `gt` table might throw a JavaScript exception and upon restart of the R session will send approximately 20 broken duplicates of the requested content to the default browser.

### Describe the behavior you expected

HTML content should show up in the viewer pane and a restart should clear the environment and any "promised" HTML content.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ ] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

mikebessuille:
@timtmok We should at least look at this for Elsbeth; Jackie thinks this might be related to something you were already looking at.

timtmok:
I've managed to reproduce. Looks like it requires running the code once before restarting R.

It looks like the `allowNavigation` check is called on web request. That may trigger opening the external browser if the URL is not on the safe list. This should be moved out to be handled by the caller and only check if the navigation is permitted.

timtmok:
The repo step requires the View pane to be showing content before restarting R. The view tries to refresh the preview with a new URL but has not yet added it to the safe list. By default, URLs not on the safe list will be opened externally and the retries causes multiple open events.

jonvanausdeln:
Verified on Desktop 2022.12.0-preview+316, Windows 11 and MacOS 12.6

With html (similar to initial repro) in the viewer, restarted R numerous times. Behaved as expected without opening many external browser windows/tabs.

PR #12274, opened 2022-11-02
Electron allowNavigation browser spawn fix

timtmok:
### Intent
Address #12256 

### Approach
When the R session restarts, it creates a new URL for the Viewer pane. GWT does set the viewer URL correctly so it is fine to load. However, the pane loads the old URL one last time just as the DOM changes the iframe to the new URL. This triggers an `allowNavigation` check, which fails. The failure causes the URL to open in an external browser.

This is a pretty quick fix to the issue but it might be simpler to allow all localhost URLs at this point. Or figure out why the old URL is loaded one last time (actually multiple times because multiple tabs open).

If Electron ever supports `will-frame-navigate` then the localhost check can be done only for iframes. I think it's still important to check the base URL's port to prevent a different UI running locally to be loaded in the IDE.

### Automated Tests
None

### QA Notes
Show anything in the Viewer pane. As long as there is content showing when R is restarted, it will load the old URL and cause the external windows to open.

### Checklist

### DIFF
--- a/src/node/desktop/src/main/desktop-browser-window.ts
+++ b/src/node/desktop/src/main/desktop-browser-window.ts
@@ -326,7 +326,11 @@ export class DesktopBrowserWindow extends EventEmitter {
 
     if (this.options.allowExternalNavigate) {
       return true;
-    } else if (isSafeHost(targetUrl.hostname)) {
+    } else if (isSafeHost(targetUrl.hostname) || isLocal) {
+      // check isLocal until it is determined why restarting R
+      // causes the old preview URL to load, even after the DOM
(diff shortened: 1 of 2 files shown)


kevinushey (approved):
LGTM!

kevinushey:
> If Electron ever supports will-frame-navigate then the localhost check can be done only for iframes. I think it's still important to check the base URL's port to prevent a different UI running locally to be loaded in the IDE.

From what I understand, the PR is basically ready but has now stalled 😞 

https://github.com/electron/electron/pull/34418

I'm not sure if there's anything we can do to help get the PR over the finish line...

Issue #12241, opened 2022-10-27
Rename rstudio.cloud to Posit Cloud in publishing


ronblum:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2022.12.0-daily+291
    OS Version      : MacOS 13
    R Version       : 4.2.1

### Steps to reproduce the problem

1. Set up publishing to Posit Cloud using https://rstudiopbc.atlassian.net/wiki/spaces/ENG/pages/125403433/Cloud+Publishing#R-Based-Outputs
2. <kbd>Tools -> Global Options -> Publishing</kbd>
3. <kbd>Connect -> rstudio.cloud</kbd> (should be "Posit Cloud")

### Describe the problem in detail

The service is called `rstudio.cloud`.

<img width="527" alt="cloud_setup_rename" src="https://user-images.githubusercontent.com/31009952/198354577-936dfb0e-cb39-483a-a996-cacd4daa1bd8.png">

### Describe the behavior you expected

The service should be called `Posit Cloud`.

Triage note: Elsbeth Geranium for rebranding purposes.

- [ ] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ ] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [ ] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
Verified in RStudio Desktop 2022.12.0-preview+314 on MacOS 13.

PR #12277, opened 2022-11-02
Posit cloud option flag

jgutman:
### Intent

Addresses [rstudio-pro#3994](https://github.com/rstudio/rstudio-pro/issues/3994)
Also addresses [#12241](https://github.com/rstudio/rstudio/issues/12241)

### Approach

Untangle logic between a new Enable Cloud publishing UI from the pre-existing Enable RStudio (Posit) Connect publishing UI (renamed the option while I'm at it)

Check behavior of that option in the new account wizards as well as the Rmd publishing wizard 

### Automated Tests

None

### QA Notes

For effective testing, we should test 
1) When publishing to Connect is enabled and publishing to Cloud is disabled (the default)
2) When publishing to Connect is disabled and publishing to Cloud is enabled
3) When publishing to both Connect and Cloud are disabled

Option 3 should test with a shiny app and an rmarkdown doc
Options 1 and 2 should test with a shiny app, a plumber api, and an rmarkdown doc, ideally

Option 4 with both publishing options enabled can be tested, but I think will be redundant


src/gwt/src/org/rstudio/studio/client/rsconnect/ui/NewPositCloudPage.java
```diff
@@ -29,7 +28,7 @@
 {
    public NewPositCloudPage()
    {
-      super(RSConnect.CLOUD_SERVICE_NAME,
+      super("Posit Cloud",
```
jgutman, line 31: We need to hardcode this because RSConnect.CLOUD_SERVICE_NAME is `rstudio.cloud`, and will eventually be changed to `posit.cloud` but we want `Posit Cloud` with capitalization and space as a display name

timtmok (approved):

PR #12279, opened 2022-11-02
Remove python dependency from electron test

timtmok:
### Intent
The `browser-window-test` was failing on RHEL9 arm64 since it doesn't have python 2 on it. Python 3 is there but probably requires calling `python3`. This should fix it for that platform and hopefully address the intermittent failures on other platforms if the server doesn't come up fast enough.

### Approach
Use node's http module to create a listen server. No need to spawn another thread so it will be ready to accept requests when trying to load the URL.

The chosen port also happens to be the same as the GWT dev server. So it would seem that it passes every time if you have it running and wouldn't encounter an intermittent failure if the Python server didn't come up fast enough. It's been changed to avoid the conflict.

### Automated Tests
Unit test passes locally

### QA Notes
n/a 

### Checklist

### DIFF
--- a/src/node/desktop/test/unit/main/browser-window.test.ts
+++ b/src/node/desktop/test/unit/main/browser-window.test.ts
@@ -17,25 +17,17 @@ import { describe } from 'mocha';
 import { assert } from 'chai';
 
 import { BrowserWindow } from 'electron';
-import { execSync, spawn } from 'child_process';
+import * as http from 'http';
 
 describe('BrowserWindow', () => {
 
   it('fires events in the expected order', async () => {
+    const server = http.createServer(function (_req, res) {
+      res.writeHead(200);
+      res.end('hello world');
+    });
 
-    const pythonVersion = execSync(
-      'python -c "import sys; print(sys.version_info[0])"',
-      { encoding: 'utf-8' }
-    );
-
-    let args : string[] = [];
-    if (pythonVersion.trim() === '2') {
-      args = ['-m', 'SimpleHTTPServer', '9876'];
-    } else {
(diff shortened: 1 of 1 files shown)


src/node/desktop/test/unit/main/browser-window.test.ts
```diff
@@ -17,25 +17,17 @@ import { describe } from 'mocha';
 import { assert } from 'chai';
 
 import { BrowserWindow } from 'electron';
-import { execSync, spawn } from 'child_process';
+import * as http from 'http';
```
jgutman, line 20: do we really need to import everything?

src/node/desktop/test/unit/main/browser-window.test.ts
```diff
@@ -54,12 +46,10 @@ describe('BrowserWindow', () => {
       });
     }
 
-    await win.loadURL('http://localhost:9876');
-    server.kill();
+    await win.loadURL('http://localhost:9875');
+    server.close();
```
jgutman, line 50: what's the reason for changing from kill to close here? is it just unnecessary?

jgutman (approved):

src/node/desktop/test/unit/main/browser-window.test.ts
timtmok: `kill` is when `server` was a child process object. It's now a node server object, which `close` will gracefully shut down the server.

src/node/desktop/test/unit/main/browser-window.test.ts
timtmok: Good point. It only needs `createServer`.

Issue #12247, opened 2022-10-28
The title bar above the menu bar is light even in the dark theme (in daily builds)


GegznaV:
### System details

    RStudio Edition : Desktop [Open Source]
    RStudio Version : 2022.12.0.292
    OS Version      : Windows 10 x64 (build 19044)
    R Version       : R version 4.2.1 (2022-06-23 ucrt)

### Steps to reproduce the problem

On Windows:

1) Open RStudio and use the dark theme (Modern).
2) Change Windows colors (color theme) to dark 
![image](https://user-images.githubusercontent.com/12725868/198636631-4d245932-4d2a-4668-abbc-9a62b96a9dca.png)
3) Look at RStudio's appearance
4) Change Windows colors (color theme) to light
5) Look at RStudio's appearance

Repeat the same with the RStudio daily builds version and the current stable version. Pay attention to the menu bar and the title bar above it.

### Describe the problem in detail

In the daily builds version, RStudio's menu bar changes color while the title bar above is always light:

![image](https://user-images.githubusercontent.com/12725868/198638296-fe58631b-0bff-4edf-bedb-de500e350a32.png)

![image](https://user-images.githubusercontent.com/12725868/198638372-c496b00f-f81b-495f-94df-688f31e13a8f.png)

That light title bar above now distracts attention. It should also become dark.

### Describe the behavior you expected

I expect the menu bar to be gray as in the current stable RStudio version:
![image](https://user-images.githubusercontent.com/12725868/198640831-bdb5e8d4-f5b7-4c46-9711-52fa07c42fc0.png)
And the title bar should change color in accordance with the Windows color theme.

If it is not possible to change the color of the title bar, it is better to keep the gray menu bar and not black to have less contrast.

ronblum:
@GegznaV Thank you for pointing this out!

@timtmok Do we still have control over this, or did we lose control due to Electron?

gtritchie:
For reference: https://www.electronjs.org/docs/latest/tutorial/dark-mode

gtritchie:
Talked about here:

https://github.com/electron/electron/issues/23479

May finally be fixed (by default) with Electron 20+ on Windows-11 (I haven't tried it to verify), but not on Windows-10 (though there is talk of a fuse that shows how to accomplish it using undocumented Windows APIs).

https://github.com/electron/electron/pull/33624

timtmok:
Based on that, it won't be supported until Electron is updated to 20+.

[...]

PR #12285, opened 2022-11-03
Use light theme on Windows

timtmok:
### Intent
Address #12247

### Approach
Electron 20 adds support for theming the title bar. Until Electron is upgraded, it looks more consistent to have the theme set to `light` on Windows. Then the menu bar and title bar will have consistent theming. Linux (Ubuntu 22) and Mac do not have this issue.

With Electron 20, we can then add in event hooks to adjust the `nativeTheme` and maybe even match it to the options that are selected in the Appearance options.

### Automated Tests
None

### QA Notes
It is noticeable on Windows when setting the system theme to dark.

Mac does not have this issue since the menu bar is always at the top and is themed according to the system.

Linux appeared fine on Ubuntu 22 but may vary depending on the windowing environment. 

### Checklist

### DIFF
--- a/src/node/desktop/src/main/application.ts
+++ b/src/node/desktop/src/main/application.ts
@@ -13,7 +13,7 @@
  *
  */
 
-import { app, BrowserWindow, dialog, Menu, screen, shell, WebContents } from 'electron';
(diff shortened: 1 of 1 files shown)


gtritchie (approved):
LGTM! There will be some complaints I'm sure but this looks better than the mixed theming we're getting now.

PR #12291, opened 2022-11-04
Prevent double slash in rserver-url -l output

melissa-barca:
### Intent

Addresses rstudio/rstudio-pro#3912

 `rserver-url -l {port}` concatenates `$RS_SERVER_URL` and `$RS_SESSION_URL`, which can lead to a the URL output including a double slash:
```
$ echo $RS_SERVER_URL; echo $RS_SESSION_URL
https://dev1.rstudio.internal/
/s/4566a3c9ab5a75ae3a3d3/

$ /usr/lib/rstudio-server/bin/rserver-url -l 8050
https://dev1.rstudio.internal//s/4566a3c9ab5a75ae3a3d3/p/3393aabd/
```

This causes issues running applications in proxied environments with prefixes like `/rstudio` in their session ID. 

### Approach

`$RS_SESSION_URL` will always begin with a leading `/` so if we find a trailing `/` in `$RS_SERVER_URL`, we remove it before concatenating the output.

### Automated Tests

The recently added FastAPI automation will fail in our proxy fuzzbucket environment. This should resolve that.

### Checklist

### DIFF
--- a/src/cpp/server/url-ports/UrlPortsMain.cpp
+++ b/src/cpp/server/url-ports/UrlPortsMain.cpp
@@ -64,6 +64,10 @@ int main(int argc, char * const argv[])
 
       char* pServerPath = std::getenv("RS_SERVER_URL");
       std::string serverPath = pServerPath != NULL ? std::string(pServerPath) : std::string();
+      if (serverPath.back() == '/')
+      {
+         serverPath.pop_back();
+      }
 
       std::cout << serverPath + sessionPath + "p/" + transformedPort + "/" << std::endl;
    }


dpastoor:
Can also confirm that manually modifying the root path provided via the snippet:

```

if __name__ == '__main__':
    path, port = '', 8001
  
    if 'RS_SERVER_URL' in os.environ and os.environ['RS_SERVER_URL']:
        path = subprocess.run(f'echo $(/usr/lib/rstudio-server/bin/rserver-url -l {port})',
    # going to be lazy and not write a regex. The issue with a simple replace is // --> / will also replace https:// to https:/ so
    # lets add an extra / there so then when we remove all // it'll go back to https://
    # ultimately this will take https://example.com//rstudio/s/... --> https://example.com/rstudio/s/...
                             stdout=subprocess.PIPE, shell=True).stdout.decode().strip().replace("://", ":///").replace("//", "/")
    uvicorn.run(app, port = port, root_path = path)
```

causes the breaking behavior to be fixed

jeffvroom (approved):
LGTM!

PR #12298, opened 2022-11-07
fix building on Macs with up-to-date brew dependencies (postgres)

gtritchie:
### Intent

Addresses ##12288

A recent change in homebrew's Postgres formula made it keg-only. This prevented the SOCI build (done when installing rstudio build dependencies) from finding the Postgres libs. SOCI still builds, just without Postgres support. That then breaks the rstudio IDE build when cmake fails to find the SOCI Postgres libs.

This manifests on recently configured Mac dev machines, but not on ones set up before this change (unless an explicit `brew update` was performed recently AND the SOCI libs rebuilt). The Mac official build machine is never `brew updated`, so it hasn't hit this problem.

### Approach

When building SOCI on a Mac, set the environment variable `POSTGRESQL_ROOT` to the location of the homebrew postgres files. This environment variable is documented in the cmake module for postgres detection included with SOCI.

Also, detect if SOCI had been built (on a Mac) without Postgres support, and rebuild it. This is to help devs who got their machines in this state; it _should_ fix itself when re-running `install-dependencies-osx`.

### Automated Tests

N/A

### QA Notes

Purely a build-time thing, and currently only an issue on dev machines (Macs).

### Checklist

### DIFF
--- a/dependencies/common/install-soci
+++ b/dependencies/common/install-soci
@@ -31,8 +31,20 @@ BOOST_DIR="$RSTUDIO_TOOLS_ROOT/boost/boost_$BOOST_VERSION"
 
 # install SOCI if it isn't already installed
 if [ -d "${SOCI_BIN_DIR}/lib" ]; then
-   echo "SOCI already installed at '${SOCI_DIR}'"
-   exit 0
+   SOCI_FOUND=true
+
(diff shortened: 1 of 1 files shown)


melissa-barca (approved):
LGTM! Thanks Gary! I just tested this out and it works great.

Issue #12321, opened 2022-11-10
Screen reader user cannot tell that Electron desktop now has basic screen reader support


gtritchie:
### System details

    RStudio Edition : Desktop
    RStudio Version : Elsbeth Geranium daily builds
    OS Version      : Any
    R Version       : Any

### Steps to reproduce the problem

Start a screen reader, then start RStudio Desktop (Electron). Try to do anything (as a blind user).

### Describe the problem in detail

Everything in the RStudio window is invisible to the screen reader software. All it can tell you is that there's a window. This is because we force Chromium accessibility support off by default. This is a carryover from an issue we had seen years ago with the Magnet application (on macOS) triggering Chromium's accessibility support and causing performance issues with Qt. #1990

The code that does this was ported over from Qt as-is (by me, for full disclosure).

https://github.com/rstudio/rstudio/blob/f8461c63eb68ea77943d488ee6c075de3977cd05/src/node/desktop/src/main/main.ts#L86-L93

As a blind user, you would need to know to go to the main menu, then Help / Accessibility / Screen Reader Support and restart the application before you could actually do anything via screen reader. That is unacceptable, and although the IDE is still far from completely usable via screen reader, this masks all the significant improvements we've made by the switch to Electron.

### Describe the behavior you expected

We should not disable Chromium's automatic accessibility support by default and let Chromium do the right thing.

In my own testing, the Magnet issue is no longer a problem; don't know if that's due to improvements in Magnet, or Chromium (via Electron), or macOS.

That said, it's possible there could be other such corner cases where a user would want to disable all Chromium accessibility (I vaguely recall there was a problem like that on Linux, too, at one point), so we should provide some mechanism to do this, but it shouldn't be the default!

### Proposed Fix

There are two aspects of screen reader support in RStudio Electron.

1. Chromium accessibility support, which enables the APIs used by accessibility tools such as screen reader software. By default, Chromium (including Electron and the Chrome Web Browser) will enable this support automatically if it detects a screen reader is in use. This is what we're currently disabling.
2. Additional RStudio-specific behaviours controlled by a user setting (Global Options / Accessibility / Screen Reader Support). This turns on additional RStudio-specific things such telling the terminal pane (xterm) to provide a DOM equivalent to usual non-DOM-based rendering (which is invisible to screen readers).

Right now, in Electron, the "screen reader support enable/disable" user preference controls both of those. It needs to only control #2. That would make it match the behaviour of RStudio Server/Workbench.

[...]

PR #12329, opened 2022-11-14
Feature/disable electron accessibility option

gtritchie:
### Intent

Addresses #12321

### Approach

Allow Chromium (via Electron) to auto-enable the internal accessibility APIs when an accessibility tool such as a screen reader is running. This is the default behaviour of Electron (and Chrome).

I added yet-another obscure option (Electron-only) named "Disable Electron Accessibility support" under Global Options / General / Advanced which can be used to disable Chromium's accessibility support by passing Chromium the `disable-renderer-accessibility` switch.

This would only be needed if a user finds that accessibility support is being enabled when it shouldn't and impacting the product's performance. There have been historical cases of this (https://github.com/rstudio/rstudio/issues/1990), but none are currently known.

I didn't put this option under Accessibility because a user needing accessibility support is never going to want to use this. It's more of an anti-accessibility setting.

screenshot of Global options showing General / Advanced with new option highlighted

Note in the diffs that there are some minor changes related to the `enable_cloud_publish_ui` setting (whitespace primarily); this indicates that when this preference was added the source files commented with dire warnings of "do not edit this file directly" were edited directly instead of being generated by the generate-prefs.R script. This should be a no-op but thought I'd point it out.

Also I removed the suffix `(RStudio Server)` from the existing Enable Accessibility option since it's no longer server-specific.

### Automated Tests

Added unit test for new setting.

### QA Notes

To test:

* Start RStudio and make sure "General / Advanced / Disable Electron accessibility support" is UNCHECKED and "Accessibility / Screen reader support" is UNCHECKED; these are the defaults on a clean install
* Exit RStudio
* Start a screen reader (on a Mac, use Cmd+F5 to start VoiceOver, on Windows, use Win+Ctrl+Enter for Narrator)
* start RStudio
* Help / Diagnostics / Show Accessibility Diagnostics

The various checkboxes under "Accessibility modes" should be checked (might vary slightly between Windows and Mac; the main thing is the checkboxes should not be disabled). If you run the above scenario on an Electron build without this fix, they will be disabled.

Screenshot 2022-11-14 at 09 39 39

There is a known issue if you toggle "Disable Electron accessibility support" via the command palette instead of the Global Options dialog -- we don't sync the desktop-side setting via that avenue. This bug exists with other such settings that have both a session-side user-preference and a desktop-side setting that needs to be kept in sync. Also, FWIW, changing settings via the command-palette doesn't trigger the necessary restart.

### Checklist

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/GeneralPreferencesPane.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/GeneralPreferencesPane.java
@@ -602,7 +602,7 @@ public class GeneralPreferencesPane extends PreferencesPane
          boolean disableRendererAccessibilityPrefValue = disableRendererAccessibility_.getValue();
          if (disableRendererAccessibilityPrefValue != initialDisableRendererAccessibility_)
          {
-            if (Desktop.hasDesktopFrame())
+            if (Desktop.hasDesktopFrame() && BrowseCap.isElectron())
                Desktop.getFrame().setDisableRendererAccessibility(disableRendererAccessibilityPrefValue);
             restartRequirement.setDesktopRestartRequired(true);
          }
--- a/src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/GeneralPreferencesPane.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/GeneralPreferencesPane.java
@@ -344,6 +344,8 @@ public class GeneralPreferencesPane extends PreferencesPane
          {
             nativeFileDialogs_ = checkboxPref(prefs_.nativeFileDialogs());
             advanced.add(nativeFileDialogs_);
+            disableRendererAccessibility_ = checkboxPref(prefs_.disableRendererAccessibility());
+            advanced.add(disableRendererAccessibility_);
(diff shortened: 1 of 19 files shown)


src/gwt/src/org/rstudio/studio/client/workbench/prefs/views/GeneralPreferencesPane.java
```diff
@@ -596,6 +599,15 @@ public RestartRequirement onApply(UserPrefs prefs)
             restartRequirement.setUiReloadRequired(true);
          }
 
+         boolean disableRendererAccessibilityPrefValue = disableRendererAccessibility_.getValue();
+         if (disableRendererAccessibilityPrefValue != initialDisableRendererAccessibility_)
+         {
+            if (Desktop.hasDesktopFrame())
```
timtmok, line 605: Would this also need `BrowseCap.isElectron()` since this is Electron only? I think this might be an issue with Desktop Pro, which wouldn't have this implemented.

gtritchie: Good catch; fixed!

timtmok (approved):

Issue #12354, opened 2022-11-17
Build pane and background job truncate useful output


hadley:
When working with quarto there are two cases where you can end up with very long output:

* Building an entire book (e.g. http://github.com/hadley/r4ds) will rebuild every chapter creating a lot of output.
* Repeatedly re-building a single page/chapter appears in the same long running background job, which never terminates.

In the first case, it's not obvious that the process is still running because it doesn't look like anything is happening in the build pane; in the second case, the process never terminates.

mikebessuille:
This should be in Elsbeth Geranium only if the fix is very low risk.  Otherwise, we'll fix it in Cherry Blossom and backport it to a patch for EG if necessary.

timtmok:
I'll see if there's a low risk fix for 2022.12.0

kevinushey:
We could also just roll back my PR.

timtmok:
@kevinushey I've got a PR to flush the saved output when it builds up to the overload limit. At least the user will be able to see more output during the build.

It also includes a fix to clear the saved output so that a subsequent build will not redisplay it.

timtmok:
PR has been built: https://dailies.rstudio.com/version/2022.12.0-preview+334/

jonvanausdeln:
Verified on 2022.12.0.334 Desktop, Windows 11 and MacOS 12.6 

Build of the original repro book succeeds, build output gets truncated once. Works as expected.  Waiting to close for a PR with some string changes.

jonvanausdeln:
Verified localized string on 2022.12.0-preview+346 Desktop Windows 11

Strings show as expected in French.

PR #12359, opened 2022-11-18
Flush build pane buffer periodically

timtmok:
### Intent
Address #12354

### Approach
Periodically flush the saved output so the user can see build progress. I don't think this would have an adverse effect on the ability to display the output since the `onCompileCompleted()` would submit everything that was saved.

### Automated Tests
None

### QA Notes
Building the Quarto book mentioned in #12354 has enough output to display a message that output would be buffered. 

### Checklist


kevinushey (approved):
LGTM!

Issue #12362, opened 2022-11-21
Keyboard shortcut help (accessible version) won't scroll


gtritchie:
### System details

    RStudio Edition : Desktop or Server
    RStudio Version : RStudio 2022.11.0-daily+192 or any build since
    OS Version      : any
    R Version       : any

### Steps to reproduce the problem

1. Start RStudio and turn on screen-reader support (Help / Accessibility / Screen Reader Support), let the app restart.
2. Help / Keyboard Shortcuts Help

### Describe the problem in detail

The plain HTML version of keyboard shortcuts help opens in web browser as it should, but it cannot be scrolled via keyboard or mouse.

This was introduced with this PR: https://github.com/rstudio/rstudio/pull/11966

### Describe the behavior you expected

Page should scroll. On the positive side, you can still read everything on this page with a screen reader since that doesn't rely on being able to scroll things into view...

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
Triage: Proposed for Elsbeth Geranium.

ronblum:
Verified in RStudio Desktop 2022.12.0-preview+339 on MacOS 13.

PR #12366, opened 2022-11-21
Fix scrolling on html pages that are much larger than the viewport

GCRev:
See: https://github.com/rstudio/rstudio/issues/12362

### Intent

If long static pages are loaded, which is usually just `keyboard.htm`, then there is no scrolling enabled. This is due to a change to the css file that fixes the viewport to the window boundaries. We do not want to hide overflowing content if it exceeds the limits of the page.

### Approach

Add `overflow: auto` to the `html, body` section so that overflowing content will receive scrollbars. The only page this really affects is the `keyboard.htm` page, as the others are already laid out relative to the viewport boundaries. I already verified the other static files by loading them in a browser to check that they were not going to be broken by this change.

### Automated Tests / QA Notes

N / A - visual pass only. I verified this in Firefox, Chrome, and Safari

### Checklist

### DIFF
--- a/src/gwt/www/rstudio.css
+++ b/src/gwt/www/rstudio.css
@@ -32,6 +32,7 @@ html, body {
   left: 0;
   height: 100vh;
   margin: 0;
+  overflow: auto;
 }
 
 #posit-logo {


ronblum (approved):
LGTM! Checked by grabbing the HTML and CSS and checking the changes locally.

gtritchie (approved):
LGTM!

Issue #12389, opened 2022-12-01
Buffered output in background jobs doesn't get displayed


ronblum:
### System details

    RStudio Edition : Desktop & Server
    RStudio Version : 2022.12.0-preview+346 & 2022.12.0+348
    OS Version      : MacOS 13 & Ubuntu 20.04.5 via Chrome on MacOS 13
    R Version       : 4.2.2

### Steps to reproduce the problem

We'll need a dataset with more than 10000 records. In this example, we'll use `nycflights13`.

1. `install.packages("nycflights13")`
2. Create a new file with the following code:
```
f <- head(nycflights13::flights, 10050)
print(f, n = 10050)
```
3. Save the file.
4. Run the file as a background job: <kbd>Source dropdown -> Source as Background Job</kbd>

### Describe the problem in detail

The output includes buffering, but once printout continues, the buffered output is never printed.

In the middle of the output, we get
```
 4997  2013     1     6     1835    1819      16    2154    2140      14 UA     
 4998  2013     1     6     1835    1815      20    2100    2034      26 UA     


[Detected output overflow; buffering the next 5000 lines of output]

 9899  2013     1    12      825     825       0    1115    1123      -8 UA     
 9900  2013     1    12      826     835      -9    1300    1340     -40 DL    
```

\<snip\>

```
 9997  2013     1    12     1020    1020       0    1358    1330      28 AA     
 9998  2013     1    12     1020    1015       5    1320    1340     -20 UA     


[Detected output overflow; buffering the next 5000 lines of output]
```

### Describe the behavior you expected

I'm not sure of the rows in the middle, but I think that at least the last lines at the end should be displayed.

This isn't an issue in the Console Pane, only Background Jobs.

Triage: This is a regression in Elsbeth Geranium, so proposing for Cherry Blossom.

- [ ] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ ] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [ ] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

timtmok:
[...]

PR #12400, opened 2022-12-06
Call completion event for background jobs

timtmok:
### Intent
Address #12389 

### Approach
Call `compileCompleted()` once the job finishes or when the output is shown to ensure the output buffer is flushed. Added a check for changing stop button visibility since it isn't available when viewing the background jobs list.

### Automated Tests
None

### QA Notes
The output buffer will be flushed when:
* watching the output as the background job runs 
* selecting a finished background job

### Checklist

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/common/compile/CompileOutputBufferWithHighlight.java
+++ b/src/gwt/src/org/rstudio/studio/client/common/compile/CompileOutputBufferWithHighlight.java
@@ -105,7 +105,8 @@ public class CompileOutputBufferWithHighlight extends Composite
    }
 
    @Override
-   public void flushOutput() {
+   public void flushOutput()
+   {
       String[] lines = savedOutput_.split("\n");
       int end = lines.length - 1;
       int start = Math.max(0, end - 100);
--- a/src/gwt/src/org/rstudio/studio/client/common/compile/CompileOutputBufferWithHighlight.java
+++ b/src/gwt/src/org/rstudio/studio/client/common/compile/CompileOutputBufferWithHighlight.java
@@ -101,16 +101,17 @@ public class CompileOutputBufferWithHighlight extends Composite
       String outputClass = styles_.output();
       console_.submit("\n", outputClass);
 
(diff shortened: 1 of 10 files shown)


src/gwt/src/org/rstudio/studio/client/workbench/views/jobs/view/JobsDisplayImpl.java
```diff
@@ -111,7 +115,8 @@ public void showJobOutput(String id, JsArray<JobOutput> output, boolean animate)
                output.get(i).type(),
                output.get(i).output()), false /* scroll */);
       }
-      
+
+      widgets_.getOutputPanel().output_.compileCompleted();
```
kevinushey, line 119: This doesn't feel right to me? `showJobOutput()` being called doesn't imply that compilation has finished.

kevinushey: Show we have a separate method like `widgets_.showBufferedOutput()` or something? Which would then be called in `compileCompleted()`?

timtmok: Right, that would make more sense. It's not really done but it needs to ensure anything buffered is displayed. I'll fix this up.

src/gwt/src/org/rstudio/studio/client/common/compile/CompileOutputBufferWithHighlight.java
```diff
@@ -101,16 +101,17 @@ public void onCompileCompleted()
       String outputClass = styles_.output();
       console_.submit("\n", outputClass);
 
-      flushOverloadBuffer(outputClass);
+      flushOutput();
    }
 
-   private void flushOverloadBuffer(String outputClass) {
+   @Override
+   public void flushOutput() {
```
kevinushey, line 108:
```suggestion
   public void flushOutput()
   {
```

kevinushey (approved):
Thanks for the changes -- LGTM!

Issue #12390, opened 2022-12-01
git clone prompts with error: cannot run rpostback-askpass


samperman:
### System details

    RStudio Edition : Workbench (running in posit.cloud)
    RStudio Version : 2022.12.0-349.pro2
    OS Version      :  Ubuntu Focal
    R Version       :  4.2.2

### Steps to reproduce the problem
Create a new project, navigate to the terminal and run `git clone`

### Describe the problem in detail
The clone prompts with errors:
```
/cloud/project$ git clone --depth=1 https://github.com/rstudio/connect
Cloning into 'connect'...
error: cannot run rpostback-askpass: No such file or directory
Username for 'https://github.com': sam@permans.com
error: cannot run rpostback-askpass: No such file or directory
```

### Describe the behavior you expected
Spotted Wakerobin would open a modal to prompt for username and password.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

jmcphers:
This will affect _any_ git operation that requires credentials. I don't think we can release with it.

samperman:
A couple other pieces of info from the environment i was using:
```
$ echo $PATH
/usr/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/opt/R/4.2.2/lib/R/bin:/opt/quarto/1.2.269/bin:/usr/lib/rstudio-server/bin/postback/postback:/usr/bin:/opt/R/4.2.2/lib/R/bin
```

```
> Sys.getenv("RS_RPOSTBACK_PATH")
[1] "/usr/lib/rstudio-server/bin/rpostback"
```

jgutman:
> This will affect _any_ git operation that requires credentials. I don't think we can release with it.

Yes, but we tested several variations of this on several different OSes with several different combinations of ssh with and without password protected key, and https cloning. So I'm a little unsure as to the scope, clearly we missed something in testing but it did pass all the tests we did run. @jmcphers @jonvanausdeln

[...]

PR #12414, opened 2022-12-09
rationalize usages of rpostback + postback script paths

kevinushey:
### Intent

Addresses (part of) https://github.com/rstudio/rstudio/issues/12390.

### Approach

- Make the development layout of postback utilities match that of package builds
- Try to rationalize usages of the `rpostback` path, plus the `postback` scripts directory

### Automated Tests

Not included.

### QA Notes

Test via various notes in https://github.com/rstudio/rstudio/issues/12390.

### Checklist

### DIFF
--- a/src/cpp/session/SessionConsoleProcess.cpp
+++ b/src/cpp/session/SessionConsoleProcess.cpp
@@ -94,7 +94,21 @@ core::system::ProcessOptions ConsoleProcess::createTerminalProcOptions(
 #else
    core::system::setHomeToUserProfile(&shellEnv);
 #endif
+   
+   // don't forward GIT_ASKPASS or SSH_ASKPASS if it's set to our rpostback
+   // handlers -- those aren't required in a terminal, as the user can be
+   // prompted for input via stdin, and it's likely that users will munge
+   // the PATH within a terminal a way that makes our utilities unavailable
+   //
+   // TODO: could we just set an absolute path to the rpostback-askpass util?
+   std::string gitAskpass = core::system::getenv("GIT_ASKPASS");
+   if (gitAskpass == "rpostback-askpass")
+      core::system::unsetenv(&shellEnv, "GIT_ASKPASS");
 
+   std::string sshAskpass = core::system::getenv("SSH_ASKPASS");
+   if (sshAskpass == "rpostback-askpass")
+      core::system::unsetenv(&shellEnv, "SSH_ASKPASS");
+   
    // amend shell paths as appropriate
    session::modules::workbench::ammendShellPaths(&shellEnv);
 
--- a/src/cpp/session/SessionModuleContext.cpp
+++ b/src/cpp/session/SessionModuleContext.cpp
@@ -108,56 +108,6 @@ namespace rstudio {
 namespace session {   
 namespace module_context {
 
-namespace {
-
-Error modifyPath(const core::FilePath& path, bool append)
-{
-   Error error;
-
-   std::string currentPath;
-   error = r::exec::RFunction("base:::Sys.getenv")
-         .addParam("PATH")
-         .call(&currentPath);
-
-   if (error)
-      return error;
-
-   std::string pathString = string_utils::utf8ToSystem(path.getAbsolutePath());
-
-#ifdef _WIN32
-   std::replace(pathString.begin(), pathString.end(), '/', '\\');
-#endif
-
-   std::string newPath = append
-         ? fmt::format("{}{}{}", currentPath, kPathSeparator, pathString)
-         : fmt::format("{}{}{}", pathString, kPathSeparator, currentPath);
-
-   // clean up duplicated path separators, if any
-   boost::regex reMultipleSeparators(kPathSeparator "+");
-   newPath = boost::regex_replace(newPath, reMultipleSeparators, kPathSeparator);
-
-   error = r::exec::RFunction("base:::Sys.setenv")
-         .addParam("PATH", newPath)
-         .call();
-
-   if (error)
-      return error;
-
-   return Success();
-}
-
-} // end anonymous namespace
-
-Error prependToPath(const core::FilePath& path)
-{
-   return modifyPath(path, false);
-}
-
-Error appendToPath(const core::FilePath& path)
-{
-   return modifyPath(path, true);
-}
-
 bool isSessionSslEnabled()
 {
    return !options().getOverlayOption(kSessionSslCertKeyOption).empty();
--- a/src/cpp/session/SessionModuleContext.cpp
+++ b/src/cpp/session/SessionModuleContext.cpp
@@ -95,12 +95,64 @@
 
 #include "session-config.h"
 
+#ifdef _WIN32
+# define kPathSeparator ';'
+#else
+# define kPathSeparator ':'
+#endif
+
 using namespace rstudio::core;
 using namespace boost::placeholders;
 
 namespace rstudio {
 namespace session {   
 namespace module_context {
+
+namespace {
+
+Error modifyPath(const core::FilePath& path, bool append)
+{
+   Error error;
+
+   std::string currentPath;
+   error = r::exec::RFunction("base:::Sys.getenv")
+         .addParam("PATH")
+         .call(&currentPath);
+
+   if (error)
+      return error;
+
+   std::string pathString = string_utils::utf8ToSystem(path.getAbsolutePath());
+
+#ifdef _WIN32
+   std::replace(pathString.begin(), pathString.end(), '/', '\\');
+#endif
+
+   std::string newPath = append
+         ? fmt::format("{}{}{}", currentPath, kPathSeparator, pathString)
+         : fmt::format("{}{}{}", pathString, kPathSeparator, currentPath);
+
+   error = r::exec::RFunction("base:::Sys.setenv")
+         .addParam("PATH", newPath)
+         .call();
+
+   if (error)
+      return error;
+
+   return Success();
+}
+
+} // end anonymous namespace
+
+Error prependToPath(const core::FilePath& path)
+{
+   return modifyPath(path, false);
+}
+
+Error appendToPath(const core::FilePath& path)
+{
+   return modifyPath(path, true);
+}
 
 bool isSessionSslEnabled()
 {
--- a/src/cpp/session/SessionModuleContext.cpp
+++ b/src/cpp/session/SessionModuleContext.cpp
@@ -96,9 +96,9 @@
 #include "session-config.h"
 
 #ifdef _WIN32
-# define kPathSeparator ';'
+# define kPathSeparator ";"
 #else
-# define kPathSeparator ':'
+# define kPathSeparator ":"
 #endif
 
 using namespace rstudio::core;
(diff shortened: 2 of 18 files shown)


src/cpp/session/SessionPostback.cpp
```diff
@@ -104,8 +120,8 @@ Error registerPostbackHandler(const std::string& name,
       return error;
                                                     
    // compute the shell command required to invoke this handler and return it
-   Options& options = session::options();
-   *pShellCommand = options.rpostbackPath().getAbsolutePath() + "-" + name;
+   *pShellCommand = rPostbackScriptPath("rpostback-" + name).getAbsolutePath();
+   std::cerr << *pShellCommand << std::endl;
```
kevinushey, line 124: Can remove this later (was for debugging)

src/cpp/session/modules/SessionGit.cpp
```diff
@@ -3354,21 +3354,12 @@ core::Error initialize()
...
 
    // add postback directory to PATH
-   FilePath postbackDir = session::options().rpostbackPath().getParent();
-   if (postbackDir.getAbsolutePath().find("session/postback") == std::string::npos) {
-      // for package builds only, postback/rpostback-askpass in same directory as rpostback itself
-      postbackDir = postbackDir.completeChildPath("postback");
-   }
-
+   FilePath postbackDir = module_context::rPostbackScriptsDir();
    core::system::addToPath(postbackDir.getAbsolutePath());
```
kevinushey, line 3363: TODO: This doesn't seem to be working on Windows (the `rsession.exe` PATH isn't updated); we might need to call `Sys.setenv()` directly?

src/cpp/session/SessionConsoleProcess.cpp
```diff
@@ -94,7 +94,21 @@ core::system::ProcessOptions ConsoleProcess::createTerminalProcOptions(
...
+   //
+   // TODO: could we just set an absolute path to the rpostback-askpass util?
+   std::string gitAskpass = core::system::getenv("GIT_ASKPASS");
+   if (gitAskpass == "rpostback-askpass")
+      core::system::unsetenv(&shellEnv, "GIT_ASKPASS");
+
+   std::string sshAskpass = core::system::getenv("SSH_ASKPASS");
+   if (sshAskpass == "rpostback-askpass")
+      core::system::unsetenv(&shellEnv, "SSH_ASKPASS");
+
```
jgutman, line 111: I think something about this is behaving differently on Unix versus Windows, maybe because on Windows the handler is considered `rpostback-askpass.exe` rather than `rpostback-askpass`. And I notice that when I use the terminal in this local build, on Mac I do actually get prompted for a username/password in stdin, while on Windows I don't get prompted for username/password. Also, on Mac, I get the value of `rpostback-askpass` when I do `Sys.getenv("GIT_ASKPASS")` or `Sys.getenv("SSH_ASKPASS")` while on Windows, these two environment variables appear to be unset, not just in the terminal but directly in the R console with `Sys.getenv`.  So I'm a little confused about the differences in behavior here

src/cpp/session/SessionModuleContext.cpp
```diff
@@ -95,13 +95,65 @@
...
+         .addParam("PATH")
+         .call(&currentPath);
+
+   if (error)
+      return error;
+
+   std::string pathString = string_utils::utf8ToSystem(path.getAbsolutePath());
+
+#ifdef _WIN32
+   std::replace(pathString.begin(), pathString.end(), '/', '\\');
```
jgutman, line 128: Is there a particular reason for changing this now in Windows? We've always had the forward slashes in the path before on Windows, and now with this we sometimes get a mix of forward and backslashes, for example:

```
 "C:\\rtools42/x86_64-w64-mingw32.static.posix/bin;C:\\rtools42/usr/bin;C:\\R\\R-4.2.1\\bin\\x64;
```
just wondering if you suddenly encountered some issue with the forward slashes that prompted this change because I think most of our codebase tends to assume that the forward slashes will work the same on Windows and we specifically don't need to convert to native path separators

src/cpp/session/SessionModuleContext.cpp
```diff
@@ -95,13 +95,65 @@
...
+#endif
+
+   std::string newPath = append
+         ? fmt::format("{}{}{}", currentPath, kPathSeparator, pathString)
+         : fmt::format("{}{}{}", pathString, kPathSeparator, currentPath);
+
```
jgutman, line 134: I don't think it particularly matters, but in some cases, at least on Windows I noticed I ended up with a double path separator, 
e.g.
```
D:\\Users\\jackie.gutman-win\\AppData\\Roaming\\python\\Python310\\Scripts;;D:\\Users\\jackie.gutman-win\\quarto-cli\\package\\dist\\bin;
```
(note the double semicolon) because the currentPath variable already terminated in a path separator, so it might be worth checking for that, even though I don't think it will cause any issues to have the double separator

src/cpp/session/postback/CMakeLists.txt
```diff
@@ -45,20 +45,15 @@ if (RSTUDIO_SERVER)
...
-
-# put rpostback in a place where it can be found for dev config
-if(NOT RSTUDIO_PACKAGE_BUILD)
-   set_target_properties(rpostback PROPERTIES RUNTIME_OUTPUT_DIRECTORY ${POSTBACK_SCRIPT_DIR})
-endif()
+configure_file(rpostback-editfile  ${POSTBACK_SCRIPT_DIR}/rpostback-editfile)
+configure_file(rpostback-pdfviewer ${POSTBACK_SCRIPT_DIR}/rpostback-pdfviewer)
+configure_file(rpostback-gitssh    ${POSTBACK_SCRIPT_DIR}/rpostback-gitssh)
+configure_file(rpostback-askpass   ${POSTBACK_SCRIPT_DIR}/rpostback-askpass)
+configure_file(rpostback-browser   ${POSTBACK_SCRIPT_DIR}/rpostback-browser)
```
jgutman, line 56: I don't know if there was a particular point to re-ordering these, but if we are reordering them then it might as well be alphabetically?

src/cpp/session/SessionModuleContext.cpp
```diff
@@ -95,13 +95,65 @@
...
+   error = r::exec::RFunction("base:::Sys.setenv")
```
jgutman, line 135: What's the point of doing it via r exec instead of using the `core::system::setenv` ? what are the differences here?

src/cpp/session/modules/SessionGit.cpp
```diff
@@ -127,7 +127,7 @@ core::system::ProcessOptions procOptions()
    // (note that we also do this on init, but we do this again for
    // child processes just to ensure any user-initiated PATH munging
    // doesn't break builtin utilities)
-   FilePath postbackDir = session::options().rpostbackPath().getParent();
+   FilePath postbackDir = module_context::rPostbackScriptsDir();
    core::system::addToPath(&childEnv, postbackDir.getAbsolutePath());
```
jgutman, line 131: again like why are we sometimes using `core::system::addToPath` and other times using the new `module_context::appendToPath`?

src/cpp/session/SessionConsoleProcess.cpp
kevinushey: > I think something about this is behaving differently on Unix versus Windows, maybe because on Windows the handler is considered rpostback-askpass.exe rather than rpostback-askpass.

Can you clarify? The various helper scripts are just regular bash scripts (without extensions) so it should be correct to use these names as-is here. Only the `rpostback` executable itself will be called `rpostback.exe` on Windows.

> And I notice that when I use the terminal in this local build, on Mac I do actually get prompted for a username/password in stdin, while on Windows I don't get prompted for username/password.

Does authentication succeed? If it does, that would probably imply the credential manager is using existing credentials; if not, what error are you seeing?

> Also, on Mac, I get the value of rpostback-askpass when I do Sys.getenv("GIT_ASKPASS") or Sys.getenv("SSH_ASKPASS") while on Windows, these two environment variables appear to be unset, not just in the terminal but directly in the R console with Sys.getenv

I think I know what the issue here is; I'm going to push a change.

src/cpp/session/SessionModuleContext.cpp
kevinushey: There isn't a strong reason, but _usually_ things on the Windows PATH should use backslashes. In practice, Windows allows both separators, so it's more hygiene than anything.

> because I think most of our codebase tends to assume that the forward slashes will work the same on Windows and we specifically don't need to convert to native path separators

That's true, but the PATH is a somewhat special case where IMHO the "correct" thing to do is to use backslashes. However, whenever we want to handle paths internally (or return those paths to the client) we should definitely use forward slashes.

src/cpp/session/SessionModuleContext.cpp
kevinushey: I'll try to take a look; I think it's worth cleaning that up

src/cpp/session/SessionModuleContext.cpp
kevinushey: This is mainly for a Windows oddity. I don't fully understand the details, but basically, environment variables set via `core::system::setenv()` won't be visible to the R session via `Sys.getenv()`. I think this might have something to do with R being built against a static copy of the libc runtime, and so it ends up with its own environment variable "block" which is separate from the one normally visible from our own session.

By using R's APIs for setting / getting environment variables, we avoid these sorts of issues.

src/cpp/session/modules/SessionGit.cpp
kevinushey: `core::system::addToPath()` should be safe to use when it's modifying an existing C++ object as opposed to the current environment. I'm not sure if there's a good way of enforcing this in the code base though :-/

src/cpp/session/postback/CMakeLists.txt
kevinushey: Will do!

src/cpp/session/SessionModuleContext.cpp
kevinushey: I was curious where the mixed slashes for rtools42 were coming from, and that's from R itself:

https://github.com/wch/r-source/blob/4ca6439c1ffc76958592455c44d83f95d5854b2a/src/library/profile/Rprofile.windows#L64-L76

I would argue that those should be backslashes instead, but it's not a huge deal ultimately.

src/cpp/session/SessionConsoleProcess.cpp
```diff
@@ -94,7 +94,21 @@ core::system::ProcessOptions ConsoleProcess::createTerminalProcOptions(
 #else
    core::system::setHomeToUserProfile(&shellEnv);
 #endif
-
+   
+   // don't forward GIT_ASKPASS or SSH_ASKPASS if it's set to our rpostback
```
kevinushey, line 98: We should still set these in non-interactive console processes, though.

kevinushey:
https://github.com/rstudio/rstudio/pull/12422 (picked wrong branch name here)

Issue #12382, opened 2022-11-29
C++ diagnostics failing with RTools 42 update


jgutman:
Using the latest version of RTools 42 on Windows ([rtools42-5355-5357](https://cran.r-project.org/bin/windows/Rtools/rtools42/news.html)) bumps the included version of gcc from 10.3 to 10.4. Unfortunately, RStudio is hardcoded to look for 10.3, leading to a bunch of misleading diagnostic warnings when using C++ code in the IDE.

User report from @AlexisDerumigny:
Hello, I had the same problems on Windows 10 with the latest version of R/RStudio/Rtools and I somehow managed to fix it.

There was a lot of meaningless diagnostic errors while the package was compiling well without any problems.
Furthermore, this only happened on my work laptop while my personal laptop did not have this problem at all (no meaningless diagnostic, no compilation errors), even though it has the latest version of R/Rstudio as well.

When running the diagnostic on my work laptop, I got
```
> .rs.setClangDiagnostics(2)
Attempting to load libclang for x86_64-w64-mingw32 
C:/Program Files/RStudio/bin/rsclang/x86_64/libclang.dll
   LOADED: 13.0.1

Restarting R session...

CLANG INDEXING: C:/Users/aderumigny/Documents/Packages/rsqlasso/rsqlasso/src/rsqlasso_fit.cpp
# RESTORING PACKAGE COMPILATION CONFIG ----
C:/Users/aderumigny/Documents/Packages/rsqlasso/rsqlasso/.Rproj.user/2F3AD38D/compilation-database/mingw64/config.json
{
    "args": [
        "-std=gnu++11",
        "-IC:/PROGRA~1/R/R-42~1.2/include",
        "-DNDEBUG",
        "-IC:/Users/aderumigny/AppData/Local/R/win-library/4.2/Rcpp/include",
        "-IC:/Users/aderumigny/AppData/Local/R/win-library/4.2/RcppEigen/include",
        "-DEIGEN_DONT_VECTORIZE",
        "-nostdinc",
        "-U_MSC_VER",
        "-D__GNUC__=10",
        "-D__GNUC_MINOR__=3",
        "-D__GNUC_PATCHLEVEL__=0",
        "-IC:/rtools42/x86_64-w64-mingw32.static.posix/lib/gcc/x86_64-w64-mingw32.static.posix/10.3.0/include/c++",
[...]

PR #12419, opened 2022-12-09
Compute gcc version automatically from installed RTools

jgutman:
### Intent

Addresses #12382 
### Approach
While we could choose to update RCompilationDatabase.cpp so that `discoverSystemIncludePaths` runs on Windows as well as Unix, it is more straightforward and requires less path/separator normalization to just update the existing RToolsInfo.cpp code so that rather than hardcoding gcc for a particular version of RTools, we infer the gcc version from the installation directory.

### Automated Tests

None
### QA Notes

See https://github.com/rstudio/rstudio/issues/12382#issuecomment-1344583816
### Checklist

### DIFF
--- a/src/cpp/core/r_util/RToolsInfo.cpp
+++ b/src/cpp/core/r_util/RToolsInfo.cpp
@@ -304,12 +304,6 @@ RToolsInfo::RToolsInfo(const std::string& name,
              gccPath.completeChildPath("include-fixed")
           };
 
-          for (auto&& stem : cStems)
-          {
-             cIncludePaths.push_back(stem.getAbsolutePath());
-             cppIncludePaths.push_back(stem.getAbsolutePath());
-          }
-
           // get C++ headers
           std::vector<FilePath> cppStems = {
              gccPath.completeChildPath("include/c++"),
@@ -317,8 +311,14 @@ RToolsInfo::RToolsInfo(const std::string& name,
              gccPath.completeChildPath("include/c++/backward")
           };
 
+          // Note: the order of these args matters. c++ headers should appear before c headers
           for (auto&& stem : cppStems)
              cppIncludePaths.push_back(stem.getAbsolutePath());
+          for (auto&& stem : cStems)
+          {
+             cIncludePaths.push_back(stem.getAbsolutePath());
+             cppIncludePaths.push_back(stem.getAbsolutePath());
+          }
       }
    }
    else
--- a/src/cpp/core/r_util/RToolsInfo.cpp
+++ b/src/cpp/core/r_util/RToolsInfo.cpp
@@ -276,34 +276,48 @@ RToolsInfo::RToolsInfo(const std::string& name,
       // not actually be defined or available in Rtools
       clangArgs.push_back("-U_MSC_VER");
 
-      // set GNUC levels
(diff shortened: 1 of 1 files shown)


src/cpp/core/r_util/RToolsInfo.cpp
```diff
@@ -276,34 +276,50 @@ RToolsInfo::RToolsInfo(const std::string& name,
...
-         "x86_64-w64-mingw32.static.posix/lib/gcc/x86_64-w64-mingw32.static.posix/10.3.0/include/c++/x86_64-w64-mingw32.static.posix",
-         "x86_64-w64-mingw32.static.posix/lib/gcc/x86_64-w64-mingw32.static.posix/10.3.0/include/c++/backward",
-         "x86_64-w64-mingw32.static.posix/lib/gcc/x86_64-w64-mingw32.static.posix/10.3.0/include",
-         "x86_64-w64-mingw32.static.posix/include",
-         "x86_64-w64-mingw32.static.posix/lib/gcc/x86_64-w64-mingw32.static.posix/10.3.0/include-fixed",
-      };
-
-      for (auto&& stem : cppStems)
-         cppIncludePaths.push_back(installPath.completeChildPath(stem).getAbsolutePath());
+      std::string toolDir = "x86_64-w64-mingw32.static.posix";
```
kevinushey, line 279: We don't have any guarantee this directory will still be used in future builds of Rtools. (what if future builds aren't static? what if MinGW changes the name for some reason?)

src/cpp/core/r_util/RToolsInfo.cpp
```diff
@@ -276,34 +276,50 @@ RToolsInfo::RToolsInfo(const std::string& name,
...
+      FilePath gccRootPath = installPath.completeChildPath(toolDir + "/lib/gcc/" + toolDir);
```
kevinushey, line 280: Similarly, I don't think we can assume this directory layout.

kevinushey (commented):
I feel pretty strongly that we should infer the compiler flags directly from the output of the gcc / g++ compilers, rather than continuing to try to hardcode them in this sort of fashion.

timtmok:
I think we can call `g++` or `c++` from `rtools42\x86_64-w64-mingw32.static.posix\bin`.  So calling `g++ -E -x c++ - -v` like it does in the non-Windows code would give the same information.

jgutman:
We could, I tried out calling the existing discoverSystemPaths function, which works correctly it just doesn't currently get called on Windows. It gets the right path, we just need some fixes for the separators since it ends up looking kind of ugly. It seemed like a bigger change to do that than to just fix the existing rtools method, but if you feel that's more robust in the long run I can do it that way -- I just ended up reverting those changes

kevinushey:
For some context: in the past, we assumed that new Rtools releases would essentially be frozen, so we could just hard-code the required information about the Rtools installation and update RStudio in tandem with R / Rtools releases.

Now that updates to Rtools are (1) decoupled from R releases (and so might appear at any time), and (2) bound to change in apparently unexpected ways, I think the extra effort to ensure we are extra robust is warranted.

This would also (hopefully) make updates for compatibility with future releases of Rtools easier, as we should be able to re-use much of the existing inference tools rather than re-writing it every time a new version of Rtools is released.

jgutman:
Abandoning in favor of #12465

PR #12422, opened 2022-12-12
rationalize usages of rpostback + postback script paths

kevinushey:
### Intent

Addresses (part of) https://github.com/rstudio/rstudio/issues/12390.

### Approach

- Make the development layout of postback utilities match that of package builds
- Try to rationalize usages of the `rpostback` path, plus the `postback` scripts directory

### Automated Tests

Not included.

### QA Notes

Test via various notes in https://github.com/rstudio/rstudio/issues/12390.

### Checklist


kevinushey:
I think this PR should be good to go now -- @jgutman, can you review once more?

src/cpp/r/RUtil.cpp
```diff
@@ -72,6 +80,79 @@ bool versionTest(const std::string& comparator, const std::string& version)
...
+   std::string newPath = prepend
+         ? fmt::format("{}{}{}", pathEntry, kPathSeparator, oldPath)
+         : fmt::format("{}{}{}", oldPath, kPathSeparator, pathEntry);
+
+#ifdef _WIN32
+   std::replace(newPath.begin(), newPath.end(), '/', '\\');
+#endif
+
+   boost::regex reDuplicateSeparators(kPathSeparator "+");
+   boost::regex_replace(newPath, reDuplicateSeparators, kPathSeparator);
```
jgutman, line 127: FYI I still see in some instances duplicate separators, so I'm not positive this bit is working

src/cpp/r/RUtil.cpp
```diff
@@ -72,6 +80,79 @@ bool versionTest(const std::string& comparator, const std::string& version)
 
 } // anonymous namespace
 
+void setenv(const std::string& key, const std::string& value)
```
jgutman, line 83: Is this something where we should potentially be replacing calls to core::system::setenv with calls to rutil::setenv in a loooooooot of places in the code?

jgutman (approved):

jgutman:
I appreciate the fix for development builds. However, please note that on MacOS, development builds, the value of `RS_RPOSTBACK_PATH` is  now `/Users/jacquelinegutman/rstudio/src/build/session/postback/rpostback.exe`, where clearly the `.exe` suffix is incorrect

kevinushey:
Thank you for catching that. I'll fix that up before merge.

src/cpp/r/RUtil.cpp
kevinushey: > I think you need to assign it to a result?

You are right; I incorrectly thought this routine took its first argument by reference and so mutated the argument in place. Thank you for catching that!

> also isn't a + needed in between kPathSeparator and "+" ?

In this case no; a feature of C / C++ is that sequential string tokens are automatically concatenated; e.g. `"a" "b" "c"` becomes `"abc"`. Since `kPathSeparator` is a macro expanding to a string we can take advantage of the same feature here.

src/cpp/r/RUtil.cpp
kevinushey: Potentially yes. 😞 

It's not completely obvious where we should be making these changes, since some calls to `core::system::setenv()` in the R session codebase run before the R session has been initialized, and I think those usages are safe (and in fact we wouldn't be able to use `r::util::setenv()` since it wouldn't be safe to call into R before it's been initialized)

kevinushey:
Thank you for the thorough review + catching my mistakes; it's much appreciated!

PR #12442, opened 2022-12-15
Sync to the latest patch release of quarto v1.2

jjallaire:
This fixes an important issue that prevents tinytex package auto-installation from working on osx.

If you end up doing another EG build this patch is worth taking (it has the change described above along with 2 other minor fixes). If you don't end up taking it we'll just need to tell users who encounter this problem to update manually (as RStudio will use the later externally installed version in preference to the internal version).

### DIFF
--- a/dependencies/common/install-quarto
+++ b/dependencies/common/install-quarto
@@ -27,7 +27,7 @@ fi
 
 # variables that control download + installation process
 # specify a version to pin for releases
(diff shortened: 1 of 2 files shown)


jgutman (approved):

jgutman:
Looking extremely unlikely we have anymore EG builds but if that changes I will merge this to that branch

PR #12465, opened 2022-12-19
use discover systemincludepaths on windows instead of hardcoded

jgutman:
Addresses https://github.com/rstudio/rstudio/issues/12382

### DIFF
--- a/src/cpp/session/modules/clang/RCompilationDatabase.cpp
+++ b/src/cpp/session/modules/clang/RCompilationDatabase.cpp
@@ -1178,10 +1178,17 @@ std::vector<std::string> RCompilationDatabase::baseCompilationArgs(bool isCpp) c
    args.push_back("-nostdinc");
 
    // add Rtools arguments
-   auto rtInfo = rToolsInfo();
-   auto rtArgs = rtInfo.clangArgs(isCpp);
-   args.insert(args.end(), rtArgs.begin(), rtArgs.end());
+   //auto rtInfo = rToolsInfo();
+   //auto rtArgs = rtInfo.clangArgs(isCpp);
+   //args.insert(args.end(), rtArgs.begin(), rtArgs.end());
 
+   // add system include headers as reported by compiler
+   std::vector<std::string> includes;
+   discoverSystemIncludePaths(&includes);
+   for (auto&& include : includes) {
+      FilePath includePath = FilePath(include);
+      args.push_back("-I" + includePath.getAbsolutePath());
+   }
    // re-generate compiler definitions
    generateCompilerDefinitions();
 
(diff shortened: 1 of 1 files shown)


src/cpp/session/modules/clang/RCompilationDatabase.cpp
```diff
@@ -1178,10 +1178,17 @@ std::vector<std::string> RCompilationDatabase::baseCompilationArgs(bool isCpp) c
    args.push_back("-nostdinc");
 
    // add Rtools arguments
-   auto rtInfo = rToolsInfo();
-   auto rtArgs = rtInfo.clangArgs(isCpp);
-   args.insert(args.end(), rtArgs.begin(), rtArgs.end());
+   //auto rtInfo = rToolsInfo();
+   //auto rtArgs = rtInfo.clangArgs(isCpp);
+   //args.insert(args.end(), rtArgs.begin(), rtArgs.end());
```
jgutman, line 1183: We can skip calling this function entirely in favor of the compiler reported paths from `where gcc.exe ...`

src/cpp/session/modules/clang/RCompilationDatabase.cpp
```diff
@@ -1178,10 +1178,17 @@ std::vector<std::string> RCompilationDatabase::baseCompilationArgs(bool isCpp) c
...
 
+   // add system include headers as reported by compiler
+   std::vector<std::string> includes;
+   discoverSystemIncludePaths(&includes);
+   for (auto&& include : includes) {
+      FilePath includePath = FilePath(include);
```
jgutman, line 1189: not strictly necessary to convert from string to file path and back to string, but it does clean up the paths and separators a bit

src/cpp/session/modules/clang/RCompilationDatabase.cpp
kevinushey: If possible, I'd prefer to keep the old code alive for older Rtools installations, just because we "know" it works and there's a possibility that this doesn't do what we expect with older Rtools installations.

In practice, that means using this code for R < 4.2.0, and the new "discover system includes" path for R >= 4.2.0 (with a comment to justify). Do you think that is reasonable?

kevinushey (approved):
LGTM! I would consider keeping the old code alive for older R installations (see comments in PR)

src/cpp/session/modules/clang/RCompilationDatabase.cpp
```diff
@@ -1177,10 +1179,21 @@ std::vector<std::string> RCompilationDatabase::baseCompilationArgs(bool isCpp) c
...
-   auto rtInfo = rToolsInfo();
-   auto rtArgs = rtInfo.clangArgs(isCpp);
-   args.insert(args.end(), rtArgs.begin(), rtArgs.end());
+   if (version < r_util::RVersionNumber(4, 2, 0))
+   {
+      // add Rtools arguments
+      auto rtInfo = rToolsInfo();
+      auto rtArgs = rtInfo.clangArgs(isCpp);
+      args.insert(args.end(), rtArgs.begin(), rtArgs.end());
+   } else {
```
kevinushey, line 1188:
```suggestion
   }
   else
   {
```

PR #12471, opened 2022-12-21
Add fixups that will set colors of vim/emacs-mode cursor to match theme

zsmeyer:
### Intent

Currently, when someone enables vim or emacs keybind mode in RStudio, the colors of the block cursor (introduced only in those keybind modes) does not reflect the current theme. Instead, the cursor is red (vim-mode) or bright green (emacs mode), no matter the theme. 

This adds some steps to the "compile themes" process that pulls colors from an already themed element, adds a bit of transparency/alpha, and then applies those colors to new elements that control the vim/emacs cursor.

I'll attach some screenshots of before/after for vim-mode below.

### Approach

I couldn't find a single element that seemed to be common across all themes (and yet was unique enough between different themes), so I just chose the fold element. This means that for a few themes that do not have fold color information (such as chrome), it will just default back to the ace editor default cursor color (red).

### Automated Tests

I don't think I remember seeing other tests in the compile-themes.R file (at least the one in gwt/tools), and I'm unfamiliar enough with testing of UI elements that I'm not sure how to write automated tests for this. I did import several of the generated themes into RStudio desktop and visually inspected the results. 

I'm happy to research some appropriate testing strategies and add them if that's helpful!

### QA Notes

One possible risk would be that this change makes RStudio less visually accessible. This would be the case if there are any themes where the new theme-related cursor colors make it harder to see the cursor and/or any symbols covered by the cursor than the default (red for vim, green for emacs). In addition, it is possible that using theme colors might make it harder to read for people who are colorblind. Perhaps the default red/green color was chosen intentionally because those colors maximize the contrast with surrounding text even for different kinds of colorblindness.

I also did not generate the themes using the source as-is. I couldn't get dependencies like the synchronization with ace to build locally (and I couldn't even get the devcontainer to build locally in VSCode; extremely likely this is my own inexperience). Instead, I just cloned the ace repo directly and pointed those parts of the script to my local version. I reverted those changes so this upstream wouldn't be referencing my local stuff. 

This means that someone (I can keep trying if that's best!) will need to actually generate the themes again and commit the newly generated .rstheme files to this repo. I could be reading this wrong, but it looks like that hasn't been done for a while, so it might be opening a big can of worms to do that now, just for this change.

### Documentation
I'm not sure I have access to modify existing documentation about theming (at least for users). Most of the information about RStudio themes seems to exist in Posit's blogs and/or forums. 

### Checklist

(I'm not sure this change aligns any better or worse with the accessibility standards than the current default. It would probably be ideal to check this by comparing the contrast ratio of the default cursor colors for each theme with the colors created by this change.)


zsmeyer:
## KR Theme before change
Screenshot from 2022-12-21 09-13-44
## KR Theme after change
Screenshot from 2022-12-21 09-16-10

## Solarized Dark before change
Screenshot from 2022-12-21 09-14-08
## Solarized Dark after change
Screenshot from 2022-12-21 09-22-47

## Tomorrow Night Blue before change
Screenshot from 2022-12-21 09-18-46
## Tomorrow Night Blue after change
Screenshot from 2022-12-21 09-22-01

## Dawn before change
Screenshot from 2022-12-21 09-13-13
## Dawn after change
Screenshot from 2022-12-21 09-17-34


## Dreamweaver before change
Screenshot from 2022-12-21 09-27-58
## Dreamweaver after change
Screenshot from 2022-12-21 09-28-39

PR #12479, opened 2022-12-21
Add logging for windows registry query / R installation search; plus fix system32 path issue

jgutman:
### Intent

Partially but sadly not at all fully address #12452

### Approach

For one user, this failed because C:\Windows\System32 wasn't on their path, so we use the absolute path to `reg.exe` rather than the shorthand `reg` to query the registry. For all other users, we haven't figured out exactly what the error is yet. Partly because logging is currently mostly useless. Adding a bunch more error logging here so we can at least get some useful feedback from users who do encounter this issue, now and in the future


jmcphers (approved):
LGTM, hope we figure out what's going on here!

Issue #12487, opened 2022-12-22
Upgrade to Electron v22 breaks at least some modals


jgutman:
RStudio was upgraded to Electron 22.0.0 in #12418, which became available in RStudio version [2023.03.0-daily+14](https://dailies.rstudio.com/version/2023.03.0-daily+14). This broke at least one modal (the choose R modal in Windows). 

In [2023.03.0-daily+13](https://dailies.rstudio.com/version/2023.03.0-daily+13):
<img width="388" alt="image" src="https://user-images.githubusercontent.com/7817881/209030778-f7615848-133f-4681-84f8-870e7f898537.png">

In [2023.03.0-daily+14](https://dailies.rstudio.com/version/2023.03.0-daily+14):
![image](https://user-images.githubusercontent.com/7817881/209030757-a1d64edb-1745-4722-aeb0-1b36f408233f.png)

I don't think there were any other changes between these 2 builds. I'm not sure if any other UI modals have been affected.

jgutman:
Moving this back into in progress to see if it's possible to refactor instead of turning on the node integration

jonvanausdeln:
Verified on 2023.03.0-daily+340 Desktop
Windows11, MacOS 12.6

Modal appears and works as expected

PR #12518, opened 2023-01-03
Fix unpopulated choose R dialog

timtmok:
### Intent
Address #12487 

### Approach
The Electron update defaults the `nodeIntegration` browser window preference to false, which prevents the preload script from running. The preload is necessary to hook up the UI listeners and populate the R versions. No other dialogs use a preload script so this is the only one affected.

### Automated Tests
None

### QA Notes
The choose R dialog is available in the preferences in the General section. This is the only dialog affected.

### Checklist

### DIFF
--- a/src/node/desktop/src/ui/modal-dialog.ts
+++ b/src/node/desktop/src/ui/modal-dialog.ts
@@ -31,6 +31,7 @@ export abstract class ModalDialog<T> extends BrowserWindow {
       height: 400,
       show: false,
       webPreferences: {
+        nodeIntegration: true,
         preload: preload,
       },
     };


jgutman (approved):
Thanks for updating this! What's the difference here between setting `nodeIntegration: true` versus setting `sandbox: false` here?

Asking because I see the release notes say:
> Beginning in Electron 20, renderers will be sandboxed by default, unless nodeIntegration: true or sandbox: false is specified.
> If your preload scripts do not depend on Node, no action is needed. If your preload scripts do depend on Node, either refactor them to remove Node usage from the renderer, or explicitly specify sandbox: false for the relevant renderers.

and I'm not sure on any tradeoffs between the approaches

kevinushey:
> Thanks for updating this! What's the difference here between setting `nodeIntegration: true` versus setting `sandbox: false` here?
> 
> Asking because I see the release notes say:
> 
> > Beginning in Electron 20, renderers will be sandboxed by default, unless nodeIntegration: true or sandbox: false is specified.
> > If your preload scripts do not depend on Node, no action is needed. If your preload scripts do depend on Node, either refactor them to remove Node usage from the renderer, or explicitly specify sandbox: false for the relevant renderers.
> 
> and I'm not sure on any tradeoffs between the approaches

I'd like to know this as well.

timtmok:
@kevinushey @jgutman 

The sandbox option enables a limited set of Node API while node integration has the full API available. I tried this with `sandbox: false` and it didn't work. It looks like enabling node integration will be necessary for the IPC calls since the dialog is based on a `BrowserWindow` which is in the main process and updating the UI needs to be done in the renderer process.

kevinushey:
Makes sense! Thanks for confirming.

gtritchie:
We should not enable node.js in renderers. The main window, for example, disables node.js and enables the sandbox, and runs a preload script (i.e. all that IPC/ContextBridge stuff for the GWT callbacks).

If the dialog's renderer code is using node.js directly, we need to refactor that to eliminate it, and instead provide those APIs via locked-down APIs invokes via the ContextBridge stuff.

Renderer code that can call node.js means javascript in the page could do things like manipulate the local filesystem, start processes, and so forth. Even for desktop-only code such as the "Choose R" dialog, we should stick to the recommended security model and turn off node.js.

kevinushey:
Does this imply Electron no longer has a way of running privileged preload scripts, even if the renderer itself is otherwise sandboxed after the preload script is run?

kevinushey:
From https://www.electronjs.org/docs/latest/tutorial/tutorial-preload, it seems like the approach described by @gtritchie is the preferred approach.

jgutman:
Okay, I had already gone ahead and merged this change because I was desperate for a temporary fix to the modal, but I'll go ahead and move Tim's issue back into in progress to see if it can be refactored instead of the quick fix of turning on node integration in renderers.

Issue #12520, opened 2023-01-03
Add support for sccache


zachhannum:
For use in PR Jenkins builds and hourlies, we would like to have the ability to use sccache. We should add sccache support to the CMake toolchain (default to off, configurable with an environment variable)

PR #12526, opened 2023-01-04
Add sccache support

zachhannum:
### Intent

Addresses #12520. Adds support for sccache to the build. Set the environment variable SCCACHE_ENABLED=1 to use sccache, otherwise sccache will not be used. The pull-request Jenkinsfile has also been updated to use sccache.

### Approach

An [S3 bucket has been created](https://s3.console.aws.amazon.com/s3/buckets/rstudio-build-cache?region=us-east-1&tab=objects) to act as a build cache for Jenkins (and developers, if they so choose). Note that this does not add sccache for Windows or OSX at this time. We should create new issues to address this if needed.

### Automated Tests

N/A

### QA Notes

N/A

### Documentation
N/A

### Checklist


cmake/modules/FindSccache.cmake
```diff
@@ -0,0 +1,26 @@
+find_program(CCACHE_PROGRAM sccache)
```
melissa-barca, line 1: This file looks very similar to sccache.cmake. Why are both necessary?

dependencies/common/install-sccache
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env bash
+
+#
+# install-sccache
+#
+# Copyright (C) 2022 by RStudio, PBC
```
melissa-barca, line 6: We 
```suggestion
# Copyright (C) 2023 by RStudio, PBC
```

We'll need to go through and update our copyrights with the new year, but best not to introduce any new 2022 Copyrights.

cmake/modules/FindSccache.cmake
zachhannum: Oops, good catch. You're right, this was a copy/paste error. Removed.

dependencies/common/install-sccache
zachhannum: Updated!

dependencies/common/install-sccache
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env bash
+
+#
+# install-sccache
+#
+# Copyright (C) 2023 by RStudio, PBC
```
melissa-barca, line 6:
```suggestion
# Copyright (C) 2023 by Posit Software, PBC
```

Sorry, I should've caught this the first time! This also needs to be Posit Software. I just `grep`-ed the repo and it looks like the majority of copyrights were updated but a few in `dependencies` were missed.

melissa-barca (approved):
LGTM!

dependencies/common/install-sccache
zachhannum: Went ahead and updated the copyright notices in `dependencies` to use Posit!

melissa-barca: Nice, thank you!

package/linux/make-package
```diff
@@ -225,6 +225,20 @@ if [ -n "${JENKINS_URL}" ]; then
    printenv
 fi
 
+# if SCCACHE_ENABLED environment variable is set, use sccache
+if [ -n "$SCCACHE_ENABLED" ]; then
+    echo "Using sccache"
+   
+    if [ -n "$AWS_ACCESS_KEY_ID" ] || [ $(aws sts get-caller-identity --query "Account" --profile ${AWS_PROFILE:-sso}) -eq 14 ]; then
```
cm421, line 232: Do any of these need to be pre-defined before this script runs? If so we'll want to add that info into the wiki

cm421 (changes requested):
Sorry for not getting these in before we merged, but I think the dependency script needs some attention

dependencies/common/install-sccache
```diff
@@ -0,0 +1,57 @@
...
+ARCH=x86_64
+SCCACHE_PACKAGE_HASH="94ea33aac8dcb358753f8240cc87345963cf83cda7c6af0395dff31ffdc88df4"
+
+# override to aarch64 if running on arm
+if [ "$(arch)" = "aarch64" ] ; then
+    ARCH=aarch64
+    SCCACHE_PACKAGE_HASH="1bf58385dc27b66324bb9ee82084e65c4d2e60baa19e3d16d2ab4da6c1ae66b2"
+fi
+
+URL=https://github.com/mozilla/sccache/releases/download/v${VERSION}/sccache-v${VERSION}-${ARCH}-unknown-linux-musl.tar.gz
```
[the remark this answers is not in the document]
zachhannum: I created #12533 and #12532 to address the sccache dependency for macOS and Linux, but I will create a small patch to move this script for now!

Issue #12532, opened 2023-01-05
Add sccache dependency for macOS builds


zachhannum:
Related to #12526, which adds sccache support, but will not work on macOS builds since the sccache dependency was only added for linux.

PR #12534, opened 2023-01-05
Prevent sccache install script on macOS

zachhannum:
This is a temporary fix to prevent the sccache install script from running on macOS until rstudio/rstudio#12532 is implemented.

Instead of moving the script, we can just prevent the install from running for now.


cm421 (approved):
LGTM! Thanks for this!

PR #12536, opened 2023-01-05
fix a bug when /usr/local/bin is available

MariaSemple:
### Intent

Fix failing pro builds.

### Approach

Sometimes `/usr/local/bin` is available. In that case, we don't set an install dir, and the last line fails because `/sentry-cli` doesn't exist. This attempts to fix that by setting a `SENTRY_PATH` variable which is either `sentry-cli` or `${INSTALL_DIR}/sentry-cli`.

Another way we could fix it is to always install sentry to `${INSTALL_DIR}`.

### Automated Tests

Build system failure.

### QA Notes

Build system failure.

### Documentation

None, build system failure.

### Checklist

### DIFF
--- a/dependencies/common/install-sentry-cli
+++ b/dependencies/common/install-sentry-cli
@@ -30,6 +30,9 @@ fi
 # specify a version
 export SENTRY_CLI_VERSION=2.9.0
 
-curl -sL https://sentry.io/get-cli/ | bash || true
+curl -sL https://sentry.io/get-cli/ | bash || trueSENTRY_PATH="sentry-cli"
+if [[ -n "${INSTALL_DIR}" ]]; then
+	SENTRY_PATH="${INSTALL_DIR}/${SENTRY_PATH}"
+fi
 
-info "Installed Sentry CLI version: `${INSTALL_DIR}/sentry-cli --version`"
+info "Installed Sentry CLI version: `${SENTRY_PATH} --version`"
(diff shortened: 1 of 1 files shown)


dependencies/common/install-sentry-cli
```diff
@@ -30,6 +30,9 @@ fi
 # specify a version
 export SENTRY_CLI_VERSION=2.9.0
 
-curl -sL https://sentry.io/get-cli/ | bash || true
+curl -sL https://sentry.io/get-cli/ | bash || trueSENTRY_PATH="sentry-cli"
```
skyeturriff, line 33:
```suggestion
curl -sL https://sentry.io/get-cli/ | bash || true
SENTRY_PATH="sentry-cli"
```

skyeturriff (approved):
I think just one small typo above, otherwise LGTM!

cm421 (approved):
Once Skye's fix is pulled in LGTM!

technocrat:
In community RStudio, the console disregards the user's $PATH and can find nothing in `/bin` or `/usr/bin`. This causes installs to fail, for lack of `tar` and knitting to fail for lack of `pandoc`, so please be aware in anything having to do with setting a $PATH for purposes of initialization doesn't continue to do this.

To reproduce, use

```
Sys.which("tar")
```

MariaSemple:
Thanks for your comment @technocrat! We are aware of the issue regarding adding `/usr/bin` to the path, but this change is only related to our build system. It doesn't affect the behaviour of the product at all.

PR #12544, opened 2023-01-09
Stub out a process creation time lookup for macOS

atheriel:
### Intent

We have an existing utility to look up the creation time of a process, but the only implementation is for Linux. This has been fine so far, as Launcher seems to be the only consumer of this API, but the compile-time guard is now generating issues for us when trying to build that project on macOS.

So this commit stubs out a macOS implementation that returns an error. Ideally, we'd actually implement this for macOS, but doing so [turns out to be nontrivial](https://stackoverflow.com/questions/31603885/get-process-creation-date-time-in-osx-with-c-c) so I don't think it's worth pursing at this time.

This should supersede https://github.com/rstudio/rstudio-pro/pull/3964. 

### Automated Tests

This change is a stub to avoid linker issues, so I don't think it needs any automated tests beyond "it compiles on macOS and Linux".

### QA Notes

This patch has near-zero risk.

### Documentation

N/A

### Checklist

### DIFF
--- a/src/cpp/core/include/core/system/System.hpp
+++ b/src/cpp/core/include/core/system/System.hpp
@@ -328,7 +328,7 @@ struct ProcessInfo
    std::string state;
    std::vector<std::string> arguments;
 
-#if !defined _WIN32 && !defined __APPLE__
+#ifndef _WIN32
    core::Error creationTime(boost::posix_time::ptime* pCreationTime) const;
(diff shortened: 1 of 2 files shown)


kfeinauer (approved):

MariaSemple (approved):
LGTM!

PR #12548, opened 2023-01-09
Update install-sccache with macOS support

zachhannum:
### Intent

Adresses #12532. Add ssccache support for macOS builds.

### Approach

Updated the sccache install script to support installing on macOS, and update the osx/make-package script to use sccache if the SCCACHE_ENABLED environment variable is set. I tested this on arm64 macOS the build seemed to succeed.

### Automated Tests

N/A

### QA Notes

N/A

### Documentation

N/A

### Checklist

### DIFF
--- a/dependencies/common/install-sccache
+++ b/dependencies/common/install-sccache
@@ -19,10 +19,6 @@ set -e
 
 source "$(dirname "${BASH_SOURCE[0]}")/../tools/rstudio-tools.sh"
 
-if is-macos; then
-   exit 0
-fi
-
 section "Installing sccache"
 
 # install dir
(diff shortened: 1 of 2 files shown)


timtmok:
The package build for Mac builds twice, once each for arm64 and x86_64. I'm not sure how that works with the caching. Does it do any caching for the x86_64 part of the build?

zachhannum:
@timtmok I did update the `make-package` script to use sccache for both the arm64 and x86_64 builds. They will both use the same cache in s3, but it is my understanding that those files will be cached separately for different architectures. See here: https://github.com/mozilla/sccache/blob/main/docs/Caching.md

timtmok:
Ah right, it should be good. I also forgot that the dependency installer scripts also install for each arch. Looks good!

timtmok (approved):

cm421 (approved):
Nice! LGTM

PR #12575, opened 2023-01-14
macOS Matrix Jenkins Pipeline

zachhannum:
### Intent

Addresses rstudio/rstudio-pro#4186. Combines `Jenkinsfile.macos` and `Jenkinsfile.macos.electron` into a single matrix build. Also add `sccache` support, integrate new version util, and new parameters (`DAILY` and `COMMIT_HASH`) for compatibility with the new job dispatcher.

### Approach

Some notes about the implementation:

#### Sequential Matrix
Since the macos agent does not run builds within docker containers, nor can we run parallel builds from the matrix across different nodes (for now), I used the `lock` mechanism to ensure that the two branches of the matrix, 'Desktop' (Qt) and 'Electron' effectively run sequentially, to prevent any "weirdness". This entire matrix will go away anyways once we move fully to Electron only on Pro. The Qt portion of the matrix does not even run for open source, only Pro.

#### Test runs

You can see test runs of this pipeline (without the upload logic) for [open source](https://build.posit.it/blue/organizations/jenkins/IDE%2Fexperiments%2Fmacos-matrix/detail/macos-matrix/17/pipeline) and [pro](https://build.posit.it/blue/organizations/jenkins/IDE%2Fexperiments%2Fmacos-matrix-pro/detail/macos-matrix-pro/2/pipeline).

### Automated Tests

N/A

### QA Notes

N/A

### Documentation

N/A

### Checklist


Jenkinsfile.macos
```diff
@@ -1,175 +1,282 @@
...
-      string(name: 'RSTUDIO_VERSION_MINOR',  defaultValue: '1', description: 'RStudio Minor Version')
-      string(name: 'RSTUDIO_VERSION_PATCH',  defaultValue: '999', description: 'RStudio Patch Version')
-      string(name: 'RSTUDIO_VERSION_SUFFIX', defaultValue: '+9999', description: 'RStudio Pro Suffix Version')
-      string(name: 'GIT_REVISION', defaultValue: 'main', description: 'Git revision to build')
-      string(name: 'BRANCH_NAME', defaultValue: 'origin/main', description: 'Branch name to build')
       string(name: 'SLACK_CHANNEL', defaultValue: '#ide-builds', description: 'Slack channel to publish build message.')
+      string(name: 'RSTUDIO_VERSION_PATCH',  defaultValue: '999', description: 'RStudio Patch Version')
+      booleanParam(name: 'DAILY', defaultValue: false, description: 'Runs daily build if true')
+      booleanParam(name: 'PUBLISH', defaultValue: true, description: 'Runs publish stage if true')
+      gitParameter defaultValue: "${env.GIT_BRANCH}",
```
zachhannum, line 20: @MariaSemple Is this what we should be doing (along with the "Checkout" stage below) for builds that are downstream of the job dispatcher? When I ran this this pipeline in my tests standalone, the `COMMIT_HASH` parameter ended up being null, so I had to comment it out to get it working, but I figured that this should work when the param is passed from an upstream job.

timtmok (approved):

cm421 (changes requested):

Jenkinsfile.macos
```diff
@@ -1,175 +1,283 @@
...
   parameters {
-      string(name: 'RSTUDIO_VERSION_MAJOR',  defaultValue: '1', description: 'RStudio Major Version')
-      string(name: 'RSTUDIO_VERSION_MINOR',  defaultValue: '1', description: 'RStudio Minor Version')
-      string(name: 'RSTUDIO_VERSION_PATCH',  defaultValue: '999', description: 'RStudio Patch Version')
-      string(name: 'RSTUDIO_VERSION_SUFFIX', defaultValue: '+9999', description: 'RStudio Pro Suffix Version')
-      string(name: 'GIT_REVISION', defaultValue: 'main', description: 'Git revision to build')
-      string(name: 'BRANCH_NAME', defaultValue: 'origin/main', description: 'Branch name to build')
       string(name: 'SLACK_CHANNEL', defaultValue: '#ide-builds', description: 'Slack channel to publish build message.')
+      string(name: 'RSTUDIO_VERSION_PATCH',  defaultValue: '999', description: 'RStudio Patch Version')
+      booleanParam(name: 'DAILY', defaultValue: true, description: 'Runs daily build if true')
```
MariaSemple, line 18: What do you think about defaulting DAILY to false? That way someone has to start a manual daily build from this job with intention (i.e. intentionally skipping the dispatcher)

Jenkinsfile.macos
```diff
@@ -1,175 +1,283 @@
...
+                    }
+                  }
+                  
+                  stage("Publish") {
+                    steps{
+                      // publish build to dailies page
+                      withCredentials([usernamePassword(credentialsId: 'github-rstudio-jenkins', usernameVariable: 'GITHUB_USERNAME', passwordVariable: 'GITHUB_PAT')]) {
+                        // publish the build
+                        sh "docker/jenkins/publish-build.sh --build ${PRODUCT}/macos --url https://s3.amazonaws.com/rstudio-ide-build/${PRODUCT}/macos/${PACKAGE_FILE} --pat ${GITHUB_PAT} --file package/osx/build/${PACKAGE_FILE} --version ${RSTUDIO_RELEASE}"
+                      }
```
MariaSemple, line 271: The string uses insecure interpolation for a secret. Also we can use `VAR = credentials('username_psw')` to get `VAR_USR` and `VAR_PSW`. `VAR` itself ends up being `${VAR_USR}:${VAR_PSW}`
```suggestion
                  stage("Publish") {
                    environment {
                      GITHUB_LOGIN = credentials('github-rstudio-jenkins')
                    }
                    steps{
                      // publish build to dailies page
                      sh 'docker/jenkins/publish-build.sh --build ${PRODUCT}/macos --url https://s3.amazonaws.com/rstudio-ide-build/${PRODUCT}/macos/${PACKAGE_FILE} --pat ${GITHUB_LOGIN_PSW} --file package/osx/build/${PACKAGE_FILE} --version ${RSTUDIO_RELEASE}'
```

MariaSemple (changes requested):
Big improvements!

Jenkinsfile.macos
MariaSemple: Yep, that looks right to me!

Jenkinsfile.macos
zachhannum: I definitely don't have a strong preference either way--- your logic definitely makes sense, though it sounds like [Craig is leaning more towards `defaultValue: true`](https://github.com/rstudio/rstudio-pro/pull/4241#discussion_r1081477819).

Jenkinsfile.macos
```diff
@@ -1,175 +1,283 @@
...
+
+                  stage("Notarize") {
+                    steps {
+                      withCredentials([usernamePassword(credentialsId: 'ide-apple-notarizer', usernameVariable: 'APPLE_ID', passwordVariable: 'APPLE_ID_PASSWORD')]) {
+                        sh "docker/jenkins/notarize-release.sh package/osx/build/${PACKAGE_FILE}"
+                      }
+                    }
+                  }
+
+                  stage("Upload") {
```
MariaSemple, line 209: We should add a when PUBLISH block here

Jenkinsfile.macos
```diff
@@ -1,175 +1,283 @@
...
+                    steps {
+                      // this job is going to run on a macOS slave, which cannot use an instance-profile
+                      withCredentials([[$class: 'AmazonWebServicesCredentialsBinding', credentialsId: 'jenkins-aws']]) {
+                        retry(5) {
+                          sh "aws s3 cp package/osx/build/${PACKAGE_FILE} ${BUILD_DEST}"
+                        }
```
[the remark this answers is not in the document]
zachhannum: Done!

Jenkinsfile.macos
```diff
@@ -1,175 +1,283 @@
...
+                    when { // See https://github.com/rstudio/rstudio-pro/commit/0590beeba8ff898b6c9ed8f524ca734bb830f930
+                      expression {
+                        return IS_PRO != true && FLAVOR == 'Electron';
+                      }
+                    }
+                    steps {
+                      // upload daily build redirects
+                      withCredentials([file(credentialsId: 'www-rstudio-org-pem', variable: 'wwwRstudioOrgPem')]) {
+                        sh 'docker/jenkins/publish-daily-binary.sh https://s3.amazonaws.com/rstudio-ide-build/${PRODUCT}/macos/${PACKAGE_FILE} ${RSTUDIO_ORG_PEM}'
+                      }
```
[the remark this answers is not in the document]
zachhannum: Done!

Jenkinsfile.macos
zachhannum: Done!

Jenkinsfile.macos
```diff
@@ -1,175 +1,283 @@
...
-      }
-
+    stage('Load Utils') {
       steps {
         script {
-          currentBuild.displayName = "${params.RSTUDIO_VERSION_MAJOR}.${params.RSTUDIO_VERSION_MINOR}.${params.RSTUDIO_VERSION_PATCH}${params.RSTUDIO_VERSION_SUFFIX}"
-        }
-        withCredentials([usernameColonPassword(credentialsId: 'github-rstudio-jenkins', variable: "GITHUB_LOGIN")]) {
-          sh 'cd dependencies/osx && RSTUDIO_GITHUB_LOGIN=$GITHUB_LOGIN ./install-dependencies-osx && cd ../..'
+          sh 'printenv'
```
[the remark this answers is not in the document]
zachhannum: Done!

Jenkinsfile.macos
```diff
@@ -1,175 +1,283 @@
...
     }
 
-    stage('tests') {
-
+    stage('Install Dependencies') {
       environment {
-        PATH = "${env.HOME}/opt/bin:${env.PATH}"
+        // boost won't compile without the brew version of openssl.
+        // only add it to the dep resolve step though, or the ide build will compile against the wrong openssl
+        PATH = '/usr/local/opt/openssl/bin:/usr/local/opt/openssl/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin'
```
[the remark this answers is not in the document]
zachhannum: Done!

Jenkinsfile.macos
```diff
@@ -1,175 +1,283 @@
...
-          try {
-            // attempt to run cpp unit tests
-            // problems with rsession finding openssl, so those tests
-            // are disabled until we solve it (#6890)
-            sh "cd package/osx/build/src/cpp && arch -x86_64 ./rstudio-tests"
-          } catch(err) {
-             currentBuild.result = "UNSTABLE"
-          }
+        withCredentials([usernameColonPassword(credentialsId: 'github-rstudio-jenkins', variable: "GITHUB_LOGIN")]) {
+          sh 'cd dependencies/osx && RSTUDIO_GITHUB_LOGIN=$GITHUB_LOGIN ./install-dependencies-osx && cd ../..'
```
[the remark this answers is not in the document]
zachhannum: Done!

Jenkinsfile.macos
zachhannum: There is a when block for the `PUBLISH` param at the 'Notarize and Upload' parent stage, so all child stages should also get skipped based on that top-level when block.

Jenkinsfile.macos
```diff
@@ -1,175 +1,283 @@
...
+          utils = load "${env.WORKSPACE}/utils.groovy"
         }
       }
-
     }
 
-    stage('build') {
+    stage ("Checkout") {
+      when {
+        expression { COMMIT_HASH != null}
```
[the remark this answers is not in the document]
zachhannum: So I went back and did some testing here, cause I could never get it to work in my experimental pipeline.

It seems to me that we are unable to use environment variables inside of the top `parameters` block. I would just end up with a `null` value and the checkout would fail.

I tried comparing against `null` with
```
when {
   expression { return COMMIT_HASH != null }
}
```

Both passing in `defaultValue: 'none'` and `defaultValue: null` would result in:
```
[2023-01-19T22:09:51.944Z] Commit_hash value: null
```
Which meant that in the "Checkout" stage, the `when` block check did not work when doing a null comparison and it ran the stage anyways.

The only solution I could get to work is passing in `defaultValue: 'none'` and doing:
```
when {
  expression { params.COMMIT_HASH != 'none' }
}
```
With that, it would correctly skip the stage. So that works, but I'm definitely open to suggestions for a better approach.

cm421 (approved):
LGTM!

Jenkinsfile.macos
zachhannum: Just to close the loop on this comment, I did in fact default `DAILY` to `false` since it seems like that's the way we are leaning.

Issue #11520, opened 2022-06-28
Code examples with scrollbar in R 4.2.0 Help Pane don't have a solid gray background


ronblum:
### System details

    RStudio Edition : Desktop and Server
    RStudio Version : 2022.07.0-preview+543
    OS Version      : MacOS 12.5, Windows 11, Ubuntu 20.04 via Chrome and Safari on MacOS 12.5
    R Version       : 4.2.0

### Steps to reproduce the problem

1. In Console Pane: `??dplyr`
2. In Help Pane: Click on the Vignette `dplyr::dplyr`
3. Scroll down till reaching a code example.
    - The code example should be on a solid background.
5. Shrink the width of the Help Pane enough that a scroll bar appears for the code.
6. Scroll right.
    - Only the rows that are too long for the window are on a solid (gray) background. The rest of the code block is on a white or black background, in accordance with the light or dark theme.

### Describe the problem in detail

The background is no longer a solid block.

<img width="311" alt="Screen Shot 2022-06-28 at 10 48 37 AM" src="https://user-images.githubusercontent.com/31009952/176211146-1bbd7ee0-c04c-4ce9-8bd3-4ec98234830c.png">

### Describe the behavior you expected

The block should remain a solid block.

- [ ] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ ] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [ ] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
Compare this to using `?data.frame` which works correctly:
<img width="312" alt="dataframe" src="https://user-images.githubusercontent.com/31009952/176431706-58616f29-40fe-4da9-918c-6f281ea71fa2.png">

kevinushey:
The issue here is that the CSS background is applied to the inner pre element, but not the div containing the code.

https://github.com/rstudio/rstudio/blob/14f8bf3517f17220d6727aa9fc9e9eeeff951b31/src/gwt/src/org/rstudio/studio/client/workbench/views/help/HelpPaneEditorStyles.css#L78-L83

We might be able to resolve this by including `div.sourceCode` in the CSS rule here.

[...]

PR #12593, opened 2023-01-18
Fix help pane scrollbar code background color

sharon-wang:
### Intent

Addresses the code background color bug in #11520.

#### Dark Theme Fix Preview
image

#### Light Theme Fix Preview
image

### Approach
Sets the `background-colour` of the `div.sourceCode` to the light or dark code background color, depending on the active theme. The `border-radius` is also set to the same value as `pre`, otherwise the code box loses the rounded corners.

**Note:** The code background grey color is defined in `HelpPaneEditorStyles.css` and there is one grey used for all dark themes and another grey used for all light themes. It's possible the grey could be incompatible with some of the dark or light themes from an accessibility standpoint. Should a code background color be defined for each theme?

### Automated Tests
N/A?

### QA Notes

- Please test with a light editor theme and a dark editor theme
- Please test with both `Modern` and `Sky` RStudio themes for light and dark editor themes
- Other code areas outside of the help panel should not be affected by this change

### Documentation
N/A

### Checklist

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/help/HelpPaneEditorStyles.css
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/help/HelpPaneEditorStyles.css
@@ -5,7 +5,7 @@
 
 @def CODE_COLOR #f0f0f0;
 
-@external rstudio-themes-dark, rstudio-themes-default;
+@external rstudio-themes-dark, rstudio-themes-light-menus, rstudio-themes-dark-menus;
 @external rstudio-themes-scrollbars;
 @external editor_dark;
 @external ace_editor_theme;
@@ -194,11 +194,11 @@ p + div[style="text-align: center;"]
    border-radius: 3px;
 }
 
-.rstudio-themes-default.ace_editor_theme div.sourceCode {
+.rstudio-themes-light-menus.ace_editor_theme div.sourceCode {
    background-color: CODE_BACKGROUND_COLOR_LIGHT;
 }
 
-.rstudio-themes-dark.editor_dark.ace_editor_theme div.sourceCode {
+.rstudio-themes-dark-menus.ace_editor_theme div.sourceCode {
    background-color: CODE_BACKGROUND_COLOR_DARK;
 }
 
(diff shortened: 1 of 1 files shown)


src/gwt/src/org/rstudio/studio/client/workbench/views/help/HelpPaneEditorStyles.css
```diff
@@ -183,5 +185,20 @@ p + div[style="text-align: center;"]
...
+/**
+ * This styling ensures the sourceCode area has a solid, continuous
+ * background color, even if the view is narrowed and a scrollbar
+ * appears to allow for horizontal scrolling.
+ */
+ .ace_editor_theme div.sourceCode {
+   border-radius: 3px;
+}
+
+.rstudio-themes-default.ace_editor_theme div.sourceCode {
```
timtmok, line 197: This has a slightly different look in light mode for some editor themes. When switching to `Chrome` or `Eclipse` for example, it doesn't cover the entire background with the colour. It seems that there's no `rstudio-themes-default` for these themes but there is a `rstudio-themes-light` class on it.

image

sharon-wang: For some reason `rstudio-themes-default` shows up for the themes `Chrome` and `Eclipse` on my end (and the background colour is covered), but I can use `rstudio-themes-light-menus` instead which also works for light themes on my end.

I've also changed the dark theme rule to use `rstudio-themes-dark-menus` instead, so it's similar to the light theme rule. I tested with various light and dark themes and seems to be working for me. Let me know if you see different behaviour!

timtmok: I tried out some different settings and `rstudio-themes-default` is tied to the Appearance -> RStudio Theme setting with the `Modern` value. Changing it to `Sky` will use `rstudio-themes-alternate`.

It feels kind of weird to use `rstudio-themes-[light|dark]-menus` for the selector. I think a CSS OR selector would work better here to handle the different classes used to represent `Modern` and `Sky`.

timtmok (approved):
Good work!

PR #12594, opened 2023-01-18
Avoid non-zero exit codes when running sessions under Launcher

atheriel:
### Intent

Users sometimes raise concerns about the number of nonzero exit codes they see from Workbench sessions on Slurm and Kubernetes (see one [recent example](https://rstudio.slack.com/archives/C7312LBPY/p1670494955085729)). This is because we use exit code 101 and 103 to signify switching between projects and suspending, respectively, both of which are common operations in these environments.

This only affects Workbench. I extracted my original patch from that repo because it only affects the open-source codebase.

### Approach

It seems that we don't rely on the exit code when running sessions under Launcher at all, so this is an easy change to make.

### Automated Tests

I don't think there are any automated tests for this. In my manual testing, switching between projects and suspending sessions continues to work just fine with zero as the exit code.

### QA Notes

This will break switching between projects and suspending (with either Launcher or without it), which I believe should be caught by existing testing.

### Documentation

N/A

### Checklist

### DIFF
--- a/src/cpp/session/SessionSuspend.cpp
+++ b/src/cpp/session/SessionSuspend.cpp
@@ -459,8 +459,18 @@ void checkForSuspend(const boost::function<bool()>& allowSuspend)
             "computations may have been interrupted.");
       }
 
+      // exit status
+      int status = EX_FORCE;
+      if (options().getBoolOverlayOption(kLauncherSessionOption))
+      {
+         // Avoid generating nonzero exit codes when running under Launcher.
+         // Error codes from normal behaviour like this are confusing for Slurm
+         // and Kubernetes administrators unfamiliar with our workloads.
+         status = EXIT_SUCCESS;
+      }
+
       // execute the forced suspend (does not return)
-      suspendSession(true, EX_FORCE);
+      suspendSession(true, status);
    }
 
    // cooperative suspend request
--- a/NEWS.md
+++ b/NEWS.md
@@ -15,6 +15,7 @@
 - "Rainbow" fenced divs, controled by Options > R Markdown > [v] Use rainbow fenced divs #12115
 - Disable argument tooltips in script editor for unknown functions #12160
 - Sessions now have lower CPU priority during suspension on macOS and Linux #12623
+- Sessions run on Kubernetes or Slurm will no longer exit with nonzero codes under normal circumstances (rstudio/rstudio-pro#3375)
 
 ### Fixed
 
--- a/src/cpp/session/SessionHttpMethods.cpp
+++ b/src/cpp/session/SessionHttpMethods.cpp
@@ -768,7 +768,15 @@ void handleConnection(boost::shared_ptr<HttpConnection> ptrConnection,
 
             // exit status
             int status = switchToProject.empty() ? EXIT_SUCCESS : EX_CONTINUE;
-            
+            if (options().getBoolOverlayOption(kLauncherSessionOption))
+            {
+               // Avoid generating nonzero exit codes when running under
+               // Launcher. Error codes from normal behaviour like this are
+               // confusing for Slurm and Kubernetes administrators unfamiliar
+               // with our workloads.
+               status = EXIT_SUCCESS;
+            }
+
             // acknowledge request & quit session
             json::JsonRpcResponse response;
             response.setResult(true);


MariaSemple (changes requested):
This is related to https://github.com/rstudio/rstudio-pro/issues/3375. We do actually check the exit code, so I don't think we can take this PR as-is. This is prioritised quite high for the upcoming release(s), so if you leave this open we can expand on the work you've started here.

In my opinion, we should be saving this information as part of the `RActiveSession` object. Eventually, that will end up in the database. For now it will be on disk.

atheriel:
After some consideration, I'm not sure this change will interact with that exit code check (at https://github.com/rstudio/rstudio-pro/blob/60c00c2de7ff8cf28d1e0898d1a4c0464c9e6fc0/src/cpp/session/SessionSourceDatabaseSupervisor.cpp#L576-L579) at all.

Let me explain, and please excuse the length.

The check in question triggers when a session is suspended and writes a different file depending on whether the exit code is `EX_SUSPEND_RESTART_LAUNCHER_SESSION || EX_CONTINUE`, or not.

This patch, in turn, changes the exit code in two places:

1. When switching projects, right before the call to `rstudio::r::session::quit()`.
2. When forcing a suspend, right before the call to `suspendSession()`.

In the case of (1), we never seem to trigger the suspend hooks at all (`quit()` [doesn't seem to call them](https://github.com/rstudio/rstudio-pro/blob/60c00c2de7ff8cf28d1e0898d1a4c0464c9e6fc0/src/cpp/r/session/RQuit.cpp#L93-L114)).

In the case of (2), we *do* end up calling that hook, but we're changing the status code from `EX_FORCE` to `EXIT_SUCCESS` -- neither of which satisfies the first branch of the ternary statement, and so results in the same behaviour.

If you think about it, this makes sense, too -- this patch does not change the behaviour around session restarts because under the Launcher we *already* emit a different exit code (and handle it in the `rsession-run` script) instead.

The only exit codes customers [report seeing](https://positpbc.slack.com/archives/C7312LBPY/p1670494955085729) in Kubernetes/Slurm are `100` (`EX_CONTINUE`) and `101` (`EX_FORCE`), and this check is specifically around restart behaviour, which is `EX_SUSPEND_RESTART_LAUNCHER_SESSION`.

When I said in the PR above that "we don't rely on the exit code when running sessions under Launcher at all", I should have clarified that I meant, specifically, that we don't rely on `EX_FORCE` or `EX_CONTINUE`.

atheriel:
> In my opinion, we should be saving this information as part of the RActiveSession object. Eventually, that will end up in the database. For now it will be on disk.

That might still be a good idea, even if my argument is correct.

jeffvroom:
I shared Maria's concern at first, but I think this is a safe change.  The only places I can see that we check the exit code are in the rsession-run script and in the onSuspend handler for the job launcher and neither of those will be affected by this change.

src/cpp/session/SessionSuspend.cpp
```diff
@@ -433,6 +433,13 @@ void checkForSuspend(const boost::function<bool()>& allowSuspend)
       }
 
       // execute the forced suspend (does not return)
+      if (options().getBoolOverlayOption(kLauncherSessionOption))
+      {
+         // Avoid generating nonzero exit codes when running under Launcher.
+         // Error codes from normal behaviour like this are confusing for Slurm
+         // and Kubernetes administrators unfamiliar with our workloads.
+         suspendSession(true, EXIT_SUCCESS);
```
jeffvroom, line 441: A minor readability point... I'd rather see a variable `statusCode` that is set to EXIT_SUCCESS or EX_FORCE and passed to one suspendSession(..) call.  That important comment "does not return" will then apply to that single suspendSession call and the reader won't be left wondering if suspendSession would ever get called twice.

atheriel: Very sensible, and done. It makes the code in both places look identical, too.

atheriel:
From my reading, this could probably considered a fix for https://github.com/rstudio/rstudio-pro/issues/3375 -- should I add that issue to the `NEWS` entry?

MariaSemple (approved):
LGTM!

MariaSemple:
Yes, go ahead and add a NEWS entry. If we test and there are still cases where this is occurring, we'll update the wording

PR #12612, opened 2023-01-23
Remove `jenkins-aws` credentials

cm421:
### Intent

Addresses https://github.com/rstudio/rstudio-pro/issues/3022.

The `jenkins-aws` credential will be turned off by CI at the end of the day today. We've removed it as part of our Jenkins restructuring, but those changes are sitting in an epic branch and it won't be merged into `main` for at least a week, so fixing this now.

### Approach

Switch to `ide-build` or `aws-build-role` instead

### Automated Tests

N/A. Build system change

### QA Notes

N/A Build system Change

### Documentation

N/A Build system change

### Checklist

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -480,8 +480,7 @@ try {
                                     bat "move package\\win32\\build\\${packageName}-RelWithDebInfo.exe package\\win32\\build\\${packageName}.exe"
                                     bat "move package\\win32\\build\\${packageName}-RelWithDebInfo.zip package\\win32\\build\\${packageName}.zip"
 
-                                    // windows docker container cannot reach instance-metadata endpoint. supply credentials at upload.
-
+                                    // Explicitly assume the IDE alias of the main `build` account
                                     withAWS(role: 'ide-build') {
                                         retry(5) {
                                             bat "aws s3 cp package\\win32\\build\\${packageName}.exe ${buildDest}/${packageName}.exe"
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -480,7 +480,7 @@ try {
                                     bat "move package\\win32\\build\\${packageName}-RelWithDebInfo.exe package\\win32\\build\\${packageName}.exe"
                                     bat "move package\\win32\\build\\${packageName}-RelWithDebInfo.zip package\\win32\\build\\${packageName}.zip"
 
-                                    // Explicitly assume the IDE alias of the main `build` role
+                                    // Explicitly assume the ide-build role in the main AWS account
                                     withAWS(role: 'ide-build') {
                                         retry(5) {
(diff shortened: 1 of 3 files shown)


melissa-barca (approved):
Overall looks good, but for my own understanding what is the difference between `ide-build` and `aws-build-role`?

cm421:
What I understand is that `aws-build-role` has more permissions and is the default account for working with our docker images in ECR (so `pullBuildPush` and `docker.withRegistery`).

`ide-build` has much fewer permissions but is enough for us to do everything else like upload to S3

Jenkinsfile
```diff
@@ -482,7 +482,7 @@ try {
 
                                     // windows docker container cannot reach instance-metadata endpoint. supply credentials at upload.
```
jspiewak, line 483: This comment is somewhat misleading. The way we have Jenkins configured is that agents have access to the AWS role that the controller runs as by default, i.e. without the need to call `withAWS` at all.

It is good to explicitly assume the `ide-build` role using `withAWS`, as we are encouraging teams to use team specific roles in order to not overload the controller's role (we had previously maxed out the size of its policy).

Jenkinsfile.macos
```diff
@@ -113,7 +113,7 @@ pipeline {
           }
 
           // this job is going to run on a macOS slave, which cannot use an instance-profile
```
jspiewak, line 115: Ditto about this being misleading.

Jenkinsfile.macos.electron
```diff
@@ -129,7 +129,7 @@ pipeline {
           }
 
           // this job is going to run on a macOS slave, which cannot use an instance-profile
```
jspiewak, line 131: Ditto.
Also, we've been working hard to get away from master/slave, use controller/agent instead please.

Jenkinsfile
cm421: Leaving the `withAWS` block to be explicitly clear about what role we're using for the benefit of the next developer that reads it but isn't familiar with the default assumed AWS role

Jenkinsfile
```diff
@@ -480,7 +480,7 @@ try {
                                     bat "move package\\win32\\build\\${packageName}-RelWithDebInfo.exe package\\win32\\build\\${packageName}.exe"
                                     bat "move package\\win32\\build\\${packageName}-RelWithDebInfo.zip package\\win32\\build\\${packageName}.zip"
 
-                                    // Explicitly assume the IDE alias of the main `build` account
+                                    // Explicitly assume the IDE alias of the main `build` role
```
jspiewak, line 483: Not quite accurate. This is an actual IAM role named `ide-build`.
The Jenkins credential `aws-build-role`, in the IDE folder, overrides the global credential and sets it to the `ide-build` IAM role.

timtmok (approved):

PR #12615, opened 2023-01-23
Update versioning for hourly builds

zachhannum:
### Intent

This PR contains a number of updates to the versioning logic to allow us to explicitly mark builds as "hourly" builds to help distinguish them from "daily" builds.

image

### Approach

There are a number of changes that are a part of this PR outlined below.

#### Updates to `rstudio-version.sh`

I've updated `rstudio-version.sh` to use keyword arguments to make what is being passed in more explicit, while adding a new `--build-type` argument, which can be used to override the build type found in `version/BUILDTYPE` for versioning purposes.

#### Updates to `utils.groovy`

I've updated the `getVersion()` function in `utils.groovy` to take a `isHourly` parameter, which defaults to `false` to retain current functionality. When `isHourly` is set to `true`, it will pass in `--build-type=hourly` to generate a version string with `hourly` as the build type.

#### Updates to CMake

Previously we were reading the build type and setting `RSTUDIO_BUILD_TYPE` based on the `version/BUILDTYPE` file, which is used for retrieving versioning information from the `rsession` (shown in the "About Rstudio" dialog in the IDE). The `CMakeRelease.txt` file has been updated to check the `RSTUDIO_VERSION_SUFFIX` environment variable first to extract the build type from there, otherwise it falls back to reading from the `BUILDTYPE` file as before.

### Checklist


MariaSemple (approved):
LGTM!

melissa-barca (approved):
LGTM!

PR #12623, opened 2023-01-24
Lower CPU priority during session suspend

atheriel:
### Intent

Large sessions can be CPU-intensive to serialize, and customers have often reported mysterious CPU spikes that were tracked down to session suspension. This CPU spike can be particularly inconvenient when it is caused by previously-dormant sessions that time out and are auto-suspended by Workbench on a busy system.

Recently, [a customer asked](https://app.prodpad.com/feedback/874d975e-9b8c-11ed-a544-2a7db0eb1d9c/canvas) if we could lower the session process's nice value right before suspending it. This seemed like a good idea to me, so this commit implements this for macOS and Linux. We could also do it on Windows with `SetPriorityClass()`, but on that OS the likelihood of being a multi-user system is much lower.

This change doesn't eliminate the CPU required to suspend, of course, (or the I/O to do so, which can be significant) but it does make suspension much less likely to interfere with other, more important workloads on the system.

I'm not 100% sure this is a unilaterally good idea, since it will affect Desktop as well. Any comments on that front are appreciated.

### Automated Tests

There are no automated tests that cover this. I did do manual testing and it works as expected.

### QA Notes

Please advise.

### Documentation

N/A

### Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,6 +14,7 @@
 - Support searching for roxygen specific information in fuzzy finder (Go to File/Function) #12190
 - "Rainbow" fenced divs, controled by Options > R Markdown > [v] Use rainbow fenced divs #12115
 - Disable argument tooltips in script editor for unknown functions #12160
+- Sessions now have lower CPU priority during suspension on macOS and Linux #12623
 
 ### Fixed
 
--- a/src/cpp/session/SessionSuspend.cpp
+++ b/src/cpp/session/SessionSuspend.cpp
@@ -17,6 +17,10 @@
 #include <signal.h>
 #include <unordered_set>
 
+#ifndef _WIN32
+#include <sys/resource.h>
+#endif
+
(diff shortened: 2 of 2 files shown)


timtmok:
I'm not entirely sure how the resume part works but it looks like there's code to set the priority again.

https://github.com/rstudio/rstudio/blob/df49a86e1f987640d4758b8996fbadae7330a89b/src/cpp/core/system/PosixSystem.cpp#L2079-L2083

But it seems to check for `!= 0` before setting the priority. My understanding is that the priority is a value from -20 to 20, with higher being lower priority. It sounds like 0 is valid but wouldn't be set if that was the priority before suspending.

atheriel:
My understanding is that the process exits after a suspend and a new process (which will have the default priority again) is always started when resuming.

jeffvroom (approved):
LGTM!  I like this.

PR #12639, opened 2023-01-27
Add support for serving static files precompressed with Gzip/Brotli

atheriel:
### Intent

At present, we compress static files at runtime to reduce the bandwidth cost of sending them over the network. Unfortunately, this compression is by no means free, and we incur significant CPU overhead to serve files this way (not to mention the latency added while we do so).

This has an especially large performance impact for the large Javascript/CSS assets on the Workbench Homepage, which are static files that never change.

It's common practice to avoid the runtime compression cost for static assets by precompressing the files beforehand, either with Gzip or, more commonly these days, with Brotli.

This commit modifies the static file handler to substitute in precompressed files when they exist, much like NGINX's popular `gzip_static` and `brotli_static` directives.

Note that this patch *does not* actually compress any of the assets that we serve. It only makes it possible to do so.

### Automated Tests

This part of the core library isn't covered by unit tests, but I did confirm it works manually in the browser.

### QA Notes

This change is low risk as it should have no effect on existing behaviour; it makes other changes possible in the future.

### Documentation

N/A

### Checklist

### DIFF
--- a/src/cpp/core/include/core/http/Response.hpp
+++ b/src/cpp/core/include/core/http/Response.hpp
@@ -23,6 +23,7 @@
 #include <boost/optional.hpp>
 #include <boost/type_traits/is_same.hpp>
 #include <boost/make_shared.hpp>
+#include <boost/algorithm/string/predicate.hpp>
 #include <boost/iostreams/copy.hpp>
 #include <boost/iostreams/concepts.hpp>
 #include <boost/iostreams/filtering_stream.hpp>
@@ -209,7 +210,8 @@ public:
    Error setBody(std::istream& is, 
                  const Filter& filter, 
                  std::streamsize buffSize = 128,
-                 bool padding = false)
+                 bool padding = false,
+                 bool enableRuntimeCompression = true)
    {
       try
       {
@@ -224,7 +226,7 @@ public:
             filteringStream.push(filter, buffSize);
 
          // handle gzip
-         if (contentEncoding() == kGzipEncoding)
+         if (contentEncoding() == kGzipEncoding && enableRuntimeCompression)
 #ifdef _WIN32
             // never gzip on win32
             removeHeader("Content-Encoding");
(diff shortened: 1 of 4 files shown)


src/cpp/core/include/core/http/Response.hpp
```diff
@@ -287,7 +289,8 @@ class Response : public Message
       // send the file from its stream
       try
       {
-         return setBody(*pIfs, filter, buffSize, padding);
+         bool precompressed = boost::algorithm::ends_with(filePath.getAbsolutePath(), ".gz");
```
GCRev, line 292: Should this also check for files ending in `.br`, or are we only ever requesting `.gz` files?

atheriel: This is an awkward consequence of how the current `Response` class methods call one another. Further down the chain `setBody()` will automatically gzip anything with a `content-encoding: gzip` header, so we need to add an escape hatch to avoid that for files that are already compressed -- we can't add the header afterwards. But we don't have an automatic Brotli step in `setBody()` (preferring Gzip for all cases), so I didn't need to add an escape hatch for it. A better variable name here would be `alreadyGzipped` rather than `precompressed`, I suppose.

src/cpp/core/include/core/http/Response.hpp
```diff
@@ -320,10 +323,13 @@ class Response : public Message
       }
       
       // set content type
-      setContentType(filePath.getMimeContentType());
+      if (contentType().empty())
+      {
+         setContentType(filePath.getMimeContentType());
```
GCRev, line 328: I'm hoping we don't inadvertently rely on the former unconditional statement somewhere else in the code -- as in the Content-Type header is set even though it shouldn't be, but it's been masked by the former logic.

We'll have to keep our eyes peeled.

MariaSemple (approved):
LGTM!

GCRev (approved):
LGTM. I'm not a C++ expert though, so you'll have to take my word with a grain of salt.

Issue #12643, opened 2023-01-27
Help pane home ugly at narrow widths


gtritchie:
### System details

    RStudio Edition : Desktop or Server
    RStudio Version : Any
    OS Version      : Any
    R Version       : Any

### Steps to reproduce the problem

Display the help pane (Ctrl+3) and resize the column to be narrow-ish.

### Describe the problem in detail

The content does not reflow at narrower widths and becomes increasingly ugly, and eventually starts to require horizontal scrolling. This is a violation of [WCAG 1.4.10 Reflow](https://www.w3.org/TR/WCAG21/#reflow).

![Screenshot 2023-01-27 at 15 50 07](https://user-images.githubusercontent.com/10569626/215221229-498946c9-86be-4535-8156-7df6e03429d5.png)

Caveat: all of the IDE's UI has extensive problems with this WCAG criteria, so fixing the help pane is a drop in the bucket. We will never be able to fully support this due to the nature of our UI. That said, in high-zoom scenarios (limited vision), having this pane continue working would still be a win.

### Describe the behavior you expected

This simple page should be responsive and handle narrower widths more appropriately.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

jonvanausdeln:
Verified with 2023.03.0-daily+280, Windows 11, Ubuntu 22

Helped pane flowed correctly for better visibility. Tested with standard rstudio help, and several different packages.

PR #12644, opened 2023-01-27
improve help pane display at narrow pane widths

gtritchie:
### Intent

Noticed during accessibility auditing that the Help pane's home page uses a multi-column layout even when the help pane is narrow (or you zoom the UI font large enough). In either case, an accessibility issue.

Targeting Cherry Blossom just because this is a simple, safe fix.

#12643

### Approach

When the pane is made narrow, the columns go away.

Did this by ripping out the table-based layout and using a simple responsive CSS layout.

Also ran the page through W3C HTML verification (it had numerous warnings) and now it verifies cleanly.

Wrapped page in a `main` element; another minor accessibility tweak.

Normal view:

help pane with multi-column output

Narrow view (before this fix):

help pane with multi-column output at narrow width showing horizontal clipping

Narrow view (after the fix):

Screenshot 2023-01-27 at 16 12 06

### Automated Tests

None

### QA Notes

This is a static HTML file bundled with the app (both for desktop and server/workbench) so very low risk.

I'm not using any bleeding-edge CSS stuff, this works in any browser we care about (including the QtWebEngine Chromium in RDP Pro).

### Documentation

None

### Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -43,4 +43,4 @@
 - Focus and keyboard-focus styles have been improved on the homepage
 - Keyboard support has been added to the job summary drop-down in session list items on the homepage
 - Improved alt-text and updated link to posit.co on sign-in page logo (rstudio-pro #4096)
-
+- Help pane home page switches to single-column display at narrow widths (#12643)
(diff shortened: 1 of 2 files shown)


timtmok (approved):
Much more accessible without the table! 👍

GCRev:
I slacked on getting a review in. I was thinking that there's probably a way to do this without using  a media query, because grid has [repeat syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/repeat) for `grid-template-columns`. This will also work, though, as an explicit declaration.

Issue #12494, opened 2022-12-22
Previewing rendered html_document causes URLs to open in browser


matt-gunther:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2022.12.0+353 
    OS Version      : MacOS 13.0.1 
    R Version       : 4.2.2

Additional details: 

```
   knitr version: 1.41
   rmarkdown version: 2.19
```

### Steps to reproduce the problem

Knit the following Rmd or Qmd document (sans the `\` escape characters in my R chunk) with **Preview in Viewer Pane** or **Preview in Window**.

````
---
output: html_document
---

```{r}
knitr::include_url("https://www.rstudio.com/")
```

````

### Describe the problem in detail

The referenced URL opens in several (4) tabs in a new browser (Chrome) window. 

*Reversion to RStudio version 2022.07.2 resolves the issue*

*Selecting **(No Preview)** resolves the issue*

### Describe the behavior you expected

Pervious versions of RStudio did not open these links in a new browser window, and I'm not sure why it opens several versions of the same tab.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

**Comment by ronblum** [comment]
@matt-gunther Thank you for raising this!
I can reproduce the problem in
- RStudio Desktop 2022.12.0+353 on MacOS 13.2

It works correctly in 
- RStudio Desktop 2022.07.2+576 on MacOS 13.2
- RStudio Server 2022.12.0+353 on Ubuntu 24.04.5 via Chrome on MacOS 13.2

The preview renders correctly, but it is annoying that the four browser windows also open. We'll review this as we continue development of RStudio.

By the way, here's a trick for including backticks in code: wrap the code with an additional backtick. That is, if you have code with three backticks, wrap it with four backticks. I edited the original post to use this:

`````
````
---
output: html_document
---

```{r}
knitr::include_url("https://www.rstudio.com/")
```

````
`````

To display this example with four backticks, I wrapped it with five backticks.

Triage: Regression from Spotted Wakerobin.

matt-gunther:
[...]

Issue #12413, opened 2022-12-08
Link opens in browser after Render with qml file when hypothes.is comments are turned on in _quarto.yml


petrbouchal:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2023.03.0-daily+3 (2023.03.0-daily+3)
    OS Version      :  MacOS 13.1 Beta (22C5050e)
    R Version       : 4.2.1 (2022-06-23)

### Steps to reproduce the problem

Create a Quarto file - e.g. index.qmd as below

````
---
title: "Test"
format: html
---

````

Create a _quarto.yml file as below:

````
project:
  output-dir: docs
  type: website
website:
  comments:
    hypothesis: true
````

Click Render

### Describe the problem in detail

When the preview opens in the Preview pane, the following link opens (twice) in the default browser:

https://hypothes.is/app.html#config=%7B%22openSidebar%22%3Afalse%2C%22showHighlights%22%3A%22always%22%2C%22origin%22%3A%22https%3A%2F%2Fhypothes.is%22%2C%22version%22%3A%221.1165.0%22%2C%22hostURL%22%3A%22http%3A%2F%2Flocalhost%3A5251%2Findex.html%3Fcapabilities%3D1%26host%3Dhttp%253A%252F%252F127.0.0.1%253A20523%22%7D

### Describe the behavior you expected

I expect the HTML file to be shown in the Preview pane and nothing to open in the browser.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
@petrbouchal Thanks for letting us know! You can probably guess my first question: Did this work in previous versions of RStudio?

Also, thank you for trying out the Cherry Blossom so early! I appreciate your trying out such an early development release.

petrbouchal:
Yes, this issue also manifests on the current stable version, but unfortunately I don't recall when it first appeared.

petrbouchal:
[...]

PR #12660, opened 2023-02-03
remove `openExternal` call from `allowNavigation` check 

sharon-wang:
### Intent

To address #12494 and #12413, this PR removes code that can lead to urls being opened multiple times.

### Approach

`allowNavigation()` checks if navigation should be allowed, but in the final `else` case, it also opens the specified url via `shell.openExternal()`.

This side effect of opening the url should be removed from the function because the function should only serve a single purpose: to check if navigation should be allowed.

Other code which calls `allowNavigation()` uses the boolean result to determine if the url should be opened or not (and opens the url accordingly), which seems to be why multiple opens are getting triggered (`allowNavigation()` may open a url and the calling code may open the url again).

### Automated Tests

N/A

### QA Notes

- Follow steps from issue description #12494 and #12413
- Test on both Chrome and Safari (I personally could not reproduce in Safari and was only getting 1 duplicate tab in Chrome)
- Only 1 tab/window should be opened per url
- Other areas which cause urls to be opened should not be affected

### Documentation

N/A

### Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -33,6 +33,7 @@
 - Removed empty spell check tooltips #11306
 - Fixed package completion tooltips #12147
 - Fixed setting rsession log level using command-line argument or logging.conf #12557
+- Fixed previewing rendered html_document causes URLs to open in browser #12494
 
 ### Accessibility Improvements
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -33,7 +33,6 @@
 - Removed empty spell check tooltips #11306
 - Fixed package completion tooltips #12147
 - Fixed setting rsession log level using command-line argument or logging.conf #12557
-- Fixed previewing rendered html_document causes URLs to open in browser #12494
 
 ### Accessibility Improvements
 
--- a/src/node/desktop/src/main/desktop-browser-window.ts
+++ b/src/node/desktop/src/main/desktop-browser-window.ts
@@ -333,7 +333,7 @@ export class DesktopBrowserWindow extends EventEmitter {
       // https://github.com/rstudio/rstudio/issues/12256
       return true;
     } else {
-      logger().logDebug('allowNavigation: no external navigation in IDE, unsafe host, open in browser');
+      logger().logDebug('allowNavigation: external navigation within IDE is not allowed and URL host is unsafe. URL must be explicitly opened in the browser.');
       return false;
     }
   }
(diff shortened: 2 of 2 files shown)


sharon-wang:
@timtmok Could you please test this on your end (steps in #12494)? When I reproduced the bug on my machine, I was only getting 2 tabs and this PR fixes the issue for me, but I'm not sure if it properly removes the duplicates in cases where 3+ tabs are getting opened.

timtmok (approved):
@sharon-wang It still opens two tabs for me but better than four. I think this is a worthwhile change until it's determined why there are still two tabs opened.

NEWS.md
```diff
@@ -33,6 +33,7 @@
 - Removed empty spell check tooltips #11306
 - Fixed package completion tooltips #12147
 - Fixed setting rsession log level using command-line argument or logging.conf #12557
+- Fixed previewing rendered html_document causes URLs to open in browser #12494
```
sharon-wang, line 36: Removing this line since the PR doesn't completely fix this issue
```suggestion
```

kevinushey:
For what it's worth, this ugly bit of code might be related: https://github.com/rstudio/rstudio/blob/21c0e8c4ff3321a0150d3c01319f28cf9ddab512/src/node/desktop/src/main/main-window.ts#L134-L152

We're waiting on an upstream Electron issue to be merged so we can make use of `will-frame-navigate`, but it's been stalled for months. 😞

sharon-wang:
@kevinushey Thanks for pointing this out! Do you happen to remember why we redirect the iframe back to the source URL?

If I remove `callback({ cancel: false, redirectURL: details.frame?.url });` or change `cancel` to `true`, I no longer hit the duplicate tab situation with iframes. I wonder if doing one of these is a reasonable fix or if we redirect back to the source URL for a particular reason?

kevinushey:
If I recall correctly, this was done as a fallback for attempts to explicitly navigate to external links from the Viewer pane; e.g. if you rendered this R Markdown document:

`````
---
title: "Untitled"
author: "Bug"
date: "`r Sys.Date()`"
output: html_document
---

<https://www.posit.co>
`````

Clicking the link in the rendered document should open the website in an external browser, without causing any navigation in the IDE itself.

Screenshot 2023-02-10 at 2 39 50 PM

Issue #12653, opened 2023-01-31
"Copy To..." deletes all files in Home directory


stibu81:
### System details

    RStudio Edition : Server
    RStudio Version : 2022.12.0 Build 353
    OS Version      : Ubuntu 20.04.5 LTS
    R Version       : 4.2.2

### Steps to reproduce the problem

* Create a new user and login into RStudio
* Create two folders `test` and `test2` in the user's home directory. Create a file in each folder. The files pane now looks like this (the folder `R` was created automatically):
![image](https://user-images.githubusercontent.com/7131000/215708104-a9d8a75f-7b81-4777-b659-603b23984fbd.png)
* Select the folder `test2` and click on "More - Copy To...".
* In the dialogue enter `~` and press return:
![image](https://user-images.githubusercontent.com/7131000/215708352-83b78e84-6b24-488e-b2ad-2fcf617edf08.png)
* RStudio shows an error dialogue "File exists" and all files and folders in the home directory will be deleted.

### Describe the problem in detail

Copying files deletes the home directory without any warning.

This seems only to happen when copying a folder. I have not managed to delete the Home directory by copying a file.

### Describe the behavior you expected

I'm not sure what the correct behaviour would be. Maybe an error message. And certainly, no files should be deleted.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

jonvanausdeln:
I was not able to reproduce on Desktop (windows and Ubuntu22).

I suspect it is because desktop uses the OS native dialogs. Windows doesn't expand `~` to a home directory. Ubuntu just gave a "failed" error message.

melissa-barca:
Thank you for reporting this. I was able to reproduce with open source server 2022.12.0+353 and workbench 2022.12.0+353.pro13.

melissa-barca:
@mikebessuille I've fixed this on `main` and notice you tagged this for backport. I believe the bug exists in all currently supported versions so should it be backported to all of them?

mikebessuille:
[...]

PR #12664, opened 2023-02-03
Update File Selector to navigate to home directory when user enters "~"

melissa-barca:
### Intent

Addresses #12653 

Prevent users from overwriting their home directory by entering a "~" in the file chooser.

### Approach

Updates the File Chooser code so that if a user enters "\~", the file chooser will navigate the user to their home directory rather than interpreting "\~" as a file name the user would like to use. The code already includes logic to navigate to the directory when a user has input a directory into this field:
https://github.com/rstudio/rstudio/blob/7f5869fcb7d263b0284fd09a7ec25afd1e3da00f/src/gwt/src/org/rstudio/core/client/files/filedialog/FileDialog.java#L120-L121
but this code doesn't handle "\~" properly so it doesn't recognize it as a directory. The new logic is added prior to this existing code because we need to determine if the full path the user has entered is "\~", and still allow them to name a directory "\~" within a different directory e.g. "\~\test\\~"

Originally I implemented a fix specific to the "Copy To" behavior that would display an error message and prevent the user from copying to "\~", but this fix will resolve confusion across all file dialogs. 

### Automated Tests

None added as part of this PR. It would be beneficial to add an E2E test for this behavior to run against server. 

### QA Notes

This logic only affects server, it doesn't get executed in Desktop mode. It affects all file selection dialogs. 

### Documentation
bug fix. 

### Checklist


kevinushey (approved):
LGTM!

timtmok (approved):

Issue #12607, opened 2023-01-20
Better error when starting on unsupported macOS


jmcphers:
RStudio currently requires macOS 11. However, when you try to run it on an older macOS, the error that comes back is very difficult to understand, and even sophisticated users don't realize that the problem is an outdated macOS. Here is one such error:

```
dyld: Library not loaded: @executable_path/../Frameworks/libssl.1.1.dylib
  Referenced from: /Applications/RStudio.app/Contents/Resources/app/bin/rsession
  Reason: no suitable image found.  Did find:
	/Applications/RStudio.app/Contents/Resources/app/bin/../Frameworks/libssl.1.1.dylib: cannot load 'libssl.1.1.dylib' (load command 0x80000034 is unknown)
	/Applications/RStudio.app/Contents/Resources/app/bin/../Frameworks/libssl.1.1.dylib: stat() failed with errno=1
```

Old macOS problems were one of the primary causes of "RStudio didn't start" reports on our community page before Electron. The Electron release, however, requires macOS 11, which has greatly exacerbated the problem to the point where it is now the *primary* cause of these reports. 

Here are 5 examples (and there are lots more).

https://community.rstudio.com/t/error-starting-r-the-r-session-failed-to-start/157858/2

https://community.rstudio.com/t/error-al-iniciar-rstudio-error-al-iniciar-r/156958/2

https://community.rstudio.com/t/im-getting-error-when-i-start-rstudio-error-starting-r/155752/2

https://community.rstudio.com/t/rstudio-is-not-working-on-my-macos/156428

https://community.rstudio.com/t/issue-with-installing-r-studio-in-mac/136704/2

We should add a macOS version test at startup. If the version is too old to be supported, we should show a friendly message suggesting that the user update macOS to use RStudio, or use an old version that still works with their operating system.

ronblum:
Triage: Suggesting backlog-1 or even sooner.

jmcphers:
Another report: https://community.rstudio.com/t/failure-to-startup-r-session-need-help/158162

jmcphers:
Here's another one: https://community.rstudio.com/t/first-time-problems/158070

gtritchie:
Another one was received via the accessibility alias from a user on High Sierra (10.13). I responded directly to them and pointed them towards the community for future questions of this sort.

[...]

PR #12670, opened 2023-02-06
Add error if using an unsupported version of macOS

melissa-barca:
### Intent

Addresses #12607 

Alert the user on startup when RStudio failed to launch due to an unsupported version. 

image

### Approach

When we detect a launch failure, check the operating system version. If we're using an unsupported version of macOS, prefix the launch error with: `You are using an unsupported operating system. RStudio requires macOS 11 or higher.` 

Reviewer note: I'm unsure if this is the correct place to add this function. I wanted to include it in the electron code rather than the C++ code since it only applies to desktop and to take advantage of node's `os` module for determining the specific OS release. Please let me know if this should be moved to somewhere in `node/desktop/src/core` instead.

### Automated Tests

No unit tests were added since the functionality depends on the system running the test. 

### QA Notes

This requires testing on a version of macOS prior to 11. I have an alternate laptop I was able to downgrade so I've verified this works as expected. 

### Documentation
The Desktop Pro Admin guide already mentioned this requirement. I've added it to the open-source user guide. 

### Checklist

- [ don't think this is necessary? ] If this PR adds a new feature, or fixes a bug in a previously released version, it includes an entry in `NEWS.md` 

### DIFF
--- a/src/node/desktop/src/main/gwt-callback.ts
+++ b/src/node/desktop/src/main/gwt-callback.ts
@@ -28,6 +28,7 @@ import {
 import { IpcMainEvent, MessageBoxOptions, OpenDialogOptions, SaveDialogOptions } from 'electron/main';
 import EventEmitter from 'events';
 import { existsSync, writeFileSync } from 'fs';
+import { platform, release } from 'os';
 import i18next from 'i18next';
 import { findFontsSync } from 'node-system-fonts';
 import path, { dirname } from 'path';
@@ -49,7 +50,6 @@ import {
   getAppPath, handleLocaleCookies, resolveAliasedPath
 } from './utils';
 import { activateWindow, focusedWebContents } from './window-utils';
-import os, { version } from 'os';
 
 export enum PendingQuit {
   PendingQuitNone,
@@ -911,10 +911,8 @@ export class GwtCallback extends EventEmitter {
   }
 
   addMacOSVersionError(): void {
-    if (os.platform() === 'darwin') {
-      const release = os.release();
-      const release_major = parseInt(release.substring(0, release.indexOf('.')));
-
+    if (platform() === 'darwin') {
+      const release_major = parseInt(release().substring(0, release().indexOf('.')));
       // macOS 11.0 uses darwin 20.0.0
       if (release_major < 20) {
         let versionError = 'You are using an unsupported operating system.' +
--- a/src/node/desktop/src/main/gwt-callback.ts
+++ b/src/node/desktop/src/main/gwt-callback.ts
@@ -917,7 +917,7 @@ export class GwtCallback extends EventEmitter {
 
       // macOS 11.0 uses darwin 20.0.0
       if (release_major < 20) {
-        let versionError = 'You are using an unsupported Operating System.' +
+        let versionError = 'You are using an unsupported operating system.' +
         ' RStudio requires macOS 11 or higher.';
         if (this.errorPageData.get('process_error')) {
           const launch_failed = this.errorPageData.get('process_error');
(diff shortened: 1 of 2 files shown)


src/node/desktop/src/main/gwt-callback.ts
```diff
@@ -902,10 +903,31 @@ export class GwtCallback extends EventEmitter {
     });
 
     ipcMain.handle('desktop_startup_error_info', async (event, varName: string) => {
+      if (varName == 'launch_failed') {
```
gtritchie, line 906: Use `===` in Javascript for comparisons. `==` does type coercion and can give unexpected results.

src/node/desktop/src/main/gwt-callback.ts
```diff
@@ -902,10 +903,31 @@ export class GwtCallback extends EventEmitter {
...
     });
   }
 
+  addMacOSVersionError(): void {
+    if (platform() === 'darwin') {
+      const release_major = parseInt(release().substring(0, release().indexOf('.')));
+      // macOS 11.0 uses darwin 20.0.0
+      if (release_major < 20) {
+        let versionError = 'You are using an unsupported operating system.' +
+        ' RStudio requires macOS 11 or higher.';
```
gtritchie, line 919: I wonder if we should say "macOS 11 (Big Sur)". A lot of Mac users use the names rather than the version numbers, I think, so including both would be helpful.

src/node/desktop/src/main/gwt-callback.ts
```diff
@@ -902,10 +903,31 @@ export class GwtCallback extends EventEmitter {
...
       return resolveTemplateVar(varName, this.errorPageData);
     });
   }
 
+  addMacOSVersionError(): void {
+    if (platform() === 'darwin') {
+      const release_major = parseInt(release().substring(0, release().indexOf('.')));
+      // macOS 11.0 uses darwin 20.0.0
+      if (release_major < 20) {
+        let versionError = 'You are using an unsupported operating system.' +
```
gtritchie, line 918: Ideally we'd make these strings localizable. We use `i18next` package for localizing Electron-specific strings. I don't know, offhand, if these status pages are localized at all, so unless they are I wouldn't worry about it now (but maybe open a tracking issue for future consideration).

gtritchie (changes requested):
I think doing this where you did (the Electron code) is the right place.

kevinushey (commented):
My main concern is that it's going to be easy to forget to update these values on future releases. Is there a way for us to make sure this isn't forgotten? We could either:

1. Add this to the release checklist, or
2. Use some kind of "global" versioning file where we declare the minimum-required versions of various operating systems, and find some way to import / use that here.

timtmok (approved):
I do like Kevin's idea to have a global versioning file to have a check like this for other OS'es.

gtritchie (approved):

src/node/desktop/src/main/gwt-callback.ts
melissa-barca: @gtritchie looking at https://github.com/rstudio/rstudio/blob/main/src/cpp/desktop/html/error.html, we're unfortunately not doing any localization for this page. 

@kevinushey I updated this to include the version (I manipulated my local code to make the error appear for a supported version for this screenshot):
image

melissa-barca:
Thanks @kevinushey, I'll be sure to update the release checklist once this is merged so I can link to the code. One thing to note about this code, which could be argued to be a feature or bug, is that it will only be executed if we get a launch failure. I think this is the right way to go because it continues to treat unsupported operating systems as "use at your own risk", but if they do completely fail to launch, we give the user a clue as to why.

src/node/desktop/src/main/gwt-callback.ts
melissa-barca: Issue for localization: https://github.com/rstudio/rstudio/issues/12672

Issue #12569, opened 2023-01-13
Uncaught Shiny error breaks RStudio's "Run in Window" feature


cpsievert:
Initially reported in https://github.com/rstudio/shiny/issues/3761

### System details

    RStudio Edition :    Desktop
    RStudio Version :   2022.12.0+353
    OS Version      :      Mac OS 13.1
    R Version       :        4.2.2

### Steps to reproduce the problem

Save the following Shiny app to disk. Then click the "Run App" button with "Run in Window" checked. When the app opens, click the "Boom" button. After the app crashes, click "Run App" again (to run the app again RStudio's Window browser). This time, the app's UI doesn't appear (and the JS console gives an error of `desktop_allow_navigation: Error invoking remote method 'desktop_allow_navigation': TypeError: Object has been destroyed`).

```r
library(shiny)

ui <- fluidPage(
  actionButton("boom", "Boom")
)

server <- function(input, output) {
  observeEvent(input$boom, {
    stop("boom")
  })
}

shinyApp(ui, server)

```  

### Describe the behavior you expected

I'd expect to be able to repeatedly run the Shiny app in RStudio's window (and click the "Boom" button to generate an uncaught error)

sharon-wang:
This hasn't been triaged, but seems to have the same root issue as #12468. PR https://github.com/rstudio/rstudio/pull/12686 should resolve this bug.

sharon-wang:
Here's a demo of the error reproduction and fix from #12686. `control+C` is used to exit the R session before re-running the app.

### Reproduction (https://dailies.rstudio.com/version/2023.03.0-daily+293/)
https://user-images.githubusercontent.com/25834218/218549918-578cb714-96b2-4b8c-a2e6-92b62d14fc4a.mov

### Behaviour after Fix from #12686
https://user-images.githubusercontent.com/25834218/218550175-94bc8f63-1f2b-4566-b8cc-16f05cf58433.mov

jonvanausdeln:
Tested on 2023.03.0-daily+316  Desktop
Windows11 and MacOS 12.4

As described, I'm able to run the app after initial run, but only after interrupting the R session (via ctrl-c or stop button). Not sure of the uncaught error should interrupt the R session by itself.

jonvanausdeln:
After some thought and discussion, I'm closing this as the initial repro is fixed.  If the viewer closing after the error is a problem, that can be opened in another issue.

Issue #12468, opened 2022-12-20
Data viewer's "Show in new window" results in empty window


uhkeller:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2022.12.0+353 and  2023.03.0-daily+31
    OS Version      : macOS 12.6.2 (arm64)
    R Version       : 4.2.2 (arm64)

### Steps to reproduce the problem

1. Open RStudio.
2. Type `View(iris)`.
3. In the data viewer pane, click the "Show in new window" button.
4. In the pop-up window showing the data, click the "Return to main window" button.
5. In the data viewer pane, click the "Show in new window" button again. The window comes up empty.

https://user-images.githubusercontent.com/2323704/208664505-fce4aade-351a-416e-aa38-d9cfd24468b2.mov

### Describe the problem in detail

All but the first clicks on "Show in new window" in the data viewer pane result in an empty window.

### Describe the behavior you expected

I expected the window to no be empty, but to show the data (like it does the first time the button is clicked).

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
@uhkeller Thank you for raising this! I can reproduce this in RStudio Desktop 2022.12.0+353 on MacOS 13.2 and Ubuntu 22.04. It works in version 2022.07.2+576

Triage: Marking for Cherry Blossom as a regression.

johninpdx:
Also occurs under Windows 10. Same steps as per @uhkeller to reproduce.

kalcan83:
I'm having the same issue on RStudio 2022.12.0+353 "Elsbeth Geranium" Release (7d165dcfc1b6d300eb247738db2c7076234f6ef0, 2022-12-03) for Windows
Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) RStudio/2022.12.0+353 Chrome/102.0.5005.167 Electron/19.1.3 Safari/537.36

nicholascarey:
[...]

PR #12686, opened 2023-02-09
Fix blank satellite window for data viewer and shiny apps

sharon-wang:
### Intent

Addresses #12468 and #12569.

Catches the error causing the gwt callback to fail partway through opening up data viewer files and shiny apps in a satellite window.

### Approach

The underlying issue appears to be that some satellite windows are not unregistered once they are returned to the main window. As a result, when we iterate through the list of owners to find the `GwtWindow` that sent the "Open in new window" event, we hit some objects that have been destroyed (`TypeError: Object has been destroyed`), but not removed from the list of owners. Since the gwt callback errors out, we don't end up populating the window with the data being popped out from the main window.

This PR is stopgap fix to catch the error and continue iterating through the loop to find the correct owner. The data is then displayed successfully. In the future, we should ensure that satellite windows returned to the main window or satellite windows otherwise dismissed (not via the "Close" X button) are unregistered accordingly. Maybe there's a particular reason why those windows are not unregistered?

<details>
    <summary>Some relevant code areas:</summary>

- anything involving `desktop_allow_navigation`
- failure occurs during:
https://github.com/rstudio/rstudio/blob/497e47f994af46050b825d9351dec1b956e9fd6b/src/node/desktop/src/renderer/desktop-bridge.ts#L557
- error `TypeError: Object has been destroyed` is thrown here
https://github.com/rstudio/rstudio/blob/497e47f994af46050b825d9351dec1b956e9fd6b/src/node/desktop/src/main/gwt-callback.ts#L987
- https://github.com/rstudio/rstudio/blob/main/src/node/desktop/src/main/satellite-window.ts
</details>

### Automated Tests

Once we determine when owners should be unregistered, we can write some tests to check the different cases where owners are unregistered, such as when a satellite window is returned to the main window.

### QA Notes

- Please test with repro steps in #12468 and #12569
- Text files such as `.qmd`, `.js`, etc should be unaffected by this change
- For shiny apps, ensure "Run in Window" is selected before clicking "Run App"

### Documentation
N/A

### Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -37,6 +37,8 @@
 - Fixed setting rsession log level using command-line argument or logging.conf #12557
 - Fixed issue that allowed users to overwrite their home directory in server mode #12653
 - Fixed "Check for updates" incorrectly reports that there are no updates (rstudio-pro #3388)
+- Fixed data viewer's "Show in new window" results in empty window #12468
+- Fixed uncaught Shiny error breaks RStudio's "Run in Window" feature #12569
 
 ### Accessibility Improvements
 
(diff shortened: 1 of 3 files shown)


melissa-barca (approved):
LGTM!

PR #12689, opened 2023-02-10
support for visual editor reference links option

jjallaire:
@timtmok This is a change intended to support the visual editor's new `editor: markdown: references: links` option (as described here: https://quarto.org/docs/visual-editor/vscode/#links)

My Homebrew installation on my dev Mac has become corrupted so I'm not actually able to compile/test this! In any case we should pair program on landing the change so you know what's going on. Message me when a good time to do this next week might be (I'll mark this PR as a draft in the meantime).

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/visualmode/VisualModeMarkdownWriter.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/visualmode/VisualModeMarkdownWriter.java
@@ -96,6 +96,7 @@ public class VisualModeMarkdownWriter
       Integer wrapAtColumnPref = null;
       String referencesLocationPref= null;
       String referencesPrefixPref = null;
+      Boolean referencesLinksPref = null;
       if (VisualModeUtil.isDocInProject(workbenchContext_, docUpdateSentinel_))
       {
          // allow any prefs defined in quarto yaml to take precedence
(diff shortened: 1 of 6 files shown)


src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/visualmode/VisualModeMarkdownWriter.java
```diff
@@ -121,6 +122,7 @@ public Options optionsFromConfig(PanmirrorPandocFormatConfig formatConfig)
                {
                   referencesLocationPref = quarto.project_editor.markdown.references.location;
                   referencesPrefixPref = quarto.project_editor.markdown.references.prefix;
+                  referencesLinksPref = quarto.project_editor.markdown.references.links;
```
timtmok, line 125: @jjallaire Everything looked good except this one part. `quarto.project_editor.markdown.references.links` is an array of length 1. If I set the project level value to `false`, the reference links are still created.

I'm not sure if this was fixed in a later Quarto release if it's parsing the yaml there. I'm using 1.2.335.

timtmok (commented):
The document level option works. It was only the project level that didn't work when setting the option to `false`.

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/visualmode/VisualModeMarkdownWriter.java
jjallaire: Okay, I'll take a closer look at this (I now have a local RStudio dev config so can actually run!)

jjallaire: I just needed to add this bit to make sure `links` was read as a scalar: https://github.com/rstudio/rstudio/pull/12689/commits/d8a5aca44251cf2eb40893603b9b3d0f1fe282fe

ronblum:
Note: I created issue #12478 to track this.

PR #12693, opened 2023-02-10
Fix initial Choose R dialog state

timtmok:
### Intent
Address some issues in #12452

### Approach
The initial dialog state could have an invalid selection such as using the default 64-bit R when none was found in the registry or the user selecting a specific version of R option but not selecting a version from the list. This results in an error that may not be obvious to the user why R could not be found.

Fixes:
* Cancelling the dialog doesn't try to show the download R prompt since the application closes before it can be read (Qt would exit without prompting the user to download R)
* Added logging after querying the registry if an R installation was found to lessen the impact of error log messages registry queries not finding R
* Add `HKEY_CURRENT_USER` hive query to the default 32/64 bit search to address users that do not have admin privileges (R would be installed to `%LOCALAPPDATA%\Programs` and registry keys written to `HKCU` hive)
* Added logging if validating selected R installation fails so it records which R was selected
* Changed initial selection when Choose R dialog is shown
   * Choose a specific version is selected if default R installations are not found so that a disabled option is not initially selected
   * OK button is disabled (as Qt behaved) if selection is invalid (i.e. specific version is selected but no version is selected in list)

### Automated Tests
None

### QA Notes
The fix list above describes specific scenarios that were addressed. Mainly, it should not be allowed to click OK if there is no R version selected.

### Documentation
None

### Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -37,6 +37,7 @@
 - Fixed setting rsession log level using command-line argument or logging.conf #12557
 - Fixed issue that allowed users to overwrite their home directory in server mode #12653
 - Fixed "Check for updates" incorrectly reports that there are no updates (rstudio-pro #3388)
+- Fixed default options for Choose R dialog on Windows (#12452)
 
 ### Accessibility Improvements
 
--- a/src/node/desktop/src/main/detect-r.ts
+++ b/src/node/desktop/src/main/detect-r.ts
@@ -105,9 +105,8 @@ export async function promptUserForR(platform = process.platform): Promise<Expec
     }
 
     // if path is null, the operation was cancelled by the user
-    // we should probably consider this to be an error
     if (data == null) {
-      return err(new Error('User must select what version of R they would like to use.'));
+      return ok(null);
     }
 
     // save the stored version of R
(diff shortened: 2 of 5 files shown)


timtmok:
@gtritchie I'm not sure if there's an accessibility issue here with disabling the OK button. There wouldn't be any feedback for why it is disabled. Perhaps it makes sense to have it always be enabled and display a validation error message if the user hasn't selected an R installation?

gtritchie:
> @gtritchie I'm not sure if there's an accessibility issue here with disabling the OK button.

@timtmok short answer: it's not a WCAG violation to disable buttons. In this situation, I think it's a reasonable approach.

That said, it's not considered an ideal approach (for accessibility) since as you point out it doesn't provide any feedback on why it's disabled, and disabled controls generally have very low contrast (but technically disabled controls are given an exception to contrast requirements).

If you are bored and want more reading with ideas on fancier patterns: https://css-tricks.com/making-disabled-buttons-more-inclusive/

melissa-barca (approved):
Overall LGTM, but this dialog seems like something we might want to document in the user guide.

Issue #12528, opened 2023-01-05
Prepare for Rtools43 release


kevinushey:
An initial revision of the Rtools43 toolchain (to be used with the upcoming R 4.3.0 release; which is typically end of April / early May) is available. https://cran.r-project.org/bin/windows/Rtools/rtools43/rtools.html

Most of the updates should happen around here:

https://github.com/rstudio/rstudio/blob/aefa76732fd47f130cd85bda44364a020923b9d3/src/cpp/core/r_util/RToolsInfo.cpp#L260-L311

In particular, now that the toolchain can evolve throughout minor releases of R, we should stop hard-coding these compiler include paths.

Other places potentially needing updates (may not be exhaustive):

https://github.com/rstudio/rstudio/blob/aefa76732fd47f130cd85bda44364a020923b9d3/src/cpp/core/r_util/RToolsInfo.cpp#L342-L346

https://github.com/rstudio/rstudio/blob/aefa76732fd47f130cd85bda44364a020923b9d3/src/cpp/session/modules/build/SessionInstallRtools.cpp#L42

https://github.com/rstudio/rstudio/blob/aefa76732fd47f130cd85bda44364a020923b9d3/src/cpp/session/modules/build/SessionInstallRtools.cpp#L50

https://github.com/rstudio/rstudio/blob/aefa76732fd47f130cd85bda44364a020923b9d3/src/cpp/session/modules/build/SessionInstallRtools.cpp#L99-L110

https://github.com/rstudio/rstudio/blob/aefa76732fd47f130cd85bda44364a020923b9d3/src/cpp/core/r_util/RToolsInfo.cpp#L430-L434

### Package Installation

Verify that users can build packages containing compiled code from sources. A stress test might be something like:

```
options(pkgType = "source")
install.packages("tidyverse")
```

### Diagnostics / Autocompletion

C++ and (hopefully) C diagnostics work within an R package containing C / C++ code. The [tidyr](https://github.com/tidyverse/tidyr) package would be a good candidate for testing; you can try opening the project in RStudio, and checking that you get valid diagnostics + autocompletion results in a file (e.g. https://github.com/tidyverse/tidyr/blob/main/src/melt.cpp).

jgutman:
> In particular, now that the toolchain can evolve throughout minor releases of R, we should stop hard-coding these compiler include paths.

This part has already been addressed

timtmok:
[...]

PR #12722, opened 2023-02-17
Support rtools 4.3

timtmok:
### Intent
Address #12528 

### Approach
Update the logic to also check for RTools 4.3 if using R 4.3.x. Renamed `findRtools42Installer` since it can be used for 4.3 (and maybe >4.3).

A companion change for the fallback link redirect: https://github.com/rstudio/rstudio-web/pull/41

I think the code could also be refactored so there's less copy/paste if we're confident that the RTools release will be hosted the same way for versions >4.3.

### Automated Tests
None

### QA Notes
This required R 4.3, which is currently available in dev builds. https://cran.r-project.org/bin/windows/base/rdevel.html

`.rs.findRtoolsInstaller("https://cran.rstudio.com/bin/windows/Rtools/rtools43/rtools.html", "some fallback link to rtools")` could be called to ensure it finds the link to the RTools installer. I don't think this is used anywhere else but the function used to be called `findRtools42Installer`. The function will return something like this: `https://cran.rstudio.com/bin/windows/Rtools/rtools43/files/rtools43-5493-5475.exe` or `some fallback link to rtools` if the page link couldn't find the RTools download link. In the IDE, the fallback link is actually to the S3 bucket where the installer is mirrored.

### Documentation
updated NEWS.md

### Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,6 +17,7 @@
 - Disable argument tooltips in script editor for unknown functions #12160
 - Sessions now have lower CPU priority during suspension on macOS and Linux #12623
 - Sessions run on Kubernetes or Slurm will no longer exit with nonzero codes under normal circumstances (rstudio/rstudio-pro#3375)
+- Add RTools 4.3 support
 
 ### Fixed
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,7 +17,7 @@
 - Disable argument tooltips in script editor for unknown functions #12160
 - Sessions now have lower CPU priority during suspension on macOS and Linux #12623
 - Sessions run on Kubernetes or Slurm will no longer exit with nonzero codes under normal circumstances (rstudio/rstudio-pro#3375)
-- Add RTools 4.3 support
+- Add RTools 4.3 support #12528 
 
 ### Fixed
 
--- a/src/cpp/core/r_util/RToolsInfo.cpp
+++ b/src/cpp/core/r_util/RToolsInfo.cpp
@@ -260,7 +260,7 @@ RToolsInfo::RToolsInfo(const std::string& name,
    else if (name == "4.2")
    {
       versionMin = "4.2.0";
-      versionMax = "5.0.0";
+      versionMax = "4.2.99";
 
       // PATH for utilities
       relativePathEntries.push_back("usr/bin");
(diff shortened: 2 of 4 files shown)


NEWS.md
```diff
@@ -17,6 +17,7 @@
 - Disable argument tooltips in script editor for unknown functions #12160
 - Sessions now have lower CPU priority during suspension on macOS and Linux #12623
 - Sessions run on Kubernetes or Slurm will no longer exit with nonzero codes under normal circumstances (rstudio/rstudio-pro#3375)
+- Add RTools 4.3 support
```
melissa-barca, line 20:
```suggestion
- Add RTools 4.3 support #12528 
```
nit: missing issue number

melissa-barca (approved):
LGTM

kevinushey:
We discussed with Jackie having this code automatically infer the compiler include paths, so that it would no longer be necessary to hard-code the compiler paths (especially since it now appears the compiler path can change over time as new releases of Rtools are published).

Are we still intending to do this?

timtmok:
@kevinushey I think I missed that bit. I suppose it should do something similar to here: 

https://github.com/rstudio/rstudio/blob/5062c370f042a49228176d741af70e4d26b571e9/src/cpp/session/modules/clang/RCompilationDatabase.cpp#L1166-L1210

This can discover the include paths and include that with the RTools info. Anything with RTools 4.2+ could call `discoverSystemIncludePaths`.

kevinushey:
That's right. If I understand correctly, the existence of https://github.com/rstudio/rstudio/blob/5062c370f042a49228176d741af70e4d26b571e9/src/cpp/session/modules/clang/RCompilationDatabase.cpp#L1191-L1197 basically means we don't need to pre-populate system include paths in this part of the PR? But we might still need to keep things like the `RTOOLS43_HOME` environment variable, and other compiler defines (+ the other things you implemented)

kevinushey:
In other words, if we're lucky, you can take this part of the PR out:

https://github.com/rstudio/rstudio/pull/12722/files?w=1#diff-8cf23987d622cf62605d86220da51d48e88ecb7d891783fe792e079a46c87f4aR333-R354

and everything should still "just work".

timtmok:
It took me some time to fix my R install but removing the include paths works. It's all good for 4.2 and 4.3.

kevinushey:
Great! Thanks for taking the time to check and verify.

kevinushey (approved):

PR #12729, opened 2023-02-21
Bugfix/hourly upload

cm421:
### Intent

Fixes the upload logic in the builds to send hourly builds to the correct channel on the Dailies Page

### Approach

Hourly builds are denoted by the `channel` metadata value for a build. Daily, Preview, and Release channels are all determined automatically by the value in `version/BUILDTYPE`. The Hourly channel needs to be provided manually to override looking at that file.

Hourly build files are also uploaded to a different location in `latest-builds` under `content/rstudio-hourly/...` instead of the regular `content/rstudio/...` location for daily builds. When the `channel` is set to `Hourly` the script will automatically upload the file to the correct location

### Automated Tests

N/A

### QA Notes

N/A

### Documentation
N/A

### Checklist

### DIFF
--- a/docker/jenkins/publish-build.sh
+++ b/docker/jenkins/publish-build.sh
@@ -23,7 +23,9 @@ if [ $# -eq 0 ]; then
     echo "--pat     The Github Personal Access Token (PAT) to be used to authorize the commit."
     echo "          May be specified in the environment variable GITHUB_PAT instead."
     echo ""
-    echo "--channel (optional) The Channel type for the build. One of Hourly, Daily, Preview, Release"
+    echo "--channel (optional) The Channel type for the build. One of Hourly, Daily, Preview, Release."
+    echo "          Required for setting to Hourly. If not set, the channel will be determined from"
+    echo "          the value in the file: version/BUILDTYPE. Channel names are case sensitive."
     exit 0
 fi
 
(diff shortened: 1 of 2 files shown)


docker/jenkins/publish-build.sh
```diff
@@ -136,7 +144,9 @@ timestamp=$(date +"%Y-%m-%dT%H:%M:%S%z")
 
 # Determine release channel (build type) and flower
 RSTUDIO_ROOT_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && cd ../.. && pwd )"
-channel="$(cat "$RSTUDIO_ROOT_DIR/version/BUILDTYPE" | tr '[ ]' '-' | tr -d '[:space:]')"
+if [ -z "$channel" ]; then
+    channel="$(cat "$RSTUDIO_ROOT_DIR/version/BUILDTYPE" | tr '[ ]' '-' | tr -d '[:space:]')"
```
zachhannum, line 148: The `hourly` build doesn't actually update the `BUILDTYPE` file, so assuming this might be incorrect. Technically it would work with the logic we have here, but it might be better to just make `channel` a required parameter? See [this pr](https://github.com/rstudio/rstudio/pull/12615) for those changes.

cm421: The intent is to continue using the `BUILDTYPE` for Daily, Preview, and Release channels as we have been. But to allow us to override it with an Hourly channel when we need to

zachhannum: With a clear help message that says we will use the `BUILDTYPE` file if `channel` is not set, I think this is fine

zachhannum (approved):
LGTM!

MariaSemple (approved):
LGTM!

zachhannum (approved):
LGTM!

timtmok (approved):

PR #12782, opened 2023-03-02
Updates to Linux Pipeline

zachhannum:
This change contains a variety of updates for the Linux pipeline. 

I've added `deleteDir` as a post step to the agents that actually do the building. I think the last post step with `deleteDir` will only clean up the workspace on the top-level agent. 

I removed `reuseNode true` from the docker block since we are actually grabbing the linux agent for the matrix build and running within docker from the same step. 

I've also added `bionic` and `rhel8` to the `arm64` exclude list, which should fix the license-manager activation linker errors on the pro pipeline. The old pipeline didn't produce `bionic` and `rhel8` `arm64` builds.

### DIFF
--- a/jenkins/Jenkinsfile.linux
+++ b/jenkins/Jenkinsfile.linux
@@ -172,7 +172,7 @@ pipeline {
           exclude {
             axis {
               name 'OS'
-              values 'centos7', 'opensuse15'
+              values 'bionic', 'centos7', 'rhel8', 'opensuse15'
             }
             axis {
               name 'ARCH'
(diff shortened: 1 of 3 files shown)


jenkins/Jenkinsfile.linux
```diff
@@ -63,7 +63,7 @@ def getAgentLabel(String arch) {
     return 'linux && arm64 && 4x'
```
MariaSemple, line 63: Do you want to remove 4x here as well?

zachhannum: Since Leticia just updated our x86 agents with more space, I've reverted this back to using the 4x agents instead!

AndrewMcClain (approved):
LGTM

MariaSemple (approved):
LGTM!

PR #12784, opened 2023-03-03
Update RELEASE Name for dev

zachhannum:
This commit updates the release name to "Cherry Blossom Dev" so that published builds will go to `cherry-blossom-dev/` on the latest builds site. This serves as a companion commit to rstudio/latest-builds#88.

## Intent

The publish builds scripts read the `RELEASE` file in order to construct the `flower-name` portion of the URL to upload dailes metadata to the `latest-builds` repo.

We'll want to revert these changes before merging back into `main`.


cm421 (approved):
I like this. 

1. It will upload to a different S3 bucket which will make clearing them out in the future easier
2. It gets put in the right place on the dailies site
3. It makes it very clear that this is different from the normal Cherry Blossom
4. Very easy to undo when we merge into `main`

PR #12786, opened 2023-03-03
Update to Electron 23

timtmok:
### Intent
Update Electron to 23.1.2. Electron 24 is in April (https://www.electronjs.org/docs/latest/tutorial/electron-timelines). There may also be patches to 23 so this may not be the last Electron update.

### Approach
Update dependencies, run `npm test`, and launched RStudio. 

### Automated Tests
None

### QA Notes
@jonvanausdeln This updates Chromium to `110.0.5481.177`. 

Electron 23 uses Node 18. There's nothing in the Electron docs about dropping support for the platforms we care about but it should be noted in case there are issues (ie. Centos 7, SLES 15.4)

### Documentation
Updated NEWS

### Checklist


sharon-wang (commented):
Do we need to update node to 18 in these places as part of this change?

https://github.com/rstudio/rstudio/blob/df42e153889de55c6025ed832105b3dd9bd5a2e6/dependencies/tools/rstudio-tools.sh#L470-L471

https://github.com/rstudio/rstudio/blob/df42e153889de55c6025ed832105b3dd9bd5a2e6/dependencies/tools/rstudio-tools.cmd#L17

https://github.com/rstudio/rstudio/blob/df42e153889de55c6025ed832105b3dd9bd5a2e6/src/gwt/build.xml#L87-L88

timtmok:
> Do we need to update node to 18 in these places as part of this change?
> 

@sharon-wang It's not necessary to upgrade the Node that's used for the builds. We can move that upgrade to later when we figure out building with Node 18 on platforms that don't support it.

sharon-wang:
> It's not necessary to upgrade the Node that's used for the builds. We can move that upgrade to later when we figure out building with Node 18 on platforms that don't support it.

Ah I see, the version of node needed to build electron is actually the version of node that is needed for electron-forge. So we don't need to match the electron-bundled version of node. But electron 23 apps can't run on platforms that don't support Node 18, so that's where the node issue comes in.

sharon-wang (approved):
Built + opened RStudio successfully for me 👍

melissa-barca (approved):
LGTM! Thanks for getting this in early in the release!

Issue #11822, opened 2022-08-23
Find in Files totally stopped working


ben1787:
### System details

    RStudio Edition : 
    RStudio Version : 
    OS Version      : 
    R Version       : 

RStudio 2022.07.1+554 "Spotted Wakerobin" Release (7872775ebddc40635780ca1ed238934c3345c5de, 2022-07-22) for Windows
Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) QtWebEngine/5.12.8 Chrome/69.0.3497.128 Safari/537.36

### Steps to reproduce the problem
Search for something using ctrl+shift+f

### Describe the problem in detail
It takes no time at all and immediately says no results. My path is correct. And I have made no other selections.

### Describe the behavior you expected
It should find the word I am searching for. It finds nothing for any search term.

- [ x] I have read the guide for [submiting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x ] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [ x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
Can you please provide some more details?

- Are you working with an RStudio project, or just in the home directory?
- Does that project directory have any non-ASCII characters, or spaces, or similar?
- Is the project managed by version control (`git`)? If so, what version of Git for Windows do you have installed?
- What filters, or options, do you have set on the Find in Files dialog? Can you share a screenshot?

You can also try setting the following environment variable:

```
Sys.setenv(RSTUDIO_GREP_DEBUG = 1)
```

and then execute a Find in Files request, and send us the resulting output printed in the console.

ben1787:
I am looking ~/GitHub/R.

Project managed by git but not through RStudio. To RStudio it is just a folder.

No filters in dialog.

![image](https://user-images.githubusercontent.com/4843501/186254166-7d899a80-4be3-464d-b5ad-bd78999c4fa6.png)

![image](https://user-images.githubusercontent.com/4843501/186254033-13b2541d-f3d9-48ec-ae06-92bcd006a7fb.png)

kevinushey:
[...]

PR #12800, opened 2023-03-06
Fix find in files with older versions of git

melissa-barca:
### Intent

Addresses #11822 

Fix find in files when using older versions of git. 

### Approach

Remove passing the `-C` argument to the `grep` command since we're already running the command from the working directory. See https://github.com/rstudio/rstudio/issues/11822#issuecomment-1452650385

### Automated Tests

I ran the existing unit tests and they continue to pass. 

### QA Notes

This was broken on CentOS7 which has an older system version of git. It only affects git projects that have the "Exclude files matched by .gitignore" selection checked in the Find in Files pane:
image
Functionality should continue to work as expected with Find In Files and newer versions of git.

### Documentation
N/A - bugfix

### Checklist


timtmok (approved):

kevinushey:
Belated LGTM; thanks!

Issue #12846, opened 2023-03-15
Low contrast focus rectangle on Sign In button (server login page) and other misc. A11y


gtritchie:
### System details

    RStudio Edition : Server
    RStudio Version : Cherry Blossom (also older versions)
    OS Version      : any
    R Version       : any

### Steps to reproduce the problem

Primary issue: Tab to focus the "Sign In" button on the login page (RStudio Server or Workbench with name/password auth).

### Describe the problem in detail

The focus rectangle has (almost) identical color to the button itself, and is drawn without any gap between it and the button, so is essentially invisible. For example, the button is focused in this screenshot:

<img width="407" alt="Screenshot of rstudio login page with focus on the Sign In button" src="https://user-images.githubusercontent.com/10569626/225426265-446384d1-8fb7-4145-8718-6c0da303909a.png">

Other minor a11y issues on this page:

* Username and password are required fields, but aren't annotated as required (so it isn't announced by a screen-reader)
* Page has a pointless skip-nav link; this was one of the first things I added when I started working on accessibility; however, it isn't necessary on this page because (1) we already have both a valid `H1` and a `main` landmark for screen-reader users; (2) For a keyboard-only user, the skip-nav link doesn't save any keystrokes:

(current): 3 tabstops to get to the username field (the skip-nav link, then the banner image, then username); or a tabstop (skip-nav) followed by "enter" key (2 keystrokes)
(without skip-nav): 2 tabstops to the username field (banner image, then username)

Here's a screenshot of the skip-nav link:

<img width="619" alt="Screenshot of skip-nav link on login page" src="https://user-images.githubusercontent.com/10569626/225428404-35b458ce-b4bd-4197-b218-eeab46078cce.png">

### Describe the behavior you expected

* visible focus rectangle for Sign In button
* remove skip-nav link
* annotate inputs with `aria-required="true"`

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
[...]

PR #12847, opened 2023-03-15
Accessibility tweaks to sign-in page

gtritchie:
### Intent

Addresses [Low contrast focus rectangle on Sign In button (server login page) and other misc. A11y #12846](https://github.com/rstudio/rstudio/issues/12846)

### Approach

- used same focus-visible styling as the workbench home page; add a tiny extra top margin to the input fields so the thicker focus rectangle doesn't bump up against the label
- removed use of the `focus-visible` polyfill from sign-in page, and just use native `focus-visible` now that it is supported by all browsers we support (the IDE itself is still using this polyfill so can't remove it altogether yet)
- removed skip-nav link from the sign-in page (and the associated styles; this was the only place using this); see the associated issue for why I'm removing it
- added `aria-required="true"` to the input fields

The button focused appearance:

Screenshot of sign-in page with improved focus rectangle on sign-in button

In dark mode:

Screenshot of rstudio server sign-in page in dark mode

Here's an input field with focus:

Screenshot of rstudio server sign-on page with username field focused

### Automated Tests

None

### QA Notes

- Verify improved visibility of focus rectangles (e.g. by tabbing through the page).
- Turn on screen reader and tab to the username and password fields; should now announce "required" as part of their descriptions
- Tab from the start of the page; the skip-nav link should no longer appear

### Documentation

None

### Checklist

### DIFF
--- a/src/gwt/www/signin.css
+++ b/src/gwt/www/signin.css
@@ -195,12 +195,12 @@ div.signinhidden {
 }
 
 :focus {
-  outline: none !important;
-  box-shadow: none !important;
+  outline: none;
+  box-shadow: none;
 }
 
 :focus-visible {
   outline-offset: 0;
   outline: 2px solid var(--primary) !important;
-  box-shadow: inset 0 0 0 2px var(--primary-fg) !important;
+  box-shadow: inset 0 0 0 2px var(--primary-fg);
 }
--- a/src/gwt/www/signin.css
+++ b/src/gwt/www/signin.css
@@ -39,6 +39,7 @@ input[type=text], input[type=password] {
   border: 1px solid #cdcecf;
   font-size: 14px;
   padding: 8px;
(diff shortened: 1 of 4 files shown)


src/gwt/www/signin.css
```diff
@@ -192,3 +193,14 @@ div.signinhidden {
 .signin-hidden {
   display: none;
 }
+
+:focus {
+  outline: none !important;
+  box-shadow: none !important;
```
GCRev, line 199: `!important` :(

 But sometimes you have to do what you must.

gtritchie: I copy and pasted this directly from your focus-visible work on the home page. :-)

GCRev (approved):
LGTM pending your findings on the `!important`s. If they're necessary then we can keep them.

PR #12864, opened 2023-03-17
Send `SIGKILL` to child processes during forced exit to avoid zombies

atheriel:
### Intent

The Package Manager team[ initially noticed this](https://positpbc.slack.com/archives/C7312LBPY/p1663804027257349) when the Kubernetes Launcher plugin would hang indefinitely but never exit. @jeffvroom has [also found the symptoms of this](https://positpbc.slack.com/archives/C7312LBPY/p1677121768420139?thread_ts=1677023476.856099&cid=C7312LBPY) in some customer logs.

### Approach

We already "force" an exit event after 30 seconds, this commit just piggybacks on that timer to send it `SIGKILL` at that point as well.

### Automated Tests

I don't think there are tests for this.

### QA Notes

Please advise.

### Documentation

N/A

### Checklist

### DIFF
--- a/src/cpp/core/system/PosixChildProcess.cpp
+++ b/src/cpp/core/system/PosixChildProcess.cpp
@@ -1547,7 +1547,10 @@ struct AsioAsyncChildProcess::Impl : public boost::enable_shared_from_this<AsioA
 
             cleanup();
 
-            exitCode_ = -1;
+            // send the process SIGKILL in an effort to avoid it hanging around
+            // as a zombie or in a broken state forever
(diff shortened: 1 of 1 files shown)


MariaSemple (approved):
The change LGTM. For the QA notes section, any details you can provide about how to reproduce the behaviour so the fix can be tested would be great.

atheriel:
My current plan is to try and test this patch against Launcher with a misbehaving plugin (which I might have to build).

cm421 (approved):
LGTM, though I wish we had a single header that contained all the exit codes we use

PR #12885, opened 2023-03-22
fix keyboard shortcut page banner layout and add tabstop

gtritchie:
### Intent

This fixes a small regression introduced by https://github.com/rstudio/rstudio/pull/12884: the keyboard shortcut help static page (only seen when screen reader support is enabled) had its logo right-justified.

Also, [Axe identified that the page might be difficult to scroll via keyboard ](https://dequeuniversity.com/rules/axe/4.6/scrollable-region-focusable?application=AxeChrome)because there are no tabstops, so put a tabstop around the entire `main` region.

Not bothering to open an issue on this due to its extremely simple nature. Here's how it looked before this fix:

Screenshot of keyboard shortcuts page with right-justified logo

After this fix:

Screenshot of fixed keyboard shortcuts page with left-justified logo

### DIFF
--- a/src/gwt/www/docs/keyboard.htm
+++ b/src/gwt/www/docs/keyboard.htm
@@ -35,8 +35,8 @@ References on default Ace shortcuts:
   https://github.com/ajaxorg/ace/blob/master/lib/ace/commands/default_commands.js
 -->
 <body>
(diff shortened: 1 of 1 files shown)


GCRev (approved):
LGTM

PR #12890, opened 2023-03-23
Bump opensuse docker version to 15.4 and update packages

cm421:
### Intent

We've been building opensuse images on 15.1 for the last 4 years. This bumps the docker image to use the 15.4 base image.

Changes/removes packages to match 15.4 packages

Fixed `OPERATING_SYSTEM` env variable typo

Once approved I'll merge `rel-cherry-blossom` forward into main

### Approach

Version bump

### Automated Tests

N/A Build change

### QA Notes

N/A Build change

### Documentation

N/A Build change

### Checklist

- [ ] If this PR adds a new feature, or fixes a bug in a previously released version, it includes an entry in `NEWS.md` 
- [ ] If this PR adds or changes UI, the updated UI meets [accessibility standards](https://github.com/rstudio/rstudio/wiki/Accessibility)
- [x] A reviewer is assigned to this PR (if unsure who to assign, check Area Owners list)
- [ ] This PR passes all local unit tests

PR #12905, opened 2023-03-24
Build on OpenSuse15.3 to use libuser package

melissa-barca:
### Intent

15.4 doesn't have `libuser` and there are currently no community packages available. Switch to using 15.3 for our builds. We just switched to 15.4 from 15.1 here: https://github.com/rstudio/rstudio/pull/12890


cm421 (approved):
LGTM

cm421:
To fix this and allow us to move to 15.4, we'll need to implement: https://github.com/rstudio/rstudio-pro/issues/4198

Issue #12923, opened 2023-03-28
Massive slowdown with Notebooks using nested tibbles


jonasfoe:
### System details

    RStudio Edition : 
    RStudio Version : 2023.03.0 Build 386
    OS Version      : Windows 10
    R Version       : 4.2.3

### Steps to reproduce the problem
Run the following code in a Notebook file (.Rmd, .qmd) in RStudio
```
library(tibble)
library(tidyr)
cars_tibble <- as_tibble(cars)
df_nested <- tibble(id = 1:1e6, nested = replicate(1e6, list(cars_tibble)))
df_renested <- df_nested %>% unnest(nested) %>% nest(nested = -id)
```
### Describe the problem in detail
I have finally found a way to reproduce the performance problem that has been crushing my productivity with RStudio for the past months. My sessions regularly slow down to evaluate no more than one line per second.
See the attached screenshot for an illustration with the supplied code:
![rstudio_slowdown_nested_tibbles](https://user-images.githubusercontent.com/9588/228309191-ed330b09-ce05-401c-9619-0a1ed6e13c95.png)
I am benchmarking chunks with tic, toc to show that once df_renested exists, RStudio fails to execute chunks in any reasonable amount of time.
In principle, df_renested should be the same as df_nested but df_nested causes no issues.
Once df_renested exists, I can watch RStudio submit line by line to the R Console at an absolute crawl.

### Describe the behavior you expected
While large dataframes may in some cases hamper performance, I don't believe it should be possible to cause such performance issues when executing unrelated chunks of code, that R has no problem evaluating in milliseconds.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
Thanks for the bug report; I was able to reproduce with the following script (copying from your example)

`````
```{r}
library(tibble)
library(tidyr)
cars_tibble <- as_tibble(cars)
df_nested <- tibble(id = 1:1e6, nested = replicate(1e6, list(cars_tibble)))
[...]

PR #12925, opened 2023-03-28
prefer unordered set

kevinushey:
### Intent

Addresses part of https://github.com/rstudio/rstudio/issues/12923.

### Approach

Use `std::unordered_set` rather than `std::set`, to avoid the cost of potentially expensive searches when the set grows very large.

### Automated Tests

N/A

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/12923.

### Documentation

N/A

### Checklist

### DIFF
--- a/src/cpp/core/system/Environment.cpp
+++ b/src/cpp/core/system/Environment.cpp
@@ -16,13 +16,13 @@
 #include <core/system/Environment.hpp>
 
 #include <algorithm>
+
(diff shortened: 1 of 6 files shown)


kevinushey:
Using an unordered set doesn't seem to help much here. I think we just need to avoid scanning very large R objects (or consider if we can do away with external pointer scanning altogether in some cases).

Issue #10652, opened 2022-02-22
reconcile differences / nits in popup windows


kevinushey:
For example, the "Save on Close" dialog.

## Before

![old](https://user-images.githubusercontent.com/1976582/155215323-f96615bd-be45-451d-a2d2-8ffa10603702.png)

## After

![new](https://user-images.githubusercontent.com/1976582/155215364-17d75cee-897f-4de1-8032-e676a957f7bd.png)

We don't necessarily need to completely match the old style, but we should at least make sure the RStudio icon is used for any popups.

ronblum:
Another example with prompt to install missing packages.

Qt
<img width="260" alt="diff_qt" src="https://user-images.githubusercontent.com/31009952/161636488-b6fb2aa4-8717-45f0-9cf3-210ad5598e15.png">

Electron
<img width="251" alt="bold_elec" src="https://user-images.githubusercontent.com/31009952/161636502-f5fa681c-6c76-4fc0-b675-bd6b2c66b275.png">

timtmok:
The dialog options might need a tweak here. Specifying an icon would force the icon that shows if we need to. I'm not sure about Windows with the application icon in the dialog title. Electron can only specify one image that is derived from the `type` or `icon` (takes precedence).
https://github.com/rstudio/rstudio/blob/e07e7bcd5133850c4cff145c343cfe7103e642dc/src/node/desktop/src/main/gwt-callback.ts#L511-L528

kevinushey:
IMHO, the main thing we should consider for this issue is the change in presentation for macOS modal prompts / dialogs, as described in https://github.com/rstudio/rstudio/issues/10652#issuecomment-1088037924. We should do the work necessary to bring the old presentation back.

kevinushey:
The originally-filed issue is apparently controlled via the `noLink` message box option:

https://www.electronjs.org/docs/latest/api/dialog

It seems like the only fix would be to change the cancel button's name from "Cancel" to something more specific, or to use `noLink: true` if we wanted the "old-fashioned" dialog display style.

<img width="379" alt="Screenshot 2023-03-31 at 3 20 21 PM" src="https://user-images.githubusercontent.com/1976582/229243506-7593ffad-55a5-4c89-a797-626af7b52136.png">

Or, we can use some clever tricks, like make a button with the label "Cancel " (note the trailing space).

[...]

PR #12928, opened 2023-03-28
fix display of macOS message box

kevinushey:
Before:

Screenshot 2023-03-28 at 3 50 07 PM

After (dev build so icon is wrong; but you get the picture):

Screenshot 2023-03-28 at 3 49 20 PM


gtritchie (approved):
LGTM!

gtritchie:
Partially addresses #10652

Issue #12932, opened 2023-03-29
mouse back / forward buttons always navigate Help pane, even when not focused


kevinushey:
### System details

    RStudio Edition : 2023.03.999-dev+999
    RStudio Version : Desktop [Open Source]
    OS Version      : macOS Ventura 13.2.1
    R Version       : R version 4.2.3 (2023-03-15)

### Steps to reproduce the problem

Run some help commands to build the Help pane with some history, e.g.

```
help(rnorm)
help(stats)
```

Then, try pressing the mouse back + forward buttons somewhere in the IDE surface (not targeting help).

### Describe the problem in detail

The Help pane navigates.

### Describe the behavior you expected

The Help pane should only navigate if the mouse was within the bounds of the Help iframe.

---

<details>

<summary>Session Info</summary>

``` r
R version 4.2.3 (2023-03-15)
Platform: aarch64-apple-darwin20 (64-bit)
Running under: macOS Ventura 13.2.1

Matrix products: default
LAPACK: /Library/Frameworks/R.framework/Versions/4.2-arm64/Resources/lib/libRlapack.dylib

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

loaded via a namespace (and not attached):
[1] compiler_4.2.3    ragg_1.2.5        tools_4.2.3       rstudioapi_0.14  
[5] textshaping_0.3.6 systemfonts_1.0.4
```

</details>

ronblum:
Note: This works correctly with the touchpad.

jonvanausdeln:
Verified on Desktop 2023.05.0-daily+180
Windows 11, Ubuntu22

Mouse forward/back buttons worked as expected on Source pane with multiple tabs open, as well as in Help pane for several different packages.

PR #12937, opened 2023-03-30
resolve issues with mouse back / forward navigation in Electron

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/12932.

### Approach

Electron appears to handle back / forward mouse buttons natively within iframes, so suppress the extra mouse handling we try to use in the iframe for Electron.

### Automated Tests

N/A

### QA Notes

Test that mouse back / forward buttons work as expected when the mouse lies over:

- The Source pane (should navigate back and forward through Source locations);
- The Help pane (should navigate back and forward through Help locations);
- Other locations (should do nothing)

You can create multiple documents with e.g. Cmd + Shift + N to test the back + forward behavior in the Source pane, and use something like:

```
help(rnorm)
help(c)
help(help)
```

to open some Help topics to test back / forward navigation in the Help pane.

NOTE: This change should only effect the desktop builds; server should be un-affected.

### Documentation

N/A

### Checklist

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/views/help/HelpPane.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/views/help/HelpPane.java
@@ -79,7 +79,7 @@ import org.rstudio.core.client.widget.ToolbarButton;
 import org.rstudio.studio.client.RStudioGinjector;
 import org.rstudio.studio.client.application.Desktop;
 import org.rstudio.studio.client.application.events.EventBus;
-import org.rstudio.studio.client.application.events.MouseNavigateEvent;
+import org.rstudio.studio.client.application.events.DesktopMouseNavigateEvent;
 import org.rstudio.studio.client.common.AutoGlassPanel;
 import org.rstudio.studio.client.common.GlobalDisplay;
 import org.rstudio.studio.client.common.SimpleRequestCallback;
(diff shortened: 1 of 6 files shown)


src/gwt/src/org/rstudio/studio/client/workbench/views/help/HelpPane.java
```diff
@@ -176,19 +176,6 @@ else if (scrollPos_ != null)
 
       prefs_.helpFontSizePoints().bind((Double value) -> refresh());
       
-      events_.addHandler(MouseNavigateEvent.TYPE, (MouseNavigateEvent event) ->
```
gtritchie, line 179: Will help pane navigation still work for RStudio Server with this code removed? The other code you added is desktop-only (DesktopApplicationHeader).

gtritchie (commented):
Question on removed code

src/gwt/src/org/rstudio/studio/client/workbench/views/help/HelpPane.java
kevinushey: Yes; it appears that event is only fired on Desktop:

https://github.com/rstudio/rstudio/blob/fbb7fafacf1e185ea2d782b0ef35ded9f613a54e/src/gwt/src/org/rstudio/studio/client/application/DesktopHooks.java#L258-L261

https://github.com/rstudio/rstudio/blob/fbb7fafacf1e185ea2d782b0ef35ded9f613a54e/src/gwt/src/org/rstudio/studio/client/workbench/views/source/Source.java#L346-L347

(and as added in this PR). Perhaps we should rename the event to make it clear it's intended to be desktop-only?

gtritchie (approved):
LGTM! Does the Plots pane also support this? Just curious; I don't have a mouse with those buttons handy to try it.

kevinushey:
Good question; it look like it doesn't.

Issue #12908, opened 2023-03-25
Changes to environment very slow when large `sf` object is loaded.


dblodgett-usgs:
### System details

    RStudio Edition :  Desktop
    RStudio Version :  2023.03.0+386
    OS Version      :  Windows 10  21H2
    R Version       :  4.2.2

### Steps to reproduce the problem

```r
sessionInfo()
options(timeout = 60000)
u <- "https://prod-is-usgs-sb-prod-publish.s3.amazonaws.com/641f0b82d34e807d39b8a1c1/sample_very_large_sf.rds"
f <- tempfile(fileext = ".rds")
download.file(u, f, mode = "wb")

test <- readRDS(f)

x <- 1

# observe very slow response when removing x
rm(x)

# also notice that operations that mutate the environment are SLOW
test2 <- test[1:100, ]
rm(test2)

# operations that mutate existing data are not slow?
test$test <- sample(1:10, nrow(test), replace = TRUE)
```

```r
> sessionInfo()

R version 4.2.2 (2022-10-31 ucrt)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 19044)

Matrix products: default

locale:
[1] LC_COLLATE=English_United States.utf8  LC_CTYPE=English_United States.utf8   
[3] LC_MONETARY=English_United States.utf8 LC_NUMERIC=C                          
[5] LC_TIME=English_United States.utf8    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

loaded via a namespace (and not attached):
 [1] compiler_4.2.2   cli_3.6.0        tools_4.2.2      withr_2.5.0      fs_1.6.0        
 [6] glue_1.6.2       rstudioapi_0.14  reprex_2.0.2     lifecycle_1.0.3  rlang_1.0.6.9000
```

Also note an error emitted while typing out the `[1:100, ]` part of the below.
```
test2 <- test[1:100, ]

Error in order(typeScores, scores) : argument lengths differ
In addition: Warning message:
In completions$type == .rs.acCompletionTypes$DATAFRAME & !completions$context %in%  :
  longer object length is not a multiple of shorter object length
```

I also notice that after running the reprex, the RStudio download monitor process name remains in Task Manager and is the process that is thrashing when the environment changes... This is likely unrelated, but was an odd observation worth noting.

### Describe the problem in detail

[...]

PR #12958, opened 2023-04-04
don't scan global environment for null pointers by default

kevinushey:
Addresses https://github.com/rstudio/rstudio/issues/12908.
Addresses https://github.com/rstudio/rstudio/issues/12923.

If we want to reintroduce this as an always-on feature, I think we need to be more targeted in terms of what R objects we choose to scan for null pointers. IIRC we did this in the past because certain R objects would cause a session crash when trying to examine those objects (due to internal null pointers) but it might be better to just explicitly filter those objects out in the cases where examining those objects cause a crash.

I vaguely recall https://cran.r-project.org/web/packages/ff/index.html being one culprit but unfortunately don't have a reproducible example.

### DIFF
--- a/src/cpp/session/SessionConsoleInput.cpp
+++ b/src/cpp/session/SessionConsoleInput.cpp
@@ -31,6 +31,7 @@
 #include "modules/jobs/SessionJobs.hpp"
 #include "modules/overlay/SessionOverlay.hpp"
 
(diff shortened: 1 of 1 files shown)


romainfrancois (approved):
The only thing is that perhaps the test could be done in `environment::isSuspendable()` but that does not really matter.

Issue #12916, opened 2023-03-27
can't select version of R when only one version installed


kevinushey:
@kevinushey here's the screenshot. Sorry for the confusion, I meant the "Ok" button is grayed out.

![image](https://user-images.githubusercontent.com/16002036/228069180-725fbc76-4317-49f3-bb6a-95286b36b1ad.png)

EDIT: just downloaded a different R version to check if it might be related to the number of version in the list and yep!

![image](https://user-images.githubusercontent.com/16002036/228069867-1c2be853-a819-4a44-99f3-5720919d6988.png)

**I closed and reopened the cmd prompt and started rstudio without any additional environment variables and it works now!**

_Originally posted by @ctm22396 in https://github.com/rstudio/rstudio/issues/12452#issuecomment-1485878161_

gtritchie:
Also noticed another quirk we should investigate and fix as part of this.

On my Windows developer machine, where I have R 3.6.3 installed into C:\R\R-3.6.3 and no other versions of R at the moment, Qt Desktop (2022.07) shows this, note the radio buttons for "machine's default version" are enabled. There's also a duplicate set of entries in the list below (same paths shown with long and short pathnames) but Electron appears to avoid this problem.

![Screenshot of Choose R dialog from Qt RStudio](https://user-images.githubusercontent.com/10569626/228651419-f736d72c-9887-4272-a06e-4da75777b52a.png)

With Electron build 2023.05.0-hourly+163, the radio buttons are disabled.

![Screenshot of Chooose R dialog in Electron](https://user-images.githubusercontent.com/10569626/228651877-d79e8a78-60a9-4ff0-8ab2-e43451d02fbb.png)

jonvanausdeln:
Verified on Desktop 2023.05.0-daily+206
Windows 11, MacOS 12.6

Tested with only one version of R installed (but with several different versions) and it behaved as expected.

PR #12967, opened 2023-04-07
Fix 

gtritchie:
### Intent

Addresses [Can't select version of R when only one version installed #12916](https://github.com/rstudio/rstudio/issues/12916)

### Approach

Two separate problems are covered by this issue (and this fix).

First, when there was only a single choice available in the list of R versions, the OK button was disabled and could not be re-enabled. Fixed by not preselecting items as they are added to the list. Now the user can click the version they want (e.g. the only version), which will trigger enabling of the OK button.

Second, when the dialog was invoked from inside the IDE (Tools / Global Options / General / R version: Change), the radio buttons for "Use your machine's default 64-bit / 32-bit version of R" were disabled when they shouldn't have been.

This was caused when we invoked the default versions of R (determined from the registry) to fetch path info, but had the R environment variables set (especially R_ARCH). The IDE sets these for its selected version of R. Trying to invoke a different version of R would fail, and the buttons would be disabled as a result. The fix was to unset those environment variables when invoking the default version of R. This could also be a problem for the dialog at startup if those environment variables happened to be set.

### Automated Tests

None

### QA Notes

Be sure to test the Choose R dialog at startup (holding down Ctrl when you launch) and from inside the IDE.

### Documentation

None

### Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -31,6 +31,7 @@
 - Fixed issues with mouse back / forward navigation in Source pane, Help pane (#12932)
 - Fixed opening files from command-line with relative paths (#12495, #12563)
 - Fixed issue with column preview with older versions of the 'pillar' package. (#12863)
+- Fixed bug where the OK button was disabled in Choose R dialog when only one version of R installed (#12916)
 
 #### Posit Workbench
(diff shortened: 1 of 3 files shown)


kevinushey (commented):
LGTM

Issue #12963, opened 2023-04-06
RStudio Desktop on Linux won't start if ~/.config/RStudio/desktop.ini isn't readable


HeenaSaini:
Hi i have installed 
RSTUDIO-2023.03.0-386-AMD64.DEB. I am getting this error:

heena@heena-HP-Z240-Tower-Workstation:~$ rstudio
A JavaScript error occurred in the main process
Uncaught Exception:
Error: EACCES: permission denied, open '/home/heena/.config/RStudio/desktop.ini'
    at Object.openSync (node:fs:600:3)
    at Object.func [as openSync] (node:electron/js2c/asar_bundle:2:1822)
    at readFileSync (node:fs:468:35)
    at t.readFileSync (node:electron/js2c/asar_bundle:2:9165)
    at c.append (/usr/lib/rstudio/resources/app/.webpack/main/index.js:8:183934)
    at new c (/usr/lib/rstudio/resources/app/.webpack/main/index.js:8:183448)
    at e.exports (/usr/lib/rstudio/resources/app/.webpack/main/index.js:8:183067)
    at new c (/usr/lib/rstudio/resources/app/.webpack/main/index.js:19:50744)
    at 7723 (/usr/lib/rstudio/resources/app/.webpack/main/index.js:19:52204)
    at __webpack_require__ (/usr/lib/rstudio/resources/app/.webpack/main/index.js:19:179183)
libva error: vaGetDriverNameByIndex() failed with unknown libva error, driver_name = (null)
[37421:0406/144416.928942:ERROR:viz_main_impl.cc(186)] Exiting GPU process due to errors during initialization
libva error: vaGetDriverNameByIndex() failed with unknown libva error, driver_name = (null)

When I run : sudo rstudio --no-sandboxup : it is still not working 

heena@heena-HP-Z240-Tower-Workstation:~sudo rstudio --no-sandboxup
[37152:0406/144221.047886:ERROR:bus.cc(399)] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")
[37152:0406/144221.047934:ERROR:bus.cc(399)] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")
Authorization required, but no authorization protocol specified
Authorization required, but no authorization protocol specified
Authorization required, but no authorization protocol specified
[...]

PR #12971, opened 2023-04-09
Enable RStudio to start even if legacy desktop.ini exists but is not readable

gtritchie:
### Intent

Addresses [RStudio Desktop on Linux won't start if ~/.config/RStudio/desktop.ini isn't readable #12963](https://github.com/rstudio/rstudio/issues/12963)

### Approach

Wrap the code in a try/catch, write an error message to stderr (too early in startup for the logging), and continue startup.

### Automated Tests

None

### QA Notes

The file ~/.config/RStudio/desktop.ini is where the Qt-based RStudio Desktop stored desktop preferences, and Electron is trying to read it to migrate those settings.

You can repro this bug by changing ownership/permissions on that file so current user can't read it, then start RStudio.

With this fix RStudio no longer crashes on startup in this situation and proceeds to startup. Of course we won't be able to migrate the settings from desktop.ini since we can't read them, but better to start than fail in an ugly fashion.

### Documentation

None

### Checklist

### DIFF
--- a/src/node/desktop/src/main/preferences/file-preferences.ts
+++ b/src/node/desktop/src/main/preferences/file-preferences.ts
@@ -34,7 +34,11 @@ class FilePreferences extends DesktopOptions {
     const desktopIni = userConfigDir.completePath(INI_FILE).getAbsolutePath().replace('rstudio', 'RStudio');
 
     if (FilePath.existsSync(desktopIni)) {
-      this.properties = PropertiesReader(desktopIni);
+      try {
+        this.properties = PropertiesReader(desktopIni);
+      } catch (err: unknown) {
+        console.error(err.message); // too early in startup for logging
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/node/desktop/src/main/preferences/file-preferences.ts
```diff
@@ -34,7 +34,11 @@ class FilePreferences extends DesktopOptions {
     const desktopIni = userConfigDir.completePath(INI_FILE).getAbsolutePath().replace('rstudio', 'RStudio');
 
     if (FilePath.existsSync(desktopIni)) {
-      this.properties = PropertiesReader(desktopIni);
+      try {
+        this.properties = PropertiesReader(desktopIni);
+      } catch (err: unknown) {
+        console.error(err.message); // too early in startup for logging
```
sharon-wang, line 40: Maybe we could add some additional text to let the user know we couldn't migrate their preferences?

Just in case they open RStudio and notice their preferences are different/missing, if they see the console error message, they'll also see the reason for the preference change in addition to the error message.

sharon-wang (approved):
lgtm! Just one suggestion

PR #12977, opened 2023-04-11
Always use the configured database.conf file

MariaSemple:
### Intent

Ensure we never check the database.conf settings without reading it at the configured location.

### Approach

Only read the config file once by separating the function that checks the driver type and passing the existing config object.

### Automated Tests

I don't think we have anything around this. It was caught by @kevinushey when he tried to start a dev rstudio-server instance on MacOS. We could test by configuring an alternate path in one or more of our fuzzbucket environments.

### QA Notes

If we configure an alternate database.conf location, we should get the settings it has, not the defaults. 

### Documentation

No documentation - we already document how to change the location, this just ensures it will work.

### Checklist

### DIFF
--- a/src/cpp/server_core/ServerDatabase.cpp
+++ b/src/cpp/server_core/ServerDatabase.cpp
@@ -406,7 +406,7 @@ Error initialize(const std::string& databaseConfigFile,
    if (error)
       return error;
    
-   if(getConfiguredDriver(options) == Driver::Postgresql)
+   if (getConfiguredDriver(options) == Driver::Postgresql)
    {
       validateMinimumPostgreSqlVersion();
    }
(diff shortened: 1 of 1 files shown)


src/cpp/server_core/ServerDatabase.cpp
```diff
@@ -402,7 +406,7 @@ Error initialize(const std::string& databaseConfigFile,
    if (error)
       return error;
    
-   if(getConfiguredDriver() == Driver::Postgresql)
+   if(getConfiguredDriver(options) == Driver::Postgresql)
```
kevinushey, line 409:
```suggestion
   if (getConfiguredDriver(options) == Driver::Postgresql)
```

kevinushey (approved):
LGTM!

zachhannum (approved):
Sorry, thought I already approved this! LGTM.

Issue #12984, opened 2023-04-12
RStudio failing to query R for path locations using R.exe on Windows but RTerm.exe works


gtritchie:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2023.05.0-daily-202
    OS Version      : Windows 10 or 11
    R Version       : Any

### Steps to reproduce the problem

Splitting this out from [RStudio not able to load selected R version on Windows #12452](https://github.com/rstudio/rstudio/issues/12452)

On certain Windows systems (haven't identified the root cause), RStudio's attempt to query for path locations by invoking R.exe fails but works if Rterm.exe is used.

@morrisseyj has reported this problem, but suspect there are many others.

Starting with build 2023.05 (202), there is no way to work around this.

### Describe the problem in detail

Users can't start RStudio on Windows after selecting an R version.

### Describe the behavior you expected

Works on all systems.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

gtritchie:
Internal research note: there is some precedent for using `rterm.exe` instead of `r.exe` when we ask the chosen version of R for some info during startup. In Qt desktop, we did that here:

https://github.com/rstudio/rstudio/blob/3a3ab49f810c1ede72e53cef39f364cfb90aaf09/src/cpp/desktop/DesktopSessionLauncher.cpp#L585

morrisseyj:
Problem appears to be solved for me.

@gtritchie indicated, in a private correspondence, that the hourly build had been updated to prefer Rterm.exe when executing Rstudio. Just tried it and 

Just tried it using the latest hourly build: 2023.05.0-hourly+216. RStudio launches as expected with no dialogue asking me to select my version of R. 

Hopefully this is all sorted. 

If you need anything else, let me know. 

Thanks @gtritchie  for all of the work to get this resolved.

PR #12986, opened 2023-04-12
use Rterm.exe instead of R.exe on Windows when checking R installation status

gtritchie:
### Intent

Addresses [RStudio failing to query R for path locations using R.exe on Windows but RTerm.exe works #12984](https://github.com/rstudio/rstudio/issues/12984)

### Approach

On Windows, when checking if an R installation is valid (and when querying an R installation for various paths) use `Rterm.exe` instead of `R.exe`. Haven't isolated why, but on some customer configurations `Rterm.exe` works and `R.exe` does not. The R documentation suggests `Rterm.exe` is preferred on Windows (though it doesn't say why), and we used `RTerm.exe` in the Qt desktop for a similar exercise, so this seems like the "right thing".

Additionally, when checking the default 32/64-bit R installations found in the registry, check using the architecture-specific Rterm.exe instead of the generic `bin\R.exe`.

Added another diagnostic environment variable: `RSTUDIO_DESKTOP_MODAL_DEVTOOLS`. If this is set when the Choose R dialog is displayed (or any other future dialogs built in a similar fashion), that dialog will show the Chrome devtools, which can be handy for debugging.

### Automated Tests

No new tests.

### QA Notes

Sanity testing of the various Choose R scenarios on Windows. Ultimately having broad customer testing will be essential since we haven't isolated the exact trigger for this problem in order to reproduce it.

### Documentation

N/A

### Checklist

### DIFF
--- a/src/node/desktop/src/ui/modal-dialog.ts
+++ b/src/node/desktop/src/ui/modal-dialog.ts
@@ -16,6 +16,7 @@
 import { BrowserWindow, BrowserWindowConstructorOptions, ipcMain } from 'electron';
 import { err, Expected, ok } from '../core/expected';
 import { safeError } from '../core/err';
+import { getenv } from '../core/environment';
 
 export abstract class ModalDialog<T> extends BrowserWindow {
   abstract onShowModal(): Promise<T>;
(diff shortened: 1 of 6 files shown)


src/node/desktop/src/ui/modal-dialog.ts
```diff
@@ -84,6 +85,11 @@ export abstract class ModalDialog<T> extends BrowserWindow {
     // show the window after loading everything
     this.show();
 
+    const showDevTools = getenv('RSTUDIO_DESKTOP_MODAL_DEVTOOLS').length !== 0;
```
kevinushey, line 88: This will be handy!

kevinushey (approved):
LGTM! Hopefully this helps at least some more Windows users get unstuck.

Issue #12988, opened 2023-04-13
Poor experience starting RStudio desktop if user-state folder inaccessible


gtritchie:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2023.03-386
    OS Version      : any
    R Version       : any

### Steps to reproduce the problem

- deny your user account access to the RStudio user-state folder:
    - (Mac/Linux): `chmod 0 ~/.local/share/rstudio`
    - (Windows): `icacls %localappdata%\RStudio /deny %USERNAME%:(OI)(CI)F`
- start RStudio

### Describe the problem in detail

- window never appears but process remains running
- if you run from a terminal, you'll see: 

```
(node:71423) UnhandledPromiseRejectionWarning: Error: Cannot create BrowserWindow before app is ready
    at t.createStandaloneErrorDialog (/Applications/RStudio.app/Contents/Resources/app-arm64/.webpack/main/index.js:19:80580)
    at Object.main (/Applications/RStudio.app/Contents/Resources/app-arm64/.webpack/main/index.js:19:37207)
(Use `RStudio --trace-warnings ...` to show where the warning was created)
```

### Describe the behavior you expected

- Behave more like Qt desktop did in this scenario: on Mac, it would display the main window with an error message so the user has some idea how to fix it; Windows is a bit different
- At a minimum, don't just sit there with no UI and a process running in the background; show SOMETHING and allow user to exit

<img width="1089" alt="example of RStudio 2022.07 behaviour" src="https://user-images.githubusercontent.com/10569626/231827619-d3dfb4db-0bed-4b2b-92dd-542df8008226.png">

### Restoring Folder Access on Windows
- `icacls %localappdata%\RStudio /grant %USERNAME%:(OI)(CI)F`

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
Verified in RStudio Desktop 2023.05.0-daily+275 on
- MacOS 13.4
- Windows 11
- Ubuntu 22.04.2

ronblum:
[...]

PR #12990, opened 2023-04-13
Don't fail silently at startup when user state folder inaccessible

gtritchie:
### Intent

Addresses [Poor experience starting RStudio desktop if user-state folder inaccessible #12988](https://github.com/rstudio/rstudio/issues/12988)

### Approach

Our logging package (Winston) was throwing an uncaught exception, preventing a bunch of initialization code from running. Fixed by catching the exception and writing to stderr. This permits startup to continue so the window appears and existing error-handling code can report the problem.

Not pretty (especially on Windows), but at least it doesn't make it appear nothing happened, and the user can close RStudio instead of accumulating processes in the background.

### Automated Tests

None

### QA Notes

On Mac and Linux, the main window should now appear and show basic error information.

The experience on Windows is still somewhat gruesome. R shows a modal dialog box with an error, which may show up under the splash screen (but you can move the splash screen), then once you dismiss that the main window appears with no particularly useful information. This is still better than it was and reasonably close to what Qt did (Qt didn't have the splash screen).

### Documentation

None

### Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -34,6 +34,7 @@
 - Fixed bug where the OK button was disabled in Choose R dialog when only one version of R installed (#12916)
 - Improved robustness when chosing custom R version in Windows desktop (#12969)
 - Fixed bug that prevented RStudio Desktop from starting on Linux if desktop.ini was unreadable (#12963)
+- Improved RStudio Desktop startup behavior when user state folder inaccessible (#12988)
 
 #### Posit Workbench
 - Fixed unlabeled buttons for screen reader users when page is narrow [Accessibility] (rstudio/rstudio-pro#4340)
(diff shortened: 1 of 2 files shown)


sharon-wang (approved):
lgtm 👌 

For my own reference, when do we use `logger().log[Error/Debug/etc]()` versus `console.[log/error/etc]()`? I think I may have been using a mix 😓

gtritchie:
> For my own reference, when do we use logger().log[Error/Debug/etc]() versus console.[log/error/etc]()? I think I may have been using a mix 😓

Answered in Slack, but just in case somebody else is wondering: generally should always use logger().log*, as that logs to a file, and pays attention to the --log-level. However, logger() doesn't work if you call it before it is initialized, so some startup code uses `console.error` and `console.log` to output errors.

We should probably tweak logger() so it can always be called (i.e. it could use `console` if invoked before it has been initialized), then you don't have to think about it.

sharon-wang:
In case anyone is still following along in this thrilling logging saga, we've added a logging section to the Electron Desktop wiki: https://github.com/rstudio/rstudio/wiki/Electron-Desktop#logging

PR #13000, opened 2023-04-14
fix windows-path-related test failures

sharon-wang:
### Intent
Fixes for the following tests:
- System test `fixupExecutablePath adds missing exe extension on Windows`
- Utils test `findComponents returns session and scripts paths`

### System test `fixupExecutablePath adds missing exe extension on Windows`
- The test previously used an incomplete file path that would cause the test to always fail on Windows
- Test was modified to check the extension matches `.exe` for Windows, and doesn't match for non-Windows

### Utils test `findComponents returns session and scripts paths`
- enhancement on `findBuildRootImpl` in utils to only use package build dirs if the app is a package build

Fixes the scenario I was experiencing:
- Run a package build (doesn't have to be successful, you can cancel it partway through)
- Run electron unit tests (`npm test`)
- Will use the package build dir since it's the most recently modified build dir, but the test isn't running a package build and there may not be an rsession.exe available, so the test fails

### Automated Tests
N/A, test fix

### QA Notes
N/A, test fix

### Documentation
N/A, test fix

### Checklist

~- [ ] If this PR adds a new feature, or fixes a bug in a previously released version, it includes an entry in `NEWS.md` ~

### DIFF
--- a/src/node/desktop/src/main/utils.ts
+++ b/src/node/desktop/src/main/utils.ts
@@ -190,11 +190,16 @@ function findBuildRootImpl(rootDir: string): string {
   const buildDirParents = [
     `${rootDir}`,
     `${rootDir}/src`,
-    `${rootDir}/src/cpp`,
-    `${rootDir}/package/linux`,
-    `${rootDir}/package/osx`,
-    `${rootDir}/package/win32`,
+    `${rootDir}/src/cpp`
   ];
+
+  if (app.isPackaged) {
+    buildDirParents.push(
+      `${rootDir}/package/linux`,
+      `${rootDir}/package/osx`,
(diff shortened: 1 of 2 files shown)


sharon-wang:
Now that this test is passing, another one is failing (only on Windows). Looking into the other test

```
 1) Utils
       findComponents returns session and scripts paths:

  Error
      at rsessionNotFoundError (rstudio\src\node\desktop\src\main\utils.ts:233:10)
      at Object.findComponents
```

sharon-wang:
@gtritchie I've added the other test fix (seems like I hit a particular scenario) so this is ready for review now.

gtritchie (approved):
LGTM!

PR #13024, opened 2023-04-20
fix `findComponents` test failure

sharon-wang:
### Intent

Fixes https://github.com/rstudio/rstudio-pro/issues/4579

Might not be a coincidence after all! Seems like this issue started happening in a build that contained #13000.

### Approach
Don't conditionally exclude the package directories from the possible build dir candidates. Just include them by default. (this was the behaviour before #13000).

### Automated Tests

The `findComponents` Electron unit test should pass on all platforms now.

### QA Notes

1. Ensure you don't have any `build*` directories in `src/cpp`
2. Create a package build
3. In `src/node/desktop`, run `npm test`
4. `findComponents` test should now pass

### Documentation
N/A

### Checklist

### DIFF
--- a/src/node/desktop/src/main/utils.ts
+++ b/src/node/desktop/src/main/utils.ts
@@ -190,16 +190,11 @@ function findBuildRootsImpl(rootDir: string): string[] {
   const buildDirParents = [
     `${rootDir}`,
     `${rootDir}/src`,
-    `${rootDir}/src/cpp`
+    `${rootDir}/src/cpp`,
+    `${rootDir}/package/linux`,
+    `${rootDir}/package/osx`,
+    `${rootDir}/package/win32`,
   ];
-
-  if (app.isPackaged) {
-    buildDirParents.push(
-      `${rootDir}/package/linux`,
-      `${rootDir}/package/osx`,
-      `${rootDir}/package/win32`
-    );
-  }
 
   // list all files + directories in root folder
   for (const buildDirParent of buildDirParents) {
--- a/src/node/desktop/src/main/utils.ts
+++ b/src/node/desktop/src/main/utils.ts
@@ -170,19 +170,19 @@ export function findRepoRoot(): string {
 }
 
 // used to help find built C++ sources in developer configurations
-function findBuildRoot(): string {
+function findBuildRoots(): string[] {
   // look for the project root directory. note that the current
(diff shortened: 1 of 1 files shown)


sharon-wang:
I was just thinking about this test failure a bit more:

It's possible that after iterating all the build directory candidates, that `rsession` still won't be found. However, that's a case the test would catch and the test would be reporting helpful info.

I'm not sure about the state of the machine where this test was failing -- why its most recently modified `build` directory didn't contain `rsession`.

Now I'm wondering if this test is actually reporting the right thing, and maybe these changes aren't needed? As in, the most recent build failed and there isn't an `rsession` executable available.

timtmok:
In a Jenkins build, wouldn't the most recent build root be the correct one to use? It runs in a container so there shouldn't be a conflict with another build. Although, it's been confusing before with which build output is being used, this change might make it more confusing since it won't always use the most recent build.

sharon-wang:
Yeah if it's isolated in a container, then it wouldn't make sense to possibly use a a different build which was successful, but isn't the latest build. I've reverted that commit.

I'm not certain why the build failed, but the remaining thing I can think of is to include the package directories which I previously conditionally included for packaged apps. I just pushed a commit to add this back in.

I would think that a non-package linux build wouldn't check the packaged app directory for `rsession`, but I'm not totally sure how the Electron tests are invoked during the build pipeline.

sharon-wang:
I looked through our Jenkinsfiles for mac, windows and linux and based on those scripts, they are all package builds. I'd expect `findComponents` to fail on all platforms if `app.isPackaged` isn't properly indicating if the app is in fact packaged. Perhaps there's some weirdness on Linux.

I'll wait for the PR build to complete (it's running on Linux) and double-check that `findComponents` passes.

GCRev (approved):
This seems fine to me. Unfortunately the PR auto-build check does not include Electron tests. If it tested fine on your linux environment then I expect it will also work in Jenkins.

sharon-wang:
> This seems fine to me. Unfortunately the PR auto-build check does not include Electron tests. If it tested fine on your linux environment then I expect it will also work in Jenkins.

Yes, this worked in my Linux environment. I'll merge. Fingers crossed we'll have some passing Linux builds now! 🤞

Issue #13025, opened 2023-04-20
Add ability to opt-in Electron crash handling via environment variable


gtritchie:
Electron's ability to generate crash dumps (when Electron itself crashes) is not enabled. So if Electron crashes, it just goes away and leaves no clues we can follow, such as in: https://github.com/rstudio/rstudio/issues/13007.

As a first step, support enabling it via an environment variable, but keep it off by default. We don't want it on by default until we've done all the work to support uploading crash dumps tracked in [Hook up crash reporter for Electron desktop #13022](https://github.com/rstudio/rstudio/issues/13022).

Suggested: `RSTUDIO_ENABLE_CRASHPAD=0|1|true|false`

Additionally, add a command (in Electron desktop only) to Help / Diagnostics / Crash RStudio Desktop that will bring up a suitably scary "Are you sure?" dialog. If they say "Yes", then invoke `process.crash()` in the main Electron process.

This will facilitate basic testing of crash dump generation (or lack thereof).

ronblum:
Verified on MacOS 13.4. Windows and Linux TBD.

ronblum:
Verified in RStudio Desktop 2023.05.0-daily+296/304 on
- MacOS 13.4
- Windows 11
- Red Hat 9.1

Checked
- Valid values (0, 1, true, false)
- Invalid value (bogus)
- Variable not defined

Issue #13022, opened 2023-04-20
Hook up crash reporter for Electron desktop


gtritchie:
Electron has a crash-reporting mechanism that we didn't hook up. So if Electron itself (aka "rstudio" process) crashes (which appears to be the case in #13007), we don't get any useful information, nor does the user get any indication that a crash took place; it just vanishes.

If running rstudio from command-line when this happens, you'll see: `ERROR:crashpad_client_win.cc(844)] not connected`

Details on hooking this up: https://www.electronjs.org/docs/latest/api/crash-reporter

We use our own fork of crashpad for other processes (rsession), so it's possible they might not get along...

PR #13026, opened 2023-04-20
Add opt-in via environment variable for Electron crash dump generation

gtritchie:
### Intent

Addresses [Add ability to opt-in Electron crash handling via environment variable #13025](https://github.com/rstudio/rstudio/issues/13025)

### Approach

Set the environment variable `RSTUDIO_ENABLE_CRASHPAD` to "1" or "true", then start RStudio Desktop (Electron-only). This will enable the electron [crashReporter feature](https://www.electronjs.org/docs/latest/api/crash-reporter), but only in local mode (it won't try to upload). If you run RStudio from a command line, you'll also see a message such as: `Starting crash logging to C:\Users\username\AppData\Roaming\RStudio\Crashpad`.

This is an interim step on the way to fully implementing Electron crash handling (tracked in https://github.com/rstudio/rstudio/issues/13022). Not comfortable enabling this by default without further testing and research, but being able to turn it on when there's a crash happening could be useful.

When Electron crash reporting is on, and Electron itself (i.e. RStudio[.exe]) has a crash, a `dmp` file will be generated. These will continue to accumulate unless manually deleted (one of the reasons we don't want this on by default at this time).

Windows: ` %appdata%\RStudio\Crashpad`
Mac: `~/Library/Application Support/RStudio/Crashpad`
Linux: `~/.config/RStudio/Crashpad`

To aid in testing this, added a new menu command: Help / Diagnostics / Crash RStudio Desktop (DANGER). This should only show up in Electron desktop, not Server or Qt desktop.

When invoked it shows a Yes/No dialog warning that work could be lost, etc. If "Yes" is selected, it invokes `process.crash()` which does as it is told.

Screenshots (Windows):

menu

dialog

### Automated Tests

None

### QA Notes

- without the environment var set to 1 or true (or set to 0 or false), no Crashpad output is generated if Electron crashes, but if the env-var is set (BEFORE starting RStudio), a .dmp file will be generated
- force a crash with the new menu command; verify that clicking "No" prevents a crash, and "Yes" triggers one
- Recommend running RStudio from command line to see the message indicating when it is enabled, and where the info will be written
- Note that when you trigger a crash WITHOUT the reporting on, you may see a message like `[16472:0420/160015.694:ERROR:crashpad_client_win.cc(844)] not connected` (this is expected)
- Check that the menu item isn't shown in Server / Server Pro / Qt Desktop (it wouldn't do anything if it was)

### Documentation

None

### Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,7 @@
 - Moved Help panel font size setting to Appearance tab in Global Options (#12816)
 - Update openssl to 1.1.1t for Windows (rstudio/rstudio-pro#3675)
 - Improve visibility of focus rectangles on Server / Workbench Sign In page [Accessibility] (#12846)
+- Added ability to enable minidump generation for Electron crashes (#13025)
 
 #### Posit Workbench
 - Added the `session-connections-block-suspend` session option, controlling whether active connections can block suspension of an R session.
--- a/src/gwt/src/org/rstudio/studio/client/application/Application.java
+++ b/src/gwt/src/org/rstudio/studio/client/application/Application.java
@@ -409,6 +409,17 @@ public class Application implements ApplicationEventHandlers
       }
    }
 
+   @Handler
+   void onCrashDesktopApplication()
+   {
+      globalDisplay_.showYesNoMessage(
+            GlobalDisplay.MSG_WARNING,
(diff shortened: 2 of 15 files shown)


kevinushey (approved):
LGTM!

Issue #11376, opened 2022-06-07
Keyboard shortcut help not displaying


gtritchie:
### System details

    RStudio Edition : Desktop
    RStudio Version : RStudio 2022.02.3+492, 2022.02.2+485
    OS Version      : macOS 11.6.6 (Big Sur / Intel)
    R Version       : Any

### Steps to reproduce the problem

Reported by @jennybc; so far I can't repro on macOS 12.4; don't have a Big Sur machine to try it on.

- Start desktop IDE
- Make sure screen-reader / accessibility mode is OFF
- (1) Help / Keyboard Shortcuts Help
- (2) invoke via Option+Shift+K
- (3) invoke via Command Palette ("Keyboard Shortcuts Help")

### Describe the problem in detail

- (1) nothing happens
- (2) inserts an Apple glyph into the console
- (3) get error:

![Screen Shot 2022-06-07 at 8 48 51 AM](https://user-images.githubusercontent.com/10569626/172425671-35f44981-d8ae-4f57-8b94-c5d93d306371.png)

With screen reader mode on, the feature is working (we show keyboard shortcut help in system web browser when accessibility is enabled).

### Describe the behavior you expected

- keyboard shortcut help shown in a popup

![Screen Shot 2022-06-07 at 9 55 58 AM](https://user-images.githubusercontent.com/10569626/172426606-f43875b4-23ec-4f99-9f40-e9b50f980954.png)

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
@gtritchie I'm unable to reproduce this issue, both with screen reader /accessibility turn off and on.

jennybc:
I am still on Big Sur 😬 and I still see exactly what was reported earlier. But I would understand if the fact that this involves a rather old OS makes this a wontfix.

gtritchie:
Last working in the 2021.09.4-403 release, stopped working in the 2022.02.0-443 release.

ronblum:
[...]

PR #13061, opened 2023-04-27
fix keyboard shortcut help not displaying when editor in Emacs or Sublime Text mode

gtritchie:
### Intent

Addresses [Keyboard shortcut help not displaying #11376](https://github.com/rstudio/rstudio/issues/11376)

### Approach

Commands that are Emacs-only or Sublime-text only were causing a variable to be null that the code assumed was not null triggering exceptions (in GWT code). Added checks.

Given the timing of when this broke, I suspect this change came with the localization work, but I didn't dive into the gory details to see why.

### Automated Tests

None

### QA Notes

Test display of keyboard shortcut help in all editor modes (default, vim, emacs, sublime), using the keyboard shortcut, the menu, and the command palette.

If you are ambitious you could spot-check that the shortcuts being displayed match expectations (e.g. for commands that differ depending on editor mode). Probably easiest to do this by comparing with 2021.09.4-403, the last version where this was working.

While you're at it, double-check that the rarely seen, possibly top-secret, vim shortcut help is also working. This is done from the editor, in vim mode, via <kbd>:help</kbd>.

### Documentation

None

### Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -44,6 +44,7 @@
 - Fixed window disappearing when restoring from maximized on macOS (#13010)
 - Fixed Project Build Tools preferences pane showing &mdash text in French UI (#11134)
 - Fixed inserting cross references in a Quarto document showing no references (#12882)
+- Fixed keyboard shortcut help not displaying in Emacs or Sublime Text mode (#11376)
 
 #### Posit Workbench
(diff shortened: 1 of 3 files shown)


timtmok (approved):

sharon-wang (approved):
⌨️ 🩳 ✂️

PR #13064, opened 2023-04-28
fix macos s3 package upload path

sharon-wang:
### Intent

[MacOS Desktop Pro build urls](https://dailies.rstudio.com/rstudio/mountain-hydrangea/electron-pro/macos/) have been erroring:
```
<Error>
    <Code>NoSuchKey</Code>
    <Message>The specified key does not exist.</Message>
    <Key>electron-pro/macos/RStudio-pro-2023.05.0-daily-312.pro2.dmg</Key>
    <RequestId>...</RequestId>
    <HostId>...</HostId>
</Error>
```

I think this PR should fix the error above!

### Approach

Currently, the package is uploaded to S3 under `"${FLAVOR.toLowerCase()}/macos/"`, where `FLAVOR` resolves to `desktop` or `electron`. This works for Open Source because there's no `-pro` suffix appended as part of the product name:

https://github.com/rstudio/rstudio/blob/a232fe149ac4611b3f84a6b449e9061b91aba696/utils.groovy#L180-L192

However, the dailies site build `link` is constructed using `PRODUCT`:

https://github.com/rstudio/rstudio/blob/a232fe149ac4611b3f84a6b449e9061b91aba696/jenkins/Jenkinsfile.macos#L232-L233

https://github.com/rstudio/rstudio/blob/a232fe149ac4611b3f84a6b449e9061b91aba696/jenkins/Jenkinsfile.macos#L253-L254

So the links will use `desktop-pro` or `electron-pro`, but the builds are getting uploaded to `desktop` or `electron`. As a result, when a user clicks on the url, they will see the `NoSuchKey` error message.

Replacing `FLAVOR.toLowerCase()` with `PRODUCT` should align the upload location with the expected build location.

### Automated Tests

N/A

### QA Notes

@ronblum The Mac Pro daily build links should work once this is merged.

### Documentation
N/A

### Checklist

~- [ ] This PR passes all local unit tests~

### DIFF
--- a/jenkins/Jenkinsfile.macos
+++ b/jenkins/Jenkinsfile.macos
@@ -182,7 +182,7 @@ pipeline {
                       withAWS(role: 'ide-build', region: 'us-east-1') {
                         retry(5) {
                           script {
-                            utils.uploadPackageToS3 "package/osx/build/${PACKAGE_FILE}", "${FLAVOR.toLowerCase()}/macos/"
+                            utils.uploadPackageToS3 "package/osx/build/${PACKAGE_FILE}", "${PRODUCT}/macos/"
                           }
                         }
                       }


timtmok:
I think this is a fix for the `latest-builds` repo (can't remember if that's the correct one). The other platforms seem to have working download links. I think changing this would break those downloads for the other platforms.

This is what Ubuntu 22 arm64 links to:
https://s3.amazonaws.com/rstudio-ide-build/electron/jammy/arm64/rstudio-pro-2023.05.0-daily-325.pro3-arm64.deb

I do see it exists in the S3 bucket as well.

image

timtmok:
https://github.com/rstudio/latest-builds/blob/ba9e63fdce31d3d49160a5dd58c23159b7c38589/content/rstudio/mountain-hydrangea/electron-pro/macos/2023-05-0-daily-325-pro3.md is a generated file from the build that has the download link. It's using `electron-pro` for macos while `electron` is used for other platforms in the download link. The builds create the file and commits it to `latest-builds`.

I think https://build.posit.it/job/sites/job/builds.rstudio.com/job/main/ is what deploys the updated download data. It runs on a schedule but it can be run manually as well if you need to update data.

sharon-wang:
This change should only affect the mac builds since the S3 path is modified only in the macos Jenkinsfile.

The `Jenkinsfile.macos` defines the S3 url where the build is uploaded to and also specifies the S3 link that is used in the dailies site (which shows up in the generated `.md` file in the `latest-builds` repo). The current issue is that the macos Jenkinsfile uploads to S3 in a different location than the dailies S3 link (it uploads to `electron` or `desktop`, but produces a link for the dailies site that uses `electron-pro` and `desktop-pro`).

In `Jenkinsfile.linux`, this issue does not occur because it defines `AWS_PATH` which is used in all places for the S3 url, so there's no mismatch. My original commit updated `Jenkinsfile.macos` to construct the same S3 upload url that the dailies site expects, but I just pushed a commit use the same approach as `Jenkinsfile.linux` (define `AWS_PATH` and use it in all three places to avoid the S3 url mismatch).

timtmok:
@sharon-wang Looks like linux uses the third parameter to set the S3 location

`utils.publishToDailiesSite PACKAGE_FILE, DAILIES_PATH, AWS_PATH`

`DAILIES_PATH` uses `PRODUCT` in it.
`AWS_PATH` uses `FLAVOR` to set the upload location

It doesn't seem like the linux builds are uploading to S3 in `electron-pro`. I think adding a third argument will keep Mac organized like the other builds. Although, past builds would need fixes to those files in `latest-builds` to get the links working.

sharon-wang:
`publishToDailiesSite` uses the 2nd param as the dailies and aws path if the 3rd param isn't provided, but I agree it makes more sense to align Mac with Linux even if we need to update a bunch of past builds in `latest-builds`.

I went through the dailies and hourlies to identify all the builds that need url fixes. Seems like all mac builds March 22 and beyond are affected. I opened an issue in `latest-builds` to detail that.

<details>
    <summary>Some additional notes for my understanding</summary>

Needed to write this out since I was getting confused between the different urls/paths 😅 

`DAILIES_PATH` impacts the directory path where the `.md` file gets created, so we want to use `PRODUCT` here because we want `pro` to be appended for pro builds. The existing implementation already does this. We are continuing to do this.

`AWS_PATH` impacts the S3 upload location, the url used for the `link` in the `.md` file and the daily redirect (I think this is the button at the top of the dailies page). The existing implementation uses `PRODUCT`, which works, but isn't consistent with Linux. We are changing this path to use `FLAVOR` which doesn't include the `pro` suffix.

MacOS builds prior to March 22, 2023 uploaded to S3 with `-pro` included in the S3 url (`electron-pro`, `desktop-pro`). Beyond this date, it seems the builds were uploaded to `electron` and `desktop`.

The existing mac builds that don't have broken URLs will continue to be fine as-is -- no changes needed because the builds are already uploaded to locations that match the urls on the dailies site.

</details>

timtmok (approved):

MariaSemple (approved):
LGTM!

PR #13108, opened 2023-05-10
don't initialize python from restored env null ptr

t-kalinowski:
closes https://github.com/rstudio/reticulate/issues/710

### Intent

Issue: when a reticulate python object is automatically loaded from a previous R session `.RData` file, the IDE causes reticulate to initialize python while the IDE is generating the object summary for presentation in the env pane.    

### Approach

Add an early return in `rs.describeObject()` that checks for the case of the object being a reticulate null ptr, this avoids running code that otherwise assumes python is already initialized (e.g., `reticulate::import_builtins()`). Reticulate objects loaded from `.RData` will always be null ptrs.

### Automated Tests

NA

### QA Notes

To test locally:

1. create a .RData containing an R object
```r
x = reticulate::py_eval("1", convert = FALSE)
save.image()
``` 
2. Start a fresh R session and restore `.RData`
```r
load(".RData")
if (reticulate::python_available()) 
  stop("Python should not be initialized from merely loading .RData")
```

### Checklist

### DIFF
--- a/src/cpp/session/modules/SessionReticulate.R
+++ b/src/cpp/session/modules/SessionReticulate.R
@@ -1568,6 +1568,14 @@ def _rstudio_html_generator_():
 #  $ contents_deferred: 'rs.scalar' logi FALSE
 .rs.addFunction("reticulate.describeObject", function(name, parent)
 {
+
+   object <- if (inherits(parent, "python.builtin.dict"))
+      reticulate::py_get_item(parent, name)
+   else if (inherits(parent, "python.builtin.object"))
+      reticulate::py_get_attr(parent, name)
+   else
+      get(name, envir = parent)
+
    # is this a null pointer? if so, handle that up-front
    if (reticulate:::py_is_null_xptr(object))
    {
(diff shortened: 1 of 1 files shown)


src/cpp/session/modules/SessionReticulate.R
```diff
@@ -1568,15 +1568,6 @@ def _rstudio_html_generator_():
...
-   
-   object <- if (inherits(parent, "python.builtin.dict"))
-      reticulate::py_get_item(parent, name)
-   else if (inherits(parent, "python.builtin.object"))
-      reticulate::py_get_attr(parent, name)
-   else
-      get(name, envir = parent)
-   
    # is this a null pointer? if so, handle that up-front
    if (reticulate:::py_is_null_xptr(object))
```
kevinushey, line 1572: `object` won't be defined here; do we need to check if `parent` itself is a null external pointer?

kevinushey (changes requested):

t-kalinowski:
Reading through the rest of `rs.reticulate.describeObject()`, it looks like `builtins` are never used, so a simpler fix is to just remove the `import_builtins()` line. Updated the PR.

kevinushey (approved):
LGTM, thanks!

PR #13114, opened 2023-05-12
Set logic to run updateDailyRedirects

timtmok:
### Intent
Run the Update Daily Build Redirects stage

### Approach
It seems that using `env.IS_PRO` was always false since it is set as a script variable instead of an environment variable. This was causing the stage to never run.

Updated the stage condition to run if `PUBLISH` and `DAILY` are set. In addition, it updates Desktop Pro to point to Qt and open source Desktop to Electron.

### Automated Tests
none

### QA Notes
https://dailies.rstudio.com/links/ all the daily links, pro and open source, should update to the latest build

### Documentation
n/a

### Checklist

### DIFF
--- a/jenkins/Jenkinsfile.linux
+++ b/jenkins/Jenkinsfile.linux
@@ -375,11 +375,13 @@ pipeline {
                       RSTUDIO_ORG_PEM = credentials('www-rstudio-org-pem') 
                     }
 
+                    // for pro, update with Qt
+                    // for open source, update with Electron
                     when { 
-                      allOf {
-                        expression { return params.PUBLISH }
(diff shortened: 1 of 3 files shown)


MariaSemple (approved):
LGTM! I almost feel like we should write a "shouldPublish" function in utils.grove that does the `(IS_PRO && Desktop) || (!IS_PRO && Electron) || (Server)`, since we do it in 3 places, but it's not a big deal either way

GCRev (approved):
I'm glad to see that the logic is now consistent across all the OS's, including that windows path deal, which seemed like it would have been clobbering the electron redirects with the QT ones. LGTM

Issue #12918, opened 2023-03-28
Error Box: Error Retrieving Help


mikecrobp:
### System details

    RStudio Edition : Desktop
    RStudio Version : RStudio 2023.03.0+386 "Cherry Blossom" Release (3c53477afb13ab959aeb5b34df1f10c237b256c3, 2023-03-09) for Windows
    OS Version      :  Windows 10
    R Version       :  4.2.2 (2022-10-31 ucrt)

### Steps to reproduce the problem
Not possible. It is an intermittent problem. But the current incarnation stops me from typing in any new code - even though I can still type to console

### Describe the problem in detail
An error box comes up saying Error Retrieving Help Error in if (nchar(current) > nchars) {: missing value where TRUE/FALSE needed
Note 2 other users report the same behaviour in Posit Community

### Describe the behavior you expected
No error box

- [x ] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x ] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [ x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
Probably from here. https://github.com/rstudio/rstudio/blob/b2d7353fdf45be8168d19db2de3dea9b6f84bc40/src/cpp/session/modules/SessionHelp.R#L465-L490

cbeesonlynch:
I'm having the same issue after I updated R Studio - I do not have anything in my script about nchars, and it's happening randomly.

drande09:
I'm getting this same error: 
![image](https://user-images.githubusercontent.com/57731834/228916677-af490a7f-76e7-490a-8ba7-2d408d72d350.png)
It happens pretty consistently when I try to press tab to get variable names in a mutate() call, or similar.

mbedward:
@drande09 does it only happen with variable names containing underscores? That seems to be the pattern for me.

kevinushey:
A reproducible example would indeed be helpful.

drande09:
Here's the information about the data frame: 
![image](https://user-images.githubusercontent.com/57731834/229796333-47fe7b7f-661f-4d2f-b68c-68090c9e3dde.png)

[...]

PR #13132, opened 2023-05-16
fix help preview error by handling `NA` values

sharon-wang:
### Intent

Addresses #12918.

Fixes error occurring when viewing help preview for data that contains `NA` character type values.
Includes last column of data in help preview (as long as it fits).

### Approach

Use the `nchar()` `keepNA = FALSE` argument to handle when the character vector contains `NA` elements. The length of `NA` (2) will be returned instead of `NA` itself.

Use `<=` in the while loop to include the last element of the formatted column data.

### Automated Tests
none

### QA Notes

Note that the error `Error in if (nchar(current) > nchars)...` is not reproducible anymore, because the errors from session help previews were suppressed (so they won't pop up anymore).

To verify this fix, source the following code:
```R
library(data.table)
library(pillar)

StartDateTime <- as.POSIXct("2023-01-01",origin="1970-01-01")
DateTime10 <- seq(StartDateTime,StartDateTime+10, length.out=11)
test.df <- as.data.table(cbind(c(1:11),DateTime10,shift(DateTime10)))
test.df <- cbind(test.df,test.df$V3)
test.df$V2 <- as.POSIXct(test.df$V2, origin = "1960-01-01")

# try typing 'test.df$' and see what autocomplete produces
```

Then type `test.df$` to loop through the autocomplete preview for each column.

If the data for each column is successfully previewed, then the fix works! In particular:
- column `V1` should should 11 elements, 1 through 11
- column `DateTime10`, `V3` and `V2` should all appear without issue

https://github.com/rstudio/rstudio/assets/25834218/ed288188-fd48-4375-bba8-f4acc5d3d7e2

### Documentation
N/A

### Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -49,6 +49,7 @@
 - Fixed issue with use of "Whole word" with Find in Files search (#13017)
 - Fixed issue with diagnostics freezing session with documents containing pipebind operator (#13091)
 - Fixed issue with stopping a Quarto Shiny doc render on Windows (#12546)
+- Fixed "Error Retrieving Help" bug when data preview column contains `NA` values (#12918)
 
 #### Posit Workbench
 - Fixed unlabeled buttons for screen reader users when page is narrow [Accessibility] (rstudio/rstudio-pro#4340)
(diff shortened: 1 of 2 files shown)


kevinushey (approved):
LGTM!

PR #13135, opened 2023-05-16
Remove the focal exclude for building arm

GCRev:
See: https://github.com/rstudio/rstudio-pro/issues/4470

### Intent / Approach

Re-enable the arm builds for Ubuntu20 Focal. 

We wish to build OS Server and OS Electron Desktop and _NOT_ OS QT Desktop. The current excludes handle this, because we already cancel the entire `Desktop` flavor from building on _any_ arm64 architecture.

Currently this will also build Pro on Focal arm64. I don't necessarily see a problem with that considering our license manager is compatible with Focal arm64, but we may want to exclude all Pro Focal arm64 builds if it struggles.

### Checklist

### DIFF
--- a/jenkins/Jenkinsfile.linux
+++ b/jenkins/Jenkinsfile.linux
@@ -2,7 +2,7 @@ def utils
 
 def ext_map = 
 [
-  'focal':     'deb',
+  'focal':      'deb',
   'jammy':      'deb',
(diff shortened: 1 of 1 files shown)


AndrewMcClain (approved):
LGTM!

Issue #13165, opened 2023-05-24
Some Windows C++ core tests have been failing silently for over a year


cm421:
### System details

    RStudio Edition : Both
    RStudio Version : 2023.05.0-hourly+367 and higher
    OS Version      : Windows
    R Version       : Any

### Steps to reproduce the problem
Build RStudio in `package/win32/` by running 
`make-package.bat` 
and then run 
`cd package/win32/build/src/cpp && rstudio-tests.bat --scope core`

The following tests fail, with `studio-tests.bat` returning exit code -1073741819:

- https://github.com/rstudio/rstudio/blob/d3f94d47d30c6580e4224327345fc260e9e48175/src/cpp/core/LoggingTests.cpp#L721
- https://github.com/rstudio/rstudio/blob/d3f94d47d30c6580e4224327345fc260e9e48175/src/cpp/core/LoggingTests.cpp#L782
- https://github.com/rstudio/rstudio/blob/d3f94d47d30c6580e4224327345fc260e9e48175/src/cpp/core/LoggingTests.cpp#L842
- https://github.com/rstudio/rstudio/blob/d3f94d47d30c6580e4224327345fc260e9e48175/src/cpp/core/system/Win32SystemTests.cpp#L260
- https://github.com/rstudio/rstudio/blob/d3f94d47d30c6580e4224327345fc260e9e48175/src/cpp/core/system/Win32SystemTests.cpp#L263

Example build failure output:
```
[2023-05-17T15:34:08.178Z] -------------------------------------------------------------------------------
[2023-05-17T15:34:08.178Z] Logging
[2023-05-17T15:34:08.178Z]   Can json log error
[2023-05-17T15:34:08.178Z] -------------------------------------------------------------------------------
[2023-05-17T15:34:08.178Z] C:\Users\jenkins\workspace\ide-os-windows\main\src\cpp\core\LoggingTests.cpp(675)
[2023-05-17T15:34:08.178Z] ...............................................................................
[2023-05-17T15:34:08.178Z] 
[2023-05-17T15:34:08.178Z] C:\Users\jenkins\workspace\ide-os-windows\main\src\cpp\core\LoggingTests.cpp(721): FAILED:
[2023-05-17T15:34:08.178Z]   REQUIRE( message == "No such file or directory" )
[2023-05-17T15:34:08.178Z] with expansion:
[2023-05-17T15:34:08.178Z]   "The system cannot find the file specified"
[2023-05-17T15:34:08.178Z]   ==
[...]

PR #13169, opened 2023-05-24
Disable failing windows tests and fix memory access violation

cm421:
### Intent

These tests have been failing silently for over a year in some cases. They're now suddenly reporting failure, holding up Windows builds. We should fix them, and an issue to do so is here: https://github.com/rstudio/rstudio/issues/13165

However, since these have been failing anyways for so long, and this is holding up certification, I'm disabling them now.

Additionally, Sharon provided a fix for a failing test causing the script to return `-1073741819 ` which translates to `C0000005` = `STATUS_ACCESS_VIOLATION`

### Checklist


AndrewMcClain (approved):
LGTM

I question the value of asserting on logging messages outside our control, but that doesn't stop this from being ready to merge in

PR #13198, opened 2023-06-02
Fixed build problem for mac with gid types (and merging into OS)

jeffvroom:
Reviewed in https://github.com/rstudio/rstudio-pro/pull/4700 

although this also fixes a build problem on the mac from that branch.

### DIFF
--- a/src/cpp/shared_core/system/PosixSystem.cpp
+++ b/src/cpp/shared_core/system/PosixSystem.cpp
@@ -43,13 +43,24 @@ namespace posix {
 
 namespace {
 
+// Apple uses int for gid_t whereas unix uses uint32
+#ifndef __APPLE__
+   typedef gid_t GIDTYPE;
+#else
+   typedef int GIDTYPE;
+#endif
+
+// Does root ever have more than one group?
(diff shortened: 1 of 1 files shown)


src/cpp/shared_core/system/PosixSystem.cpp
```diff
@@ -43,13 +43,24 @@ namespace posix {
 
 namespace {
 
+// Apple uses int for gid_t whereas unix uses uint32
+#ifndef __APPLE__
+   typedef gid_t GIDTYPE;
+#else
+   typedef int GIDTYPE;
```
kevinushey, line 50: I think you could also bring in `<unistd.h>` to get `gid_t`, which woudl be slightly cleaner? I believe we do this in other source files using `gid_t`.

kevinushey: Ah, but that doesn't help because ::getgrouplist takes an `int*`, as per https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/getgid.2.html.

Issue #12972, opened 2023-04-10
Menu commands not disabled when a modal dialog is displayed (Electron desktop)


gtritchie:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2023.03.0-386
    OS Version      : any
    R Version       : any

### Steps to reproduce the problem

- In RStudio desktop (electron), show a modal dialog such as Tools / Global Options
- Now open one of the main menus and examine the state of commands

### Describe the problem in detail

- The commands are still enabled, and you can trigger at least some of them

<img width="1407" alt="Screenshot 2023-04-10 at 10 10 37" src="https://user-images.githubusercontent.com/10569626/230943088-39847aa5-6630-4703-a9c5-8386a5499597.png">

- Compare with RStudio Qt (e.g. 2022.07); on Windows and Linux (at least desktops where the main menu is displayed in the app window) the entire menubar is disabled; on Mac, the menus can still be opened but all the commands are disabled

<img width="1198" alt="Screenshot 2023-04-10 at 09 58 59" src="https://user-images.githubusercontent.com/10569626/230942306-621dc829-70d4-4e0a-b09e-f1b49f1ef103.png">

### Describe the behavior you expected

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

gtritchie:
Note to self: https://github.com/rstudio/rstudio/pull/4790

sharon-wang:
While looking into this, I was wondering why there was some weirdness with the Window/Fenêtre menu bar item (shows up in placeholder menu, but doesn't show up in the regular menu), but I just saw https://github.com/rstudio/rstudio/issues/12972 which explains the situation!

timtmok:
Validated in 2023.08.0-daily+32.pro6

I don't think this functionality can be automated since it requires interaction with the menu (native).

timtmok:
I'm reopening since Linux shows a menubar now in the license dialog.



[...]

PR #13199, opened 2023-06-02
Disable main menu bar when modals are open

sharon-wang:
### Intent

Addresses #12972 

#### demo
https://github.com/rstudio/rstudio/assets/25834218/fd80ae6b-e5ef-4913-a59d-fb97a9bfb823

### Approach
If a modal is opened, we save the current menu bar and stub the menu bar with a placeholder (it's the same placeholder menu that's used during app startup before the IDE is ready). If any additional (nested or otherwise) modals are opened, we do nothing, since the menu bar is already stubbed. Once all modals are closed, the saved menu bar is restored.

Adds open modal tracking to the app state, so that both GWT and non-GWT modals are considered when restoring the menu bar (only restore the menu bar if all modals are closed).

### Automated Tests
none

#### Test Scenarios
- GWT modals
    - single modal `Tools > Global Options...`
    - nested modals `Tools > Global Options...` (modal 1), then `Packages > Add...` (modal 2)
- non-GWT modals
    - open choose R or license dialog modals
- electron native modals
    - Check for Updates modal
- combination of GWT, non-GWT and electron native modals open
    - I'm not sure how a user can end up in a state where GWT, non-GWT and native modals are open, but I can provide a custom build to test this if needed (I tested these changes locally with a custom build to check that the modal tracking works)

For each test case, the menu bar should be disabled when the modals are open and should be restored to its pre-modal-opening state after all open modals are closed.

### Documentation
N/A

### Checklist


src/node/desktop/src/main/gwt-callback.ts
```diff
@@ -114,10 +114,8 @@ export class GwtCallback extends EventEmitter {
             }),
           ),
         ].sort((a, b) => a.localeCompare(b));
-        this.proportionalFonts = [
-          ...new Set<string>(
-            findFontsSync({ monospace: false }).map((fd) => 
-              fd.family))].sort((a, b) => a.localeCompare(b),
+        this.proportionalFonts = [...new Set<string>(findFontsSync({ monospace: false }).map((fd) => fd.family))].sort(
+          (a, b) => a.localeCompare(b),
```
sharon-wang, line 118: just a note: this was auto-applied by linting/formatting

src/gwt/src/org/rstudio/core/client/widget/ModalDialogTracker.java
```diff
@@ -44,8 +48,14 @@ public static boolean isTopMost(PopupPanel panel)
    public static void onHide(PopupPanel panel)
    {
       dialogStack_.removeIf(panel::equals);
-      if (Desktop.hasDesktopFrame() && numModalsShowing() == 0)
-         DesktopMenuCallback.setMainMenuEnabled(true);
+      
+      if (Desktop.hasDesktopFrame()) {
+         int numModals = numModalsShowing();
+         if (BrowseCap.isElectron())
```
timtmok, line 54: I suppose this check is just to be safe since the menu would be disabled when a modal is displayed. That would make it impossible to show a GWT modal and an Electron modal at the same time.

timtmok: I also checked using Cmd+Shift+P to try and open another dialog but disabling the menu is probably preventing the quick access from executing the command.

gtritchie (approved):
LGTM!

src/gwt/src/org/rstudio/core/client/widget/ModalDialogTracker.java
sharon-wang: Yeah I'm not sure there are ways for the user to directly open another dialog while the menu is disabled, but I think it could be possible for a dialog to be opened:
- if a user does some things in the background which impacts the IDE state
- if the IDE has some timed background tasks that cause dialogs to open due to an error state

Your comment also makes me realize I haven't handled Electron dialogs (only GWT dialogs and dialogs based on `ModalDialog` typescript class)...I'd need to maybe wrap our calls to `dialog.<showMessage/etc>` and detect when those dialogs are closed

sharon-wang: Hmm I'll need to modify `src/gwt/src/org/rstudio/core/client/widget/ModalDialogBase.java` so that those modals are tracked as well.

So, what I currently have in this PR only covers non-native modal dialogs. Native dialogs currently don't disable the menu bar. I just checked a QT build and native dialogs _do_ disable the menu bar.

I'll look further into this.

sharon-wang: I've added some handling for electron modals now. I'm not super stoked about the verbosity, but Electron doesn't have a dialog event that can be listened to (eg. show/hide events), so I've opted to create wrappers around the `dialog.<dialog function>` calls which enable/disable the menu bar.

timtmok (approved):

PR #13203, opened 2023-06-05
Additional source of user startup files for PWB

jgutman:
### Intent

Addresses [rstudio-pro#4584](https://github.com/rstudio/rstudio-pro/issues/4584)
Follow-up to [PR #9205](https://github.com/rstudio/rstudio/pull/9205)

### DIFF
--- a/src/cpp/session/resources/terminal/bash/.bash_profile
+++ b/src/cpp/session/resources/terminal/bash/.bash_profile
@@ -5,12 +5,15 @@ _FAKEHOME="${HOME}"
 HOME="${_REALHOME}"
 
 # source the user startup file, if any
+STARTUP_FILE="/etc/profile.d/*.sh"
 if [ -f ~/.bash_profile ]; then
 	source ~/.bash_profile
 elif [ -f ~/.bash_login ]; then
 	source ~/.bash_login
 elif [ -f ~/.profile ]; then
 	source ~/.profile
+elif ls $STARTUP_FILE 1> /dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


src/cpp/session/resources/terminal/bash/.bash_profile
```diff
@@ -5,12 +5,15 @@ _FAKEHOME="${HOME}"
...
 
 # source the user startup file, if any
+STARTUP_FILE="/etc/profile.d/*.sh"
 if [ -f ~/.bash_profile ]; then
 	source ~/.bash_profile
 elif [ -f ~/.bash_login ]; then
 	source ~/.bash_login
 elif [ -f ~/.profile ]; then
 	source ~/.profile
+elif ls $STARTUP_FILE 1> /dev/null 2>&1; then
```
kevinushey, line 15: Iterating over the output of `ls` is dangerous; instead prefer using something like:

```
for PROFILE in /etc/profile.d/*.sh; do
  if [ -e "${PROFILE}" ]; then
    source "${PROFILE}"
  fi
done
```

See also: https://superuser.com/a/418490/317257

jgutman: Good point, thanks! Any particular reason to use `-e` over `-f` here?

kevinushey: Good point -- this should probably be `-f` since you can't source a directory. 🙂 Thanks!

kevinushey (approved):

PR #13224, opened 2023-06-09
Avoid crash for empty input string in base64 decode

jeffvroom:
### Intent

Addresses:  https://github.com/rstudio/rstudio-pro/issues/4757

### DIFF
--- a/src/cpp/shared_core/system/Crypto.cpp
+++ b/src/cpp/shared_core/system/Crypto.cpp
@@ -244,6 +244,10 @@ Error base64Encode(const unsigned char* in_data, int in_length, std::string& out
 
 Error base64Decode(const std::string& in_data, std::vector<unsigned char>& out_decoded)
 {
+   // this crashes if passed an empty string so avoid that altogether
+   if (in_data.length() == 0)
+      return Success();
+
    // allocate b64 BIO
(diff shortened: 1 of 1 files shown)


src/cpp/shared_core/system/Crypto.cpp
```diff
@@ -244,6 +244,10 @@ Error base64Encode(const unsigned char* in_data, int in_length, std::string& out
 
 Error base64Decode(const std::string& in_data, std::vector<unsigned char>& out_decoded)
 {
+   // this crashes if passed an empty string so avoid that altogether
+   if (in_data.length() == 0)
+      return Success();
```
cm421, line 249: I wasn't sure if returning success here was the right thing to do at first, but after searching around it looks like `base64Decode` of an empty string is, in fact, an empty string. (Some test examples I found: https://www.rfc-editor.org/rfc/rfc4648#section-10)

cm421 (approved):
LGTM

src/cpp/shared_core/system/Crypto.cpp
AndrewMcClain: Thanks for this extra context!

AndrewMcClain (approved):
LGTM

PR #13233, opened 2023-06-13
Increase read buffer size from 512 to 8192

jeffvroom:
### Intent

Addresses: https://github.com/rstudio/rstudio-pro/issues/4764

### Approach

Prepare is the way to tell the streambuf how much to read from the socket by default. Is 8K the right value?  I might have chosen larger but noticed nginx uses 4K. 

### Automated Tests

Covered by automation for regressions.  

### QA Notes

Would be nice to compare download latency for > 100K downloads. You can do this with developer tools on the IDE or VS code and just look for the largest static js files downloaded. 

Use strace on rserver or other processes to check the value of the size parameter to recvfrom.

### DIFF
--- a/src/cpp/core/include/core/http/AsyncClient.hpp
+++ b/src/cpp/core/include/core/http/AsyncClient.hpp
@@ -101,6 +101,8 @@ public:
         closed_(false),
         requestWritten_(false)
    {
+      // Make sure we read at least 8192 bytes from the socket at a time. The default ends up as 512.
+      responseBuffer_.prepare(8192);
    }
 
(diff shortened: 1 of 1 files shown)


kfeinauer (approved):
This LGTM. I noticed we also use 8192 in `AsyncConnection`, so that seems fine as well.

I would be doubtful that this would fix the customer issue you saw, but I think your analysis of the potential root causes (busy rserver threads or slow-sending rsession) are likely culprits.

zachhannum (approved):
LGTM! I see a unit test failure but it's unrelated (GWT)

PR #13238, opened 2023-06-13
wrap `cacheKey` usages with zero length check

sharon-wang:
### Intent

Addresses #13188.

avoids hitting the following error:
```
Error in exists(cacheKey, where = .rs.WorkingDataEnv, inherits = FALSE) : 
  invalid first argument
Error in assign(cacheKey, frame, .rs.CachedDataEnv) : 
  attempt to use zero-length variable name
```

Backport issue: https://github.com/rstudio/rstudio/issues/13241

### Approach

Wraps usages of `cacheKey` in `SessionDataViewer.R` with a zero-length check.

In particular, the usage in `assignCachedData` was problematic (causing #13188).

### Automated Tests

none

### QA Notes

[Simplest repro steps](https://github.com/rstudio/rstudio/issues/13188#issuecomment-1589910695):
1. create a dataframe (any # rows/columns) eg. `df <- data.frame(matrix(rnorm(1), nrow=1, ncol=1))`
2. trigger autocomplete: type `df` + tab key
3. try to reassign the same variable eg. `df <- 'hi'`
4. see error

### Documentation
none

### Checklist

### DIFF
--- a/src/cpp/session/modules/SessionDataViewer.R
+++ b/src/cpp/session/modules/SessionDataViewer.R
@@ -584,7 +584,8 @@
       }
    }
    
-   if (.rs.isNonEmptyScalarString(cacheKey)) {
+   if (.rs.isNonEmptyScalarString(cacheKey))
+   {
       # if the object exists in the cache environment, return it. objects
       # in the cache environment have already been coerced to data frames.
       if (exists(cacheKey, where = .rs.CachedDataEnv, inherits = FALSE))
@@ -603,6 +604,7 @@
             return(get(cacheKey, envir = .rs.CachedDataEnv, inherits = FALSE))
       }
    }
+
    # failure
    return(NULL)
 })
@@ -823,7 +825,8 @@
    if (Encoding(cacheDir) == "unknown")
       Encoding(cacheDir) <- "UTF-8"
    
-   if (.rs.isNonEmptyScalarString(cacheKey)) {
+   if (.rs.isNonEmptyScalarString(cacheKey))
+   {
       # remove data from the cache environment
       if (exists(".rs.CachedDataEnv") &&
          exists(cacheKey, where = .rs.CachedDataEnv, inherits = FALSE))
--- a/src/cpp/session/modules/SessionDataViewer.R
+++ b/src/cpp/session/modules/SessionDataViewer.R
@@ -584,7 +584,7 @@
       }
    }
    
-   if (!.rs.isZeroLength(cacheKey)) {
+   if (.rs.isNonEmptyScalarString(cacheKey)) {
       # if the object exists in the cache environment, return it. objects
       # in the cache environment have already been coerced to data frames.
       if (exists(cacheKey, where = .rs.CachedDataEnv, inherits = FALSE))
@@ -753,7 +753,7 @@
    # save a copy into the cached environment
    cacheKey <- .rs.addCachedData(force(x), name)
    
-   if (.rs.isZeroLength(cacheKey))
+   if (!.rs.isNonEmptyScalarString(cacheKey))
       return(invisible(NULL))
 
    # call viewData 
@@ -786,7 +786,7 @@
 
 .rs.addFunction("viewDataFrame", function(x, title, preview) {
    cacheKey <- .rs.addCachedData(force(x), "")
-   if (!.rs.isZeroLength(cacheKey))
+   if (.rs.isNonEmptyScalarString(cacheKey))
       invisible(.Call("rs_viewData", x, "", title, "", emptyenv(), cacheKey, preview))
 })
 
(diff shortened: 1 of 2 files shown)


src/cpp/session/modules/SessionDataViewer.R
```diff
@@ -890,3 +901,8 @@
    invisible("")
 })
 
+.rs.addFunction("isZeroLength", function(x)
```
kevinushey, line 904: This doesn't feel quite right to me, since all R objects have a length and e.g.

```
> length(NA)
[1] 1
```

Could we instead reverse the check, and just confirm something like:

```
.rs.addFunction("isNonemptyScalarString", function(x)
{
  is.character(x) && length(x) == 1 && nzchar(x)
})
```

since that's more-or-less what `get()` and `assign()` require in their usages?

kevinushey (commented):
Approach LGTM but I'd recommend an alternate function name / implementation.

We should also file an issue to remind us to backport this fix.

src/cpp/session/modules/SessionDataViewer.R
sharon-wang: Thanks for this suggestion! My lack of R knowledge is showing 😅 

TIL all objects have a length. I think this is going to be a helpful fact for me to carry forward!

Makes more sense to check for non-empty string and this covers our bases if something like `NA` or `NULL` is passed.

src/cpp/session/modules/SessionDataViewer.R
```diff
@@ -584,24 +584,27 @@
...
+         status <- try(load(cacheFile, envir = .rs.CachedDataEnv), silent = TRUE)
+         if (inherits(status, "try-error"))
+            return(NULL)
+         
+         if (exists(cacheKey, where = .rs.CachedDataEnv, inherits = FALSE))
+            return(get(cacheKey, envir = .rs.CachedDataEnv, inherits = FALSE))
+      }
    }
-   
+
```
sharon-wang, line 607: this diff is confusing -- the existing code was just wrapped with `if (.rs.isNonEmptyScalarString(cacheKey))`

sharon-wang: This might be easier to read (left = `main`, right = this PR)

image

sharon-wang:
Created backport issue: https://github.com/rstudio/rstudio/issues/13241

kevinushey (approved):
LGTM!

src/cpp/session/modules/SessionDataViewer.R
kevinushey: For what it's worth, the diff renders clearly for me if I use "hide whitespace" in the file view, e.g. https://github.com/rstudio/rstudio/pull/13238/files?w=1. 

Screenshot 2023-06-14 at 10 35 41 AM

sharon-wang: ah yes that's so much better 🙌

sharon-wang:
Just a note: CI build failure is the same test failure as on `main` (not a new failure)

```
DummyCommands.java:6: error: DummyCommands is not abstract and does not override abstract method copilotStatus() in Commands

[2023-06-14T15:50:41.877Z]     [javac] public class DummyCommands extends Commands {

[2023-06-14T15:50:41.877Z]     [javac]        ^

[2023-06-14T15:50:41.877Z]     [javac] 1 error
```

src/cpp/session/modules/SessionDataViewer.R
```diff
@@ -890,3 +904,8 @@
    invisible("")
 })
 
+.rs.addFunction("isNonEmptyScalarString", function(x)
+{
+   is.character(x) && length(x) == 1 && nzchar(x)
```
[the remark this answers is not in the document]
sharon-wang: Kevin enlightened me that all objects have length 1, so even `length('hello') == 1`.

So `length(x) == 1` in this case just checks that it isn't `NULL`, because `length(NULL) == 0`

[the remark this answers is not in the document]
kevinushey: `cacheKey` should be a character vector of length one, whose string is non-empty. So, a normal string like `"hello"`, but not `""` or `c("a", "b")`.

kevinushey:
> Just a note: CI build failure is the same test failure as on `main` (not a new failure)
> 
> ```
> DummyCommands.java:6: error: DummyCommands is not abstract and does not override abstract method copilotStatus() in Commands
> 
> [2023-06-14T15:50:41.877Z]     [javac] public class DummyCommands extends Commands {
> 
> [2023-06-14T15:50:41.877Z]     [javac]        ^
> 
> [2023-06-14T15:50:41.877Z]     [javac] 1 error
> ```

Thanks for catching that; that should hopefully be fixed up on main with https://github.com/rstudio/rstudio/commit/930ebfadbab961276ee339d50aebf954050d27fb.

PR #13262, opened 2023-06-19
Hoping to avoid deparse by not calling length on language objects

jeffvroom:
Attempt to fix rsession-tests that are crashing


sharon-wang (approved):
lgtm!🤞

Issue #13244, opened 2023-06-15
Backport #13220 to Mountain Hydrangea 2023.06.1


ronblum:
Backport #13220 to Mountain Hydrangea 2023.06.1.

**Note: This issue is to record the need to backport the linked issue only. Please add comments or discussion relating to the original issue onto it directly.**

```[tasklist]
### PRs
- [x] backport https://github.com/rstudio/rstudio/pull/13270
- [x] backport https://github.com/rstudio/rstudio/pull/13275
```

Issue #13235, opened 2023-06-13
Docs Describe navigating Data Viewer with more than 50 columns


ronblum:
Add something like the following and a screenshot to [the data viewer page in the RStudio User Guide](https://docs.posit.co/ide/user/ide/guide/data/data-viewer.html) to describe what to do when there are more than 50 columns:

> When the data frame you are viewing has more than 50 columns, you will be limited to 50 in the Viewer. To view others columns use the "Cols" arrow options to shift left or right (with image showing where to find this) (edited)

ronblum:
Verified in the 2023.09.0 RStudio User Guide revision 22 (7/31/2023) at https://docs.posit.co/ide/user/daily/ide/guide/data/data-viewer.html#scrolling

Issue #13220, opened 2023-06-09
RStudio Dataframe Column Navigation Limited to 50 Columns after Updating


BradyHorton:
### System details

    RStudio Edition :  Desktop
    RStudio Version :  RStudio 2023.06.0+421 "Mountain Hydrangea" 
    OS Version      :  Windows 11
    R Version       :  4.3

### Steps to reproduce the problem

Create a data frame with more than 50 columns and attempt to navigate with the "Cols" arrows after viewing the data frame.

### Describe the problem in detail

I am currently working on a project and analyzing a dataset that has more than 50 columns of data, and after updating to the latest version of RStudio today, I can't seem to navigate to the later columns in the data frame. I should also note that before updating this morning, the column navigation worked just fine. In other words, when viewing the data, the column navigation arrows do not seem to be working. Here is what I am referring to in RStudio:

![image](https://github.com/rstudio/rstudio/assets/135742177/099991eb-317c-410f-a0da-6af2d40354e3)

I also tried manually updating the column numbers, but even though it displayed the correct numbers, only the lower bound was updated and I could not access the data in the later columns (Beyond column 50).

### Describe the behavior you expected

Before updating to the latest version of RStudio today, the column navigation worked like normal.  I simply expected the column navigation to work as it previously has in older versions of RStudio.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

lbenz730:
I had the same issue with Rstudio version RStudio 2023.06.0+421 on R4.1.0 and had to revert to previous version of the IDE because of this issue.

ronblum:
Thank you for raising this! Confirmed that this is failing in
- RStudio Desktop 2023.06.0+421 on MacOS .13.4 
- RStudio Desktop 2023.06.0+421 on Ubuntu 22.04
- RStudio Server 2023.06.0+421 on Ubuntu 20.04 via Chrome on MacOS 13.4

Triage: We should patch this.

JMSteinberg:
[...]

PR #13270, opened 2023-06-21
fix data viewer column navigation bug

sharon-wang:
### Intent

Address #13220 

This will be backported to MH as part of https://github.com/rstudio/rstudio/issues/13244.

### Approach

- only retrieve data for the visible slice of column data, not the entire data frame (helps us avoid slow loading for large data sets)
    - search for `getColSlice`, `describeColSlice`
- separate the usage of `max_cols` and `max_display_columns` to avoid the confusing mixed usage
    - `max_cols`/`maxCols` is for specifying the maximum number of columns we want to retrieve from a data frame
        - e.g. sometimes we only want the first 100 columns, to avoid slow loading with large data sets
    - `max_display_columns`/`maxDisplayColumns` is the maximum number of columns to display in a view of the data frame (in autocomplete/help or in the data viewer) and the user preference `data_viewer_max_columns`/`dataViewerMaxColumns` sets the default for this value
        - e.g. the default is viewing 50 columns at a time
- return the total number of columns `total_cols` of the data frame as part of the `rownames` column data
    - search for `totalCols`
    - this is a bit awkward because the `rownames` column is meant to hold data specific to that column, but I wanted to avoid changing the response structure/adding a new endpoint to request the total columns. Perhaps we can rename that column type to be something like `metadata`?

It's possible we no longer need `subsetData` and could refactor that out, but I didn't want to complicate this PR too much!

### Automated Tests

- Added an R test for the new `describeColSlice` function
- Added an id to the column range input element for automated test use -- id = `data-viewer-column-input`

### QA Notes

As noted by Kevin below:
> This has more changes than we would typically want to include in a patch release, so extra scrutiny is warranted in verification.

#### Suggested Test Cases
- autocomplete preview displays the correct data
```
# set a variable to a data frame
df <- data.frame(matrix(rnorm(300), nrow=5, ncol=60))

# type the variable name and hit the tab key to trigger the autocomplete preview
df
```
- test with a large dataset such as https://github.com/rstudio/rstudio/issues/12964#issuecomment-1561916287
    - the data preview should load pretty quickly (shouldn't be longer than 1s or so)
- test reassigning an already viewed dataframe to a different dataframe

```
df1 <- data.frame(matrix(rnorm(300), nrow=5, ncol=60))

df2 <- data.frame(matrix(1, nrow=1, ncol=100))

View(df1)

# After `View`ing, run the below line to reassign `df1` a different data frame value
# the open data viewer of df1 should refresh with the new data, since df1 was assigned the value of df2
df1 <- df2
```
- change the `data_viewer_max_columns` user preference and verify that the data viewer displays the right number of columns
    - note that the autocomplete data viewer has a default maximum display of 100 columns and should not be affected by this user pref

### Documentation
Related documentation issue https://github.com/rstudio/rstudio/issues/13235

### Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,7 @@
 ### Fixed
 - Fixed issue where Electron menubar commands are not disabled when modals are displayed (#12972)
 - Fixed bug causing invalid/empty `cacheKey` error when accessing a dataframe variable (#13188)
+- Fixed bug preventing dataframe column navigation past the default number of display columns (#13220)
 
 ### Accessibility Improvements
 -
--- a/src/cpp/session/modules/SessionDataViewer.R
+++ b/src/cpp/session/modules/SessionDataViewer.R
@@ -135,7 +135,9 @@
    if (!is.character(colLabels)) 
       colLabels <- character()
    
-   totalCols <- if (totalCols < 0) ncol(x) else totalCols
+   # we pass totalCols in the rownames col so we can pass this information
+   # along when we retrieve column data, without changing the response format
+   totalCols <- if (totalCols > 0) totalCols else ncol(x)
 
    # the first column is always the row names
    rowNameCol <- list(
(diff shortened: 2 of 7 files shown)


kevinushey (approved):
Fantastic work; this LGTM!

This has more changes than we would typically want to include in a patch release, so I think extra scrutiny is warranted in verification.

One other thing worth considering in this PR: should we respect the "max display columns" preference in the data preview presented in the Help popup?

https://github.com/rstudio/rstudio/blob/4f7bab22cddc3ff12bc2abbc4bb43c90366a9adc/src/cpp/session/modules/SessionHelp.R#L592

(maybe we're lucky and it's as simple as removing the `max_cols` item there?)

sharon-wang:
Thanks for the reviews!

> This has more changes than we would typically want to include in a patch release, so I think extra scrutiny is warranted in verification.

💯 

> One other thing worth considering in this PR: should we respect the "max display columns" preference in the data preview presented in the Help popup?

Makes sense to me! I retrieved the user pref and passed it as `max_cols` so we can limit our column retrieval to the first `max_display_columns` number of columns (removing `max_cols` defaults it to `-1` server-side, which results in all of the columns being loaded)

PR #13276, opened 2023-06-22
Fix terminal symlink test on Windows

timtmok:
### Intent
Fix Windows terminal symlink test 

### Approach
Add a symlink path for Windows

### Automated Tests
n/a

### QA Notes
n/a 

### Documentation
n/a

### Checklist


kevinushey:
FWIW Windows only supports symlinks in certain scenarios, and isn't usually enabled by default. This is discussed a bit in https://learn.microsoft.com/en-us/windows/security/threat-protection/security-policy-settings/create-symbolic-links.

That is to say; we should probably just disable any symlink-related tests (and avoid trying to use it, if we are) on Windows?

PR #13279, opened 2023-06-23
Guard against missing edge case

MichaelChirico:
I encountered this somewhat puzzlingly just now:

```
Error in exists(cacheKey, where = .rs.CachedDataEnv, inherits = FALSE) : 
  invalid first argument
Error in if (maxRows != -1 && nrow(data) > maxRows) data <- head(data,  : 
```

This PR addresses the latter error. It could conceivably be obviated by debugging the first error, but I'm clueless on that for now and this PR seems benign enough.

RStudio version info:

```
RStudio
2023.06.0 Build 421
© 2009-2023 Posit Software, PBC
"Mountain Hydrangea" Release (583b465e, 2023-06-06) for Ubuntu Focal
Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) rstudio/2023.06.0+421 Chrome/110.0.5481.208 Electron/23.3.0 Safari/537.36
```

### DIFF
--- a/src/cpp/session/modules/SessionDataViewer.R
+++ b/src/cpp/session/modules/SessionDataViewer.R
@@ -23,10 +23,10 @@
 
 .rs.addFunction("subsetData", function(data, maxRows = -1, maxCols = -1)
 {
-   if (maxRows != -1 && nrow(data) > maxRows)
+   if (!is.na(maxRows) && maxRows != -1 && nrow(data) > maxRows)
       data <- head(data, n = maxRows)
    
-   if (maxCols != -1 && ncol(data) > maxCols)
+   if (!is.na(maxCols) && maxCols != -1 && ncol(data) > maxCols)
       data <- data[1:maxCols]
    
    data


kevinushey:
Seems more likely that `nrow(x)` or `ncol(x)` might be returning `NA`? Isn't that the case for "remote" data frames; e.g. if you're using `dplyr` to connect to a database or something?

MichaelChirico:
You're right that `nrow()` can return NA, though I'm not doing any remote connection in my section.

How about changing to `isTRUE(maxRows != -1 && nrow(data) > maxRows)`, then?

sharon-wang:
@MichaelChirico This error looks to be the same issue reported in https://github.com/rstudio/rstudio/issues/13188, which will be fixed in our upcoming patch.

In the meantime, you can try out the fix in our daily builds ([2023.06](https://dailies.rstudio.com/rstudio/mountain-hydrangea/) or [2023.08](https://dailies.rstudio.com/)), but please note these builds are for testing purposes and are not recommended for production use.

Just wanted to chime in with some test build info! We may still want to guard against situations where the `data` getting passed in may not be assigned as expected.

MichaelChirico:
Thanks Sharon! This PR definitely reflects a deeper issue, glad there's already work to fix it! My hope is with safer code like this it would make issues like this easier to debug

kevinushey (approved):
LGTM!

Issue #8932, opened 2021-02-09
Hide Menu Bar


392781:
Allow users to hide the jarring unthemed menu bar.  Nearly every editor out there has this option, RStudio should too.  As an example, VS Code allows for convenient unhiding by pressing "Alt" key.

How it would look in RStudio:

**Before:**
![image](https://user-images.githubusercontent.com/25296775/107336746-5500e180-6a6e-11eb-94e9-c3b76ac33f5c.png)

**After:**

![image](https://user-images.githubusercontent.com/25296775/107337147-d2c4ed00-6a6e-11eb-81cf-64cee20a517e.png)

My personal version information:
```
RStudio Edition : Desktop
RStudio Version : 1.4.1103
R Version       : 4.0.3
OS Version      : Linux Mint 20
```

ronblum:
@392781 Thank you for the suggestion! We'll review it as we continue development of RStudio. By the way, what version of RStudio and OS are you using?

392781:
@ronblum Added it to original issue request.  Also this request is tangentially related to #3243.

gtritchie:
We should get this very inexpensively with Electron, see:

```
autoHideMenuBar Boolean (optional) - Auto hide the menu bar unless the Alt key is pressed. Default is false.
```

https://www.electronjs.org/docs/all#new-browserwindowoptions

gtritchie:
For Electron, on Linux and Windows (but not Mac):

- add a checkbox to Global Options / General / Other: "Hide menu bar until Alt-key pressed"
- default should be unchecked (menu always visible, current behaviour)
- the value of this setting is used to set the `autoHideMenuBar` flag when constructing the main-window
- changing the value in Global Option requires a restart to take effect

gtritchie:
Moving this out of the SW Electron milestone. This is a new feature so not critical for the preview release. There's some WIP on the feature/autohide-electron-menubar branch. Problem needing further investigation is why tapping Alt key isn't hiding/showing menu, presumably we're eating that key sequence in the GWT code.

gtritchie:
Nice to have; moving out of Elsbeth Geranium.

timtmok:
I did add my fix to this branch: https://github.com/rstudio/rstudio/tree/feature/autohide-electron-menubar

It seemed to work for me but it didn't work for Gary. The branch is probably quite stale now. I'll target this for next release to see if there's time to figure out what's going on.

ronblum:
[...]

PR #13280, opened 2023-06-23
use `setMenuBarVisibility` instead of `setMenu`

sharon-wang:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/12972#issuecomment-1591996748

### Approach

`setMenu(null)` seems to trigger the placeholder menu to show on electron modal dialogs, however using `setMenuBarVisibility(false)` hides the menu bar.

This only affects linux and windows.

### Automated Tests
none

### QA Notes

This change should ensure that the License Management modal doesn't have a menu bar on Windows, Linux and Mac and the Choose R dialog doesn't have a menu bar on Windows.

### Documentation
none

### Checklist

### DIFF
--- a/src/node/desktop/src/ui/modal-dialog.ts
+++ b/src/node/desktop/src/ui/modal-dialog.ts
@@ -61,7 +61,7 @@ export abstract class ModalDialog<T> extends BrowserWindow {
     });
 
     // make this look and behave like a modal
-    this.setMenu(null);
+    this.setMenuBarVisibility(false);
     this.setMinimizable(false);
     this.setMaximizable(false);
     this.setFullScreenable(false);


timtmok (commented):
I wonder if the menu disablement is only necessary for MacOS. I suppose a Linux user might have a window manager with a global menu like MacOS.

I'm not sure of a way around this since `setMenuBarVisibility(false)` still allows showing the menu when pressing `Alt`.

sharon-wang:
> I wonder if the menu disablement is only necessary for MacOS. I suppose a Linux user might have a window manager with a global menu like MacOS.
> 
> I'm not sure of a way around this since `setMenuBarVisibility(false)` still allows showing the menu when pressing `Alt`.

For the `BrowserWindow`-based modals (Choose R + License Management), Windows and Linux handle the modal correctly -- when the modal is open, interaction with RStudio (including the main window menu bar) is disabled.

However, GWT-based and Electron-native-based modals don't disable interaction with RStudio by default on Windows + Linux + Mac, so the modal tracking introduced in https://github.com/rstudio/rstudio/pull/13199 helps us enable/disable the menu bar across the different types of modals we use.

Following that change, `setMenu(null)` now causes a menu bar to show inside the License Management modal, but replacing that with `setMenuBarVisibility(false)` resolves the issue. As long as we don't set the menu bar to "auto-hide" (we don't set that in our modal-dialog constructor), then `Alt` won't have an effect on the menu bar.

As you pointed out offline, we have some menu bar auto-hide WIP (https://github.com/rstudio/rstudio/issues/8932), so we'll have to make sure that the auto-hiding implemented there only affects regular IDE windows, and not the `BrowserWindow`-based modal windows.

timtmok (approved):

Issue #13105, opened 2023-05-09
Error Saving File on Windows when `HOME` env var set with trailing slash


sharon-wang:
### System details

    RStudio Edition : 
    RStudio Version : 2023.03.0 Build 386, 2023.03.1 Build 446, 2023.05.0 Build 329
    OS Version      : Windows 10 x64
    R Version       : 4.2.2

### Steps to reproduce the problem
1. Go to environment variables and create new User variable `HOME` which is set to `C:\Users\<username>\` (note trailing backslash)
2. Create new file in RStudio (I created a new R Script)
3. Save the file and navigate in file browser to `C:\Users\<username>\Desktop` , then save the file
4. See error message
<img width="311" alt="image" src="https://github.com/rstudio/rstudio/assets/25834218/05ca102c-60a3-48d5-89a5-12a866bb81d0">

Full error message in session log:

```
2023-05-09T14:21:56.697764Z [rsession] ERROR system error 3 (The system cannot find the path specified) [path: C:/Users/<username>/~Desktop/a.R]; OCCURRED AT auto __cdecl rstudio::core::FilePath::openForWrite::<lambda_acfc80dd3014ccea1eb874618c499a42>::operator ()(void) const src/cpp/shared_core/FilePath.cpp:1520; LOGGED FROM: class rstudio::core::Error __cdecl rstudio::core::`anonymous-namespace'::openFileForWritingWithRetry(const class rstudio::core::FilePath &,bool,int,class std::shared_ptr<class std::basic_ostream<char,struct std::char_traits<char> > > *) src/cpp/core/FileSerializer.cpp:73
```

### Describe the problem in detail
When setting a user environment variable `HOME` to your user's home directory with a trailing backslash included, an error occurs when saving a new file to a child directory of your home directory, such as the `Desktop` directory and prevents the user from saving the file in that location.

Seems to be an issue with path aliasing.

### Describe the behavior you expected

Regardless if a trailing slash is included in the `HOME` variable or not, it should be possible to save a file to a child directory without erroring.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
~- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).~
[...]

PR #13304, opened 2023-06-29
Handle trailing slashes in user's `HOME` path

sharon-wang:
### Intent

Addresses #13105

### Approach
Remove trailing slashes when we retrieve the user's home path in Electron. In the desktop bridge, we now retrieve this trimmed version of the user's home path instead of retrieving the env var directly.

We already handle trailing slashes in SessionMain.cpp, which ensures `Sys.getenv("HOME")` doesn't contain trailing slashes:
https://github.com/rstudio/rstudio/blob/e444633ccd5c7b12fa4052868ca946e96fed2155/src/cpp/session/SessionMain.cpp#L2018-L2031

However, we were missing this equivalent handling on the desktop side.

### Automated Tests

- includes tests for various trailing slash cases
- modified an unrelated wait-utils test because it was failing during my testing on a slower machine

### QA Notes

- Although this is a fix to address an issue on Windows, the code is executed on all OS's
- when saving a file, ensure that the file is being saved at a subdirectory of the `HOME` path
    - e.g. if `HOME=C:/my/path/`, then the file may be saved in this directory `C:/my/path/hello`

### Documentation
N/A

### Checklist


src/node/desktop/src/main/utils.ts
```diff
@@ -482,15 +482,18 @@ export const handleLocaleCookies = async (window: BrowserWindow, isMainWindow =
...
+    .then(async (cookies) => {
+      if (cookies.length === 0) {
+        await updateLocaleFromCookie({ name: 'LOCALE', value: 'en' } as any, window);
+      } else {
+        cookies.forEach(async (cookie) => {
+          await updateLocaleFromCookie(cookie, window);
+        });
+      }
+    })
+    .catch((err) => logger().logError(err));
```
sharon-wang, line 496: this was auto-formatted -- no functional changes were made here

src/node/desktop/src/main/utils.ts
```diff
@@ -600,3 +603,36 @@ export function getNumericEnvVar(envVarName: string): number | undefined {
...
+const TRAILING_SLASH_REGEX = /[\\/]+$/;
+
+/**
+ * Removes trailing slashes from a string, optionally keeping the starting slash
+ * or double-slash for UNC paths.
+ * @param str The string to remove trailing slashes from
+ * @param keepStartSlash Whether to keep the starting slash if it exists
+ * @returns The string with trailing slashes removed
+ */
+export function removeTrailingSlashes(str: string, keepStartSlash = false): string {
```
kevinushey, line 616: IMHO we can remove `keepStartSlash`; we should only remove trailing slashes from strings which have some non-slash components within. From what I can see in the usages in this PR, that parameter is always set to `true`, other than in tests?

src/node/desktop/src/main/utils.ts
```diff
@@ -600,3 +603,36 @@ export function getNumericEnvVar(envVarName: string): number | undefined {
...
+  const trailingSlashes = str.match(TRAILING_SLASH_REGEX);
```
kevinushey, line 617: Could we simplify some of the logic below by detecting the "only slashes" case first? E.g.

```
if (/^[\\/]+$/.test(str)) {
  return str;
}
```

That would remove the need to detect single slashes / UNC paths below.

src/node/desktop/src/main/utils.ts
sharon-wang: yeah in our existing usages, I've set the `keepStartSlash` to true to cover cases where `HOME=/` or similar. I wasn't sure if we'd more widely use `removeTrailingSlashes()` in the future, with strings that aren't paths? So I added `keepStartSlash` so it could be more general-purpose. But maybe unnecessary, and we can keep it simple/specific to paths (including simplifying the logic as you mentioned below)

src/node/desktop/src/core/user.ts
```diff
@@ -19,21 +19,28 @@ import { getenv } from './environment';
...
-  const home = getenv('HOME');
-  if (checkPath(home)) return new FilePath(home);
+  const pathsToCheck = [() => getenv('HOME')];
   if (process.platform === 'win32') {
-    const currentHome = desktop.currentCSIDLPersonalHomePath();
-    if (checkPath(currentHome)) return new FilePath(currentHome);
-    const defaultHome = desktop.defaultCSIDLPersonalHomePath();
-    if (checkPath(defaultHome)) return new FilePath(defaultHome);
+    pathsToCheck.push(
+      () => getenv('R_USER'),
```
kevinushey, line 28: This check should become before `HOME`, as R will check that before `R_USER`. See https://github.com/wch/r-source/blob/9d27b1c9ad917f0c8fbe8f6476927a1bd8ed9d9d/src/gnuwin32/sys-win32.c#L76-L121 for how this is handled in R on Windows (we try to stay true to pre-existing behaviors in R whenever possible)

kevinushey (approved):
LGTM; only recommendation is to move the `R_USER` check so it comes first in the list of candidates (and Windows only)

src/node/desktop/src/core/user.ts
sharon-wang: thanks for pointing this out!

PR #13314, opened 2023-06-30
use renv lockfile when installing dependencies

kevinushey:
### Intent

This PR allows us to opt-in to installing packages from an [renv](https://rstudio.github.io/renv/) lockfile when available for that platform. This allows us to pin our R package dependencies when building + testing on a particular platform. Not implemented on Windows.

### Approach

Mostly straightforward wiring into our `install-packages` script.

### Automated Tests

### QA Notes

No need for QA; just need to determine if this does survive on CI.

### Documentation

N/A

### Checklist


dependencies/common/install-packages
```diff
@@ -92,25 +100,31 @@ mv $PACKAGE_ARCHIVE $PACKAGE_ARCHIVE_SHA1
...
+}
+
+lockfile <- "${lockfile:=}"
+if (nzchar(lockfile)) {
+  # if we were provided a lockfile, use it
+  writeLines(paste("Using lockfile:", lockfile))
+  renv::restore(lockfile = lockfile)
+} else {
+  # otherwise, install preset list of packages from CRAN
+  renv::install("devtools", "rmarkdown", "renv", "testthat", "xml2", "yaml")
```
kevinushey, line 124: Is this too aggressive? Maybe we should just use a regular `install.packages` call in this case, as we did before?

kevinushey: (The nice thing about using `renv` is we would get to re-use the `renv` global package cache, and it would also ensure we're using the latest versions of packages locally)

dependencies/common/lockfiles/bionic/renv.lock
```diff
@@ -0,0 +1,1521 @@
+{
+  "R": {
+    "Version": "4.3.0",
```
kevinushey, line 3: Note that the R version here is ignored during restore; it's only used to notify the user when "loading" an renv project if the version doesn't match the one in the lockfile.

dependencies/common/install-packages
timtmok: I think this is ok for builds and the cache is a plus for dev environments!

timtmok (approved):

kevinushey:
Do we have a way of testing this PR on CI before merge?

timtmok:
There were PR builds at some point but I'm not sure why they aren't running on OS PR's. In any case, I'm not even sure if it would build bionic. It might be jammy only.

cm421:
Is this ready to give a whirl in Jenkins? We technically could build this branch in Jenkins, but it gets a little messy since you want to disable binary uploading so they don't conflict with the real binaries being created on `main`. IMHO it's easier just to merge it and then tweak it as it hits problems

kevinushey:
I think this should be ready to go.

PR #13315, opened 2023-06-30
Use cached api to lookup users in several more places

jeffvroom:
Make sure all getUser passwd entry database lookup calls use the cached api 

### Intent

Addresses https://github.com/rstudio/rstudio-pro/issues/4800

- in processInfo call for looking up process owner (main issue)
- at startup time for the server-user
- session metadata rpc for authorizing metadata lookups
- for authorizing server session rpc requests
- for authorizing local stream requests authenticated by a uid

### Approach

Found the rest of the places we make these getUserFromIdentifier calls and replaced them with the cached equivalent: either getUserFromUserId or getUserFromUsername. 

### QA Notes

All of these code paths will be tested by automation. I don't think manual testing is necessary.


cm421 (approved):
LGTM!

PR #13320, opened 2023-07-03
Speedup and bugfix for rsession metadata

jeffvroom:
### Intent

Addresses part of https://github.com/rstudio/rstudio-pro/issues/4746 and 
part of https://github.com/rstudio/rstudio-pro/issues/4085

### Approach

* Don't create the session directory as part of creating the storage object as this implicitly creates a session that's not valid just by looking to see if a session exists with a given id. We don't ordinarily do that in OS but for workbench, that happens occasionally for rworkspaces and that causes a delay while that session is determined to be invalid. 

* Don't use ensureDirectory for the properties directory on each property access. This adds overhead and improperly creates a session just by trying to read a property from it. Instead, catch the "no such file or directory" error when trying to write the first property and create the directory at that point.

* Bonus - speedup of test for web-socket upgrade (don't create the regex on every request) - this showed up in the flame-graph I did of the rserver under load.

### Automated Tests

Our automation will cover all of the relevant code paths for regressions.

### QA Notes

To validate that the performance part is included, run strace against the rworkspaces process. Capture 30 seconds with a homepage open that has several sessions listed (suspended, running, or whatever) (make sure at least one workspaces/model request gets handled)

Grep for `sessions/active` in that strace and further grep for `statx` calls. You should see 2 or 3X fewer with the fix for each model refresh captured in the strace.  You can probably tell when there's a gap in the timestamp for the statx calls.

### DIFF
--- a/src/cpp/core/r_util/RActiveSessionStorage.cpp
+++ b/src/cpp/core/r_util/RActiveSessionStorage.cpp
@@ -46,9 +46,11 @@ Error createError(const std::string& errorName, const std::string& preamble,
 FileActiveSessionStorage::FileActiveSessionStorage(const FilePath& scratchPath) :
    scratchPath_ (scratchPath)
 {
-   Error error = scratchPath_.ensureDirectory();
-   if(error)
-      LOG_ERROR(error);
+   // Do not create the directory here or else any attempt to look for a session ends up creating it. The API is designed
+   // to return an empty object if it does not exist.
+   //Error error = scratchPath_.ensureDirectory();
+   //if(error)
+   //   LOG_ERROR(error);
 }
 
 const std::map<std::string, std::string> FileActiveSessionStorage::fileNames =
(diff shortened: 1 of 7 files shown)


src/cpp/core/r_util/RActiveSessionStorage.cpp
```diff
@@ -112,7 +114,9 @@ Error FileActiveSessionStorage::readProperties(std::map<std::string, std::string
    std::vector<FilePath> files{};
    std::vector<FilePath> failedFiles{};
    pValues->clear();
-   propertyDir.getChildren(files);
+   Error error = propertyDir.getChildren(files);
```
zachhannum, line 117: nit: Is this change really needed? I think we would already return `Success()` if the propertyDir is not found, since `files` will be empty.

zachhannum (approved):
LGTM! Thanks for getting this fix in.

src/cpp/core/r_util/RActiveSessionStorage.cpp
jeffvroom: Oh yeah, good point. I like the code a little better where we are testing the error just from the readability perspective.

zachhannum: You're right, that makes sense to me from a readability perspective.

PR #13334, opened 2023-07-07
Feature/mh renv dependency

cm421:
### Intent

Applies https://github.com/rstudio/rstudio/pull/13314 but onto the MH branch. 

Also includes the `libjpeg-devel` package name tweak on opensuse.


timtmok (approved):

kevinushey (approved):
Thanks for taking care of this!

Issue #13328, opened 2023-07-06
Filtering and sorting doesn't work for columns great than 50 in the data viewer


docal555:
Loaded a dataframe with 89 columns. When you skip past the first 50 columns nothing works. Cannot filter, cannot sort. If you have a dataframe with e.g. 89 columns and you skip, numbering starts with 40. Like counting from 1 to 89, once you reach 50 you have to resume at 40. This is kind of frustrating. Past the first 50 columns everythings seems to be at random. Random sorting, random filtering

### System details

    RStudio Edition : Version 2023.06.1+438 (2023.06.1+438)
    RStudio Version : 
    OS Version      : 13.4.1 (22F82)
    R Version       : R version 4.3.1

### Steps to reproduce the problem
Load a dataframe with more than 50 columns, try to sort or filter. Nothing works

### Describe the problem in detail
Load a dataframe with more than 50 columns, try to sort or filter. Nothing works.

### Describe the behavior you expected
Skip to 51 and the column starts with 51 and not 40. Let´s me filter or sort. Maybe concentrate more on R again and not on Python integration

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
There are three issues here:

1) Filtering past 50 columns: I can reproduce this—filtering doesn't work in RStudio Desktop 2023.06.1+524. This worked correctly in version 2023.03.2+454.
2) Sorting past 50 columns: I cannot reproduce this, as sorting works correctly for me. Can you provide a sample dataset with which this is happening?
3) Numbering of the last frame: This is addressed in issue #13301.

Note: Tested in part using a dataset of random numbers:
`df <- as.data.frame(matrix(runif(n=3290, min=1, max=20), nrow=10))`

Triage: I recommend this for Desert Sunflower, but we should consider this for the next patch of 2023.06.2.

docal555:
[...]

PR #13339, opened 2023-07-08
fix a variety of sorting / filtering issues

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/13328.

### Approach

We made some recent changes to the Data Viewer, such that only the data actually presented to the user is sent over the wire. This necessitated a number of changes to how column offsets were handled both on the server + client side. It looks like we might've missed a couple spots around sorting / filtering, and also in displaying column titles.

### Automated Tests

N/A

### QA Notes

Test that sorting + filtering work even in paginated tables, when paginating to later columns. Here's an example to play with:

```
.rs.writeUiPref("data_viewer_max_columns", 15L)
data <- replicate(100, 0:9, simplify = FALSE)
for (i in seq_along(data)) {
  data[[i]] <- (data[[i]] + i) %% 10
}

names(data) <- paste0("V", 1:100)
df <- as.data.frame(data)
View(df)

# .rs.writeUiPref("data_viewer_max_columns", 50L) when you're done
```

### Documentation

N/A

### Checklist

### DIFF
--- a/src/cpp/session/modules/data/DataViewer.cpp
+++ b/src/cpp/session/modules/data/DataViewer.cpp
@@ -572,6 +572,10 @@ json::Value getData(SEXP dataSEXP,
 
    std::string cacheKey = http::util::urlDecode(
          http::util::fieldValue<std::string>(fields, "cache_key", ""));
+   
+   // Parameters from the client to delimit the column slice to return
+   int columnOffset = http::util::fieldValue<int>(fields, "column_offset", 0);
+   int maxDisplayColumns = http::util::fieldValue<int>(fields, "max_display_columns", 0);
 
    // loop through sort columns
    std::vector<int> ordercols;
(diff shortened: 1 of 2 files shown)


kevinushey:
Converting to draft for now as there still seem to be some issues.

sharon-wang (approved):
LGTM!

Tested the case you provided and sorting/filtering is now working for each slice of columns in the data frame.

Thank you for fixing this up 🙌

sharon-wang (commented):
Oh! I just remembered, we'll want a NEWS.md entry for this

kevinushey:
> Oh! I just remembered, we'll want a NEWS.md entry for this

IIUC since we want to include this in the patch release, we'll plan to include a NEWS entry there?

sharon-wang:
I think we still want a NEWS.md entry in `main` because the issue is targeted for Desert Sunflower, but will be backported to the patch

But I'm not totally sure! @ronblum probably knows the answer to this

Issue #13347, opened 2023-07-10
"Update Available" dialog not popping up


sharon-wang:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2023.06.0
    OS Version      : **UPDATE** this probably happens on all OS's, as long as the user has never ignored an update before, or has removed their ignored updates list (original: Ubuntu 22, Windows 11)
    R Version       : any

### Steps to reproduce the problem

Pre-req: you have never ignored updates on the machine or you've removed your list of ignored updates

1. Install a `2023.06.0` build ~for Ubuntu 22 or Windows 11 (doesn't seem to occur with Mac)~
2. Open the IDE
3. Expect to get an "Update Available" dialog once the IDE starts up, to update to `2023.06.1`

### Describe the problem in detail

The "Update Available" dialog doesn't show on Linux/Windows automatically on startup, nor does it show when manually checking for updates via Help > Check for Updates.

Upon inspecting the network requests, the IDE successfully retrieves the information regarding the new update, but fails to display the dialog.

On Linux, the console shows:

```
ERROR: Processing client events
ConsoleLogger.java:55 Exception: com.google.gwt.core.client.JavaScriptException: (TypeError) : Cannot read properties of undefined (reading 'updates')
ConsoleLogger.java:33 TypeError: Cannot read properties of undefined (reading 'updates')
    at $getIgnoredUpdates_0_g$ (IgnoredUpdates.java:29:1)
    at DesktopApplicationHeader$IgnoredUpdatesStateValue_1_g$.getIgnoredUpdates_1_g$ [as getIgnoredUpdates_2_g$] (DesktopApplicationHeader.java:228:1)
    ...
```

We might be missing a step to create the persistent file for updates, or otherwise something is going wrong with the ignored updates file. This errors and the IDE can't continue to displaying the new update, despite having the information to do it!

### Describe the behavior you expected

When a new major/minor/patch version is related, the user should be presented with the "Update Available" dialog.

If a user is manually checking for updates, we should bypass any code related to ignored updates, because they are not relevant during a manual check.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
~- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).~
[...]

PR #13350, opened 2023-07-11
Fix update available ignored update error

sharon-wang:
### Intent

Addresses: #13347 
Backport issue: #13351

### Approach

- Ensure we initialize the `ignoredUpdates_` object
- Only run code to "ignore update" if we haven't already ignored the update

### Automated Tests

none

### QA Notes

I will create custom builds so we can test this (because the dailies/hourlies are considered "up to date" already, and won't trigger the Update Available dialog).

The first released version we can test this in will be Desert Sunflower, as long as we backport this to the 2023.06.2 release.

### Documentation
none

### Checklist

`- [ ] If this PR adds or changes UI, the updated UI meets [accessibility standards](https://github.com/rstudio/rstudio/wiki/Accessibility)`

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/application/ui/impl/DesktopApplicationHeader.java
+++ b/src/gwt/src/org/rstudio/studio/client/application/ui/impl/DesktopApplicationHeader.java
@@ -233,7 +233,7 @@ public class DesktopApplicationHeader implements ApplicationHeader,
          ignoredUpdatesDirty_ = true;
       }
 
-      private IgnoredUpdates ignoredUpdates_;
+      private IgnoredUpdates ignoredUpdates_ = IgnoredUpdates.create();
       private boolean ignoredUpdatesDirty_ = false;
    }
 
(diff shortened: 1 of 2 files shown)


src/gwt/src/org/rstudio/studio/client/application/ui/impl/DesktopApplicationHeader.java
```diff
@@ -549,15 +549,22 @@ public void execute()
...
             elementIds.add(ElementIds.DIALOG_CANCEL_BUTTON);
             buttonOperations.add(new Operation() {
                @Override
                public void execute()
                {
-                  ignoredUpdatesState_.addIgnoredUpdate(result.getUpdateVersion());
-                  // Trigger an update to the persistent updates state file
-                  eventBus_.fireEvent(new PushClientStateEvent(true));
+                  // only run the following code if we didn't already ignore the update
+                  if (!finalIgnoredUpdate)
```
timtmok, line 562: Is there a case where an update is available and we don't want the ignore button to do anything? It looks like updating the ignored versions might be good to do.

timtmok: I think this makes sense now. Just had to think about how `ignoredUpdate` is set. The ignore action doesn't have to do anything if the version has already been ignored.

timtmok (approved):

Issue #13369, opened 2023-07-14
Error message regression when reloading the session using keyboard shortcut


jeffvroom:
### System details

    RStudio Edition : Server
    RStudio Version: Mountain Hydrangea (2023.06)

### Steps to reproduce the problem

In the server version, login to the IDE and use the keyboard shortcut a couple of times to "Restart R".  Look in the rserver.log file for the error:

```
2023-07-13T23:49:06.440784Z [rserver] ERROR RStudio session start failed for: jvroom: in 0.19s error: rpc error: asio.misc error 2 (End of file) [host: localhost:8787, uri: /rpc/get_environment_state, path: /tmp/rstudio-server/rstudio-rsession/jvroom-d, user-id: 501]; OCCURRED AT void rstudio::core::http::AsyncClient<rstudio_boost::asio::basic_stream_socket<rstudio_boost::asio::local::stream_protocol>>::handleReadStatusLine(const boost::system::error_code &) [SocketService = rstudio_boost::asio::basic_stream_socket<rstudio_boost::asio::local::stream_protocol>] src/cpp/server/session/ServerSessionProxy.cpp:513; LOGGED FROM: void rstudio::server::SessionManager::removePendingLaunch(const r_util::SessionContext &, const bool, const std::string &) src/cpp/server/session/ServerSessionManager.cpp:431
```

### Describe the problem in detail

For this issue: https://github.com/rstudio/rstudio-pro/pull/4644 I added some additional session start time diagnostics that includes logging an error whenever "removePendingLaunch" finds a pending session when handing an rpc or content error. 

This trips over a weird situation though where the IDE is making two rpc calls in a row - one to restart the session, the second to get_environment_state - to refresh the environment pane.  That second call will never make it but the error handler will end up removing the pending launch for the new session, and the new behavior is to log an error (I was thinking for sure this meant the new session failed and the error was important - doh!) 

### Describe the behavior you expected

No error message logged and ideally we would not remove the pending launch.  

Possibly eliminate the get_environment_state call (that is made here: 
https://github.com/rstudio/rstudio/blob/6442230be31c1a67e877b3fe8758681d1ef2e1ec/src/gwt/src/org/rstudio/studio/client/workbench/views/environment/EnvironmentPane.java#L882)

kevinushey:
Confirmed that this resolves the issue on my end -- thanks @jeffvroom!

PR #13370, opened 2023-07-14
Don't log pending launch errors for handleRpcError and handleContentError

jeffvroom:
### Intent

Addresses: https://github.com/rstudio/rstudio/issues/13369

### Approach

Removes a recently added error that shows up during restart session process. The IDE sends a get_environment_state request right after the message to restart which will always generate an error and remove the pending launch. This eliminates the error part of the bug only.

### QA

Reproduces easily in server version.  See issue for details.


kevinushey (approved):
LGTM!

Issue #10308, opened 2022-01-05
Changing LANGUAGE env var inside RStudio IDE console does not have expected effect


cderv:
### System details

    RStudio Edition : Desktop [Open Source]
    RStudio Version : 2022.2.0.345
    OS Version      : Windows 10 x64 (build 19043)
    R Version       : R version 4.1.2 (2021-11-01)

### Steps to reproduce the problem

* Open RStudio

````r
> Sys.setenv(LANGUAGE="fr")
> mean[[1]]
Error in mean[[1]] : object of type 'closure' is not subsettable
````

* Open R Gui

````r
> Sys.setenv(LANGUAGE="fr")
> mean[[1]]
Erreur dans mean[[1]] : objet de type 'closure' non indiçable
````

* Open R Term

````r
> Sys.setenv(LANGUAGE="fr")
> mean[[1]]
Erreur dans mean[[1]] : objet de type 'closure' non indiçable
````

### Describe the problem in detail

Following investigation for an issue in **testthat** (https://github.com/r-lib/testthat/issues/1534), it seems there could be an issue with RStudio IDE as the behavior is different with RGui and R Term - This was reported first in https://github.com/r-lib/withr/issues/189

Is there anything specific with how translation are handled for inside RStudio ?

### Describe the behavior you expected

I would expect the same behavior than in R Gui or R Term from within RStudio IDE console. 

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

**Comment by kevinushey** [comment]
See also: https://github.com/rstudio/rstudio/issues/4127

**Comment by kevinushey** [comment]
Even weird is that some error messages are properly translated, but others are not. For example:

```
> Sys.setenv(LANGUAGE = "en")
> oops
Error: object 'oops' not found
> vignette("oops", package = "nope")
Error in find.package(package, lib.loc) : 
  there is no package called ‘nope’
> 
> Sys.setenv(LANGUAGE = "fr")
> oops
Error: object 'oops' not found
> vignette("oops", package = "nope")
[...]

Issue #9824, opened 2021-09-07
R 4.2: support UTF-8 code page with UCRT builds of R


kevinushey:
The next release of R (4.2.0) will likely be built using UCRT with UTF-8 as the active code page. In order for RStudio to support this, the `rsession` executable will need to be built to use this code page as well. (Currently, RStudio is compatible with such UCRT builds but because we don't declare support for a UTF-8 code page we fall back to the "default" system code page.)

In theory we just need to set this in the application manifest:

https://docs.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page

but there might be other work required e.g. in CMake to ensure some requisite compiler flags are passed down during compilation as well.

mikebessuille:
Hadley says 4.2 might release in June 2022.   In PT, we need to test that this works (doesn't just crash) and identify what, if anything, we need to do for PT.  It's possible that we might determine some of this work is not urgent and could be pushed out to Elsbeth Geranium.

mikebessuille:
based on https://github.com/rstudio/rstudio/issues/10065 this will need to be addressed.

kevinushey:
The crash appears to be the same thing I saw previously:

```
 # Child-SP          RetAddr           Call Site
00 000000f5`727fa0d0 00007ffd`f37924b1 ucrtbase!invoke_watson+0x18
01 000000f5`727fa100 00007ffd`f3792379 ucrtbase!_invalid_parameter+0x12d
02 000000f5`727fa140 00007ffd`f37c0b3c ucrtbase!invalid_parameter_noinfo+0x9
03 000000f5`727fa180 00007ffd`f377c40e ucrtbase!_wsetlocale+0x43e8c
04 000000f5`727fa200 00007ffd`f377c32f ucrtbase!<lambda_2116bde18c9e5f34230805ea4a4660ed>::operator()+0xae
05 000000f5`727fa280 00007ffd`f377c2e7 ucrtbase!__crt_seh_guarded_call<char *>::operator()<<lambda_e9ce07acdb0b138e546925ae9f1a2c9c>,<lambda_2116bde18c9e5f34230805ea4a4660ed> &,<lambda_5d037afbfc54bf1ca80d3d1ee4062886> >+0x3b
*** ERROR: Symbol file could not be found.  Defaulted to export symbols for R.dll - 
06 000000f5`727fa2b0 00000000`6c822a74 ucrtbase!setlocale+0x47
[...]

PR #13378, opened 2023-07-15
ensure default translation domain is set on startup

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/10308.

### Approach

R uses [libintl](https://github.com/wch/r-source/tree/trunk/src/extra/intl) for message translation. The default domain for translations is "messages", but R normally tries to set the default domain to "R" here:

https://github.com/wch/r-source/blob/e3d6c3fdd8666ae96bee913e1c0edab7bbaa088d/src/main/main.c#L737

And that only gets called in a few places, e.g.

https://github.com/wch/r-source/blob/e3d6c3fdd8666ae96bee913e1c0edab7bbaa088d/src/gnuwin32/system.c#L1073

So we need to make sure to call `::cmdlineoptions()` so that the default text domain is set. Note that we previously had to avoid using this function as R would crash on startup when we tried to call this, but this issue has now been alleviated by virtue of the invalid parameter handler we install (which just ignores such errors, as R and MinGW do).

### Automated Tests

N/A

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/10308. In particular, we should verify that translations work in this scenario:

```
Sys.setenv(LANGUAGE = "fr")
ouch
```

You should see a message like:

```
Erreur : objet 'ouch' introuvable
```

We should also confirm this works across a variety of versions of R on Windows; e.g. R 3.6.3; R 4.0.3, R 4.1.3, R 4.2.3 and R 4.3.1. This is necessary because this code removes an old workaround for an issue that was causing R to crash on startup, but that issue should now be alleviated via a separate fix.

### Documentation

N/A

### Checklist

### DIFF
--- a/src/cpp/r/session/REmbeddedWin32.cpp
+++ b/src/cpp/r/session/REmbeddedWin32.cpp
@@ -183,80 +183,6 @@ int askYesNoCancel(const char* question)
    }
 }
 
-bool initializeMaxMemoryDangerously()
-{
-   Error error;
-
-   void* pLibrary = nullptr;
-   error = core::system::loadLibrary("R.dll", &pLibrary);
-   if (error)
-      return false;
-
-   // first, see if we can load the 'R_max_memory' symbol directly
-   size_t* p_R_max_memory = nullptr;
-   error = core::system::loadSymbol(
-            pLibrary,
-            "R_max_memory",
-            (void**) &p_R_max_memory);
-
-   if (error)
-   {
-      // terrible, terrible hack -- for typical builds of R from CRAN,
-      // the memory address for R_max_memory lies just before the
-      // Rwin_graphicsx symbol, so find that symbol, and compute the
-      // position of R_max_memory offset from that
-      char* p_Rwin_graphicsx = nullptr;
-      error = core::system::loadSymbol(
-               pLibrary,
-               "Rwin_graphicsx",
-               (void**) &p_Rwin_graphicsx);
-
-      if (error)
-         return false;
-
-      // get memory address for R_max_memory
-      p_R_max_memory = (size_t*) (p_Rwin_graphicsx - sizeof(size_t));
-   }
-
-   // newer versions of R initialize R_max_memory to SIZE_MAX, while
-   // older versions use INT_MAX. allow either value here when checking
-   bool ok =
-         *p_R_max_memory == SIZE_MAX ||
-         *p_R_max_memory == INT_MAX;
-
-   if (!ok)
-      return false;
-
-   // we found the memory address! let's fill it up
-   MEMORYSTATUSEX status;
-   status.dwLength = sizeof(status);
-   ::GlobalMemoryStatusEx(&status);
-   *p_R_max_memory = status.ullTotalPhys;
-
-   return true;
-}
-
-bool initializeMaxMemoryViaCmdLineOptions()
-{
-   static const int rargc = 2;
-   static const char* rargv[] = {"R.exe", "--vanilla"};
-   ::cmdlineoptions(rargc, (char**) rargv);
-
-   return true;
-}
-
-void initializeMaxMemory(const core::FilePath& rHome)
-{
-   // no action required with newer versions of R
-   const char* dllVersion = getDLLVersion();
-   core::Version rVersion(dllVersion);
-   if (rVersion >= core::Version("4.2.0"))
-      return;
-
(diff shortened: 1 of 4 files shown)


src/cpp/session/SessionMain.cpp
```diff
@@ -1941,9 +1941,7 @@ int main(int argc, char * const argv[])
       //
       // https://github.com/mirror/mingw-w64/blob/eff726c461e09f35eeaed125a3570fa5f807f02b/mingw-w64-crt/crt/crtexe.c#L98-L108
 #ifdef _WIN32
-      std::string enabled = core::system::getenv("RSTUDIO_WIN32_ENABLE_INVALID_PARAMETER_HANDLER");
-      if (core::string_utils::isTruthy(enabled, true))
-         _set_invalid_parameter_handler(win32InvalidParameterHandler);
+      _set_invalid_parameter_handler(win32InvalidParameterHandler);
```
kevinushey, line 1944: We now always set this -- R may pass invalid parameters to different CRT functions in ways that we have no control over, so we need to ensure those calls don't terminate the process.

src/cpp/r/session/REmbeddedWin32.cpp
```diff
@@ -183,80 +183,6 @@ int askYesNoCancel(const char* question)
    }
 }
 
-bool initializeMaxMemoryDangerously()
```
jmcphers, line 186: Were these all removed because they only run for R < 4.2? (Do we now require R 4.2+?)

src/cpp/r/session/REmbeddedWin32.cpp
```diff
@@ -301,8 +227,14 @@ void runEmbeddedR(const core::FilePath& rHome,
...
-   // initialize R_max_memory
-   initializeMaxMemory(rHome);
+   // setup command line options
+   // note that R does a lot of initialization here that's not accessible
+   // in any other way; e.g. the default translation domain is set within
+   //
+   // https://github.com/rstudio/rstudio/issues/10308
+   static const int rargc = 2;
+   static const char* rargv[] = { "R.exe", "--vanilla" };
+   ::cmdlineoptions(rargc, (char**) rargv);
```
jmcphers, line 237: Are these arguments consumed by R in some way or do we just need to pass in placeholders? (seems like if they are actually consumed it could lead to e.g. R not running startup scripts, and/or getting confused on non-Windows platforms?)

src/cpp/r/session/REmbeddedWin32.cpp
kevinushey: These are also normally initialized as part of `cmdlineoptions()`, but we stopped doing that with R 4.2 because of crashes we were seeing. https://github.com/rstudio/rstudio/issues/9824

Previously, we avoided using `cmdlineoptions()` because we were seeing an "invalid parameter" crash. However, we later learned that MinGW by default ignores "invalid parameter" errors, so we now resolve the issue by also setting a no-op "invalid parameter" handler.

For the curious:

https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/set-invalid-parameter-handler-set-thread-local-invalid-parameter-handler?view=msvc-170

https://github.com/mirror/mingw-w64/blob/eff726c461e09f35eeaed125a3570fa5f807f02b/mingw-w64-crt/crt/crtexe.c#L261

https://github.com/mirror/mingw-w64/blob/eff726c461e09f35eeaed125a3570fa5f807f02b/mingw-w64-crt/crt/crtexe.c#L98-L108

src/cpp/r/session/REmbeddedWin32.cpp
kevinushey: They are read + consumed, and are used to set some default values for the R startup parameters; e.g. things like here:

https://github.com/wch/r-source/blob/e3d6c3fdd8666ae96bee913e1c0edab7bbaa088d/src/gnuwin32/system.c#L1165-L1176

But we do later set the parameters ourselves here:

https://github.com/rstudio/rstudio/blob/9d3b5a4f4c5abbceee1e91efb7bb90ec4727018d/src/cpp/r/session/REmbeddedWin32.cpp#L336-L370

I think I need to remove `--vanilla` here since we do want to process Renviron / Rprofile and such, though. I think I'm going to do one more pass over `cmdlineoptions()` to see if it does anything that we need to undo afterwards.

PR #13383, opened 2023-07-17
identify callable python objects from reticulate as not suspendable

t-kalinowski:
### Intent

Currently, the IDE will silently trigger a python exception, overwriting `reticulate::py_last_error()`. This occurs while the IDE is introspecting the R session to determine if the session is suspendable. The `isSuspendalImpl` function already has an early check for reticulate objects, but it is too narrow currently, missing callable python objects that present in the R session as R functions, not R environments. 

### Approach

update the check 
```r
if (is.environment(x) && inherits(x, "python.builtin.object))`
```
to
```r 
if (is.environment(value) || is.function(value)) && value, "python.builtin.object")
```

### DIFF
--- a/src/cpp/session/modules/SessionEnvironment.R
+++ b/src/cpp/session/modules/SessionEnvironment.R
@@ -20,7 +20,7 @@
       .rs.isFunctionMasked(functionName, packageName),
       error = function(e) FALSE
    )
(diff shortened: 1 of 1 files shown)


kevinushey:
Thanks!

PR #13373, opened 2023-07-14
Update text for Posit Cloud dialogs as requested

jgutman:
### Intent

Addresses https://github.com/rstudio/lucid-server/issues/5040
Use language as proposed by @Alex-Chisholm here:
https://docs.google.com/document/d/1pi8qof6UPIlJW914d6XQsBV2Ko9vgDeiBXXgTlERtjc/edit

PR #13440, opened 2023-08-01
Update Posit Cloud text and add French translations

jgutman:
Follow-up to #13373
Edit text so we don't have any truncation happening
image


timtmok (approved):

ronblum:
I'll tag Tom for thoughts about language.

PR #13498, opened 2023-08-15
Add separate flag to force docker step

jeffvroom:
Need FORCE_BUILD=true for nightly builds but don't want to force docker rebuilds in that case so adding a new flag.


MariaSemple (approved):
LGTM!

Issue #13365, opened 2023-07-13
Can't paste into New project from various pop-up dialogues e.g. Git


mine-cetinkaya-rundel:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2023.09.0-daily+217
    OS Version      : Apple M1 Pro
    R Version       : 4.3.0

### Steps to reproduce the problem

- Copy the URL of a Git repo.
- In RStudio, go to New Project > From version control > Git
- Paste the URL (and observe that you can't paste it)

### Describe the problem in detail

In the last few dailies you can't paste into this dialogue, as well as some other pop-up ones.

<img width="675" alt="Screenshot 2023-07-13 at 1 18 43 PM" src="https://github.com/rstudio/rstudio/assets/5965649/06c2da34-20a4-4f93-978b-0c8c932e1e5b">

### Describe the behavior you expected

I expected to be able to paste the repo URL.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
Thanks @mine-cetinkaya-rundel! Looks like this is a regression, as it works in RStudio Desktop 2023.06.1+524 on MacOS 13.4.

ronblum:
Other examples:
- <kbd>Files -> Blank File -> [Any file type)</kbd>
- <kbd>Environment -> Import Dataset -> From Text (base) -> (Choose file) -> Name</kbd>

etc.

kevinushey:
This seems like a weird focus issue. I can reproduce the issue, but it seems to go away if I click away from RStudio (focus a different window), then click back into RStudio. After doing that, I can then copy + paste as normal.

@ronblum, can you confirm whether the issue is macOS-specific?

ronblum:
MacOS only—it works correctly on Windows 11 and Ubuntu 22.04.

timtmok:
I'll check if this was a regression due to the Electron update and verify after taking the latest Electron (25.5.0).

timtmok:
[...]

PR #13507, opened 2023-08-16
Fix modals causing edit commands to be disabled

timtmok:
### Intent
Address #13365

### Approach
Rebuilds the menus with some commands enabled for editing and hiding the window. This allows the commands to still work. I think Electron shouldn't normally do this but we have dummy commands for these edit actions that might be a GWT necessity.

The recursive menu builder was refactored so it could be called when building the disabled menu.

### Automated Tests
Added a unit test

### QA Notes
There is a menu refresh that occurs when the IDE gains focus. This should maintain the menu state if a modal dialog is open.

### Documentation
n/a

### Checklist

### DIFF
--- a/src/node/desktop/src/main/menu-callback.ts
+++ b/src/node/desktop/src/main/menu-callback.ts
@@ -285,99 +285,102 @@ export class MenuCallback extends EventEmitter {
    *
    */
   updateMenus(): void {
-    /*
-     * This function will remove items that has a submenu array with no items
-     */
-    const removeItemsWithEmptySubmenuList = (item: MenuItemConstructorOptions) => {
-      if (Object.prototype.hasOwnProperty.call(item, 'submenu')) {
-        if (Array.isArray(item.submenu)) {
-          if (item.submenu.length > 0) {
-            item.submenu = item.submenu.filter((submenu) => {
-              return removeItemsWithEmptySubmenuList(submenu);
-            });
-            return true;
-          } else {
-            return false;
-          }
-        }
-      }
-
-      return true;
-    };
-
-    /**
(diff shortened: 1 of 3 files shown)


src/node/desktop/src/main/menu-callback.ts
```diff
@@ -485,16 +488,25 @@ export class MenuCallback extends EventEmitter {
...
+        const disabledMenu = Menu.buildFromTemplate(this.recursiveCopy(this.mainMenuTemplate));
+        disabledMenu?.items?.forEach((item) => {
+          item.submenu?.items.forEach((subItem) => {
+            // keep some commands enabled
+            subItem.enabled = ['cut', 'copy', 'paste', 'redo', 'hide', 'hideOthers', 'unhide', 'selectAll'].includes(subItem.role ?? '');
+          });
+        });
+        Menu.setApplicationMenu(disabledMenu);
+        this.mainMenu = disabledMenu;
+      }
```
sharon-wang, line 503: IIUC, this means we can remove `showPlaceholderMenu()` altogether?

timtmok: Oh, that's something that you had added before to handle the modals. I'll remove it!

timtmok: And I'm already mixing myself up. It looks like it was refactored for that modal change. Gary had initially added the placeholder and I'm pretty sure it's so that the top-level menus show up during startup until GWT sends the menu items.

sharon-wang: ah right, the menu items aren't yet ready during startup, so we need the placeholder 👍

PR #13513, opened 2023-08-17
Fix package build

timtmok:
### Intent
Address rstudio/rstudio-pro#4943

### Approach
Add exclude rules for cmake install. A separate PR will handle the part of the package that is pro only. https://github.com/rstudio/rstudio-pro/pull/4966

### Automated Tests
n/a

### QA Notes
The package should not include `.gitignore` at all. This includes all platforms since any common parts like Quarto or GWT would have this unnecessary file included.

### Documentation
n/a

### Checklist

### DIFF
--- a/src/cpp/session/CMakeLists.txt
+++ b/src/cpp/session/CMakeLists.txt
@@ -626,7 +626,9 @@ if(NOT RSTUDIO_SESSION_WIN32 AND NOT RSESSION_ALTERNATE_BUILD)
            DESTINATION ${RSTUDIO_INSTALL_SUPPORTING}/resources)
    # templates
    install(DIRECTORY "resources/templates"
-           DESTINATION ${RSTUDIO_INSTALL_SUPPORTING}/resources)
+           DESTINATION ${RSTUDIO_INSTALL_SUPPORTING}/resources
+           PATTERN ".gitignore"
+           EXCLUDE)
    # pandoc
    install(DIRECTORY "resources/pandoc"
            DESTINATION ${RSTUDIO_INSTALL_SUPPORTING}/resources)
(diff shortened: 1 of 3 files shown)


src/cpp/session/CMakeLists.txt
```diff
@@ -695,15 +697,21 @@ if(NOT RSTUDIO_SESSION_WIN32 AND NOT RSESSION_ALTERNATE_BUILD)
...
+                  PATTERN ".gitignore"
+                  EXCLUDE)
          else()
             install(DIRECTORY "${RSTUDIO_DEPENDENCIES_QUARTO_DIR}"
                   DESTINATION "${RSTUDIO_INSTALL_BIN}"
                   USE_SOURCE_PERMISSIONS
-                  PATTERN "*/share" EXCLUDE)
+                  PATTERN "*/share"
+                  PATTERN ".gitignore"
+                  EXCLUDE)
```
sharon-wang, line 709: I'm not super familiar with CMake exclude syntax, but does each `PATTERN` need to be followed by `EXCLUDE`? Or maybe the single `EXCLUDE` covers both `PATTERN`s?

```suggestion
                  PATTERN "*/share" EXCLUDE
                  PATTERN ".gitignore" EXCLUDE)
```

src/gwt/CMakeLists.txt
```diff
@@ -138,13 +138,17 @@ if(GWT_COPY)
...
+      EXCLUDE)
 
    # install compiled GWT artefacts
    install(
       DIRECTORY "${GWT_WWW_DIR}"
       DESTINATION "${RSTUDIO_INSTALL_SUPPORTING}"
-      PATTERN "WEB-INF" EXCLUDE)
+      PATTERN "WEB-INF"
+      PATTERN ".gitignore"
+      EXCLUDE)
```
sharon-wang, line 151: Same comment here as above

sharon-wang (approved):
Just a syntax comment, otherwise LGTM!

⛔ .gitignore ⛔

src/cpp/session/CMakeLists.txt
kevinushey: Based on the example at https://cmake.org/cmake/help/latest/command/install.html#installing-directories, I think you're right -- each exclude pattern needs to be followed by `EXCLUDE`.

timtmok: I think I read those docs wrong. I'll update it!

Issue #13134, opened 2023-05-16
File opened twice in RStudio with cli `{.file` hyperlink


olivroy:
See r-lib/cli#604

### System details

    RStudio Edition : Desktop
    RStudio Version : 2023.03
    OS Version      : Windows
    R Version       : 4.2.3

Hello,

I was wondering if it was possible to open files with a relative path. (Windows)

https://cli.r-lib.org/reference/links.html#files

I usually open files with  5 ways, the first 4 are consistent, but {.file } is not. (when in Project)

1. Opening a file from Ctrl + .,
2. Files pane in RStudio,
3; Find in Files RStudio
4. `usethis::edit_file()` or `rstudioapi::documentOpen()`
5. With cli {.file } hyperlinks

to 4. open with a relative path "~/path/to/file.R" (when hovering over the tab) in RStudio, so avoid opening the same file in 2 different tabs.

### Steps to reproduce the problem

Clicking on
```r
cli::cli_text("{.file file.R}
```
opens a 'new' file (duplicate tab)
When hovering over it, I see that the path is "C:/Users/.../path/to/file.R"

RStudio doesn't recognize that they are the same file, so they are open twice.

### Describe the problem in detail

The file opens twice, by Methods 1 to 4 as "~/path/to/file.R", and when clicking on the cli hyperlink as "C:/Users/.../path/to/file.R", which means the same file is open twice.

### Describe the behavior you expected

I would expect the cli hyperlink to just change the focus of the source, instead of opening 2 instances of the file.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

olivroy:
Edit: I investigated a little bit more!
Running `rstudioapi::documentId()` in 
1. a document with the `~/` path gives "F31EC12D"
 2. the same document wit the `C:/.../`path gives   "9AB3555D".

It means that RStudio really sees the same document as a different document.

olivroy:
Maybe it's not related, but on Windows, when writing a path with `~/`, RStudio autocompletes from R's default locations

[...]

PR #13515, opened 2023-08-17
ensure forward slashes used when normalizing hyperlink path

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/13134.

### Approach

It seems like our 'createAliasedPath()' helper has an implicit assumption about the path separators used. I'm not sure if we should fix this so that both `/` and `\` separators are properly handled on Windows?

### Automated Tests

N/A

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/13134. Note that the issue should only reproduce on Windows.

### Documentation

N/A

### Checklist


olivroy:
If possible (and if not already planned) , I think this could be tested on network drives (UNC paths) too on Windows.. Many paths problems also occur there too on Windows.

Issue #13512, opened 2023-08-17
diagnostics system fails to detect 'mutate' as NSE function


kevinushey:
From @DavisVaughan:

Just a note that this exact code doesn't work right for `dplyr::mutate()` because `dplyr:::mutate.data.frame` actually uses `dplyr_quosures(...)` rather than something like `enquos(...)` to capture the vars.

<img width="242" alt="Screenshot 2023-08-17 at 12 35 07 PM" src="https://github.com/posit-dev/amalthea/assets/19150088/de69abe9-bb04-440f-9071-0582c3b3655c">

Also, this "symbol in scope" diagnostic is off in rstudio by default

---

I imagine there are also other cases where the `...` are passed untouched through various layers of function calls before finally being captured with `enquos(...)`. It works a little better with named variables where you have to do `enquo(x)` or `{{ x }}` to pass them through - even at the top level

Although even with named variables there are weird cases, like ggplot2's capturing of `x` and `y` here

```r
> ggplot2:::aes
function (x, y, ...) 
{
    xs <- arg_enquos("x")
    ys <- arg_enquos("y")
    dots <- enquos(...)
```

ronblum:
Verified the first part, that `mutate` no longer throws a warning, in
- RStudio Desktop 2023.09.0-daily+380 on MacOS 14
- RStudio Server 2023.09.0-daily+380 on Ubuntu 20.04.5 via Chrome on MacOS 14

However, in this example from the PR, `pivot_long` is fine in RStudio Desktop, but a warning appears in RStudio Server:

```
library(tidyr)
df <- tibble(x = c("a", "b", "a"))
pivot_longer(df, x)
```

![Image](https://github.com/rstudio/rstudio/assets/31009952/1fa91bef-d111-404c-b1ea-2bc95845a10b)

ronblum:
Sometimes `mutate` still shows a warning, but will later go away (Not certain of the conditions):
<img width="293" alt="mutate_is_back" src="https://github.com/rstudio/rstudio/assets/31009952/4057e84e-ad88-4885-88ea-3e300d89b3d8">

kevinushey:
The issue should go away as soon as the `tidyr` and `dplyr` packages are loaded.

ronblum:
Oops.... Ahem.

The `mutate`/`mpg` warning goes away when I remember to load the libraries. The warning for `pivot_longer` doesn't go away, though.

ronblum:
[...]

PR #13516, opened 2023-08-17
turn off symbol diagnostics in mutate()

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/13512.

Screenshot 2023-08-17 at 3 02 56 PM

### Approach

Since `dplyr::mutate.data.frame` uses `dplyr_quosures` for quasi-quotation and tidy evaluation, add that to our NSE primitives list.

### Automated Tests

N/A

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/13512.

### Documentation

N/A

### Checklist

### DIFF
--- a/src/cpp/r/RSexp.cpp
+++ b/src/cpp/r/RSexp.cpp
@@ -1736,12 +1736,17 @@ std::set<std::string> makeNsePrimitives()
 
       // rlang tidy evaluation
       "enexpr",
+      "enexprs",
       "enquo",
       "enquos",
       "eval_tidy",
+      "quo",
+      "quos",
 
-      // dplyr
-      "dplyr_quosures"
+      // dplyr -- not really primitives
+      "dplyr_quosures",
+      "group_by",
+      "group_by_prepare"
       
    };
 }
--- a/src/cpp/r/RSexp.cpp
+++ b/src/cpp/r/RSexp.cpp
@@ -1710,28 +1710,40 @@ bool inherits(SEXP object, const char* S3Class)
    return Rf_inherits(object, S3Class);
 }
 
+// Keep in sync with 'nse.primitives' in SessionCodeTools.R
 std::set<std::string> makeNsePrimitives()
 {
-   std::set<std::string> nsePrimitives;
-   nsePrimitives.insert("quote");
-   nsePrimitives.insert("substitute");
-   nsePrimitives.insert("match.call");
-   nsePrimitives.insert("library");
-   nsePrimitives.insert("require");
-   nsePrimitives.insert("enquote");
-   nsePrimitives.insert("bquote");
-   nsePrimitives.insert("expression");
-   nsePrimitives.insert("evalq");
-   nsePrimitives.insert("subset");
-   nsePrimitives.insert("eval.parent");
-   nsePrimitives.insert("sys.call");
-   nsePrimitives.insert("sys.calls");
-   nsePrimitives.insert("sys.frame");
-   nsePrimitives.insert("sys.frames");
-   nsePrimitives.insert("sys.function");
-   nsePrimitives.insert("sys.parent");
-   nsePrimitives.insert("lazy_dots");
-   return nsePrimitives;
+   return {
+      
+      // base R primitives
+      "bquote",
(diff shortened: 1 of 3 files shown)


sharon-wang (commented):
Just one question for my understanding, otherwise LGTM!

Is the removal of some of the primitives because we no longer expect non-standard evaluation for them? E.g. `lazy_dots`, `select_vars`, etc. Or perhaps those primitives are the symbols we want to turn off diagnostics for?

src/cpp/r/RSexp.cpp
```diff
@@ -1710,28 +1710,40 @@ bool inherits(SEXP object, const char* S3Class)
...
+      "sys.calls",
+      "sys.frame",
+      "sys.frames",
+      "sys.function",
+      "sys.parent",
+
+      // rlang tidy evaluation
+      "enexpr",
+      "enquo",
+      "enquos",
```
DavisVaughan, line 1740: Also `quos()`, which sometimes is used instead of `enquos()`

src/cpp/r/RSexp.cpp
```diff
@@ -1710,28 +1710,40 @@ bool inherits(SEXP object, const char* S3Class)
...
+      "substitute",
+      "sys.call",
+      "sys.calls",
+      "sys.frame",
+      "sys.frames",
+      "sys.function",
+      "sys.parent",
+
+      // rlang tidy evaluation
+      "enexpr",
```
DavisVaughan, line 1738: Very rare, but `enexprs()`

src/cpp/session/modules/SessionCodeTools.R
```diff
@@ -1488,10 +1488,41 @@
...
+   "sys.call",
+   "sys.calls",
+   "sys.frame",
+   "sys.frames",
+   "sys.function",
+   "sys.parent",
+   
+   # rlang tidy evaluation
+   "enexpr",
+   "enquo",
```
DavisVaughan, line 1519: I think `tidyr::pivot_longer()` will still be flagged with "symbol not in scope"

```r
library(tidyr)

df <- tibble(x = c("a", "b", "a"))

pivot_longer(df, x)
```

If you look at `tidyr:::pivot_longer.data.frame()`, the `cols` argument is passed along using `{{ cols }}` which is semantically the same as `!!enquo(cols)` but `enquo()` doesn't appear.

src/cpp/r/RSexp.cpp
```diff
@@ -1710,28 +1710,40 @@ bool inherits(SEXP object, const char* S3Class)
...
+      "enquo",
+      "enquos",
+      "eval_tidy",
+
+      // dplyr
+      "dplyr_quosures"
+      
+   };
```
DavisVaughan, line 1746: I think a simple `group_by()` is still going to get flagged

`dplyr:::group_by.data.frame` passes the dots through to `group_by_prepare()` and that is the one that `enquos()` it. That's the main thing I worry about with this kind of approach (but don't have another good option at the moment)

DavisVaughan: Same with `dplyr:::nest_by.data.frame` which just forwards dots to `group_by()`

DavisVaughan: Same with `tidyr:::nest.data.frame` which just forwards dots to `tidyr:::nest.tbl_df` which also just forwards dots to `nest_info()`

kevinushey:
> Is the removal of some of the primitives because we no longer expect non-standard evaluation for them? E.g. lazy_dots, select_vars, etc. Or perhaps those primitives are the symbols we want to turn off diagnostics for?

Those were used with an older-fashioned version of tidy evaluation (called lazy eval), which has been largely phased out in the tidyverse. I think it's okay to drop those usages now.

src/cpp/session/modules/SessionCodeTools.R
kevinushey: Perhaps we could also detect usages of `{{}}`? I'm guessing those are going to be relatively uncommon outside of tidy evaluation.

Issue #12791, opened 2023-03-06
Error when running x86_64 binaries on M1 Mac without rosetta installed


timtmok:
> Hello.
> 
> I leave this comment in case someone else gets the same error. I'm working on a mac mini running macOS ventura 13.1 with an M1 chip. I have **R 4.2.2** and **RStudio 2022.12.0 Build 353** installed. Everything works fine, except that when I tried to create a .qmd file, I got the error window "State not recoverable". The issue was resolved by installing rosetta with `/usr/sbin/softwareupdate --install-rosetta --agree-to-license`.

_Originally posted by @armelbrizuela in https://github.com/rstudio/rstudio/issues/10871#issuecomment-1387501651_

Users that haven't installed rosetta yet on M1 may encounter an error when RStudio running these executables. Normally, MacOS would present a prompt to guide the user to install rosetta.

It may be helpful to surface this guidance or notify the user that the error is due to needing rosetta.

sharon-wang:
Is there a good place for us to list general M1 Mac troubleshooting instructions? In addition to errors like the above, sometimes users run into architecture mismatch situations which cause errors to occur in RStudio or other dependencies, and those errors don't always indicate that an architecture-related discrepancy needs to be resolved.

For example, someone might have x86_64 R/RStudio, but through `brew` they might install an aarch64 dependency, which is incompatible with their R/RStudio. In the case of the `odbc` driver library installed via `brew`, they might get a "file not found" error or some sort of "failed connection" error from `odbc`, but it's not obvious from those errors that they need the x86_64 version of `odbc`, or alternatively, that they need to install aarch64 R/RStudio to work with their aarch64 `brew` installation of `odbc`.

I'm thinking some generic instructions like:
- ensure you have rosetta installed, here's how: `/usr/sbin/softwareupdate --install-rosetta --agree-to-license`
- try installing the x86_64 version of the library -- install x86_64 `brew`, then use that `brew` to install your library
- ensure you're using either the aarch64 version for each piece of software you're trying to connect together, or the x86_64 version for each piece, but don't mix architectures

ronblum:
@sharon-wang Perhaps as [a support article](https://support.posit.co/hc)?

gtritchie:
Be aware we still ship some Intel-only components with RStudio Desktop, so they require Rosetta: https://github.com/rstudio/rstudio/issues/12572

If you write a support article, consider including a link to https://support.apple.com/en-us/HT211861.

[...]

Issue #12572, opened 2023-01-13
Intel-only components (diagnostics, rpostback) in Universal RStudio.app (Mac)


gtritchie:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2022.12.0+353
    OS Version      : M1/M2 Mac (this is an ARM Mac-only issue)
    R Version       : any

### Steps to reproduce the problem

- install RStudio.app, then in terminal:

```
$ cd /Applications/RStudio.app/Contents/Resources/app/bin
$ file *
diagnostics:    Mach-O 64-bit executable x86_64
postback:       directory
r-ldpath:       Bourne-Again shell script text executable, ASCII text
rpostback:      Mach-O 64-bit executable x86_64
rsession:       Mach-O 64-bit executable x86_64
rsession-arm64: Mach-O 64-bit executable arm64
```

### Describe the problem in detail

RStudio.app (electron) is supposed to be a universal app, meaning it can be run on either Intel or ARM Macs, and will use only native binaries on each. The above shows we still ship some Intel-only components (diagnostics, rpostback), so related features would require user to have Rosetta2 installed on their Macs for it to work.

### Describe the behavior you expected

We either need to make `diagnostics` and `rpostback` be actual universal apps (that would be the cleanest) or do what we did with rsession / rsession-arm64 and have separate binaries and select the correct one at runtime.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

gtritchie:
- Tahoe (macOS 26) is the last macOS to support Intel Macs
- Rosetta2 being removed from macOS 28 (2027)

gtritchie:
Newer release of macOS Tahoe are now showing this, occasionally, when starting RStudio. This is (presumably) due to the presence of Intel-only binaries.

<img width="350" height="112" alt="Image" src="https://github.com/user-attachments/assets/16b9e114-1187-4cbd-bf71-c31318621740" />

PR #13518, opened 2023-08-18
Warn user to install rosetta on apple silicon if not installed

sharon-wang:
### Intent

- Addresses: #12791

This warning is needed because we still ship with Intel components (#12572), including Quarto. As a result, Apple silicon users may run into errors when they hit Intel components while using the IDE.

### Approach

When RStudio is opened on a Mac with Apple silicon, check to see if Rosetta is installed. If Rosetta is installed, open RStudio as usual. If Rosetta is not installed, show a dialog warning the user that Rosetta is not installed (and provide link to docs with installation instructions), but let the user proceed with opening the IDE.

Currently, there is not a mechanism to opt out of showing the warning dialog from showing if the user is on Apple silicon and doesn't have Rosetta installed.

#### Preview

rosetta_not_installed_warning

### Automated Tests

Includes a unit test to confirm the Rosetta check is only run on Apple silicon.

#### Suggested test cases
1. Mac without Rosetta installed. Upon opening RStudio, a warning dialog should show. (During my testing, I used a VM running macOS Monterey 12.3.1).
2. Ensure RStudio is closed. Install Rosetta on the Mac from `1`, then reopen RStudio. No warning should show.
3. Open RStudio on a Mac that already has Rosetta installed. No warning should show. (I tested on macOS Ventura 13.4.1)

To check if Rosetta is installed or not:
- Run `/usr/bin/pgrep oahd` in a Terminal.
- If no output is provided, then Rosetta is not installed.
- If a number if outputted, the number indicates the process id of `oahd`, which indicates that Rosetta is installed and running.

### Documentation

The IDE Installation instructions in the Pro Desktop Admin Guide will be updated with instructions on how to install Rosetta. See PR: rstudio/rstudio-pro#4968

### Checklist


timtmok (approved):

src/node/desktop/src/main/session-launcher.ts
```diff
@@ -586,6 +587,10 @@ export class SessionLauncher {
       }
     }
 
+    if (process.platform === 'darwin' && process.arch === 'arm64') {
+      checkForRosetta();
```
kevinushey, line 591: Can we add an escape hatch just in case users need to turn off our Rosetta check for some reason? Perhaps via a desktop option, or an environment variable?

src/node/desktop/src/main/detect-rosetta.ts
```diff
@@ -0,0 +1,123 @@
...
+  [
+    DetectRosettaStatus.ROSETTA_INSTALL_WARNING,
+    {
+      type: 'warning',
+      buttons: ['OK'],
+      defaultId: 0,
+      title: 'Rosetta Is Not Installed',
+      message: 'Rosetta is not installed on your machine. You may experience issues running RStudio.',
+      detail:
+        'To avoid these issues, please install Rosetta and reopen RStudio.\n\nFor instructions, please visit https://docs.posit.co/ide/desktop-pro/getting_started/installation.html#apple-silicon-mac-m1m2.',
```
kevinushey, line 47: If I understand correctly, one downside to using a desktop dialog here is that users won't be able to copy + paste the URL here, or click on it to navigate.

What do you think about adding a button like "More Information" that, on click, would navigate to the above page?

Alternatively, we could use a GWT web dialog, where we'd have more control over the content of the dialog (and so could display clickable links).

src/node/desktop/src/main/detect-rosetta.ts
```diff
@@ -0,0 +1,123 @@
...
+      detail: 'Please reopen RStudio to try again.',
+    },
+  ],
+  [
+    DetectRosettaStatus.ROSETTA_INSTALL_WARNING,
+    {
+      type: 'warning',
+      buttons: ['OK'],
+      defaultId: 0,
+      title: 'Rosetta Is Not Installed',
```
kevinushey, line 44:
```suggestion
      title: 'Rosetta is not Installed',
```

kevinushey (commented):
Overall LGTM; main comment is around making it easier to navigate to the associated URL from the dialog.

src/node/desktop/src/main/detect-rosetta.ts
sharon-wang: Users will be able to copy+paste, but since it's a pretty long URL, it's not great to have to select all that text, then copy+paste into a browser. I like the "More Information" button idea!

I've modified the buttons to now be: `More Information`, `Don't Remind Me Again` and `Remind Me Later`. I updated the screenshot in the description with the new buttons.

src/node/desktop/src/main/session-launcher.ts
sharon-wang: I've added an ElectronDesktopOption `checkForRosetta` which is set to true by default.

If the user clicks the `Don't Remind Me Again` button, it will set that option to false.

kevinushey (approved):
Looks perfect -- thank you for adding the new buttons!

Issue #13446, opened 2023-08-02
CoPilot: Toggle between standard intellisense and the new copilot recommendations


slodge:
I'm loving the new CoPilot experience.

However, I'm also missing the old code completion/intellisense experience.

In particular I am missing completions of e.g. function/member lists when I type `rlang::`

e.g. with "legacy" RStudio I frequently use:
![image](https://github.com/rstudio/rstudio/assets/533662/8fbf3421-ad65-4d91-a41c-40f1382324f8)

while with CoPilot I only seem to get:
![image](https://github.com/rstudio/rstudio/assets/533662/da253d24-bb1d-4ac4-9ca9-348ec20f3f71)

Is there any way to allow users to benefit from both sets of recommendations? Or at least to quickly toggle between the two systems?

ronblum:
@slodge Thank you for raising this!

@kevinushey Is there any way around this?

There's a cumbersome workaround: Hit <kbd>Tab</kbd> to accept the Copilot suggestion, delete the suggested code, then hit <kbd>Tab</kbd> again. Note that <kbd>Ctrl-Z</kbd> won't help.

slodge:
Ctrl+SPACE - I found it - maybe just needs publicising?

kevinushey:
The intention is indeed to allow Ctrl + Space to retrieve "regular" completions. There might be other tweaks worth making too; we're still figuring out what the best way to allow these two systems to co-mingle is.

jthomasmock:
I've moved towards a longer delay for Copilot suggestions (> 1,000 ms) and being more aggressive to initiate auto-completion manually via `Tab`.

kevinushey:
Here's a question -- suppose you're in this state, where you have both ghost text and a completion list visible.

<img width="332" alt="Screenshot 2023-08-16 at 2 29 55 PM" src="https://github.com/rstudio/rstudio/assets/1976582/ca4258eb-2f27-487e-9f87-c583b312cce6">

What would you expect:

- The Tab key to do?
- The Enter key to do?

jthomasmock:
It looks like there's some existing write-ups about dev UX on this:

- https://github.com/orgs/community/discussions/9430
- https://visualstudiomagazine.com/articles/2023/02/17/copilot-intellisense.aspx

One takeaway I saw was "I turned off inline editor suggestions and use <kbd>Alt-\</kbd> to trigger Copilot"

[...]

PR #13521, opened 2023-08-18
allow user to control behavior of autocomplete + Tab key

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/13446.

##### TODO

- [x] Confirm that Tab works as expected here.
- [x] Provide a command / keyboard shortcut for requesting Copilot suggestions.
- [x] Prefer displaying the code completion panel on top, so it doesn't obscure ghost text.

### Approach

This PR gives the user some options controlling how GitHub Copilot's code suggestions will live + work together with the existing autocompletion system.

Screenshot 2023-08-18 at 12 55 53 PM

The intention here is that, for users who'd still like to get as-you-type autocompletions, they can enable the preference, and then when a user is in this state:

Screenshot 2023-08-18 at 12 57 38 PM

they can still use Tab to insert the Copilot code suggestion, rather than the selected completion from the list. (The Enter key can be used to insert that completion.)

### Automated Tests

### QA Notes

### Checklist

### DIFF
--- a/src/cpp/session/include/session/prefs/UserPrefValues.hpp
+++ b/src/cpp/session/include/session/prefs/UserPrefValues.hpp
@@ -424,6 +424,10 @@ namespace prefs {
 #define kDisableRendererAccessibility "disable_renderer_accessibility"
 #define kCopilotEnabled "copilot_enabled"
 #define kCopilotCompletionsDelay "copilot_completions_delay"
+#define kCopilotAllowAutomaticCompletions "copilot_allow_automatic_completions"
+#define kCopilotTabKeyBehavior "copilot_tab_key_behavior"
+#define kCopilotTabKeyBehaviorSuggestion "suggestion"
+#define kCopilotTabKeyBehaviorCompletions "completions"
 
 class UserPrefValues: public Preferences
 {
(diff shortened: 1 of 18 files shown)


kevinushey:
Example of the display:

Screenshot 2023-08-29 at 1 18 32 PM

In this state, pressing Tab will accept the Copilot code suggestions, while pressing Enter will accept the completion candidate. If the user wants to dismiss the completion popup so they can view the ghost text, they can use the Escape key.

jgutman:
Just taking a look at this now, but I notice that I can enable the copilot code completions setting from the Command Palette, but it does not display in Global Options, as you showed in the above screenshot. Is there a reason this isn't showing there?

image
image

jgutman:
One note, and I think this is unavoidable, but this will have weird behavior if the user has enabled `Code > Completion > Show code completions: Manually (tab)`

With that setting enabled, you would get a scenario where pressing tab prior to the Copilot keyboard idle delay displays the code completions, but pressing tab after the Copilot delay once the Copilot suggestion has appeared will accept the suggestion. I don't really see a way around that, since the manual tab completions don't really play nicely with the Copilot suggestions, but wanted to point that out

jgutman:
Also, it seems like if the setting is set to Tab inserts completions, then neither tab nor enter inserts the Copilot suggestion. How do you get the Copilot suggestion if you've set Tab to prefer the completion?

kevinushey:
> Just taking a look at this now, but I notice that I can enable the copilot code completions setting from the Command Palette, but it does not display in Global Options, as you showed in the above screenshot. Is there a reason this isn't showing there?

I'm sort of thinking (after playing with this for a bit) we shouldn't expose those options in the UI, as they make it a bit too easy to make Copilot awkward to use.

That is, I think we should:

1. Respect the user's preference for as-you-type code completions, and continue showing those as normal;
2. Adjust the behaviour of Tab so that it prefers inserting Copilot suggestions (when visible) as opposed to a selected code completion.

We can leave them behind as accessible from the command palette if we find out we're wrong and users _do_ want to adjust this behavior?

> Also, it seems like if the setting is set to Tab inserts completions, then neither tab nor enter inserts the Copilot suggestion. How do you get the Copilot suggestion if you've set Tab to prefer the completion? By deactivating/escaping and then tabbing?

That's right. It's not a particularly useful configuration...

Issue #13178, opened 2023-05-30
Clicking links in Terminal doesn't automatically open them in browser


etiennebacher:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2023.03.1 build 446
    OS Version      : Windows 10 x64 (build 19044)
    R Version       : 4.3.0

### Steps to reproduce the problem

* Open a terminal (Tools / Terminal / New Terminal)
* Do something like `echo "https://rstudio.com"`
* Click on the link in the terminal 

### Describe the problem in detail

I think this is similar to #6621. When I click on the link in the terminal, it doesn't open it automatically in the browser. Instead it opens a small Windows 10 window "You'll need a new app to open this about link" and it only proposes me to go on the Microsoft Store, which is unexpected since Firefox and Chrome are installed on my laptop. 

I checked in "Tools > Global options > Clickable web links" but it is already ticked.

I can't remember whether opening links printed in Terminal directly in the browser was available before but I think it was (and if it wasn't I think it'd make a good feature request). Unfortunately I can't install nightly versions or previous versions of RStudio on this laptop.

### Describe the behavior you expected

Automatically open links in default web browser.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ ] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

snowpong:
I am running the latest version of RStudio on Windows 10 and this is still an issue.
Click a link, get the small dialog with "You'll need a new app to open this about link" and "Search in Microsoft Store"
Does nobody care? It's a regression.

```
RStudio 2023.06.1+524 "Mountain Hydrangea" Release (547dcf861cac0253a8abb52c135e44e02ba407a1, 2023-07-07) for windows
Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) RStudio/2023.06.1+524 Chrome/110.0.5481.208 Electron/23.3.0 Safari/537.36
```

jonvanausdeln:
[...]

PR #13539, opened 2023-08-22
patch web-links.js to fix up window opening

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/13178.

### Approach

The xterm.js web links addon tries to open pages using the following code:

https://github.com/xtermjs/xterm.js/blob/master/addons/xterm-addon-web-links/src/WebLinksAddon.ts#L23-L35

However, this attempt to first open a blank window, and then update its href, doesn't seem to work as expected. This PR side-steps the issue by just calling `window.open(<url>)`, by patching the generated JavaScript.

### Automated Tests

N/A

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/13178.

### Documentation

N/A

### Checklist


kevinushey:
The diff is gross since it's just minified JavaScript, but the key point is that

> function o(e,t){var n=window.open();n?(n.opener=null,n.location.href=t):console.warn("Opening link blocked as opener could not be cleared")}

is replaced with

> function o(e,t){window.open(t)}

timtmok (approved):

Issue #13527, opened 2023-08-21
Turn off Copilot by default in RStudio Server


ronblum:
Turn off Copilot by default in RStudio Server. This will give administrators more control over it, who can still turn it on if they wish. (Especially relevant for the initial rollout.)

ronblum:
It looks like this is getting turned off for RStudio Desktop, as well. It should be on for Desktop but off for Server.

kevinushey:
Should be fixed via https://github.com/rstudio/rstudio/commit/355c79d82345a9cfe09384d925767165547e1551; hopefully to be live in the next daily builds.

ronblum:
Verified that RStudio Desktop 2023.09.0-daily+380 on MacOS 14 is back in business. Now to check that Copilot is still disabled in RStudio Server.

ronblum:
Verified in RStudio Server 2023.09.0-daily+380 on Ubuntu 20.04.5 via Chrome:
- Copilot is turned off by default.
- Copilot is turned off with `copilot-enabled=0` in /etc/rstudio/rsession.conf
- Copilot is enabled with `copilot-enabled=1` in /etc/rstudio/rsession.conf, authorization works, and Copilot returns suggestions

Note that while setting `copilot-enabled=1` will work turn on Copilot, is unsupported yet, as it hasn't been tested and certified. This would be strictly for testing purposes only, for the current release (Desert Sunflower).

PR #13545, opened 2023-08-23
disable copilot by default in RStudio Server

kevinushey:
Addresses https://github.com/rstudio/rstudio/issues/13527.

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/prefs/model/UserStateAccessor.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/prefs/model/UserStateAccessor.java
@@ -205,6 +205,10 @@ public class UserStateAccessor extends Prefs
          return this && this.windows || {"rBinDir":"","preferR64":true,"rExecutablePath":""};
       }-*/;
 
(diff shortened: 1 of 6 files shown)


src/gwt/src/org/rstudio/studio/client/workbench/prefs/model/UserStateAccessor.java
```diff
@@ -205,6 +205,10 @@ public final native JavaScriptObject getWindows() /*-{
          return this && this.windows || {"rBinDir":"","preferR64":true,"rExecutablePath":""};
       }-*/;
 
+      public final native JavaScriptObject getMacos() /*-{
```
timtmok, line 208: Is this used for the Copilot preference?

kevinushey: This was generated when I ran the `generate-prefs` script -- perhaps it was missed with https://github.com/rstudio/rstudio/pull/13518?

Sspirax:
Using the dailies build with this commit disabled copilot for my desktop install as well… Using previous builds works.

Issue #13548, opened 2023-08-24
Add French translations for Rosetta installation warning


sharon-wang:
Follow-up to #12791 to include French translations for the strings used in the warning dialog.

Strings in this file will need French translations: https://github.com/rstudio/rstudio/blob/main/src/node/desktop/src/main/detect-rosetta.ts

sharon-wang:
A note on my progress so far:

I have the string translations themselves ready, but I'm running into issues detecting the locale at the point that `detectRosetta` is called. This needs to be resolved because we won't be able to support a language other than the default (English) if we can't detect the LOCALE setting.

Possible workarounds:
- Move the `detectRosetta` call to a later point, when the main window is ready. With access to the main browser window, we can check the cookies for the `LOCALE` value and `changeLanguage()` so the i18n instance switches to the right strings
- Don't initialize a new instance of the i18n manager, which won't know about the `LOCALE`. Where can the strings live so that an existing instance of i18n can set the strings properly?

I've tried working in both of the above workarounds, but keep running into issues around not being able to access the LOCALE or i18n early enough.

WIP branch: https://github.com/rstudio/rstudio/tree/add-fr-translations-rosetta-install-warning

kevinushey:
> Move the detectRosetta call to a later point, when the main window is ready. With access to the main browser window, we can check the cookies for the LOCALE value and changeLanguage() so the i18n instance switches to the right strings

This seems like the right solution. Alternatively, something similar would be to have some kind of way of registering "on load" callbacks with the i18n manager, so you could have something like:

```
i18nManager.onReady(() => {
  // display the Rosetta window
})
```

And then that `onReady()` helper would either run the callback immediately if the i18n manager is already initialized, or register the callback to be run later when it is initialized.

jonvanausdeln:
Verified on [2023.09.0-daily+414](https://dailies.rstudio.com/version/2023.09.0-daily+414)

French displayed as expected (MacOS only)

PR #13596, opened 2023-09-06
Add fr translations rosetta install warning

sharon-wang:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/13548

#### Preview
image

### Approach
- add English and French strings for the detect rosetta dialog text
- relocate the `detectRosetta` call from the session launcher to the initialization of the desktop header, so that the i18n framework is initialized and set to the correct language by the time we call `detectRosetta`
    - one tradeoff of this approach is that the IDE window shows up in a loading state behind the Detect Rosetta dialog, where previously the IDE window would not open until the user selects a choice from the Detect Rosetta dialog
- removed a call to initialize i18next from the detect-rosetta code, which was interfering with i18n initialization and language detection

#### Sidenote
We seem to be using i18next incorrectly (?) based on [the docs](https://www.i18next.com/overview/api):
> The default export of the i18next module is an i18next instance ready to be initialized by calling init. You can create additional instances using the [createInstance](https://github.com/overview/api#createinstance) function.
>
> Do not call init multiple times.
> To change language use [changeLanguage](https://github.com/overview/api#changelanguage). If you need complete different configs use [createInstance](https://github.com/overview/api#createinstance) or [cloneInstance](https://github.com/overview/api#cloneinstance).

We call `i18next.init` in several places, but the docs instead suggest to create separate instances if we need a different config, or call `init` once and wait until it initialized. I also tried using `i18next.on('initialized', callback)`, but my callback wasn't getting called. Maybe has to do with the multiple calls to `i18next.init` confusing the listener?

### Automated Tests

none

### QA Notes

Same testing notes as https://github.com/rstudio/rstudio/pull/13518, with an added step of changing the language to French in global settings, then restarting the IDE to ensure you are prompted to install rosetta in French.

Note that this dialog should not appear on Linux or Windows, since Rosetta is Mac-only. I have tested in a Linux Ubuntu 22.04 VM that the check does not occur.

### Documentation
none

### Checklist

- [x] If this PR adds a new feature, or fixes a bug in a previously released version, it includes an entry in `NEWS.md` already covered in https://github.com/rstudio/rstudio/pull/13518


timtmok (approved):
It would be good to test on Linux if the dialog appears behind the main window due to the [Electron bug with dialogs](https://github.com/electron/electron/issues/32857)

sharon-wang:
> It would be good to test on Linux if the dialog appears behind the main window due to the [Electron bug with dialogs](https://github.com/electron/electron/issues/32857)

If the dialog appears at all on Linux or Windows, then something has gone wrong 😜 We only do the Rosetta check on Mac!

I'll add a testing note that the dialog shouldn't appear on Linux or Windows at all. I've tested on a Linux VM that the Rosetta check doesn't occur.

kevinushey (approved):
LGTM!

Issue #13433, opened 2023-07-31
Cloud publishing text is too long for its box


ronblum:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2023.09.0-daily+285
    OS Version      : 13.5
    R Version       : 4.3.1

### Steps to reproduce the problem

Try to publish an R Markdown document. 

### Describe the problem in detail

The text for Posit Cloud is too long and gets cut off.

<img width="574" alt="new_cloud_publish" src="https://github.com/rstudio/rstudio/assets/31009952/128ab735-692a-4631-b24a-bb4f9c2fa2bb">

### Describe the behavior you expected

Shorten the text so that it fits on two lines. I recommend leaving out the phrase "Create your free account today." It comes across as an ad, even though it's for a free (or freemium) account.

- [ ] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ ] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [ ] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
Everything looks fine on my side:

<img width="596" alt="Screenshot 2023-09-05 at 3 14 00 PM" src="https://github.com/rstudio/rstudio/assets/1976582/74cfed19-4528-46c8-8b43-0029c633fedd">

I sort of wonder if we should just remove the "Create your free account today." marketing blurb though.

ronblum:
I concur with @kevinushey regarding removing the marketing blurb—it feels like an explicit upsell.

jgutman:
See https://github.com/rstudio/rstudio/pull/13440 where I addressed the original issue

ronblum:
@jthomasmock What are your thoughts about "Create your free account today?"

jthomasmock:
Chatting w/ Alex Chisholm internally re: Cloud. Let's hold on merging the PR until I can confirm w/ him.

Alex-Chisholm:
Hi @ronblum. I think mentioning free is needed - as it is done with rpubs - but I'm okay softening the language. 

How about changing "Create your free account today" to "Free plans available".

jonvanausdeln:
[...]

PR #13599, opened 2023-09-06
tweak language

kevinushey:
Addresses https://github.com/rstudio/rstudio/issues/13433.


jgutman (approved):
I thought this language was pretty weird too, but this was literally the language the Posit Cloud PM handed me. Then I noticed it being truncated, and I asked for a shorter version, and I was given the current version, which shows up correctly for me as it does for you. @ronblum seems to have the older version, and given the issue was from July I think it was before my changes/fix to the language

Not sure if we should just double check in with the Cloud team and PM again though because they did provide the language and review/sign off on it after my changes earlier this summer

jgutman:
Right, I merged this change on August 1, unaware that Ron filed this issue on Jul 31. But if we want to remove that language anyway I suppose we can

PR #13631, opened 2023-09-14
Fix no space left on disk creating dmg

skyeturriff:
mac builds are still failing on `main`, and only on `main`. The installer disk image seems to run out of space

```bash
[2023-09-12T16:45:47.726Z] [I] Creating 'RStudio-2023.12.0-hourly-8.dmg ...'

[2023-09-12T16:46:03.310Z] could not access /Volumes/RStudio-2023.12.0-hourly-8/RStudio.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework - No space left on devicehdiutil: create failed - No space left on device
```

This is an attempt to fix that following what we did last time, though I'm not sure why other branches are okay. It seems to only affect electron builds, as [non-electron flavour](https://build.posit.it/blue/organizations/jenkins/IDE%2FPro-Builds%2FPlatforms%2Fmacos-pipeline/detail/main/445/pipeline) seems to build okay.

### DIFF
--- a/package/osx/make-package
+++ b/package/osx/make-package
@@ -454,7 +454,7 @@ if [ "${build_package}" = "1" ]; then
 
       # allocate oversized dmg, which will be compressed
       hdiutil create                       \
-         -size 1g                          \
+         -size 2g                          \
          -fs "APFS"                        \
          -volname "${RSTUDIO_BUNDLE_NAME}" \
          -srcfolder "install"              \


timtmok (approved):
Electron builds are almost 2x bigger due to a second set of arm64 binaries. Electron builds are around 350 MB so I'm not sure what's happening with the DMG size.

skyeturriff:
Aah!... are we only building that second set of arm64 binaries on `main` currently?

timtmok:
> Aah!... are we only building that second set of arm64 binaries on `main` currently?

I'm not really sure what changed with `main` to cause a space issue. The Electron build has been compiling the arm64 binaries for probably over a year now.

Issue #13635, opened 2023-09-14
Copilot: don't store project preferences in .Rproj file


kevinushey:
Right now, Copilot project settings are serialized to the project `.Rproj` file. This has two big consequences:

1. It forces all users of the project to opt-in to the same Copilot project preferences,
2. The changes become visible in version control.

However, because different users working on the same project could potentially want to have differing Copilot project preferences here, I think we need to change this to instead store these within the `.Rproj.user` folder.

ronblum:
Verified that data is stored, but no longer in `.Rproj`, in RStudio Desktop 2023.09.0+445 on MacOS 14.

PR #13638, opened 2023-09-14
use private settings location for copilot settings

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/13635.

### Approach

Rather than serialize the Copilot project options as part of the project file, we instead serialize them to their own settings file within the `.Rproj.user` state folder. While there are a lot of changes required to make this happen, the changes are relatively straightforward.

### Automated Tests

N/A

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/13635. In particular, confirm that the `.Rproj` file is not updated when changing these project options.

### Documentation

N/A

### Checklist

### DIFF
--- a/src/cpp/core/r_util/RProjectFile.cpp
+++ b/src/cpp/core/r_util/RProjectFile.cpp
@@ -1005,21 +1005,6 @@ Error readProjectFile(const FilePath& projectFilePath,
       pConfig->spellingDictionary = it->second;
    }
    
-   // extract copilot fields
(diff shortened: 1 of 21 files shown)


jgutman:
This looks good to me and works as expected. I don't love that we write out the settings as 0 1 2 rather than default yes no, but I realize that's what we've always done elsewhere and the .Rprojuser folder is not intended to be inspected by users so it doesn't matter.

Issue #13577, opened 2023-08-31
Update RStudio dependencies


kevinushey:
Typically, with new releases, we try to stay relatively current on our dependencies.

This is also important as others will attempt to package RStudio themselves for different Linux distributions, and so typically build against the latest versions of dependencies (even if this is something we don't explicitly support).

## Compile-time Dependencies
- [x] Boost
- [x] CMake
- [x] Soci (already up-to-date)
- [x] yaml-cpp

## Runtime Dependencies
- [x] Node / npm
- [x] Electron

petetronic:
See also https://github.com/rstudio/rstudio/issues/12424

jgutman:
We also have the following ticket for updating Ace: #13592

gtritchie:
Another one to consider is xterm (the terminal emulator); we're many, many years out of date on this. https://github.com/rstudio/rstudio/issues/9195

This is a non-trivial upgrade; a lot of the APIs have changed and some wiseguy used a lot of internal implementation details of xterm to work around limitations at the time and that needs to be unravelled and switched to official APIs when possible.

Seems like suitable punishment for me to do it, if I get the opportunity!

ronblum:
Verified that Electron has been updated to 26.2.4 in RStudio Desktop 2023.09.1+470 and 2023.12.0-daily+131 on MacOS 14.

petetronic:
Moving the remaining work to Chocolate Cosmos

jgutman:
What is the remaining work here? I think the only things not done in Ocean Storm mentioned here were:
yaml-cpp upgrade (currently 0.6.3, latest available 0.8.0)
Upgrade to Node 18 (currently 16.20.2)
Upgrade to openssl 3 (tracked in #12855)
Upgrade xterm (tracked in #9195)

And I believe we determined we won't or couldn't do:
Upgrade to GWT 2.10.0 (tracked in #11505 - currently in the Later milestone)
Allow IDE to be built with JDK 17/18 (tracked in #11242 - currently in the Chocolate Cosmos milestone but I believe the latest update was that it may be intractable)

gtritchie:
I haven't completely given up on building IDE GWT with JDK 17, but so far I've not been successful (the issue is all related to the Gin dependency injection framework, which uses the Java Guice framework at build time). The theory is if we upgrade the version of Guice to one that supports JDK 17 it _might_ work. But I couldn't get it to work at all with any version of Java.

kevinushey:
[...]

PR #13641, opened 2023-09-15
Update to Boost 1.83.0

kevinushey:
Part of https://github.com/rstudio/rstudio/issues/13577.

Did some light testing locally; confirmed things compile and run as expected on macOS + Windows.


dependencies/windows/install-boost/install-boost.R
```diff
@@ -122,7 +144,7 @@ b2_build_args <- function(bitness) {
       "--abbreviate-paths",
       sprintf("variant=%s", variant),
       sprintf("link=%s", link),
-      sprintf("runtime-link=shared", link),
```
kevinushey, line 125: The formatting argument was being ignored here, so in fact we were always using shared linking to the runtime. Because we do ship the required runtime libraries (necessary for Electron) we don't need to perform static runtime linking here.

dependencies/windows/install-boost/install-boost.R
```diff
@@ -91,10 +92,31 @@ invisible(lapply(docs, function(doc) {
 
 # bootstrap the boost build directory
 section("Bootstrapping boost...")
+Sys.setenv(MSVCDir = "C:/Program Files/Microsoft Visual Studio/2022/Community/VC")
+Sys.unsetenv("MSVCDir")
```
jgutman, line 96: What does this do?

src/cpp/core/zlib/zconf.h
jgutman, line 1: Why is this being removed now?

jgutman (commented):
Do we usually add the boost update to NEWS? Do you know how long we have before ocean storm release to kick the tires on this upgrade?

dependencies/windows/install-boost/install-boost.R
```diff
@@ -91,10 +92,31 @@ invisible(lapply(docs, function(doc) {
...
+
+# TODO: Boost has trouble finding the vcvarsall.bat script for some reason?
+# We set this environment variable here to help it find the tools.
+if (is.na(Sys.getenv("B2_TOOLSET_ROOT", unset = NA))) {
+   
+   candidates <- c(
+      "C:/Program Files/Microsoft Visual Studio/2022/Community/VC/"
```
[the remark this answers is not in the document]
kevinushey: I can check; I think we do something similar in other code paths...

src/cpp/core/zlib/zconf.h
kevinushey: It's generated by CMake here:

https://github.com/rstudio/rstudio/blob/a423cc85cc50356d9b1f2c3b0b580e0670ccf01f/src/cpp/core/zlib/CMakeLists.txt#L59-L60

And interestingly enough, it's part of the directories `.gitignore`:

https://github.com/rstudio/rstudio/blob/a423cc85cc50356d9b1f2c3b0b580e0670ccf01f/src/cpp/core/zlib/.gitignore#L1

so it seems like it must've been committed by accident some time long ago.

PR #13645, opened 2023-09-15
Fix a 502 on login when upgrade downgrade upgrade occurs

MariaSemple:
### Intent

Addresses [Pro 5138](rstudio/rstudio-pro/issues/5138).

This was caused by a null value in the posix_name field. The AlterTables files set the posix_name and display_name fields, but they only apply on the first upgrade.

### Approach

If we find a null posix_name, update the posix_name and the display_name fields to the user's posix username.

### Automated Tests

We don't currently have automated tests for the upgrade/downgrade/upgrade scenario.

### QA Notes

May be worthwhile to see what happens if the user is actively logged in when the server is upgraded.

### Documentation
N/A

### Checklist

### DIFF
--- a/src/cpp/server/auth/ServerAuthHandler.cpp
+++ b/src/cpp/server/auth/ServerAuthHandler.cpp
@@ -373,6 +373,28 @@ void removeExpiredCookies()
    }
 }
 
+Error updateUser(const boost::shared_ptr<IConnection>& connection,
+                 const system::User& user)
+{
+   LOG_DEBUG_MESSAGE("Updating user in database: " + user.getUsername());
+
+   const std::string& username = user.getUsername();
+   Query updateQuery = connection->query("UPDATE licensed_users SET posix_name = :pn, display_name = :dn WHERE user_id = :uid")
+         .withInput(username)
+         .withInput(username)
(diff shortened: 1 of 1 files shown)


jeffvroom (approved):
LGTM!

src/cpp/server/auth/ServerAuthHandler.cpp
```diff
@@ -373,6 +373,28 @@ void removeExpiredCookies()
    }
 }
 
+Error updateUser(const boost::shared_ptr<IConnection>& connection,
+                 const system::User& user)
+{
+   LOG_DEBUG_MESSAGE("Updating user in database: " + user.getUsername());
+
+   const std::string& username = user.getUsername();
+   Query updateQuery = connection->query("UPDATE licensed_users SET posix_name = :pn, display_name = :dn WHERE user_id = :uid")
```
zachhannum, line 382: Since the `posix_name` column doesn't exist in open source, would it make more sense to make these changes in Pro?

For the future, we can make an effort to get this into the overlay code that Skye is introducing in https://github.com/rstudio/rstudio/pull/13610

jgutman:
Merged in Pro

PR #13647, opened 2023-09-16
Limit rate of writes to 'executing' property metadata

jeffvroom:
### Intent

Addresses: https://github.com/rstudio/rstudio-pro/issues/4922

### Approach

In Posit Workbench, updating the 'executing' metadata property causes a file system write. If that happens too frequently it causes performance issues, especially with some customers that have network file systems that limit the number of writes to a file per second. 

In this PR the rate of writes it not controlled by the number of console-input statements but instead limited to a few times a second. The transition to the executing state happens right away, but the transition from executing to idle happens in the offline service at no more than 4X second, so no more than 8X/second.

### Automated Tests

I don't think we have any automated tests for the 'executing' state so that might be worth doing?  

### QA Notes

Capture an strace of the rsession while running a script like: 

```
system("sleep 0.01")
system("sleep 0.01")
system("sleep 0.01")
...
(200 or 300 times)
```

Count the number of 'open' calls to the executing file in the properites directory.  Should see no more than 8X/second with the PR. Without it, you'll see one per line.

### DIFF
--- a/src/cpp/session/SessionConsoleInput.cpp
+++ b/src/cpp/session/SessionConsoleInput.cpp
@@ -51,17 +51,55 @@ ConsoleInputQueue s_consoleInputBuffer;
 
 // manage global state indicating whether R is processing input
 volatile sig_atomic_t s_rProcessingInput = 0;
+// the saved version of s_rProcessInput - the 'executing' property in the session metadata
+volatile sig_atomic_t s_sessionExecuting = 0;
+// Controls access to s_sessionExecuting and the file system
+boost::mutex s_sessionExecutingMutex;
 
 // last prompt we issued
 std::string s_lastPrompt;
 
+void setSessionExecuting(bool executing)
+{
+   LOCK_MUTEX(s_sessionExecutingMutex)
+   {
+      if (s_sessionExecuting != executing)
+      {
+         s_sessionExecuting = executing;
+         // Not ideal to be saving the file while holding the lock but
+         // it's only used from two threads - the main thread, and an offline
+         // thread that runs infrequently. If we do this outside the mutex, the writes
+         // could be reordered and the saved executing state would be incorrect until it changed again.
+         module_context::activeSession().setExecuting(executing);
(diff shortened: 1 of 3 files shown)


kevinushey (commented):
I feel like there's a simpler solution here where we take this line of code:

https://github.com/rstudio/rstudio/blob/1fb8c970fb212301ce2e91834f2b68deb7827169/src/cpp/session/SessionConsoleInput.cpp#L64

and move it elsewhere, using something like `scheduleIncrementalWork()` or `schedulePeriodicWork()` to synchronize the `s_rProcessingInput` flag with the filesystem at periodic intervals. I think that would eliminate the need for a second global variable (and maybe the mutex could then go away as well?)

I'm not sure what the right place for that to live is, though... maybe just `SessionModuleContext.cpp`?

jeffvroom:
Yeah, good point.  We could use schedulePeriodicWork rather than the offline request thread to save of the `executing` file.  

There are two reasons I didn't do it that way:
* "scheduledCommands" is a vector and we iterate through that pretty often. 
* It's less reliable because it runs on the main thread (and requires pollHandler to be called by the R code)

But It would let us get rid of the mutex so maybe it would be faster in the long run (and a little less complicated)?  

If we keep the code that eagerly sets executing=1, I don't think there's a risk the session gets stuck with executing=0 by some poorly behaved R code.  I do think we need that 2nd variable though to shadow what we last saved, or else read/write the file each time in the poll handler.

I'm happy to switch to using schedulePeriodicWork if you think that's the better way to go!

kevinushey:
> It's less reliable because it runs on the main thread (and requires pollHandler to be called by the R code)

That's a good point. That implies we might never see a process running `system("sleep 3600")` as "busy", since the R polled events handler doesn't get called in that scenario. To wit, I think we do need to allow an independent thread to update this state.

I think your PR is its current state is good -- thanks!

kevinushey (approved):

Issue #13534, opened 2023-08-22
RStudio checks if session is suspendable too eagerly


kevinushey:
Brought in from https://github.com/rstudio/rstudio/pull/13529; currently, we check if the session is suspendible quite eagerly:

```
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 1.1
  * frame #0: 0x0000000102ea6dd8 rsession`rstudio::session::console_input::(anonymous namespace)::canSuspend(prompt="> ") at SessionConsoleInput.cpp:106:58 [opt]
    frame #1: 0x00000001030ac3cc rsession`rstudio::session::suspend::checkForTimeout(rstudio_boost::function<bool ()> const&) [inlined] rstudio_boost::function0<bool>::operator(this=<unavailable>)() const at function_template.hpp:763:14 [opt]
    frame #2: 0x00000001030ac3b4 rsession`rstudio::session::suspend::checkForTimeout(allowSuspend=0x000000016cf8cf08)> const&) at SessionSuspend.cpp:361:22 [opt]
    frame #3: 0x00000001030ac910 rsession`rstudio::session::suspend::checkForSuspend(allowSuspend=0x000000016cf8cf08)> const&) at SessionSuspend.cpp:491:4 [opt]
    frame #4: 0x0000000102f637e4 rsession`rstudio::session::http_methods::waitForMethod(method="console_input", initFunction=0x000000016cf8ce98, allowSuspend=0x000000016cf8cf08, pRequest=0x000000016cf8cf68)> const&, rstudio_boost::function<bool ()> const&, rstudio::core::json::JsonRpcRequest*) at SessionHttpMethods.cpp:473:7 [opt]
    frame #5: 0x0000000102ea5808 rsession`rstudio::session::console_input::rConsoleRead(prompt=<unavailable>, addToHistory=true, pConsoleInput=<unavailable>) at SessionConsoleInput.cpp:342:24 [opt]
[...]

PR #13648, opened 2023-09-17
Reduce number of calls to canSuspend/allowSuspend

jeffvroom:
### Intent

Addresses: https://github.com/rstudio/rstudio/issues/13534

### Approach

Split the checkForSuspend method into two pieces:
  1) run before processing each request (like today) to check for explicit suspension - it only
calls allowSuspend if the user requested suspension
  2) another that's run to check for idle suspension that's run periodically.
The rsession waitForMethod, poll handler has a 50ms timeout for background processing while waiting for r console input.  Only if 20 of these go by now do we check for an idle session (i.e. 1/second).

Are there scenario where console input is being processed and we need to allow suspension?  Could use a timer here but trying to avoid extra system calls since this is an inner loop for running R code.

Or we could a check in the offline polling thread so it's done regularly, not based on the rate of console input.

### QA Notes

The key things to test here are that sessions still suspend when idle. This is an important use case that we ignore certain types of background activity but reliably suspend when the user has not interacted with the system for a while.

This PR doesn't change the core suspension logic, but reduces the rate of the 'check for timeout' calls so we need to make sure those calls are still happening in the edge cases for session suspension.

### DIFF
--- a/src/cpp/session/SessionConsoleInput.cpp
+++ b/src/cpp/session/SessionConsoleInput.cpp
@@ -101,21 +101,20 @@ void consolePrompt(const std::string& prompt, bool addToHistory)
 
 bool canSuspend(const std::string& prompt)
 {
-   bool suspendIsBlocked = session::suspend::checkBlockingOp(main_process::haveDurableChildren(), suspend::kChildProcess);
-   if (!suspendIsBlocked)
-      suspendIsBlocked = session::suspend::checkBlockingOp(!modules::jobs::isSuspendable(), suspend::kActiveJob);
-   if (!suspendIsBlocked)
-      suspendIsBlocked = session::suspend::checkBlockingOp(!rstudio::r::session::isSuspendable(prompt), suspend::kCommandPrompt);
-
-   if (!suspendIsBlocked && session::options().sessionConnectionsBlockSuspend())
-      suspendIsBlocked = session::suspend::checkBlockingOp(!modules::connections::isSuspendable(), suspend::kConnection);
+   bool suspendIsBlocked = false;
    
-   if (!suspendIsBlocked && session::options().sessionExternalPointersBlockSuspend())
-      suspendIsBlocked = session::suspend::checkBlockingOp(!modules::environment::isSuspendable(), suspend::kExternalPointer);
-   
-   if (!suspendIsBlocked)
-      suspendIsBlocked = !modules::overlay::isSuspendable();
+   suspendIsBlocked |= session::suspend::checkBlockingOp(main_process::haveDurableChildren(), suspend::kChildProcess);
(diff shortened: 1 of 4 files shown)


src/cpp/session/SessionConsoleInput.cpp
```diff
@@ -101,20 +101,21 @@ void consolePrompt(const std::string& prompt, bool addToHistory)
 
 bool canSuspend(const std::string& prompt)
 {
-   bool suspendIsBlocked = false;
```
kevinushey, line 104: I thought we intentionally used `|=` below because we do want to know (for each of these operations) whether it's blocking suspend or not, even if there are multiple operations blocking suspend. @cm421 does that sound true?

jeffvroom: Oh, that makes sense. We are using those methods to track why it won't suspend?  I should have dug into that code in more depth.  I will revert this optimization as it's probably not helpful anyway since most of the time, they all return false.

src/cpp/session/SessionSuspend.cpp
```diff
@@ -486,9 +487,6 @@ void checkForSuspend(const boost::function<bool()>& allowSuspend)
       // errors will be logged/reported internally and we will move on
       suspendSession(false);
    }
-
-   // timeout suspend
-   checkForTimeout(allowSuspend);
```
cm421, line 491: It's been so long, I can't remember exactly, but moving this from the beginning of the `waitForMethod` loop to the end is raising a red flag in my head. I vaguely recall hitting issues when I moved this line around.

If you've been testing this locally and it seems to be running fine then maybe we're okay, but once this builds I'll want to run through some tests with it.

cm421 (changes requested):

src/cpp/session/SessionSuspend.cpp
```diff
@@ -356,14 +356,14 @@ bool suspendSession(bool force, int status)
    return r::session::suspend(force, status, session::options().ephemeralEnvVars());
 }
 
-void checkForTimeout(const boost::function<bool()>& allowSuspend)
+void checkForIdleSuspend(const boost::function<bool()>& allowSuspend)
 {
-   bool canSuspend = allowSuspend();
```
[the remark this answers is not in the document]
jeffvroom: Thanks, I reverted that then!

src/cpp/session/SessionSuspend.cpp
jeffvroom: Yes, it seems to suspend an idle session just fine and I've tested one of the ways to block suspension. I could try more ways to block suspension, but we didn't really change that logic.  I also used temporary debug messages to be sure that key canSuspend() is called at the right times and with a lower volume.

cm421 (approved):
LGTM!

kevinushey (approved):
LGTM!

Issue #13636, opened 2023-09-14
knitr chunk option `file` is not respected by interactive rendering


cderv:
### System details

    RStudio Edition : Desktop [Open Source]
    RStudio Version : 2023.12.0.14
    OS Version      : Windows 11 x64 (build 22621)
    R Version       : R version 4.3.1 (2023-06-16 ucrt)

### Steps to reproduce the problem

* Write the Rmd below using a `file` chunk option to load content of external file  (https://bookdown.org/yihui/rmarkdown-cookbook/option-code.html)

````markdown
---
title: "my project"
output: html_document
---

```{r}
xfun::write_utf8("library(glue)", "setup.R")
```

```{r, file = "setup.R"}
```

```{r}
glue("1 + 1 = {1 + 1}")
```

```{r}
unlink("setup.R")
```
````

* Knit the document => this is working fine
* Use Run All for interactive rendering and the chunk with **glue** will error as `library(glue)` as not be called because `setup.R` not be inserted in chunk. 

cc @topepo

### Describe the problem in detail

RStudio IDE does not know probably how to handle this chunk option, neither in R Markdown or in Quarto (using the in-chunk YAML syntax). 

Probably related to other issues where some **knitr** option are not known or handled correctly when running chunks interactively. 

### Describe the behavior you expected

The same behavior as in rendering where 
* `file` option is seen 
* R file are source as chunk content
* No error in rendering

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
Verified in RStudio Desktop 2023.12.0-daily+84 on MacOS 14 using the original example.

Issue #13525, opened 2023-08-21
Enable RStudio IDE to account for a global database connection set with knitr opts_chunks$set() for interactive execution of a SQL chunk.


rressler:
The goal is to have the RStudio IDE recognize the knitr chunk option for setting the database connection so it is recognized when a chunk is executed individually without having to identify the connection in each chunk. Reference is the R Markdown Definitive Guide. 

The solution would allow for each chunk to use the document set up option as the default connection unless overridden by a connection option within the chunk. Thus a document that uses only a single connection would only need it specified in the set up chunk.

Currently the global option works for a rendering but not interactive. The following picture shows an example where the interactive error message for the SQL chunk is on the left but the rendered document is on the right.
![image](https://github.com/rstudio/rstudio/assets/59033325/78e2a971-045b-4b63-831c-349239032b78)

@cderv provided an additional example and comments (See issue 13240).
Example:

`````
---
output: html_document
---

```{r setup}
library(DBI)
# Create an ephemeral in-memory RSQLite database
con <- dbConnect(RSQLite::SQLite(), ":memory:")
dbWriteTable(con, "mtcars", mtcars)
dbListTables(con)
knitr::opts_chunk$set(connection = con)
```

```{sql}
SELECT * FROM mtcars WHERE cyl = 4
```
`````

Running interactively the second chunk will throw

> Error in .rs.runSqlForDataCapture("SELECT * FROM mtcars WHERE cyl = 4",  : 
>  The 'connection' option (DBI connection) is required for sql chunks.

RStudio IDE is not using knitr when doing interactive execution, and it is not aware of the knitr globally set chunk option.

Rendering the document works because knitr will account for the globally set chunk options.

ronblum:
@jthomasmock Can you provide your thoughts?

jthomasmock:
Thanks for the feature request! Parsing `knitr` options interactively is a fairly complex task but we'll evaluate that as we continue to develop RStudio.

While this is a bit of a workaround, I have used code snippets to make this easier and I always use `con` as the name of the connection.

You can create your own snippets as seen at: https://docs.posit.co/ide/user/ide/guide/productivity/snippets.html

Note, that in RMarkdown or Quarto you will insert them with the <kbd>Shift+Tab</kbd> shortcut.

Here is the snippet I use, and you can save it under the Edit Snippets > Markdown.

````
snippet sql
  ```{sql, connection=con}
  ${1}
  ```
````

[...]

Issue #13326, opened 2023-07-06
Warnings are thrown when running a R chunks interactively in Quarto document


cderv:
Initially asked at 
* https://github.com/yihui/knitr/issues/2267 
* https://github.com/quarto-dev/quarto/issues/211

Other related issues

* https://github.com/quarto-dev/quarto-cli/issues/6341

### System details

    RStudio Edition : Desktop [Open Source]
    RStudio Version : 2023.8.0.196
    OS Version      : Windows 11 x64 (build 22621)
    R Version       : R version 4.3.1 (2023-06-16 ucrt)

### Steps to reproduce the problem

* Take this Qmd document

````markdown
---
title: Table
format: html
---

```{r}
#| label: tbl-tab
#| tbl-cap: "mtcars table"
#| tbl-subcap:
#|  - "head = 5"
#|  - "tail = 5"
knitr::kable(head(mtcars, 5))
knitr::kable(tail(mtcars, 5))
```

Look at Table @tbl-tab, just look at it.

````

* Put in a .qmd document 
* Render interactively the chunk  => Warnings
* Render the document 

Then adapt the chunk opions removing `=`
````yaml
#| tbl-subcap:
#|  - "head is 5"
#|  - "tail is 5"
````

Then run interactively, and no more warnings. 

So it seems the parsing with the characters like `=` is causing wrong parsing. 

### Describe the problem in detail

Running the chunk will throw warnings where I don't think it should 
````r
Warning messages:
1: In value[[3L]](cond) : Failed to parse chunk options in body:
Error in parse(text = code, keep.source = FALSE): <text>:1:24: unexpected symbol
1: alist( label: tbl-tab1 tbl
                           ^

2: In opts$engine <- unlist(strsplit(rmdChunkOpts, split = "(\\s|,)+"))[[1]] :
  Coercing LHS to a list
````

One of the warning is definitely from IDE code base 
https://github.com/rstudio/rstudio/blob/a9a0652e3feac02a05caeead5d54e370f3ec4209/src/cpp/session/modules/SessionRmdNotebook.R#L686

Rendering the document is not throwing any warnings, so it seems this is because of some IDE processing probably

### Describe the behavior you expected

No warning shown in the IDE 

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
[...]

Issue #13240, opened 2023-06-14
Special engine chunk does not account for specific comment character in interactive mode


cderv:
### System details

    RStudio Edition : Desktop [Open Source]
    RStudio Version : 2023.8.0.33
    OS Version      : Windows 10 x64 (build 22621)
    R Version       : R version 4.2.2 (2022-10-31 ucrt)

### Steps to reproduce the problem

Takes this document  (you need to install `Lahman`,  `dbplyr` packages)

````
---
title: "SQL Cell Issue"
format: html
---


```{r}
#| include: false
# requires Lahman package
lahman_con <- dbplyr::lahman_sqlite()
```

```{r}
#| label: working-db
DBI::dbListTables(lahman_con)
```

```{sql}
--| label: join-using-example
--| connection: lahman_con
SELECT yearID, playerID, HR, salary
FROM Batting JOIN Salaries USING(yearID, playerID)
LIMIT 10
```

```{r}
#| include: false
DBI::dbDisconnect(lahman_con)
```
````

* You can render using `quarto render` without issue (or using the button) 

* Try running interactively the chunks - the `sql` chunk will error

### Describe the problem in detail

The `sql` chunk won't be working interactively. It will give this error 

````
Error in .rs.runSqlForDataCapture("--| label: join-using-example\n--| connection: lahman_con\nSELECT yearID, playerID, HR, salary\nFROM Batting JOIN Salaries USING(yearID, playerID)\nLIMIT 10",  : 
  The 'connection' option (DBI connection) is required for sql chunks.
Error in .rs.runSqlForDataCapture("--| label: join-using-example\n--| connection: lahman_con\nSELECT yearID, playerID, HR, salary\nFROM Batting JOIN Salaries USING(yearID, playerID)\nLIMIT 10",  : 
  The 'connection' option (DBI connection) is required for sql chunks.
````

This is because RStudio IDE is handling the `sql` engine itself, and does not account for the special comment characters for chunk options. 

In R Markdown and Quarto, the special in-chunk YAML syntax is using the comment chars of the language usually. 
This means that for a sql chunk this could be don using `--|`. In **knitr** and Quarto we are keeping a list of those possible value 

* https://github.com/yihui/knitr/blob/8403ca57946713bb532d340eed766a7e731125a3/R/parser.R#L216-L236
[...]

PR #13653, opened 2023-09-18
tweak detection of YAML chunk options in R Markdown documents

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/13326.
Addresses https://github.com/rstudio/rstudio/issues/13240.
Addresses https://github.com/rstudio/rstudio/issues/13525.
Addresses https://github.com/rstudio/rstudio/issues/13636.

### Approach

Previously, we used a rather ad-hoc heuristic to try and guess whether inline chunk options were YAML or R code. This PR takes a more permissive approach, whereby we first attempt to parse the code as YAML, and then fall back to R code if that fails.

For reference, the associated knitr code: https://github.com/yihui/knitr/blob/768db6f45525e15399724451df4153e900461e12/R/parser.R#L301-L312

### Automated Tests

N/A

### QA Notes

To be tested via community.

### Documentation

N/A

### Checklist

### DIFF
--- a/src/cpp/session/modules/SessionRmdNotebook.R
+++ b/src/cpp/session/modules/SessionRmdNotebook.R
@@ -614,31 +614,42 @@ assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
    .rs.fromJSON(.rs.base64decode(encoded))
 })
 
+
+.rs.addFunction("parseYamlOptImpl", function(opt)
+{
+   # read as YAML
+   opts <- .rs.fromYAML(opt)
+   
+   # parse truthy and falsy style yaml options
+   opts <- lapply(opts, function(value)
+   {
+      if (!is.character(value))
+         return(value)
+      if (tolower(value) %in% c("y", "yes", "on"))
+         return(TRUE)
+      if (tolower(value) %in% c("n", "no", "off"))
+         return(FALSE)
+      else
+         return(value)
+   })
+   
+   opts
+})
+
 .rs.addFunction("parseYamlOpt", function(opt)
 {
-    opt <- sub("^#\\|\\s*", "", opt)
-    if (any(grepl(".+=.+", opt))) # R-style chunk options
-    {
-        opts <- paste(opt, collapse = " ")
-        opts <- knitr:::parse_params(opts)
-    }
-    else
-    {
-        opts <- .rs.fromYAML(opt)
-        # parse truthy and falsy style yaml options
-        opts <- lapply(opts, function(value)
-           {
-             if (!is.character(value))
-               return(value)
-             if (tolower(value) %in% c("y", "yes", "on"))
-               return(TRUE)
-             if (tolower(value) %in% c("n", "no", "off"))
-               return(FALSE)
-             else
-               return(value)
-           })
-    }
-    return(opts)
+   # remove comment prefix
+   opt <- sub("^#\\|\\s*", "", opt)
+   
+   # first, attempt to parse the code as YAML, then fall
+   # back to the regular knitr params parser
+   tryCatch(
+      .rs.parseYamlOptImpl(opt),
+      error = function(cnd) {
+         opts <- paste(opt, collapse = " ")
+         knitr:::parse_params(opts)
+      }
+   )
 })
 
 .rs.addFunction("evaluateChunkOptions", function(code)
--- a/src/cpp/session/modules/SessionRmdNotebook.R
+++ b/src/cpp/session/modules/SessionRmdNotebook.R
@@ -15,12 +15,13 @@
 assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
 
 .rs.addFunction("extractRmdFromNotebook", function(notebook) {
+   
    # parse the notebook to get the text of the contained R markdown doc
    parsed <- try(rmarkdown::parse_html_notebook(notebook), silent = TRUE)
    if (inherits(parsed, "try-error") || is.null(parsed$rmd)) {
-    return("")
+      return("")
    }
-
+   
    # as this string will be eventually written to (and compared with) files
    # on disk, use native line endings
    return(paste(parsed$rmd,
@@ -38,9 +39,9 @@ assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
    "^[\t >]*```+\\s*$"
 })
 
-.rs.addFunction("reYamlOptChunkBegin", function()
+.rs.addFunction("reYamlOptChunkBegin", function(commentPrefix = "#")
 {
-   "^#\\| .*"
+   sprintf("^%s\\| .*", commentPrefix)
 })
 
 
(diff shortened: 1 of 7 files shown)


src/cpp/session/modules/SessionRmdNotebook.R
```diff
@@ -614,142 +615,197 @@ assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
...
-    opts <- lapply(opts, function(opt) {
-      if (identical(opt, as.name("T")))
-        TRUE
-      else if (identical(opt, as.name("F")))
-        FALSE
-      else
-        opt
-    })
+   # remove comment prefix
+   commentPrefixPattern <- sprintf("^%s\\|\\s*", commentPrefix)
```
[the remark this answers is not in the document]
jgutman: Also, should this be `%s+` rather than `%s*`

jgutman, line 643: Ah I see now, this is `%s` as in `#` and then `\\s*` as in some zero or more number of spaces. I think (just tested in Quarto) at least one space before the pipe is required so maybe this should be `^%s\\|\\s+`

src/cpp/session/modules/SessionRmdNotebook.R
```diff
@@ -614,142 +615,197 @@ assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
...
+         return(FALSE)
+      else
+         return(value)
+   })
+   
+   opts
 })
 
-.rs.addFunction("evaluateChunkOptions", function(code)
+.rs.addFunction("parseYamlOpt", function(opt, commentPrefix = "#")
```
jgutman, line 640: is the comment character ever not `#`?

src/cpp/session/modules/SessionRmdNotebook.R
```diff
@@ -614,142 +615,197 @@ assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
...
-    })
+   # remove comment prefix
+   commentPrefixPattern <- sprintf("^%s\\|\\s*", commentPrefix)
+   opt <- sub(commentPrefixPattern, "", opt)
+   
+   # check for YAML entry, mimicing code in knitr
+   if (grepl("^[^ :]+:($|\\s)", opt[[1]])) {
+      meta <- .rs.tryCatch(.rs.parseYamlOptImpl(opt))
+      if (!inherits(meta, "error"))
+         return(meta)
```
jgutman, line 650: So if it matches the regex but then fails to parse with `.rs.parseYamlOptImpl`, it will fall back to rmd style chunk params? does that make sense?

src/cpp/session/modules/SessionRmdNotebook.R
```diff
@@ -614,142 +615,197 @@ assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
...
+   # strip chunk indicators if present
+   matches <- unlist(regmatches(rmdChunkOpts, regexec(.rs.reRmdChunkBegin(), rmdChunkOpts)))
+   if (length(matches) > 1)
+      rmdChunkOpts <- matches[[2]]
+   
+   # attempt to parse parameters (swallow errors)
+   .rs.tryCatch({
+      
+      # if this is the setup chunk, it's not included by default
+      if (.rs.startsWith(rmdChunkOpts, "r setup"))
```
jgutman, line 678: just realizing, couldn't the setup chunk also be `r, setup`? it seems like we don't check for the comma?

src/cpp/session/modules/SessionRmdNotebook.R
```diff
@@ -614,142 +615,197 @@ assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
...
+      )
+      
+      opts <- .rs.mergeLists(opts, parsedParams)
+                             
+      # if a chunk option is provided in both body (yaml chunk options) and header (rmdChunkOpts)
+      # the body should override the header option
+      opts <- opts[unique(names(opts))]
+      
+      # convert T, F for R code chunk options to TRUE, FALSE as appropriate
+      opts <- lapply(opts, function(opt) {
```
jgutman, line 698: I realize this is from the pre-existing code, but this comment implies that we only convert T and F for R chunk options, but I don't see us actually checking the engine here?

src/cpp/session/modules/SessionRmdNotebook.R
jgutman: Okay, I see below we have alternatives for some engines like `"cc", "cpp", "go", "js", "node", "rcpp", "stan"` etc (which I haven't used, only seen the r python julia examples) but wondering why down below we set this as `#+` but here the default arg we use `#`.  Are multiple `#` allowed in this context?

src/cpp/session/modules/SessionRmdNotebook.R
```diff
@@ -614,142 +615,197 @@ assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
...
+         opts$connection,
+         knitr::opts_chunk$get("connection")
+      )
+      
+      # our chunk execution tools expect a character connection,
+      # so use a temporary variable to store the connection value
+      if (!is.null(conn) && !is.character(conn)) {
+         .rs.setVar("chunkConnection", conn)
+         opts$connection <- ".rs.chunkConnection"
+      }
```
jgutman, line 788: These changes were kind of hard to find, especially amidst all the whitespace changes -- this probably could have been its own PR, but thanks for fixing this

src/cpp/session/modules/SessionRmdNotebook.R
kevinushey: You're right, thanks for catching that!

src/cpp/session/modules/SessionRmdNotebook.R
kevinushey: Good question... it looks like only a single instance of the comment character is allowed, so e.g.

```
#| label: foo     # okay
##| label: foo    # not okay
```

I think I need to tweak the other usages then.

src/cpp/session/modules/SessionRmdNotebook.R
kevinushey: Hmm, maybe not... here's what `knitr` does:

https://github.com/yihui/knitr/blob/42f6b3ad75f8245d26c1de3d3e0b9fa018566af3/R/parser.R#L301-L312

So we should probably just fail in a similar way.

src/cpp/session/modules/SessionRmdNotebook.R
kevinushey: Good idea -- instead of checking the chunk text itself, we should just check if it has a `setup` label.

src/cpp/session/modules/SessionRmdNotebook.R
kevinushey: I think this is intentional, since the chunk options would either be specified as R code, or (typically, for inline chunk options) as YAML. That's independent of the engine that was chosen for the code in the chunk

src/cpp/session/modules/SessionRmdNotebook.R
kevinushey: Sorry about that -- since I was in this part of the code, I wanted to dig through other related pieces.

PR #13654, opened 2023-09-18
sprintf_linter() fixes

MichaelChirico:
Found during an exploratory {lintr} search: https://github.com/r-lib/lintr/issues/2183

Took a guess at what belongs in `%s`, PTAL to confirm.


kevinushey:
Thanks!

Issue #13657, opened 2023-09-19
Popup or detachable panels do not load


andresrcs:
### System details

    RStudio Edition : Server
    RStudio Version : 2023.09.0 Build 446 for arm64
    OS Version      : Ubuntu Server 22.04 LTS for arm64
    R Version       : 4.3.1

### Steps to reproduce the problem
On a git enabled project open the Diff panel or detach any panel from the original browser window for the session.

### Describe the problem in detail
The popup or detachable panel stays loading forever
![image](https://github.com/rstudio/rstudio/assets/14981085/bd0b2e49-2a4a-4b77-afe0-0b5d4e6ea07a)

### Describe the behavior you expected
I expect to see the panel loaded like with the stable RStudio Server version (2023.06.2 Build 561 for x86_64)
![image](https://github.com/rstudio/rstudio/assets/14981085/9636d1e3-63ae-4d3f-a8bb-e63125066d06)

- [X] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [X] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [X] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
Verified in
- RStudio Desktop 2023.09.0+448 on Debian 12
- RStudio Desktop 2023.09.0+448 on MacOS 14
- RStudio Desktop 2023.09.0+448 on Windows 10
- RStudio Server 2023.09.0+448 on Ubuntu 20 via Chrome on MacOS 14

Tested with Git Diff pane and rendering an R Markdown file with preview in a window.

andresrcs:
The fix is confirmed, thanks!

PR #13660, opened 2023-09-19
ensure satellite prefs are sync'ed on init

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/13657.

### Approach

Ensure that user preferences are loaded when the session is initialized.

As to why this change is needed... my only guess is that something changed with the order that GWT is trying to initialize some of the static / singleton classes we have, and now we're ending up in a state where the UserPrefs class is being initialized before the preferences are actually available, since the constructor seems to assume the preferences are already available:

https://github.com/rstudio/rstudio/blob/8f63e7b629740429f07d6683bdb9e13331f92f5a/src/gwt/src/org/rstudio/studio/client/workbench/prefs/model/UserPrefs.java#L72-L75

There's an (unfortunately implicit) contract that anything which wants to make use of certain elements like user preferences need to wait for the session init event. Since the session init event is fired relatively early, we _usually_ get away without requiring sequencing based on that event, but it does look like things can go wrong sometimes.

For reference:

https://github.com/rstudio/rstudio/blob/8f63e7b629740429f07d6683bdb9e13331f92f5a/src/gwt/src/org/rstudio/studio/client/common/satellite/Satellite.java#L254-L256

### Automated Tests

Should be tracked by existing automation.

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/13657. 

### Documentation

N/A

### Checklist

### DIFF
--- a/src/gwt/src/org/rstudio/studio/client/workbench/prefs/model/UserPrefs.java
+++ b/src/gwt/src/org/rstudio/studio/client/workbench/prefs/model/UserPrefs.java
@@ -14,12 +14,6 @@
  */
 package org.rstudio.studio.client.workbench.prefs.model;
 
-import com.google.gwt.core.client.JsArray;
-import com.google.gwt.core.client.Scheduler;
-import com.google.inject.Inject;
-import com.google.inject.Singleton;
-import com.google.gwt.core.client.GWT;
-
 import org.rstudio.core.client.CommandWithArg;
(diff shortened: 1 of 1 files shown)


sharon-wang (approved):
LGTM!

Satellite Windows better not sneak in calls to UserPrefs before SessionInit! 🤨😅

PR #13668, opened 2023-09-20
use in-memory cache for chunk console outputs

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/8034.

### Approach

Rather than eagerly writing each piece of console output as it's received, we now instead buffer those console outputs, and only write the console output to disk after chunk execution has finished. We also make sure those outputs are flushed to disk when a new client has connected, so that those clients will see an up-to-date view of any currently-running executions.

### Automated Tests

N/A

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/8034.

### Documentation

N/A

### Checklist

### DIFF
--- a/src/cpp/session/SessionClientInit.cpp
+++ b/src/cpp/session/SessionClientInit.cpp
@@ -179,6 +179,9 @@ Error makePortTokenCookie(boost::shared_ptr<HttpConnection> ptrConnection,
 void handleClientInit(const boost::function<void()>& initFunction,
                       boost::shared_ptr<HttpConnection> ptrConnection)
 {
+   // notify that we're about to initialize
+   module_context::events().onBeforeClientInit();
+   
    // alias options
    Options& options = session::options();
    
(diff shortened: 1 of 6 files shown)


jeffvroom (commented):
Thanks for doing this!  I think it will make a big difference, especially on those file systems that introduce delays when repeatedly writing the same file (ala that "set executing" issue). Am I missing a case in your fix as it seems like the output will only be flushed when a client disconnects, and later when one connects again.  Don't we want to periodically save that info, like once per second?  What if the session suspends while disconnected?

kevinushey:
> Am I missing a case in your fix as it seems like the output will only be flushed when a client disconnects, and later when one connects again.

We also flush outputs whenever we finish executing code in a chunk -- that gets handled here.

https://github.com/rstudio/rstudio/pull/13668/files#diff-5aa31fb701cd5557b582471ef213443a5cf4259100b318c2835c4fa05a204a93R505

In other words, for a typical use case, execution of a chunk might produce quite a bit of console output but those now all get collected during execution, and then later written out to disk at once after the chunk has finished execution. However, because chunk outputs are still streamed to the client as they are produced, users will still see console output presented as it is generated as well. Does that make sense?

jeffvroom (approved):
I see now. LGTM!

PR #13687, opened 2023-09-25
Fix issue with session protocol debug

jeffvroom:
### Intent

Addresses: https://github.com/rstudio/rstudio-pro/issues/5095

Speedup for LOG_DEBUG_MESSAGE macro

Apis need by an admin page (in development) to turn on/off logging in workbench. 

### Approach

Don't evaluate message unless debugging is enabled.  The macro now uses a ? operator to hide the logDebugMessage call while still making a valid statement.  I tried a few ways to do this and this turned out to be the only one that's robust inside of if/else statements.

Apis to get/set log level for file/stderr destinations

Exposed name to LogLevel mapping for an upcoming rserver admin to change levels on the fly

### Automated Tests

No new automation required. 

### QA Notes

* With this fix, turning on "Session protocol debug" in a session should enable debug logging in ~/.local/share/rstudio/log/rsession-username.log.  
* Debug logging should be unaffected by the macro change

News TBD.

### DIFF
--- a/src/cpp/core/include/core/Log.hpp
+++ b/src/cpp/core/include/core/Log.hpp
@@ -82,7 +82,9 @@ std::string errorAsLogEntry(const Error& error);
 #define LOG_INFO_MESSAGE_NAMED(logSection, message) rstudio::core::log::logInfoMessage(message, \
                                                                                        logSection)
 
-#define LOG_DEBUG_MESSAGE(message) rstudio::core::log::logDebugMessage(message)
+
+#define LOG_DEBUG_MESSAGE(message) (rstudio::core::log::isLogLevel(rstudio::core::log::LogLevel::DEBUG_LEVEL) ? rstudio::core::log::logDebugMessageReturn(message) : false)
+
 
 #define LOG_DEBUG_MESSAGE_WITH_PROPS(message, props) rstudio::core::log::logDebugMessage(message, \
                                                                                          std::string(), \
--- a/src/cpp/shared_core/Logger.cpp
+++ b/src/cpp/shared_core/Logger.cpp
@@ -912,16 +912,16 @@ bool hasLogDestination()
 {
    Logger& log = logger();
    WRITE_LOCK_BEGIN(log.Mutex)
-      {
-         LogMap* logMap = &log.DefaultLogDestinations;
+   {
+      LogMap* logMap = &log.DefaultLogDestinations;
 
-         const auto destEnd = logMap->end();
-         for (auto iter = logMap->begin(); iter != destEnd; ++iter)
-         {
-            if (std::dynamic_pointer_cast<T, ILogDestination>(iter->second) != nullptr)
-               return true;
-         }
+      const auto destEnd = logMap->end();
+      for (auto iter = logMap->begin(); iter != destEnd; ++iter)
+      {
+         if (std::dynamic_pointer_cast<T, ILogDestination>(iter->second) != nullptr)
+            return true;
       }
+   }
    RW_LOCK_END(false);
    return false;
 }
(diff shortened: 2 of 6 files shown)


src/cpp/shared_core/include/shared_core/Logger.hpp
```diff
@@ -92,7 +92,8 @@ enum class LogLevel
    ERR = 1,       // Error messages will be logged.
    WARN = 2,      // Warning and error messages will be logged.
    INFO = 3,      // Info, warning, and error messages will be logged.
-   DEBUG = 4      // All messages will be logged.
+   DEBUG = 4,     // All messages will be logged.
+   DEBUG_LEVEL = 4// All messages will be logged - use to avoid name conflict with DEBUG macro
```
kfeinauer, line 96: I don't understand the need for `DEBUG` and `DEBUG_LEVEL`.

src/cpp/core/include/core/Log.hpp
```diff
@@ -82,7 +82,9 @@ std::string errorAsLogEntry(const Error& error);
 #define LOG_INFO_MESSAGE_NAMED(logSection, message) rstudio::core::log::logInfoMessage(message, \
                                                                                        logSection)
 
-#define LOG_DEBUG_MESSAGE(message) rstudio::core::log::logDebugMessage(message)
+
+#define LOG_DEBUG_MESSAGE(message) (rstudio::core::log::isLogLevel(rstudio::core::log::LogLevel::DEBUG_LEVEL) ? rstudio::core::log::logDebugMessageReturn(message) : false)
```
kfeinauer, line 86: I would recommend adding some tests of this new macro as you mentioned - you tried several ways, would be good to see them used to make sure we support the style we want.

src/cpp/shared_core/include/shared_core/Logger.hpp
jeffvroom: In Macros.hpp we have a `#define DEBUG` macro that's used all over the place in rsession. So if you include that header and Logger.hpp you can't refer to LogLevel::DEBUG without a compile error.  Since I needed to use that constant from LOG_DEBUG_MESSAGE, I needed a constant that would not conflict.

src/cpp/core/include/core/Log.hpp
jeffvroom: Sure, good idea.

jeffvroom: Ok I added unit tests for the key cases.

src/cpp/shared_core/include/shared_core/Logger.hpp
kfeinauer: I don't love this as it seems like a pollution of the interface to get around the `rsession` specific code. Is it possible we can tweak the DEBUG macro? If not, no worries.

kevinushey: It might make sense to rename that macro to something less likely to collide, like DBG?

kfeinauer (approved):

src/cpp/shared_core/include/shared_core/Logger.hpp
MariaSemple: What does the macro do? Could we give it a more descriptive name, like "ONLY_IN_DEBUG" or something?

src/cpp/shared_core/Logger.cpp
```diff
@@ -47,6 +47,52 @@ namespace log {
 typedef std::map<std::string, std::shared_ptr<ILogDestination>> LogMap;
 typedef std::map<std::string, LogMap> SectionLogMap;
 
+LogLevel logLevelFromStr(const std::string& in_str)
+{
+   if (in_str == "ERROR")
```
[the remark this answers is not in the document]
jeffvroom: Looks like the code that's currently using it is parsing it from a log file so presumably spaces will be removed there.  Same with how it will be used in workbench. I'm happy to add that code though if you'd prefer. It does make life easier for the caller.

src/cpp/shared_core/Logger.cpp
```diff
@@ -884,18 +926,82 @@ bool hasLogDestination()
...
+         {
+            if (std::dynamic_pointer_cast<T, ILogDestination>(iter->second) != nullptr)
+            {
+               iter->second->setLogLevel(in_newLevel);
+               if (in_newLevel > log.MaxLogLevel)
+                  log.MaxLogLevel = in_newLevel;
+            }
+         }
+      }
+   RW_LOCK_END(false);
```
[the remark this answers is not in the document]
jeffvroom: Thanks! Fixed the spacing in all 3 of those methods.

src/cpp/shared_core/include/shared_core/Logger.hpp
jeffvroom: I originally went and changed ~20 files from using DEBUG(...) to RSTUDIO_DEBUG(...) but it felt like overkill for a name conflict on a symbol like this that's not useful very often.  Those are valuable messages for debugging details of various components like RMarkdown, and other stuff in rsession but pretty sure they are compiled out by default.

PR #13694, opened 2023-09-26
Prefer system zlib.h over local

zachhannum:
### Intent

When `zconf.h` was removed in rstudio/rstudio#13641, it revealed an issue with the way we are including `zlib.h` in our `zlib.cpp` wrapper. We only want to use the local zlib.h if zlib is not installed on the system. This should only happen on Windows as we only build the embedded zlib _for_ Windows.

Otherwise, we were essentially including the local `zlib.h` file, even though the actual definitions would be defined by the system library (and in the case of centos7, was actually incompatible).

### Approach

`z_size_t` is not expected to exist on the version of zlib on centos7 (and is incompatible with the embedded version of zlib we have locally).

### Automated Tests

N/A

### QA Notes

N/A

N/A

### Checklist

### DIFF
--- a/src/cpp/core/zlib/zlib.cpp
+++ b/src/cpp/core/zlib/zlib.cpp
@@ -19,7 +19,7 @@
 
 #include <core/Log.hpp>
 
-#include "zlib.h"
+#include <zlib.h>
 
 namespace rstudio {
 namespace core {


GCRev (approved):
LGTM

kevinushey:
This breaks the Windows builds, since RStudio expects to find `zlib.h` in the same directory there.

https://github.com/rstudio/rstudio/blob/main/src/cpp/core/zlib/zlib.h

Note also that this built version of zlib is only used on Windows:

https://github.com/rstudio/rstudio/blob/e32fb38f34af7cf3fd0fff630dba41cf79b69df5/src/cpp/session/CMakeLists.txt#L539-L540

so the correct fix here is to avoid compiling zlib on non-Windows platforms altogether:

https://github.com/rstudio/rstudio/blob/e32fb38f34af7cf3fd0fff630dba41cf79b69df5/src/cpp/core/CMakeLists.txt#L135

That line can probably be removed, or just moved to a Windows-specific block.

kevinushey:
See also the comments here: https://github.com/rstudio/rstudio/blob/e32fb38f34af7cf3fd0fff630dba41cf79b69df5/src/cpp/core/zlib/CMakeLists.txt#L16-L20

zachhannum:
I think the confusion here is that `zlib.cpp` is not a part of the zlib library (even though it's in the same folder). `zlib.cpp` was written by us as a wrapper around some of the zlib functionality. The change to `#include <zlib.h>` was so that the compiler would search the system path first instead of the local directory, but AFAIK it should still find the header file in the local directory as well.

kevinushey:
Ahh! Thanks, you're right -- that threw me off.

PR #13699, opened 2023-09-27
only compile zlib on Windows

kevinushey:
Follow-up to https://github.com/rstudio/rstudio/pull/13694.

### DIFF
--- a/src/cpp/core/CMakeLists.txt
+++ b/src/cpp/core/CMakeLists.txt
@@ -132,8 +132,11 @@ set(CORE_SOURCE_FILES
    text/TextCursor.cpp
    text/TemplateFilter.cpp
    text/TermBufferParser.cpp
-   zlib/zlib.cpp
 )
+
+if(WIN32)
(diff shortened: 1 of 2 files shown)


zachhannum:
I don't think we actually use this file at all in the rstudio or rstudio-pro code, but `zlib.cpp` is actually used by the launcher slurm plugin:

https://github.com/rstudio/launcher/blob/e824b5879503d42980a4a40133a3166135c1503b/src/cpp/job_launcher/impls/slurm/SlurmObjectHelpers.cpp#L267

We may be able to just remove `zlib.cpp` altogether from our builds?

kevinushey:
Ah, thanks for the pointer -- I wasn't aware of that. That makes sense.

Then perhaps we should use the local `"zlib.h"` only on Windows there?

zachhannum:
No worries, it really confused me too.

> Then perhaps we should use the local "zlib.h" only on Windows there?

Yeah that's a good point, a potential improvement here might be to do something like:

```cpp
#ifdef _WIN32
#include "zlib.h"
#else
#include <zlib.h>
```

PR #13702, opened 2023-09-27
move zlib.cpp to ZlibUtil.cpp; update zlib.h include

kevinushey:
cc: @zachhannum; let me know if this seems reasonable!


zachhannum:
Looks good to me, do we also want to move the associated unit tests? https://github.com/rstudio/rstudio/blob/main/src/cpp/core/zlib/zlibTests.cpp

Issue #13685, opened 2023-09-25
Visual editor fails to load and pandoc error with hello tutorial


astayleraz:
### System details

    RStudio Edition : Server and workbench
    RStudio Version : 2023.12.0-daily+14 and 2023.12.0-daily+8.pro1
    OS Version      : ubuntu20
    R Version       : 4.3.0

### Steps to reproduce the problem

1. Create a markdown, rmarkdown, or quarto file.
2. Click to go into visual mode.
or 
1. click the tutorial tab and try and run the hello world tutorial.

### Describe the problem in detail
The user will just see a spinner or an error with the spinner. If you try the tutorial, you get a pandoc error. 
```
Error: pandoc version 1.12.3 or higher is required and was not found (see the help page ?rmarkdown::pandoc_available).
Execution halted
```

![Screenshot 2023-09-25 at 12 54 54 PM](https://github.com/rstudio/rstudio/assets/1482677/e6b11274-0847-45f5-8642-e025dfeeb595)
![Screenshot 2023-09-25 at 12 56 06 PM](https://github.com/rstudio/rstudio/assets/1482677/54bce708-4f2e-42b0-89a5-17b5b65833c6)

Kevin noticed the path with pandoc has changed but RSTUDIO_PANDOC shows a different path.

```
kevin@ip-172-98-22-79:/usr/lib/rstudio-server/bin/quarto/bin/tools/x86_64$ ll
total 291584
drwxr-xr-x 4 root root      4096 Sep 25 20:28 ./
drwxr-xr-x 3 root root      4096 Sep 25 20:27 ../
drwxr-xr-x 3 root root      4096 Sep 25 20:28 dart-sass/
-rwxr-xr-x 1 root root 113436728 May 18 17:09 deno*
drwxr-xr-x 2 root root      4096 Sep 25 20:28 deno_dom/
-rwxr-xr-x 1 root root   9261056 Oct 26  1985 esbuild*
-rwxr-xr-x 1 root root 148600824 Aug 31 20:02 pandoc*
-rwxr-xr-x 1 root root  27260184 Aug  7 16:30 typst*
```
Note the x86_64 suffix; I don’t think that existed before. However, RSTUDIO_PANDOC still has:
```
> Sys.getenv("RSTUDIO_PANDOC")
[1] "/usr/lib/rstudio-server/bin/quarto/bin/tools"
```

### Describe the behavior you expected
Visual mode and tutorials should work and run with no issues or errors.

[...]

PR #13703, opened 2023-09-27
update pandoc path resolution

kevinushey:
Addresses https://github.com/rstudio/rstudio/issues/13685.


kevinushey:
I'm going to merge this opportunistically since right now Quarto / R Markdown documents don't work in the daily builds without this.

Issue #13592, opened 2023-09-05
Update Ace to get latest a11y fixes


gtritchie:
### System details

    RStudio Edition : Desktop and Server
    RStudio Version : 2023.09.0-daily+397
    OS Version      : any
    R Version       : any

RStudio 2023.09 was upgraded to Ace 1.19; there have been several a11y improvements made to Ace since then, and we should consider upgrading Ace again soon to pick these up.

## v1.20.0
Add gutter controls to keyboard accessibility mode ([#5146](https://github.com/ajaxorg/ace/issues/5146)) ([5f2face](https://github.com/ajaxorg/ace/commit/5f2face8febe7c260a42bc30c67195f6a5a46085))
Hide Ace content from assistive technologies to improve a11y by [@akoreman](https://github.com/akoreman) in [#5160](https://github.com/ajaxorg/ace/pull/5160)
## v1.23.0
signal gutterkeydown in keyboard accessibility mode ([#5202](https://github.com/ajaxorg/ace/issues/5202)) ([538b18c](https://github.com/ajaxorg/ace/commit/538b18cf86e21c810910d286c820bac399927cd7))
Improve ARIA labelling fold controls ([#5205](https://github.com/ajaxorg/ace/issues/5205)) ([dad5e6f](https://github.com/ajaxorg/ace/commit/dad5e6f6e445b805f7b470ef7892442a83d342e5))
## v1.23.2
bug(autocomplete): fix a11y violations by [@oykuyilmaz](https://github.com/oykuyilmaz) in [#5241](https://github.com/ajaxorg/ace/pull/5241)
## v1.24.0
Allow more lines in hidden textarea to improve screen reader experience on Windows ([#5225](https://github.com/ajaxorg/ace/issues/5225)) ([bccff5a](https://github.com/ajaxorg/ace/commit/bccff5ae21b86004c2a4ab4a86a989c61e728ebf))

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

jonvanausdeln:
[...]

Issue #12370, opened 2022-11-23
fix missing css scope for rmarkdown footnotes


resistor4u:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2022.07.2+576 "Spotted Wakerobin" Release (e7373ef832b49b2a9b88162cfe7eac5f22c40b34, 2022-09-06) 
    OS Version      : macOS 10.15.7
    R Version       : 4.2.2 (2022-10-31)

### Steps to reproduce the problem
Issue #10128 is still a problem.

### Describe the problem in detail
Rstudio is missing a css selector for footnotes. Reading through issues with ShinyAce and Ace editor, some reference was made to licensing incompatibilities preventing an expanded syntax highlighting feature set in RStudio. Either way, the lack of a css selector (or footnote highlighting more generally) should be documented in the online material about modifying themes, if for no other reason than to save users the trouble of trying to implement an unsupported feature.
### Describe the behavior you expected
Providing in-line footnote reference (e.g., `[^note]`) and the actual footnote (e.g., `[^note]: Ahoy.`) in an rmarkdown document in any given rstudio theme should trigger syntax highlighting of one sort or another (defined in the rstheme).

- [x ] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x ] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x ] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
@resistor4u Thank you for raising that this is still an issue! I'll mark it for review.

resistor4u:
Still an issue!! Any movement?

PR #13708, opened 2023-09-28
Upgrade Ace to v1.28

jgutman:
### Intent
Addresses #13592 -- upgrading Ace to get some bugfixes and accessibility improvements

Please see https://github.com/rstudio/rstudio-ace/commits/rstudio/ocean-storm for further info


kevinushey:
LGTM but let's resolve the comment re: Ace's built-in autocompletion threshold before merge?

jgutman:
Done -- removed both autocompletion threshold and autocompletion delay patches as discussed on the branch

jgutman:
Do Ace upgrades need to get mentioned in NEWS, or no?

kevinushey:
Good question; in the past I don't think we've mentioned them consistently. I think we should mention it, though -- especially if we can confirm that we'll be benefiting from their work on accessibility.

resistor4u:
Not necessarily the place, but any thoughts on whether upgrading Ace impacts #12370 ?

Issue #13710, opened 2023-09-28
Screen reader mode always announced as disabled and shown as disabled in menu, same with "Toggle Tab Key Always Moves Focus"


gtritchie:
### System details

    RStudio Edition : Desktop or Server
    RStudio Version : 2023.09.0-463 (Desert Sunflower)
    OS Version      : any
    R Version       : any

### Regression Notes

Worked in 2023.06.2+561 "Mountain Hydrangea" so is a regression in Desert Sunflower.

I repro'd on Mac, open-source Server, and Windows.

It was a late regression. Some quick manual bisecting shows it broke with https://github.com/rstudio/rstudio/pull/13638 which first showed up in build 2023.09.0-443.

### Steps to reproduce the problem

- Under **Help / Accessibility** menu the first item says "Screen reader support (disabled)" which is the correct default state
- Bring up Command Palette and type "screen reader" and you'll see both the command text with (disabled) and the actual checkbox state (unchecked/off)

<img width="628" alt="Screenshot of command palette with screen reader typed into it" src="https://github.com/rstudio/rstudio/assets/10569626/39f2c7b2-df10-40f7-a734-7aeaefb61ffc">

- Now click that "Screen reader support" command and a modal dialog asks if you really want to enable it, say Yes, then the IDE restarts
- Look at Help / Accessibility menu again
- Look at Command Palette and type "screen reader" again

NOTE, the same problem happens with the Help / Accessibility / Focus / "**Toggle Tab Key Always Moves Focus**" command. It always shows as "off" (unchecked) in the menu, even when it is turned on.

### Describe the problem in detail

- Menu command still says it is disabled

<img width="562" alt="Screenshot showing the Help / Accessibility menu" src="https://github.com/rstudio/rstudio/assets/10569626/ebfae132-f499-49c6-9262-5491b9e3db7c">

- Command Palette has a mixed state (text says disabled, checkbox shows enabled)

<img width="645" alt="Screenshot of command palette showing inconsistent state for screen reader mode" src="https://github.com/rstudio/rstudio/assets/10569626/bb12af6e-a815-411d-97c0-790df268cf96">

- Global Options also shows it being checked:

<img width="632" alt="Screenshot of Global Options Accessibility pane" src="https://github.com/rstudio/rstudio/assets/10569626/371ff47a-7a8d-46a0-b796-08a828c91119">

### Describe the behavior you expected

- It should switch to "enabled" as it does in 2023.06.2 and earlier

[...]

PR #13715, opened 2023-09-28
fix "screen reader mode" toggle and "tab key always moves focus" toggle

gtritchie:
### Intent

Addresses #13710

### Approach

Due to other changes, the previous approach for checking the state of the screen reader flag at startup was no longer getting the current value (I believe it was getting the default value). This code is responsible for deciding if it should announce "screen reader not enabled" and for setting the menu text.

Code that checks the pref later was getting the correct value, hence the mismatches.

There was a second symptom of this problem, also related to accessibility. The "tab key moves focus" toggle (Help / Accessibility / Focus / Tab Key Always Moves Focus) was also not showing the correct state (via a checkmark).

Fixed by moving that startup code to the end of onSessionInit at which point preferences are happy.

### Automated Tests

None that I'm aware of; it would be nice to have a test to check that this screen reader flag is working, but not sure we have the capability to test something that requires a reload of the IDE.

### QA Notes

Test that the menu text both in Help / Accessibility and in the Command Palette now correctly show the current state of the screen reader toggle.

Also test with a screen reader running: when the screen reader flag is off, you should hear an announcement at startup (often after the Console is read) warning that screen reader mode isn't enabled. Once you enable it and restart you should no longer hear that announcement.

Finally, spot-check that the Tab Key Always Moves Focus command is showing the correct state in the menu Help / Accessibility / Focus : Tab Key Always Moves Focus. The text never changes but there should be a checkmark when it is enabled.

### Documentation

N/A

### Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,0 +1,17 @@
+
+## RStudio 2023.09.1 "Desert Sunflower" Release Notes
+
+### New
+#### RStudio IDE
+- 
+
+#### Posit Workbench
+- 
+
+### Fixed
+#### RStudio IDE
+- Fixed issue preventing enabling screen reader mode (#13710)
+
+#### Posit Workbench
+- 
+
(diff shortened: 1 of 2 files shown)


kevinushey (approved):
LGTM! This sounds similar to another recent commit:

https://github.com/rstudio/rstudio/commit/06570c9dfa69f13d7e074a32239070f23f020971

I think the Copilot class I created must've done something that changed the order in which some of these singleton classes are instantiated, bringing some of these ordering issues into light.

If I understand correctly, we can't safely access preferences until we receive the session init event, so this looks like the right change to me.

Issue #13676, opened 2023-09-22
Remove postgres dependency from open source


kevinushey:
This is required only for Workbench (via SOCI). To resolve this, we should:

1. Produce a build of SOCI that connects only to SQLite;
2. Use that build of SOCI in our open-source builds.

kevinushey:
Duplicate of https://github.com/rstudio/rstudio/issues/8968.

ronblum:
Please see #13676 for any updates.

ronblum:
Tested in Red Hat 9.2 with RStudio Desktop 2023.12.0-daily+105. `libpq` is still getting installed:


```
Dependencies resolved.
================================================================================================================
 Package         Architecture   Version                          Repository                                Size
================================================================================================================
Installing:
 rstudio         x86_64         2023.12.0_daily+105-1            @commandline                             165 M
Installing dependencies:
 libpq           x86_64         13.5-1.el9                       rhel-9-for-x86_64-appstream-rpms         208 k
 sqlite          x86_64         3.34.1-6.el9_1                   rhel-9-for-x86_64-appstream-rpms         750 k

Transaction Summary
================================================================================================================
Install  3 Packages

Total size: 166 M
Total download size: 957 k
Installed size: 694 M
Is this ok [y/N]: y
Downloading Packages:
(1/2): libpq-13.5-1.el9.x86_64.rpm                                              385 kB/s | 208 kB     00:00    
(2/2): sqlite-3.34.1-6.el9_1.x86_64.rpm                                         1.1 MB/s | 750 kB     00:00    
----------------------------------------------------------------------------------------------------------------
Total                                                                           1.4 MB/s | 957 kB     00:00     
Running transaction check
Transaction check succeeded.
Running transaction test
Transaction test succeeded.
Running transaction
  Preparing        :                                                                                        1/1 
  Installing       : sqlite-3.34.1-6.el9_1.x86_64                                                           1/3 
  Installing       : libpq-13.5-1.el9.x86_64                                                                2/3 
  Installing       : rstudio-2023.12.0_daily+105-1.x86_64                                                   3/3 
  Running scriptlet: rstudio-2023.12.0_daily+105-1.x86_64                                                   3/3 
  Verifying        : libpq-13.5-1.el9.x86_64                                                                1/3 
  Verifying        : sqlite-3.34.1-6.el9_1.x86_64                                                           2/3 
  Verifying        : rstudio-2023.12.0_daily+105-1.x86_64                                                   3/3 
[...]

PR #13721, opened 2023-09-29
don't include libpq in open-source builds

kevinushey:
Addresses https://github.com/rstudio/rstudio/issues/13676#issuecomment-1740589971.


jeffvroom (approved):
LGTM!

PR #13739, opened 2023-10-03
prefer 'true' and 'false' for logical YAML values

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio-pro/issues/5227.

### Approach

The R `yaml` package attempts to conform to the YAML 1.1 specification, and chooses to emit `yes` and `no` for logical values. YAML 1.2 requires these constants to instead be called `true` and `false`. Since these constants are also recognized by YAML 1.1, we just choose to always emit `true` and `false` for logical vectors.

### Automated Tests

N/A

### QA Notes

Test via notes in https://github.com/rstudio/rstudio-pro/issues/5227.

### Documentation

N/A

### Checklist

### DIFF
--- a/src/cpp/session/modules/SessionRMarkdown.R
+++ b/src/cpp/session/modules/SessionRMarkdown.R
@@ -357,8 +357,16 @@
       })
    }
    
+   # define handler for logical values
+   handlers <- list(
+      logical = function(x) {
+         value <- ifelse(x %in% TRUE, "true", "false")
+         structure(value, class = "verbatim")
+      }
+   )
+   
    # substitute ticks and convert to yaml
-   yaml <- yaml::as.yaml(tick_sub(input))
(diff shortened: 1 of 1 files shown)


src/cpp/session/modules/SessionRMarkdown.R
```diff
@@ -357,8 +357,16 @@
       })
    }
    
+   # define handler for logical values
+   handlers <- list(
+      logical = function(x) {
+         value <- ifelse(x %in% TRUE, "true", "false")
```
kevinushey, line 363: `x %in% TRUE` also helps us handle `NA` values, which we just emit as `false`.

jgutman: nice!

jgutman (approved):

Issue #13637, opened 2023-09-14
publishing attempts to use an empty directory, causing linting and publishing errors


aronatkins:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2023.09.0-daily+384
    OS Version      : macOS Ventura 13.5.2
    R Version       : 4.3.1

### Steps to reproduce the problem

Create a Quarto project with the following structure:

```
_quarto.yml
index.qmd
project.Rproj
materials/project/05_connectwidgets/project_curation.qmd
```

The contents of the two `*.qmd` does not matter.

The content of `_quarto.yml` must be:

```yaml
project:
  type: website
  render:
    - "index.qmd"
```

In the IDE, open the `project_curation.qmd` file and then click the publishing button. Target some Connect server.

### Describe the problem in detail

When attempting to publish, the pre-flight lint check errs:

<img width="264" alt="image" src="https://github.com/rstudio/rstudio/assets/362187/53a89d52-1ffb-4732-966c-fc633f8c996c">

This error is because the lint workflow is passed an empty target string:

https://github.com/rstudio/rstudio/blob/main/src/cpp/session/modules/SessionRSConnect.R#L246-L273

If you try to "Publish Anyway", you receive an error in the Deploy pane:

```
Error in `check_file()`:
! `x` must be a single string, not the empty string "".
Backtrace:
    ▆
 1. └─rsconnect::deployApp(...)
 2.   └─rsconnect:::check_directory(appDir)
 3.     └─rsconnect:::check_file(x = x)
 4.       └─rsconnect:::check_string(...)
 5.         └─rsconnect:::stop_input_type(...)
 6.           └─rlang::abort(message, ..., call = call, arg = arg)
Execution halted
```

This second error is because `deployApp()` is passed an empty string for `appDir`, which is also the reason that the linting failed.

### Describe the behavior you expected

Linting passed and deployment succeeded.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
[...]

PR #13741, opened 2023-10-03
Permit standalone Quarto file publishing non-root of project

jgutman:
### Intent

Addresses #13637

We basically had two checks for locating metadata for (parts of) a Quarto book, manuscript, or website. One was looking at the output of `quarto inspect`, and the other was looking at the `_quarto.yml` file, which we locate by [looking for any config file in a parent of the target file](https://github.com/rstudio/rstudio/blob/b52a4c2ef54937226aec5c0d4c66b9810d0ce435/src/cpp/session/modules/quarto/SessionQuarto.cpp#L273-L299) until we get to the user's home. But we only actually use the metadata we get from the `_quarto.yml` file. And we don't need to be at the root of the quarto project in order to publish it. So this check basically seems to only throw unnecessary errors? 

I can publish both the standalone `about.qmd` file by itself, or the root `index.qmd` file and entire project contents successfully without the check, just based on finding the `_quarto.yml` file. If there's no `_quarto.yml` file found, we will get the same error as before when attempting to deployApp.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -23,6 +23,7 @@
 - Fixed an issue where inline YAML chunk options were not properly parsed from SQL chunks. (#13240)
 - Fixed an issue where help text in the autocompletion popup was not selectable. (#13674)
 - Fixed an issue where a file could be opened twice when debugging functions sourced from other directories. (#13719)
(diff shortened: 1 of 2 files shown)


kevinushey (approved):
LGTM! Should we update NEWS as well?

Issue #13451, opened 2023-08-03
Autocomplete for `@` / `.AtNames()`


t-kalinowski:
Autocompletion for `@` powered by `.AtNames()` doesn't work in the IDE. This *does* work outside the IDE, e.g., when running R in a terminal. 

Minimal example: 
```r
`@.myclass` <- function(x, name) {
  attr(x, name)
}

.AtNames.myclass <- function(x, pattern) {
  utils::findMatches(pattern, names(attributes(x)))
}

x <- structure(list(), class = "myclass", attr1 = "attr1val")
```
Typing: 
```r
x@<tab>
``` 
In the IDE this shows a "No Matches" popup. 
![image](https://github.com/rstudio/rstudio/assets/8462255/e79a3faf-77b9-49f6-b31c-8e38a4773e6f)

In the terminal this gives the completions list: 
![image](https://github.com/rstudio/rstudio/assets/8462255/68f7c18c-a7be-4daa-a774-948ccb009d51)

jtlandis:
I'm going to throw my support for this being patch as well. For a long time I had thought this was an issue with `.AtNames` not working but it is just the RStudio IDE. Not having the tab complete work in the IDE makes programing with custom classes leveraging `@` S3 generic significantly harder.

PR #13743, opened 2023-10-03
support '@' completions via .AtNames

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/13451.

### Approach

If `.AtNames` is defined and available, use it when requesting completions following an `@`.

### Automated Tests

N/A

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/13451.

### Documentation

N/A

### Checklist


jgutman (dismissed):
Looks good! How important is this -- should this be backported?

kevinushey:
I don't think it's necessary to backport; this just adds support for a new completion option introduced in R 4.3.0.

jgutman:
@kevinushey it looks like they also added similar support for `.DollarNames`.

```
`$.myclass` <- function(x, name) {
   attr(x, name)
 }
 
 DollarNames.myclass <- function(x, pattern) {
   utils::findMatches(pattern, names(attributes(x)))
}

x <- structure(list(), class = "myclass", attr1 = "attr1val")

x$
```
i.e. this tab completes in R but not in RStudio and likely needs a similar addition

src/cpp/session/modules/SessionRCompletions.R
```diff
@@ -1526,16 +1526,30 @@ assign(x = ".rs.acCompletionTypes",
...
+            code <- substitute(
+               utils::.AtNames(object, pattern = ""),
+               list(object = object)
+            )
+            allNames <- eval(code, envir = globalenv())
+            names <- .rs.selectFuzzyMatches(allNames, token)
+            type <- attr(names, "types")
+         })
+      }
+      else if (isS4(object))
```
jgutman, line 1541: Actually on second thought I don't get why this is an else if? So we only do the s4 slot completions on older R versions where AtNames is not defined?

jgutman: Oh I see now that S4 objects also use the .AtNames so this works for standard S4 object definitions too

jgutman (approved):
LGTM, but can we add the `.DollarNames` autocompletion as well?

Issue #13744, opened 2023-10-04
stop indexing 'package' directory in RStudio project


kevinushey:
Since we might end up editing the wrong source file and quietly losing changes...

<img width="350" alt="Screenshot 2023-10-03 at 5 23 09 PM" src="https://github.com/rstudio/rstudio/assets/1976582/5afc3032-b01e-4e86-bb10-7be7d6709b14">

jgutman:
I have done this before! 🤦‍♀️

PR #13748, opened 2023-10-04
tweaks for file indexer

kevinushey:
Closes https://github.com/rstudio/rstudio/issues/13744.


src/cpp/session/modules/SessionCodeSearch.cpp
```diff
@@ -122,7 +122,7 @@ bool isWithinIgnoredDirectory(const FilePath& filePath, const std::vector<FilePa
          return true;
 
       // cmake build directory
-      if (parentPath.completeChildPath("cmake_install.cmake").exists())
```
kevinushey, line 125: Since `cmake_install.cmake` is only written out for CMake build directories that actually have build commands. `CMakeCache.txt` is always written out.

Issue #11708, opened 2022-08-05
YAML-style chunk options for plots are ignored when using hyphenated options (fig-height, fig-width)


jthomasmock:
### System details

    RStudio Edition : Desktop
    RStudio Version : '2022.7.1.554'
    OS Version      :  Mac OS X 12_4_0
    R Version       :  4.2.0

### Steps to reproduce the problem

Using source mode or visual mode, modify plot dimensions with traditional `knitr` chunk options and YAML-styled `#|` options:

````
---
format: html
editor_options: 
  chunk_output_type: inline
---

This modifies appearance of plot
```{r, fig.width=12, fig.height = 5}
plot(mtcars$disp, mtcars$mpg)
```

But this doesn't:

```{r}
#| fig-width: 12
#| fig-height: 5
plot(mtcars$disp, mtcars$mpg)
```
````

### Describe the problem in detail

When using `knitr` chunk options only using the older syntax `r, fig.width=12, fig.height=5` modifies the appearance of inline plots in Quarto. Hashpipe/YAML style: `#| fig-width: 12` doesn't modify the appearance.

![image](https://user-images.githubusercontent.com/29187501/183095733-c2a25e50-0453-4241-938a-b56b5eafbd40.png)

### Describe the behavior you expected

Both `knitr` style and YAML style chunk options do the same thing when rendering - would be ideal if that also affected interactive behavior in RStudio.

Tagging @debruine who reported the behavior initially - I was able to reprex.

https://twitter.com/lisadebruine/status/1555506264519630849?s=21&t=9xOZw-tSkKLd8IayCK0ajw

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

andresrcs:
Since you are already into this, have you noticed a big difference in proportions between the plots shown in the code editor and the plots that get rendered into HTML? I first thought it was because of this issue but even using knitr chunk options I get a noticeable difference as if they were using different units. For "difference in proportions" I mean, for example, the text being way bigger in the HTML output relative to the rest of the plot than in the code editor.

[...]

PR #13749, opened 2023-10-04
convert dashes to dots in yaml style chunk options

jgutman:
### Intent

Addresses #11708

### DIFF
--- a/src/cpp/session/modules/SessionRmdNotebook.R
+++ b/src/cpp/session/modules/SessionRmdNotebook.R
@@ -620,7 +620,14 @@ assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
 {
    # read as YAML
    opts <- .rs.fromYAML(opt)
-   
+
+   # knitr version 1.44 introduced a function `knitr:::dot_names`
+   # which converts dashes to dots in chunk options
+   if (packageVersion("knitr") >= "1.44") {
+      opts <- knitr:::dot_names(opts)
+   } else {
+      names(opts) <- gsub("-", ".", names(opts))
+   }
    # parse truthy and falsy style yaml options
    opts <- lapply(opts, function(value)
    {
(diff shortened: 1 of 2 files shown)


src/cpp/session/modules/SessionRmdNotebook.R
```diff
@@ -620,7 +620,14 @@ assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
 {
    # read as YAML
    opts <- .rs.fromYAML(opt)
-   
+
+   # knitr version 1.44 introduced a function `knitr:::dot_names`
+   # which converts dashes to dots in chunk options
+   if (packageVersion("knitr") >= "1.44") {
+      opts <- knitr:::dot_names(opts)
```
kevinushey, line 627: Since we're calling an unexported function which might change (or not exist) in a future release, it might actually be better to do something like this:

```
knitr <- asNamespace("knitr")
if (is.function(knitr$dot_names)) {
  opts <- knitr$dot_names(opts)
}
```

I realize I made the initial recommendation to call it directly but I didn't realize it was an unexported function as first.

src/cpp/session/modules/SessionRmdNotebook.R
```diff
@@ -620,7 +620,14 @@ assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
 {
    # read as YAML
    opts <- .rs.fromYAML(opt)
-   
+
+   # knitr version 1.44 introduced a function `knitr:::dot_names`
+   # which converts dashes to dots in chunk options
+   if (packageVersion("knitr") >= "1.44") {
```
kevinushey, line 626:
```suggestion
   if (packageVersion("knitr") >= "1.44")
   {
```

src/cpp/session/modules/SessionRmdNotebook.R
```diff
@@ -620,7 +620,14 @@ assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
...
+      opts <- knitr:::dot_names(opts)
+   } else {
```
kevinushey, line 628:
```suggestion
   }
   else
   {
```

kevinushey (approved):
LGTM!

Issue #9861, opened 2021-09-20
markdown package is no more required to render a Rmd document but still asked by Knit button


cderv:
Filling this here as an issue following @ronblum request

I know the IDE is keeping track of required R dependencies and force to install missing or outdated versions.
I believe this is tracked here: https://github.com/rstudio/rstudio/blob/main/src/cpp/session/resources/dependencies/r-packages.json

Just so you know, [last version of knitr 1.34](https://github.com/yihui/knitr/releases/tag/v1.34) has moved the markdown package from Import to Suggest. So **markdown** package is no more a hard dependency. 

I believe this is causing the behavior I am seeing:
* Create a new renv project with rmarkdown document
* Install packages with last CRAN versions - **markdown** package won't be install anymore
* Restart the project using the isolated renv library where markdown is not there.
* Try to knit the document using the IDE button - a message from the IDE ask me to install the markdown package.
* This package is no more required and `rmarkdown::render()` works ok to build the doc.

This is not really an issue because the **markdown** can be installed safely (by anwsering yes to install when IDE is asking). 
However I don't know how the dependency graph is working but it could be suprising / puzzling for user to have a missing package after a successful `install.packages("rmarkdown")`.  So maybe in a future version, the requirement could be often when knitr 1.34 is used. 
Or maybe it used in another place in the IDE. 

Anyway, just wanted to inform you of this change in **knitr**

cderv:
Found at least on occurence of painful experience for a user : https://community.rstudio.com/t/updated-version-of-markdown-package-required/132176/4

eitsupi:
Hello.
New users may not know that the `markdwon` package exists (I didn't before either), so this popup may cause them to fear that they have failed to install the `rmarkdown` package. (e.g. rocker-org/rocker-versioned2#525)
It would be great to see this issue resolved.

cderv:
> New users may not know that the markdwon package exists (I didn't before either), so this popup may cause them to fear that they have failed to install the rmarkdown package


new occurrence of this  situation at https://community.rstudio.com/t/error-when-opening-creating-new-r-markdown-document/171067

eliocamp:
Another recorder occurrence: https://github.com/rstudio/renv/issues/1708

jeremy-allen:
Happens to me daily. Would love a fix. Happened during a live demo this week and caused many questions from the audience.

ronblum:
Verified, using the steps in the initial post, that
[...]

PR #13755, opened 2023-10-05
rmarkdown no longer requires markdown

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/9861.

### Approach

The `markdown` package is no longer required by the `rmarkdown` package, so we removed it.

### Automated Tests

N/A

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/9861.

### Documentation

N/A

### Checklist

### DIFF
--- a/src/cpp/session/resources/dependencies/r-packages.json
+++ b/src/cpp/session/resources/dependencies/r-packages.json
@@ -395,7 +395,6 @@
                 "jsonlite",
                 "knitr",
                 "magrittr",
-                "markdown",
                 "mime",
                 "rmarkdown",
                 "stringi",


eliocamp:
Is this list of dependencies supposed to be the same as rmarkdown's actual dependencies on CRAN? There are dependencies in the file that are not currently (direct) rmarkdown dependencies, like digest, stringi, base64enc and some are missing, like fontawesome and tinytex.

``` r
tools::package_dependencies("rmarkdown")
#> $rmarkdown
#>  [1] "bslib"       "evaluate"    "fontawesome" "htmltools"   "jquerylib"  
#>  [6] "jsonlite"    "knitr"       "methods"     "stringr"     "tinytex"    
#> [11] "tools"       "utils"       "xfun"        "yaml"
```

jgutman:
It's not necessarily meant to be the same. This list includes packages that are required for the RStudio IDE more broadly, not necessarily for rmarkdown. For example, `shiny` and `rsconnect` are required for other IDE functionality. We also don't necessarily have all dependencies of rmarkdown included because there is no requirement that users have the latest version of rmarkdown. For example the requirement for `fontawesome` was added only 8 months ago in rmarkdown 2.21, but it's fine to use an older version. Adding these explicit recursive dependencies when not necessary makes it more likely we need to remove or change something when these package dependencies change, as in this instance.

jgutman (approved):

Issue #13799, opened 2023-10-16
debugging breakpoints and source references encounter srcfile$wd errors


aronatkins:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2023.12.0-daily+170
    OS Version      : macOS Sonoma 14.0
    R Version       : 4.3.1

### Steps to reproduce the problem

While working on the rsconnect package, first use the IDE to Build>Install the package. Next, navigate to some source file, let's say the `deployApp()` definition inside `R/deployApp.R`.

Attempt to add a breakpoint on some executable line of code. The IDE shows an error in the console and a notification in the editor pane.

<img width="668" alt="image" src="https://github.com/rstudio/rstudio/assets/362187/919d6e5d-2c24-4131-a110-0834f82db5a3">

```
Error in enc2utf8(srcfile$wd) : argument is not a character vector
```

Run a Build>Install of the package again. After that completes and the package is reloaded, the breakpoint icon goes solid red. Restarting the R session alone is not enough to change the breakpoint icon.

Try running the `rsconnect::deployApp()` function such that it hits the breakpoint.

```r
rsconnect::deployApp()
```

When the breakpoint is hit, the IDE shows the same errors and tells me that source is not available.

<img width="670" alt="image" src="https://github.com/rstudio/rstudio/assets/362187/37579fcc-4f77-42ef-b185-ac61a3ebb416">

This might be related to https://github.com/rstudio/rstudio/pull/13723

### Describe the behavior you expected

Debugging with breakpoints shows no warnings and no errors.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.posit.co/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

jgutman:
@aronatkins Putting in a fix right now -- the error message was raised incorrectly, so you can ignore it, but I'm making a change so the error will not be raised

jonvanausdeln:
Verified with [2023.12.0-daily+201](https://dailies.rstudio.com/version/2023.12.0-daily+201)
Windows 11

The error messages are gone while debugging and things appear to function as expected

jthomasmock:
[...]

PR #13801, opened 2023-10-16
don't create wd until we know we actually need it

jgutman:
### Intent

Address #13799 

In PR #13723, we tried to get the value of `srcfile$wd` even before we knew we needed it, raising unnecessary errors in cases where `srcfile$filename` contains the absolute path and `srcfile$wd` is unset, which raises errors and makes it look like the debugger couldn't find the source code even when it did


kevinushey (approved):
LGTM; thanks for jumping on this!

PR #13802, opened 2023-10-16
add fixes to parsing of in-body chunk options in various styles

jgutman:
### Intent
Addresses the rest of https://github.com/rstudio/rstudio/issues/13240 including some regressions introduced in #13653 

Sorry for all the IDE whitespace changes, the changes are basically limited to L741-L755 and L804-L815 

Basically needed to test that all of the following in-body chunk option styles work when executing chunks inline:

~~~~
---
output: html_document
---

```{r setup-db}
library(DBI)
db = dbConnect(RSQLite::SQLite(), dbname = "sql.sqlite")
```

# SQL chunk with -- comment character and tag: value (works in RStudio)
```{sql dash-and-yaml}
--| connection: db

SELECT 'test'
```

# SQL chunk with # comment character and tag: value (fails in RStudio)
```{sql hash-and-yaml}
#| connection: db

SELECT 'test'
```

# SQL chunk with -- comment character and tag=value (fails in RStudio)
```{sql dash-and-equals}
--| connection=db

SELECT 'test'
```

# SQL chunk with # comment character and tag=value (fails in RStudio)
```{sql dash-and-equals}
#| connection=db

SELECT 'test'
```
~~~~

Also checked that the following still works
~~~~
---
output: html_document
---

```{r setup}
library(DBI)
# Create an ephemeral in-memory RSQLite database
con <- dbConnect(RSQLite::SQLite(), ":memory:")
dbWriteTable(con, "mtcars", mtcars)
dbListTables(con)
knitr::opts_chunk$set(connection = con)
```

```{sql}
SELECT * FROM mtcars WHERE cyl = 4
```
~~~~

### DIFF
--- a/src/cpp/session/modules/SessionRmdNotebook.R
+++ b/src/cpp/session/modules/SessionRmdNotebook.R
@@ -15,13 +15,13 @@
 assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
 
 .rs.addFunction("extractRmdFromNotebook", function(notebook) {
-   
+
    # parse the notebook to get the text of the contained R markdown doc
    parsed <- try(rmarkdown::parse_html_notebook(notebook), silent = TRUE)
    if (inherits(parsed, "try-error") || is.null(parsed$rmd)) {
       return("")
    }
-   
+
    # as this string will be eventually written to (and compared with) files
    # on disk, use native line endings
    return(paste(parsed$rmd,
@@ -54,10 +54,10 @@ assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
    # nothing to coalesce if < 2 outputs
    if (length(chunkData) < 2)
       return(chunkData)
-   
+
    # keep all output by default
    keep <- rep(TRUE, length(chunkData))
-   
+
    # coalesce contents for consecutive csv entries
    csvs <- .rs.endsWith(names(chunkData), ".csv")
    for (i in seq.int(1, length(csvs) - 1)) {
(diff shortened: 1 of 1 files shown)


src/cpp/session/modules/SessionRmdNotebook.R
```diff
@@ -735,20 +735,29 @@ assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
...
+         # if the language-specific comment prefix character was not used,
+         # fallback and check for starting with #| instead before bailing
+         match <- unlist(regmatches(line, regexec(.rs.reRmdInlineChunk("#"), line)))
+         if (length(match) == 0)
+            break
+         # if we found a match, we also need to update what gets passed to .rs.parseInlineChunk
+         commentPrefix <- "#"
+         reCommentPrefix <- .rs.reRmdInlineChunk(commentPrefix)
+      }
+      else if (length(match) == 0)
```
jgutman, line 755: basically, the behavior in https://github.com/yihui/knitr/blob/607148df836902b8d1c2f496df97f2f83159cac4/R/parser.R#L279-L282

if the language-specific comment prefix is not `#`, but the user used `#` anyway, we update `commentPrefix` and `reCommentPrefix` accordingly and add the match to inline chunk options. We don't really allow a mix of `#|` and `--|` in a single chunk, but I don't think we should (and I doubt anyone is mixing and matching within a chunk).

src/cpp/session/modules/SessionRmdNotebook.R
```diff
@@ -759,83 +768,89 @@ assign(".rs.notebookVersion", envir = .rs.toolsEnv(), "1.0")
...
+      # to store the connection value and pass the temp variable name
       if (!is.null(conn) && !is.character(conn)) {
-         .rs.setVar("chunkConnection", conn)
-         opts$connection <- ".rs.chunkConnection"
+         if (is.name(conn)) {
+            opts$connection <- as.character(conn)
+         } else {
+            .rs.setVar("chunkConnection", conn)
+            opts$connection <- ".rs.chunkConnection"
+         }
```
jgutman, line 815: basically, if we're using in-body chunk options in csv-format (tag=value), then we call `knitr:::parse_params` to parse those chunk options for us. which will end up making `conn` a symbol (e.g. `conn=as.name("db")`), so calling as.character here allows us to set `opts$connection <- "db"

jgutman:
Might be worth doing this as well: https://github.com/rstudio/rstudio/issues/13240#issuecomment-1764821834

kevinushey (approved):
LGTM!

Issue #9798, opened 2021-09-01
View blame on Github opens master branch even when there is not master (instead of local branch or main branch)


cderv:
### System details

    RStudio Edition : Desktop [Open Source]
    RStudio Version : 2021.11.0.23 (daily)
    OS Version      : Windows 10 x64 (build 19043)
    R Version       : R version 4.1.0 (2021-05-18)

### Steps to reproduce the problem

1. Open RStudio version (daily) 
2. Create a new project from Git using : https://github.com/rstudio/rmarkdown/ (place it in a folder without name conflict)
3. Create a new branch locally that won't be pushed / tracked on remote 
   `git checkout -b dummy` 
3. Open README.md (`file.edit("README.md")`)
4. Click on the button to view blame for the file (or CTRL + MAJ + P and search "blame" for "View Git Blame on Github"
5. This will go to https://github.com/rstudio/rmarkdown//blame/master/README.md which does not exist instead of https://github.com/rstudio/rmarkdown//blame/main/README.md 
   (not the double slash also but that is not part of the problem)
6. Click "View file on Github" will open https://github.com/rstudio/rmarkdown/blob/master/README.md which is also wrong, but will work because Github redirect to https://github.com/rstudio/rmarkdown/blob/main/README.md
7. When doing the same on a synced branch, the correct branch name is used and correct url is opened.

Cleaning: Remove the created project from your computer.

### Describe the problem in detail

In a project using Git, with a remote available, using `main` and not `master` as main branch (like rstudio/rmarkdown), when a file is opened, one can click on "Blame on Github" button. 
When the file is from a branch that is not yet think, looking at the blame will send to a url using `master` by default, which no more exist. It does not link to the `main` branch in this case. 

I think `master` is used as default when no remote branch is tracked with the current local branch. 

### Describe the behavior you expected

The blame view should not be opened on `master` because it could not exist (renamed `main`). It should be opened on the default branch called `main` (or else ?)
Same for "View File". 

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

[...]

PR #13803, opened 2023-10-16
Fix constructing github url for file

timtmok:
### Intent
Address #9798

### Approach
Remove using `master` as upstream branch. If this fails to get the upstream branch name then fallback to using the last commit id for the file. This would provide a url to the exact state of the file.

One case would still fail if the file has a commit that doesn't exist on the remote. Maybe it is better to always use `HEAD` as the branch name instead of the commit id? However, the commit id has the advantage of showing the exact state as seen in RStudio.

The repo url retrieval was changed to use `ls-remote` so that it doesn't need to query by the upstream name. This command was introduced in `git@1.7.5`.

### Automated Tests
none

### QA Notes
The one case this would still fail is making a change to a file and committing it. Do not push the commit and try to use the view on github action. The commit doesn't exist on github so it cannot be viewed.

### Documentation
n/a

### Checklist


kevinushey (approved):
LGTM; nice work!

Issue #13809, opened 2023-10-17
Stray `-ne` text in the Terminal


mine-cetinkaya-rundel:
### System details

    RStudio Edition : Desktop
    RStudio Version : RStudio 2023.12.0-daily+193 "Ocean Storm" Daily (799b50d74505cbb7a13c67ddb7a4d570edcb11fa, 2023-10-16) for macOS
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) RStudio/2023.12.0-daily+193 Chrome/116.0.5845.190 Electron/26.2.4 Safari/537.36
    OS Version      : Mac OS 13.5
    R Version       : 4.3.1

### Steps to reproduce the problem

Go to the Terminal, it starts with `-ne` on top.

Type any command, it keeps printing `-ne` at each new line.

This has been happening for the last few versions, it's not a brand new issue.

### Describe the problem in detail

<img width="643" alt="Screenshot 2023-10-17 at 5 04 01 PM" src="https://github.com/rstudio/rstudio/assets/5965649/4dcf9aa5-0018-4f55-920f-3839a1348200">

### Describe the behavior you expected

I expected to not see the stray character string.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.posit.co/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
Do you have a `~/.bashrc` or `~/.bash_profile`? Is this using bash or zsh?

The output from `Tools -> Terminal -> Terminal Diagnostics...` may be helpful.

mine-cetinkaya-rundel:
Here is the terminal diagnostics:

```bash
Global Terminal Information
---------------------------
Loaded TerminalSessions: 1
Handle: '877A8FBD' Caption: 'Terminal 1'

Terminal List Count: 1
Handle: '877A8FBD' Caption: 'Terminal 1' Session Created: true

Global Terminal Information
---------------------------
Caption:     'Terminal 1'
Title:       '~/Desktop'
Cols x Rows  '80 x 14'
Shell:       'Bash'
[...]

PR #13815, opened 2023-10-18
fix handling of shell startup scripts on macOS with default Bash

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/13809.

### Approach

- On macOS, `/bin/bash --login --posix` does not respect the `ENV` environment variable, so we have no choice but to use a non-login shell. Made this a macOS-only change just to avoid fallout in Workbench installations.
- Instead of using the opaque "Custom" shell type for custom shells, try to guess if the Shell is instead Bash or Zsh or another shell we provide specific support for.

We also use a slightly more portable prompt command for Bash. In my shells, it seems like `~` needs to be escaped in replacements, e.g.

```
$ NAME=kevin; echo ${NAME/kevin/~}
/Users/kevin
$ NAME=kevin; echo ${NAME/kevin/\~}
~
```

but because I don't know what affects this (Bash version? some other shell option?) I figure it's best to skirt the issue altogether.

### Automated Tests

N/A

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/13809.

### Documentation

N/A

### Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -33,6 +33,7 @@
 - Fixed a bug where project options updated in the Project Options pane were not properly persisted in RStudio 2023.09.0. (#13757)
 - Improved screen reader support when navigating source files in the editor. [accessibility] (#7337)
 - Fixed viewing or blaming file on GitHub for a newly created branch. (#9798)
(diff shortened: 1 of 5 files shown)


gtritchie (approved):
LGTM!

PR #13819, opened 2023-10-19
Add EnvironmentFile support to systemd service

bschwedler:
Set EnvironmentFile to `/etc/default/rstudio-server`

This supports an optional EnvironmentFile in the systemd service
definition for rstudio-server.
Prefixing this with `=-` will allow the service to start when the file
is not present.

### Intent

There are cases that we want to sent environment variables that are specific to a system. These can be added to a file that will be read when then `systemd` service starts up.

### Approach

Add `EnvironmentFile=-` to the systemd service definitions. The `-` prefix will allow the service to start if the file is not present.

### Automated Tests

n/a

### QA Notes

From my perspective, the largest risk here is adding another location where we would need to look for configuration.

### Checklist


kfeinauer:
Worth noting that Connect follows this particular pattern for adding environment variables to the service.

NEWS.md
```diff
@@ -8,6 +8,18 @@
 - RStudio no longer highlights `\[ \]` and `\( \)` Mathjax equations; prefer `$$ $$` and `$ $` instead (#12862)
 - Added cmake option to build RStudio without the check for updates feature (#13236)
 - Allow choosing R from non-standard location at startup (#14180; Windows Desktop)
+- Updated Ace to version 1.28. (#13708)
```
kevinushey, line 11: These NEWS updates look unintentional?

bschwedler: Thanks! For catching this. I'll fix up my rebase.

gtritchie (approved):
LGTM! Only question is if this should be documented in the admin guide (I don't know the current process for flagging things for a docs update or who would write it, etc.).

PR #13824, opened 2023-10-19
fix paths in image paste

kevinushey:
Addresses https://github.com/rstudio/rstudio/issues/12690.


donboyd5:
My first test of this in RStudio-2023.12.0-daily-212.exe (Windows desktop) works perfectly:
- screenshot pastes properly from clipboard into visual editor
- renders properly
- all good when loaded to web 

Having this feature is a big productivity boost for me.

Many thanks, @ronblum and @kevinushey.

Issue #13835, opened 2023-10-24
Click on file (not *.R) in File pane does not open it if file name in cyrillic


ariel32:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2023.09.1 Build 494
    OS Version      : Windows 10 Pro x64
    R Version       : R version 4.3.0 (2023-04-21 ucrt)

### Steps to reproduce the problem
Open in File pane any folder with files, some filenames in english, some with cyrillic.
![image](https://github.com/rstudio/rstudio/assets/7060961/d1929e97-90f9-462b-b5bf-94fc909f1bb1)

### Describe the problem in detail
If I click on file with name 'тест.docx' nothing happens. But click on 'test.docx' opens it. 

### Describe the behavior you expected
File opens in Word.

I attached empty word file with russian name:
[тест.docx](https://github.com/rstudio/rstudio/files/13113802/default.docx)

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ ] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.posit.co/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
@ariel32 Thank you for raising the issue! I can reproduce this in RStudio Desktop 2023.06.2+561 on Windows 11. It works correctly in 2022.07.2+576, as well as on MacOS 14. Also tested with Chinese, `测试.rtf`.

jonvanausdeln:
Verified with [2023.12.0-daily+266](https://dailies.rstudio.com/version/2023.12.0-daily+266)
Windows 11

Files open as expected from files pane.

Issue #12467, opened 2022-12-20
PDF (and xlsx/ods) files cannot be opened from Files pane when their filenames/paths include non-ASCII characters


k1-ito:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2022.12.0-353 (Elsbeth Geranium)
    OS Version      : Windows 10/11 Pro 22H2
    R Version       : 4.2.2

### Steps to reproduce the problem
1. Update Rstudio Desktop from 2022.07.2-576 (Spotted Wakerobin) to 2022.12.0-353.
1. In Files pane of RStudio, move to a folder where a PDF file is.
1. Click the PDF file in Files pane.

### Describe the problem in detail
When its filename or path includes non-ASCII character(s), the default PDF viewer does not launch and the PDF file is not opened.
In the case below, only `London.pdf` can be opened by clicking it.
(In contrast, any Rmd file is properly shown in Source pane.)
![Files](https://user-images.githubusercontent.com/121043466/208591154-837511cd-17af-449f-915c-5c4a0dfc1d6e.png)

### Describe the behavior you expected
Regardless of its filename or path, a PDF file is opened by clicking it in Files pane.

### Additional information
- Under the identical conditions mentioned above (non-ASCII characters are included in the filename or path), 
  - when clicking an xlsx file and selecting `View File` in Files pane, Excel does not launch and the file is not opened.
  - when clicking an ods file in Files pane, LibreOffice Calc does not launch and the file is not opened.
- In the previous version 2022.07.2-576 (Spotted Wakerobin), all of these files (PDF/xlsx/ods) are opened as expected.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
@k1-ito Thank you for raising this! We've run into issues with non-ASCII characters in the past. In this case, do the .Rmd file open, or do they fail to open, as well?

k1-ito:
[...]

PR #13846, opened 2023-10-25
handle non-ASCII characters in openExternal

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/13835.
~Addresses https://github.com/rstudio/rstudio/issues/12467~ EDIT: Nope, that appears to affect macOS; this PR targets Windows only.

### Approach

`shell.openExternal()` does not appear to handle Unicode paths correctly. Work around this the old fashioned way -- convert the path to a so-called "short" path name, and use that path instead.

### Automated Tests

N/A

### QA Notes

Test via notes in the associated issues.

### Documentation

N/A

### Checklist

### DIFF
--- a/src/node/desktop/src/native/desktop/desktop.cc
+++ b/src/node/desktop/src/native/desktop/desktop.cc
@@ -502,6 +502,7 @@ Napi::Value shortPathName(const Napi::CallbackInfo& info)
       }
 
       wShortPath.assign(buffer);
+      free(buffer);
    }
 
    // Set the resulting path.
--- a/src/node/desktop/src/native/desktop/desktop.cc
+++ b/src/node/desktop/src/native/desktop/desktop.cc
@@ -496,6 +496,10 @@ Napi::Value shortPathName(const Napi::CallbackInfo& info)
       }
 
       wchar_t* buffer = (wchar_t*) malloc(requiredSizeInBytesIncludingTerminator);
+      if (buffer == nullptr) {
+         return path;
+      }
+
       int numBytesWritten = ::GetShortPathNameW(wPath.data(), buffer, requiredSizeInBytesIncludingTerminator);
       if (numBytesWritten == 0) {
          return path;
--- a/src/node/desktop/src/native/desktop/desktop.cc
+++ b/src/node/desktop/src/native/desktop/desktop.cc
@@ -474,54 +474,40 @@ Napi::Value shortPathName(const Napi::CallbackInfo& info)
 
 #ifdef _WIN32
 
-   // Get path as UTF-8 text
-   std::string utf8Path = info[0].As<Napi::String>().Utf8Value();
+   // Get path as UTF-16 text.
+   // Note that, on Windows, std::wstring and std::u16string are interchangable.
+   // With the exception that char16_t is unsigned, and wchar_t is signed.
+   std::u16string u16Path = info[0].As<Napi::String>().Utf16Value();
+   std::wstring wPath(u16Path.begin(), u16Path.end());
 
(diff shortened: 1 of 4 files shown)


src/node/desktop/src/native/desktop/desktop.cc
```diff
@@ -468,6 +468,67 @@ Napi::Value cleanClipboard(const Napi::CallbackInfo& info)
...
+      int size = ::MultiByteToWideChar(
+         CP_UTF8, 0,
+         utf8Path.data(), utf8Path.size(),
+         &widePath[0], widePath.size()
+      );
+
+      if (size == 0) {
+         return path;
+      } else {
+         widePath.resize(size);
```
gtritchie, line 492: Windows allows paths of up to something like 32767 Unicode characters these days (depending on system configuration), so ideally we'd not assume 1024 as the max. Maybe we won't encounter anything in the real world using paths > 1024, but it's not impossible.

Best to first query for the size (via passing 0 for the last argument), then set the necessary size for the buffer and call again. Does add a bit of overhead but I doubt enough to impact our use of it.

src/node/desktop/src/native/desktop/desktop.cc
```diff
@@ -468,6 +468,67 @@ Napi::Value cleanClipboard(const Napi::CallbackInfo& info)
...
+      }
+   }
+
+   // Convert to a short path name
+   std::wstring wShortPath(1024, 0);
+   {
+      int size = ::GetShortPathNameW(widePath.data(), &wShortPath[0], wShortPath.size());
```
gtritchie, line 499: Similar comment about 1024 being shorter than theoretically allowable path lengths. For complete robustness here we'd first want to query for the necessary size, adjust the buffer if necessary, then call it again.

src/node/desktop/src/native/desktop/desktop.cc
```diff
@@ -468,6 +468,67 @@ Napi::Value cleanClipboard(const Napi::CallbackInfo& info)
...
+         return path;
+      } else {
+         wShortPath.resize(size);
+      }
+   }
+
+   // Convert back to a UTF-8 path
+   std::string aShortPath(1024, 0);
+   {
+      int size = ::WideCharToMultiByte(
```
gtritchie, line 510: Ditto on path length assumptions.

gtritchie (changes requested):
Don't ABSOLUTELY need to make these changes (especially if this is a pattern we're using in a bunch of other places).

kevinushey:
Getting this right was a bit trickier than I thought!

gtritchie (approved):
One small tweak requested just for best-practice purposes, otherwise looks good!

src/node/desktop/src/native/desktop/desktop.cc
```diff
@@ -468,6 +468,54 @@ Napi::Value cleanClipboard(const Napi::CallbackInfo& info)
...
+      // return value is the size of the buffer, _in TCHARs_, that is required to
+      // hold the path _and the terminating null character_.
+      //
+      // Hence, the return value here does include a null terminator.
+      int requiredSizeInBytesIncludingTerminator = ::GetShortPathNameW(wPath.data(), nullptr, 0);
+      if (requiredSizeInBytesIncludingTerminator == 0) {
+         return path;
+      }
+
+      wchar_t* buffer = (wchar_t*) malloc(requiredSizeInBytesIncludingTerminator);
```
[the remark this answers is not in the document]
kevinushey: Good catch, thanks!

PR #13850, opened 2023-10-26
fixups for short path conversion

kevinushey:
Follow-up to https://github.com/rstudio/rstudio/pull/13846.

### DIFF
--- a/src/node/desktop/src/native/desktop/desktop.cc
+++ b/src/node/desktop/src/native/desktop/desktop.cc
@@ -498,7 +498,7 @@ Napi::Value shortPathName(const Napi::CallbackInfo& info)
    }
 
    // Copy into that buffer.
-   int numBytesWritten = ::GetShortPathNameW((LPWSTR) wPath.data(), wShortPath, wShortPathSize);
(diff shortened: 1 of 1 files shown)


src/node/desktop/src/native/desktop/desktop.cc
```diff
@@ -477,41 +477,37 @@ Napi::Value shortPathName(const Napi::CallbackInfo& info)
...
+   if (wShortPath == nullptr) {
+      return path;
+   }
 
-      wShortPath.assign(buffer);
-      free(buffer);
+   // Copy into that buffer.
+   int numBytesWritten = ::GetShortPathNameW((LPCWSTR) wPath.data(), wShortPath, wShortPathSize);
+   if (numBytesWritten == 0) {
+      return path;
```
gtritchie, line 503: Nit: memory leak (Real Code, none of those sissy smart pointers).

gtritchie (approved):
One tiny nitpick otherwise LGTM!

src/node/desktop/src/native/desktop/desktop.cc
kevinushey: Doh. This is why I prefer C++ -- the compiler is a lot better at remembering what memory I've allocated than I am!

Issue #13118, opened 2023-05-12
Unexpected preview of named and hexadecimal colors in the console


zekiakyol:
### System details

    RStudio Edition : Desktop
    RStudio Version : RStudio 2023.03.1+446
    OS Version      : Windows 10, 22H2
    R Version       : R 4.2.3

---

Because this is an obvious case, I am not sure whether this is an intended behavior but I describe the problem below.
Thank you!

### Steps to reproduce the problem

1. RStudio Desktop > Tools > Global Options > Code > Display > Uncheck "Enable preview of named and hexadecimal colors".
   There is no problem in this step.
3. Crate test.R and write the code below, then save it.
   ```
   plot(mtcars$mpg, mtcars$hp, col = "red")
   plot(mtcars$mpg, mtcars$hp, col = "#2B750E")
   ```
   There is no problem in this step.
4. Copy and paste the same code in Step 2 to console.
5. **Problem:** color preview for `"red"` and `"#2B750E"` is enabled until we run the code by hitting the enter button.
6. When we run the code, color preview disappears. There is no problem in this step.

---

### Steps to reproduce the problem [Gif Version]

![issue_color](https://github.com/rstudio/rstudio/assets/40212849/9f236af5-6d83-49e2-ae80-6ee7556d9993)

- [X] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [X] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [X] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

---

### Describe the behavior you expected

When we uncheck the "Enable preview of named and hexadecimal colors" option, we expect to have no color preview in the console both before and after running the code.

collinn:
This issue persists in R Studio Desktop version 2023.09.01+494 on Ubuntu 22.04

kevinushey:
Thanks for the bug report! More concretely, the issue here is that the color preview option is not respected for console input, when highlight is enabled.

jonvanausdeln:
Just tried with [2023.12.0-daily+234](https://dailies.rstudio.com/version/2023.12.0-daily+234) on Windows, and the code is still highlighted in the console regardless of the Global options setting.

[...]

PR #13851, opened 2023-10-27
respect color preview pref in console

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/13118.

### Approach

We normally try to respect preference changes for editor instances here:

https://github.com/rstudio/rstudio/blob/main/src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTargetPrefsHelper.java#L128-L132

However, this only applies to Ace instances in the Source pane; not in the Console pane. This PR makes sure we apply the preference for Ace instances in the Console as well.

### Automated Tests

N/A

### QA Notes

N/A

### Documentation

N/A

### Checklist

### DIFF
--- a/src/node/desktop/envvars.bat
+++ b/src/node/desktop/envvars.bat
@@ -4,6 +4,7 @@ REM Run this script with 'call envvars.bat' to update the PATH in your
 REM cmd.exe shell, so that the appropriate versions of node are found
 set NODE_PATHS=^
 	..\..\..\dependencies\common\node\16.20.2^
+	..\..\dependencies\common\node\16.20.2^
 	c:\rstudio-tools\dependencies\common\node\16.20.2
 
 call :SetNodePath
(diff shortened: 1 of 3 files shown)


src/node/desktop/envvars.bat
```diff
@@ -4,6 +4,7 @@ REM Run this script with 'call envvars.bat' to update the PATH in your
 REM cmd.exe shell, so that the appropriate versions of node are found
 set NODE_PATHS=^
 	..\..\..\dependencies\common\node\16.20.2^
+	..\..\dependencies\common\node\16.20.2^
```
jgutman, line 7: Why is this here? seems like the dependencies folder 3 levels up is correct

jgutman (approved):

src/node/desktop/envvars.bat
kevinushey: Because I needed to source this from the `src/gwt` folder, because the build tasks were failing to find node. So really ... just a way to be lazy.

Issue #13857, opened 2023-10-27
RStudio IDE hangs / crashes on startup when rendering very large console outputs


kevinushey:
We had one user whose RStudio IDE session would freeze, and then crash, on startup. After analysis, it appeared to be due to an overly-large amount of text in the Console history -- the `console_actions` file had:

```
$ ll | grep console_actions
-rw-rw-r--   1 kevin  staff    34M Oct 27 14:22 console_actions
```

So, I suspect the IDE barfed when trying to shove 34MB of text into the Console history.

In this case, the large output was produced by `knitr::kable(<large data>, format = "html")` so that might be useful for making a reproducible example.

kevinushey:
To reproduce, with RStudio Server -- first, create an R file with the contents:

```
library(nycflights13)
knitr::kable(flights, format = "html")
```

Then, execute that code, and wait for the output to be printed in the console. Then, suspend the R session.

Now, come back and reload the tab. RStudio will start attempting to load, and then barf as it tries to load 296MB of Console history 🤮

kevinushey:
https://github.com/rstudio/rstudio/blob/ad0e5854c7a0abed5728c476ec3611f458e52d82/src/cpp/r/session/RConsoleActions.cpp#L95-L111

Perhaps we should split outputs by newlines, or do some other kind of chunking for overly-large outputs? The problem here is that we try to serialize all of the HTML (newlines included!) as one single enormous string; if we instead split on newlines and dropped earlier pieces of output, we'd be in much better shape.

jonvanausdeln:
Verified with [2023.12.0-daily+313](https://dailies.rstudio.com/version/2023.12.0-daily+313)
Ubuntu22 

IDE doesn't hang nor crash

PR #13858, opened 2023-10-27
improve chunking / buffering of console actions

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/13857.

### Approach

This PR adjust how we handle buffering of console outputs for console actions, which is used to populate the Console history when a suspended session is restored.

In particular, this PR makes a couple of changes:

1. We maintain a single `action_` action, which is used to accumulate outputs of the same type (imagine: lots of output fired into `stdout`);
2. As output is produced, we commit chunks of the `action_` buffer into the circular buffers that host the actual Console history.

This approach ensures that we always properly chunk console outputs, and avoids issues where a single-fire gigantic output was not properly chunked when committing that file to disk.

### Automated Tests

N/A

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/13857. In particular, we want to make sure the Console history representation remains correct when a suspended session is restored.

### Documentation

N/A

### Checklist

### DIFF
--- a/src/cpp/r/session/RConsoleActions.cpp
+++ b/src/cpp/r/session/RConsoleActions.cpp
@@ -96,21 +96,21 @@ void ConsoleActions::add(int type, const std::string& data)
          return;
       }
 
-      // check for type change, and push into buffer on type change
+      // if we've received more output of the same type, we can append that data
       bool isOutputType = type == kConsoleActionOutput || kConsoleActionOutputError;
-      if (isOutputType && type == pendingAction_.type)
+      if (isOutputType && type == action_.type)
       {
-         pendingAction_.data.append(data);
+         action_.data.append(data);
       }
       else
       {
          // the output type has changed; push the pending action
-         if (!pendingAction_.data.empty())
-            actions_.push_back(pendingAction_);
+         if (!action_.data.empty())
+            actions_.push_back(action_);
          
          // update the action with the newly-provided data
-         pendingAction_.type = type;
-         pendingAction_.data = data;
+         action_.type = type;
+         action_.data = data;
       }
       
       // consume chunks of data if available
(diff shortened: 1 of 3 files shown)


src/cpp/r/session/RConsoleActions.cpp
```diff
@@ -85,29 +88,51 @@ void ConsoleActions::add(int type, const std::string& data)
...
-
-      // automatically combine consecutive output actions (up to 512 bytes)
-      // we enforce a limit so that the limit defined for our circular buffer
-      // (see setCapacity above) implies a content size limit as well (if we
-      // didn't cap the size of combined output then the output actions could
-      // grow to arbitrary size)
-      if (type == kConsoleActionOutput &&
-          actionsType_.size() > 0      &&
-          actionsType_.back().getInt() == kConsoleActionOutput &&
-          actionsData_.back().getString().size() < 512)
```
kevinushey, line 103: Note that the previous code only set this restriction on the already-existing data, not the newly-incoming data, so a single large piece of data could still cause us to overflow here.

jeffvroom (approved):
LGTM!  Nice, that fix cleans things up quite a bit and the new code is easy to follow.

src/cpp/r/session/RConsoleActions.cpp
```diff
@@ -85,29 +88,51 @@ void ConsoleActions::add(int type, const std::string& data)
...
+      
+      // consume chunks of data if available
+      std::size_t offset = 0;
+      while (true)
+      {
+         // the remaining data in the buffer is less than the chunk size;
+         // now, substring that data to remove any data we've committed
+         if (action_.data.length() < offset + kChunkSize)
+         {
+            action_.data = action_.data.substr(offset);
```
kevinushey, line 124:
```suggestion
            if (offset > 0)
               action_.data = action_.data.substr(offset);
```

Issue #13860, opened 2023-10-30
Error when saving document while chunk is rendering


kevinushey:
Code:

```r
library(ggplot2)
data("diamonds")
ggplot(diamonds, aes(x = carat, y = price)) +
  geom_point()
```

To replicate error:
1. Put this code into a rmarkdown file. Make sure changes have not been saved.
2. Run chunks and press ctrl+s to save while the second chunk is executing.
3. If there were changes to be saved while the chunk was evaluating, I see an error something like:
```
Warning in gzfile(file, "wb") :
  cannot open compressed file '/basepath/.local/share/rstudio/notebooks/BD1727E3-troubleshooting_rmarkdown_figure_errors/1/0B0015E71ef6a3dc/ccpql3zezl1k3_t/d9df9a6f565d4ac79600b99612e834ab.snapshot', probable reason 'No such file or directory'
Error in gzfile(file, "wb") : cannot open the connection
```

RStudio info:
RStudio 2023.06.0+421 "Mountain Hydrangea" Release (583b465ecc45e60ee9de085148cd2f9741cc5214, 2023-06-06) for Ubuntu Jammy
Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) rstudio/2023.06.0+421 Chrome/110.0.5481.208 Electron/23.3.0 Safari/537.36

ggplot2 version: 3.4.2 (I had 3.4.0 until a few days ago and I haven't noticed any change in behavior since updating)

Note that I'm not certain whether this is the only time that I get a similar error, as I noticed it happening a lot from within map() or a for loop previously (and the error message here is a bit different from the one I reported in my previous post). However, the error's occurrence during a loop could have been a coincidence since I often save the rmarkdown file while waiting for a bunch of figures to be created.

_Originally posted by @HRodenhizer in https://github.com/rstudio/rstudio/issues/6260#issuecomment-1785756390_

jonvanausdeln:
Verified with [2023.12.0-daily+313](https://dailies.rstudio.com/version/2023.12.0-daily+313)
Windows 11

No errrors while trying to save and the chunk is executing.

PR #13861, opened 2023-10-30
always overwrite chunk migration files

kevinushey:
### Intent

Addresses #13860.

### Approach

Avoid treating an already-existing migration file as an error -- such files can exist if an R Markdown document is saved multiple times while a document is running.

### Automated Tests

N/A

### QA Notes

N/A

### Documentation

N/A

### Checklist


kevinushey:
@jeffvroom, requesting your review just because this might be related to the offline service, since I think that would be running when attempting to save a document while the R session is busy executing chunks. Does that sounds plausible?

jeffvroom (approved):
Yeah, this could be run from the offline thread if R is busy, so most likely that's why users started seeing this problem.  LGTM!

PR #13862, opened 2023-10-30
Note that `SessionInfo` may be out-of-date

sharon-wang:
### Intent

Just a little FYI for anyone using `SessionInfo` that the data stored may be stale if accessed anytime after session startup. Totally didn't run into this myself or anything 😜


kevinushey (approved):
LGTM; I've definitely made this mistake more than once before!

Issue #13856, opened 2023-10-27
CNTRL+ENTER automatically Views a matrix


canderson318:
### System details

    RStudio Edition : Desert Sunflower

    RStudio Version :RStudio 2023.09.1+494 "Desert Sunflower" Release (cd7011dce393115d3a7c3db799dda4b1c7e88711, 2023-10-16) for macOS Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) RStudio/2023.09.1+494 Chrome/116.0.5845.190 Electron/26.2.4 Safari/537.36

    OS Version      : macOS Sonoma 14.0

    R Version       :  4.3.1 (2023-06-16)

### Steps to reproduce the problem

rm(list = ls())
gc()

library(tidyr)
library(dplyr)
library(ggplot2)
library(reshape2)
library(tidyverse)
library(ggpubr)
library(vcd)
library(UpSetR)
library(gridExtra)
library(ComplexHeatmap)
library(viridis)
library(BiocGenerics)
library(mice)
library(Metrics)
library(purrr)
library(SingleCellExperiment)
library(corrplot)

m <- matrix(1, 10, 10)
m <- m %>% t()

### Describe the problem in detail

When running a function on a matrix or dataframe using CNTRL+ENTER, R will sometimes View() the object. This switches to the new data viewer tab away from the script I am working on.  

### Describe the behavior you expected

I expect the function to run without Viewing the result. If I want to View() the object, I expect to have to write that explicitly. 

- [x ] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.posit.co/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [ x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
A smaller reproducible example:

```
m <- matrix(1:4, nrow = 2)
View(m)
m <- matrix(5:8, nrow = 2)
```

kevinushey:
[...]

PR #13863, opened 2023-10-30
tweak how changes in viewed data are detected

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/13856.

### Approach

This PR removes some logic that tried to transition views from Data Viewer => Object Explorer, when the associated type of the underlying object changed. The old implementation appeared to be overly-sensitive to changes in the underlying object's type.

We could consider adding this back in the future, but given that this is a relatively painful bug, I felt it was worth fixing.

### Automated Tests

N/A

### QA Notes

N/A

### Documentation

N/A

### Checklist

### DIFF
--- a/src/cpp/session/modules/data/DataViewer.cpp
+++ b/src/cpp/session/modules/data/DataViewer.cpp
@@ -246,6 +246,7 @@ struct CachedFrame
       }
 
       // cache number of columns
+      nrow = safeDim(sexp, DIM_ROWS);
       ncol = safeDim(sexp, DIM_COLS);
    };
 
@@ -255,10 +256,9 @@ struct CachedFrame
    std::string envName;
    std::string objName;
 
-   // The frame's columns; used to determine whether the shape of the frame has
-   // changed (necessitating a full reload of any displayed version of the
-   // frame)
-   int ncol;
+   // The number of rows and columns; used to determine whether the shape of the frame has changed
+   // (necessitating a full reload of any displayed version of the frame)
+   int nrow, ncol;
    std::vector<std::string> colNames;
 
    // The current search string and filter set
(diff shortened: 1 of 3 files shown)


src/cpp/session/modules/data/DataViewer.cpp
```diff
@@ -246,6 +246,7 @@ struct CachedFrame
       }
 
       // cache number of columns
+      nrow = safeDim(sexp, DIM_ROWS);
```
kevinushey, line 249: I wonder if we should only do this for `data.frame` or `matrix` objects -- for database tables, I think this could be expensive.

src/cpp/session/modules/data/DataViewer.cpp
```diff
@@ -324,28 +324,6 @@ SEXP findInNamedEnvir(const std::string& envir, const std::string& name)
    return obj == R_UnboundValue ? nullptr : obj;
 }
 
-SEXP getNamedEnvir(const std::string& envir)
```
kevinushey, line 327: No longer used.

src/cpp/session/modules/data/DataViewer.cpp
```diff
@@ -1080,26 +1058,56 @@ void onDetectChanges(module_context::ChangeSource source)
...
+               if (error)
+                  LOG_ERROR(error);
+
+               // emit client event
+               json::Object changed;
+               changed["cache_key"] = i->first;
+               changed["type_changed"] = typeChanged;
+               changed["structure_changed"] = structureChanged;
+               changed["object_exists"] = true;
+               changed["object_class"] = json::toJsonArray(objectClass);
```
kevinushey, line 1104: Not currently used, but figure it could be useful if we later want to try and transition the view from Data Viewer to Object Explorer.

jgutman (approved):
This actually seems to work much better than before, not just the issue you linked here, but now datasets seem to actually refresh when you drop rows and when you drop columns, which they didn't before

Compare 
```
iris <- iris
View(iris)
iris <- dplyr::select(iris, Species)
```
This now actually refreshes with your current changes, which surprisingly before it did not, even though in theory the code appeared to be checking for a change in ncol and colNames.

It's perhaps a bit surprising (at least for the user) that the change detection is kind of sensitive to changes in namespace, for example that 
```
View(iris)
iris <- iris %>% select(Species)
```
doesn't trigger a refresh whereas this does:
```
iris <- datasets::iris # or even iris <- iris in a fresh session
View(iris)
iris <- iris %>% select(Species)
```
 but that's a separate issue

I think keeping nrow constant at -1 for remote tables makes sense, I wouldn't expect automatic refresh whenever the underlying tbl.src_dbi changed rows

src/cpp/session/modules/data/DataViewer.cpp
```diff
@@ -1080,26 +1062,56 @@ void onDetectChanges(module_context::ChangeSource source)
...
+            bool typeChanged = false;
+            if (i->second.observedSEXP != nullptr)
+            {
+               SEXP oldClass = Rf_getAttrib(i->second.observedSEXP, R_ClassSymbol);
+               SEXP newClass = Rf_getAttrib(sexp, R_ClassSymbol);
+               typeChanged = !R_compute_identical(oldClass, newClass, 0);
+            }
+            
+            bool structureChanged =
+                  i->second.nrow != newFrame.nrow ||
```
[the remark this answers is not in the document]
kevinushey: Yeah, I thought so too! My main guess was that row computation could be expensive for remote data (e.g. database tables) but that doesn't mean we can't do the computation for local tables.

kevinushey:
Thanks for the review + testing!

> It's perhaps a bit surprising (at least for the user) that the change detection is kind of sensitive to changes in namespace

Perhaps we should file an issue to track this?

Issue #12657, opened 2023-02-01
Find in Files doesn't load


ynraza:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2022.12.0+353 (2022.12.0+353)
    OS Version      : macOS Catalina 10.15.6
    R Version       : 4.2.2 (2022-10-31)

### Steps to reproduce the problem

Use 'find in files' function to search for a word.

### Describe the problem in detail

Find in files function doesn't work. It continues to "search" without finding anything or displaying results, even if searching in a path that contains specific files that would have the matching word.

<img width="790" alt="Screenshot 2023-02-01 at 15 47 17" src="https://user-images.githubusercontent.com/75734732/216092505-ab4666e8-6571-4c9d-944c-95c7f32ffce9.png">

Enabling RSTUDIO_GREP_DEBUG shows the following:
> Sys.setenv(RSTUDIO_GREP_DEBUG = 1)
grep '--binary-files=without-match' '-rHn' '--color=always' '--devices=skip' '-i' '-F' '-f' '/var/folders/0v/yk1jj65d6msbbx6m_kzn25c80000gn/T/Rtmpuo1FMS/rs_grep8f253f88b7dd.txt' '--include=*.r' '--include=*.R'

### Describe the behavior you expected

Find in files was working in previous versions of R Studio. Would display results/files.  

- [X] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [X] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [X] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

kevinushey:
Can you please share a screenshot of what your search query looks like? For example:

<img width="416" alt="Screenshot 2023-02-01 at 9 40 34 AM" src="https://user-images.githubusercontent.com/1976582/216120631-720d99bd-c8f4-4b81-a7c4-792f8b507fda.png">

The include statement here looks incorrect:

```
'--include=.r' '--include=.R'
```

as that should be `--include=*.r --include=*.R`, but I wasn't able to reproduce that.

ynraza:
Sorry just to follow up on this I have also tried using custom filter to include *.r  *.R and grep debug shows the following:

[...]

PR #13871, opened 2023-10-31
tolerate out-of-order RPCs in Find in Files

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/12657.

### Approach

It was observed that, in some scenarios, the resulting output could be received before a response from the `begin_find` RPC was returned. In this scenario, the output was silently dropped, because we relied on that RPC to generate a handle that mapped Find output to the current Find session.

The PR resolves this by letting the client generate the Find session handles, so that the current handle is known even in the scenario where Find output is received before the `begin_find` RPC response is received.

### Automated Tests

N/A

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/12657.

### Documentation

N/A

### Checklist


jgutman (commented):
This looks reasonable to me, I'm still trying to understand if there are certain circumstances that make this order more likely (would help with testing). How does this affect all the checks we have for `event.getHandle()` equal to `currentFindHandle_`? Don't we set the event handle in `initialize` https://github.com/rstudio/rstudio/blob/efb86c2b01fd2ec66c3c2d10769bad87a55ba710/src/gwt/src/org/rstudio/studio/client/workbench/views/output/find/FindOutputPresenter.java#L484

and then change the `currentFindHandle_` in the calls to `generateHandle()`? Is the mismatch then expected? For example if I hit the refresh button to re-run the search, then event handle and current find handle no longer match

Also it seems like I can crash RStudio by hitting the refresh button while a replace event is happening, but that appears to not be a new issue -- we disable the refresh and replace it with a stop button while a find is in progress, but I don't think we do that while a replace is in progress, instead we create a stop replace button but the refresh button remains clickable. But this I believe is a pre-existing issue

kevinushey:
> Don't we set the event handle in initialize

That `initialize` method is used only at startup; it's used to populate the Find in Files state in case the user connects to an R session with Find in Files already running. In effect, that means the handle produced there would match something that was generated from an earlier Find in Files session started by the user.

> Is the mismatch then expected? For example if I hit the refresh button to re-run the search, then event handle and current find handle no longer match

I think this makes sense, because that state here is specifically the initial state when the session was first started.

> Also it seems like I can crash RStudio by hitting the refresh button while a replace event is happening, but that appears to not be a new issue -- we disable the refresh and replace it with a stop button while a find is in progress, but I don't think we do that while a replace is in progress, instead we create a stop replace button but the refresh button remains clickable. But this I believe is a pre-existing issue

Thanks for catching that and filing https://github.com/rstudio/rstudio/issues/13873!

kevinushey:
> This looks reasonable to me, I'm still trying to understand if there are certain circumstances that make this order more likely (would help with testing).

I wasn't able to figure out how to reproduce this consistently either. My only guess is that this would be more likely to happen if the Find in Files search were to execute really fast, so perhaps it's more likely to occur in small projects with specific file filters?

jgutman (approved):

PR #13866, opened 2023-10-30
Download node archives from buildtools

gtritchie:
### Intent

Addresses #13773

### Approach

Downloaded node packages for Mac (x64 and arm), Linux (x64 and arm), and Windows (x64), uploaded them to S3 buildtools/node/vX.Y.Z, then tweaked the dependency scripts to download from there.

Also created the `RSTUDIO_BUILDTOOLS` variable in dependency scripts to hold the base URL for buildtools. I will do a separate PR to use this variable instead of hardcoding the location all over the dependency scripts (the URL may be changing with our move to a CDN).

### Automated Tests

N/A

### QA Notes

Nothing to test; this is purely a build / dev machine thing to (hopefully) eliminate the occasional glitches we have downloading node during builds.

### Documentation

N/A

### Checklist

- [x] If this PR adds a new feature, or fixes a bug in a previously released version, it includes an entry in `NEWS.md` 
- [x] If this PR adds or changes UI, the updated UI meets [accessibility standards](https://github.com/rstudio/rstudio/wiki/Accessibility)
- [x] A reviewer is assigned to this PR (if unsure who to assign, check Area Owners list)
- [x] This PR passes all local unit tests

PR #13872, opened 2023-10-31
use RSTUDIO_BUILDTOOLS instead of hardcoded URL in dependency scripts

gtritchie:
In a [previous PR](https://github.com/rstudio/rstudio/pull/13866) I added `export RSTUDIO_BUILDTOOLS="https://rstudio-buildtools.s3.amazonaws.com"` to `rstudio-tools.sh` (and `rstudio-tools.cmd`) and used it in the changes I was making.

This PR uses it in more places in the dependency scripts, so it will be easier to update if we need to when we move to using a CDN.

There are still hardcoded URLs in several R scripts, and also in some scripts used manually to upload things to S3.


kevinushey (approved):
LGTM!

Issue #13522, opened 2023-08-20
No ability to accept copilot suggestions word-by-word


Robinlovelace:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2023.09.0-daily+347 "Desert Sunflower" Daily (6b14ae05ab615040a0e7a1e66ae1737e6aef3f00, 2023-08-18) 
    OS Version      : Ubuntu Jammy
    R Version       : R version 4.3.1 (2023-06-16)

```
RStudio 2023.09.0-daily+347 "Desert Sunflower" Daily (6b14ae05ab615040a0e7a1e66ae1737e6aef3f00, 2023-08-18) for Ubuntu Jammy
Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) rstudio/2023.09.0-daily+347 Chrome/114.0.5735.289 Electron/25.5.0 Safari/537.36
```

### Steps to reproduce the problem

Authenticate copilot, write something, and try to accept only the first word by pressing Right arrow + Tab.

### Describe the problem in detail

You can only accept all or nothing, as far as I can tell.

### Describe the behavior you expected

It should work like VS Code's copilot where you can accept one word at a time by pressing Ctrl+Right: https://stackoverflow.com/questions/75114315/how-to-accept-only-the-next-word-in-github-copilot-suggestion

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

jonvanausdeln:
I tested with [2023.12.0-daily+313](https://dailies.rstudio.com/version/2023.12.0-daily+313) on Windows 11 and Ubuntu22

I can only seem to get full completion, not word-by-word.  I tried several combinations of the right arrow key combinations including `Right Arrow + Tab` as suggested by OP.

Example of what I'm seeing
![image](https://github.com/rstudio/rstudio/assets/6433789/ff53d1c7-8828-4a5f-baea-69cab121b0e3)

I can only get the full line to be accepted.

@kevinushey what is the key combination that should trigger this?

[...]

PR #13879, opened 2023-11-01
add command for accept next word in copilot

kevinushey:
Addresses https://github.com/rstudio/rstudio/issues/13522.


jgutman (commented):
The build is now failing with 

```
src/gwt/test/org/rstudio/studio/client/workbench/commands/DummyCommands.java:6: 
error: DummyCommands is not abstract and does not override abstract method 
               copilotAcceptNextWord() in Commands
```

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/TextEditingTargetCopilotHelper.java
```diff
@@ -295,7 +297,42 @@ else if (event.getKeyCode() == KeyCodes.KEY_BACKSPACE)
    @Handler
    public void onCopilotRequestCompletions()
    {
-      completionTimer_.schedule(0);
+      if (display_.isFocused())
```
jgutman, line 300: good catch

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/AceEditor.java
```diff
@@ -1192,14 +1192,15 @@ public void scrollToY(int y, int animateMs)
       fireEvent(new ScrollYEvent(Position.create(getFirstVisibleRow(), 0)));
    }
 
-   public void insertCode(String code)
+   public void insertCode(String code, boolean unused)
```
jgutman, line 1195: We don't have like a naming convention for this in Java, like `_unused` ?

jgutman (approved):
LGTM, pending fixing the issue causing the build failure. Does this just create the command, or does this also create the way to trigger it via keyboard shortcut ? (it doesn't seem like it does the latter, but maybe I'm missing something) We should add to NEWS (at least once this can be triggered by the user)

src/gwt/src/org/rstudio/studio/client/workbench/views/source/editors/text/AceEditor.java
kevinushey: As far as I know, we don't, at least for the Java sources.

Issue #13015, opened 2023-04-18
Find in Files error "Not a git repository" on Windows but not on Mac


eipi10:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2023.3.0.386
    OS Version      : Windows: Server 2016 Standard version 1607, Build 14393.5850
                      Mac: Ventura 13.2.1
    R Version       : 4.2.3
    git Version     : Windows: 2.40.0.windows.1 
                      Mac: 2.39.2

### Steps to reproduce the problem

On Windows, open an R project or package, select `Find in Files` from the `Edit` menu. Type in any search term. Displays error message "Not a git repository".

The problem only occurs on the Windows server I remote into and when the package or project was loaded from a separate network drive. The problem does not occur on the Windows server with a package or project stored locally on the server, and it does not occur at all on my Macbook Pro laptop, regardless of whether the package/project folder is local or on the network drive.

### Describe the problem in detail

I remote into a Windows server to do most of my work. My R project and package folders are stored on a separate network drive that the Windows server can access. All of these projects have git repositories. Until a couple of weeks ago, "Find in Files" worked as expected. I don't think I made any software changes just the before the problem started occurring. After I realized the error was occurring with every project/package I opened, I upgraded to the latest version of git, but the error still occurred.

The error message is:
```
Not a git repository
`\\path\to\project\folder`
CMD.EXE with the above path as the project directory.
UNC paths are not supported. Defaulting to Windows directory.
fatal. not a git repository (or any of the parent directories): .git
```
This issue happens only on the Windows machine and with a package/project folder on the network drive. It does not happen on my Mac laptop (tested with the same projects and packages loaded from the same network drive), or with project/package folders stored locally on the Windows server. 

### Describe the behavior you expected

I expected Find in Files to return text matches as it normally does.

- [ x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [ NA] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.rstudio.com/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [ x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

ronblum:
@eipi10 Thank you for raising this! We've run into multiple issues with shared and network drives—some fixed, some persist. I'll tag this for review.

[...]

PR #13904, opened 2023-11-09
run search as program instead of command

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/13015.

### Approach

Use `runProgram()` rather than `runCommand()`, so that we can support UNC paths on Windows. This also helps us obviate the need to escape paths and other shell-specific characters.

### Automated Tests

### QA Notes

To test locally, you can add 'localhost' (that is, your own machine) as a network location. To do this, open the File Explorer, click on "This PC" on the left bar, then use the More (...) dropdown and select Add a Network Location.

image

Then, as the network location, type `\\localhost\c$`:

image

Give that the default name. Now, when you use File -> New Project or File -> Open Project in RStudio, you can use that network location (which is really just your own machine). Then, try and test Find in Files in one of these projects, both with and without git enabled.

### Documentation

N/A

### Checklist


gtritchie (approved):
LGTM!

PR #13915, opened 2023-11-09
update to Node 18.18.2

kevinushey:
Addresses https://github.com/rstudio/rstudio-pro/issues/5275.

TODO: Test on CentOS 7, and disable if necessary.


timtmok (commented):
https://github.com/rstudio/rstudio/blob/8f6dc26013936ab3e84263cd0eed6418ba133975/CMakeGlobals.txt#L237

Looks like another location to update

kevinushey:
Also confirming this version of Node won't work on CentOS 7:

```
[root@9b931bd6180d bin]# ./node --version
./node: /lib64/libm.so.6: version `GLIBC_2.27' not found (required by ./node)
./node: /lib64/libc.so.6: version `GLIBC_2.25' not found (required by ./node)
./node: /lib64/libc.so.6: version `GLIBC_2.28' not found (required by ./node)
./node: /lib64/libstdc++.so.6: version `CXXABI_1.3.9' not found (required by ./node)
./node: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.20' not found (required by ./node)
./node: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by ./node)
```

kevinushey:
For CentOS 7, we'll still need to use Node 16 since we need that as part of building and installing the desktop components. I think we can accomplish this (and perhaps simplify things a bit) by using a symlink (or, junction point on Windows).

- In the sources, we'll stop hard coding the path to Node in various locations, and instead use a generic "node" path.
- We'll allow for multiple Node installations, but use a symlink to the different versions as appropriate.

timtmok (approved):

Issue #13918, opened 2023-11-12
Terminals don't start when Terminal pane is set to use WSL


gtritchie:
### System details

    RStudio Edition : Desktop
    RStudio Version : All builds since early 2022
    OS Version      : Windows specific
    R Version       : Any

### Steps to reproduce the problem
- Ensure you have WSL (Windows Subsystem for Linux) enabled and working
- In RStudio Terminal options, set "New terminals open with" to "Bash (Windows Subsystem for Linux)"
- Try to start a new terminal in the Terminal pane

### Describe the problem in detail
- Some text briefly flashes in the terminal pane then it goes blank
- A new terminal is not started

### Describe the behavior you expected
- A wsl terminal session starts

This appears to have regressed due to https://github.com/rstudio/rstudio/pull/10283/ (early 2022!). I have reproduced this issue with several older builds back to that point, and it worked on builds before that.

The fix is a very simple 1-liner; don't pass `-l` to `bash.exe` when starting the wsl terminal. Given WSL sessions are running in a separate subsystem and a different operating system (unlike a git-bash terminal) I don't think conda integration is meaningful for WSL terminals.

I don't know if this has been 100% broken since then, or if it actually worked on wsl at that time, and has stopped working since then. I've only tested on the latest Windows-11 and WSL,

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.posit.co/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

gtritchie:
Found mentions of this on community forum:

https://community.rstudio.com/t/setting-the-terminal-to-use-wsl-bash-in-windows-10/97904
https://community.rstudio.com/t/terminal-instant-crash-when-set-as-wsl-windows-subsystem-for-linux/157874

jonvanausdeln:
Verified with [2023.12.0-daily+313](https://dailies.rstudio.com/version/2023.12.0-daily+313)
Windows 11, WSL2

WSL terminal opens up as expected.

kamalmdmostafa:
Some text briefly flashes in the terminal pane then it goes blank

gtritchie:
[...]

PR #13919, opened 2023-11-12
fix WSL terminals so they start

gtritchie:
# Don't merge; waiting to hear from QA if they want to take this for Ocean Storm or a future release

### Intent

Addresses #13918

### Approach

Reverts the WSL-specific part of https://github.com/rstudio/rstudio/pull/10283 by not passing any arguments to `system32\bash.exe` when we try starting a WSL terminal pane.

The conda integration mentioned in that issue was aimedat git-bash based terminals. The technique used for that won't work with WSL terminals, plus they are running in a separate subsystem and operating system so even if the WSL terminal was starting the integration would never have worked, near as I can tell.

### Automated Tests

I doubt we have WSL-specific terminal automation (would require running the tests on Windows Desktop with WSL installed).

### QA Notes

This change is very narrowly scoped to just the WSL terminal codepath, so test setting the terminal to use WSL, and see if you can start a terminal (you will need WSL installed, of course).

### Documentation

N/A

### Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -46,6 +46,7 @@
 - Fixed an issue that could cause errors to occur if an R Markdown document was saved while a chunk was running. (#13860)
 - Fixed an issue where console output could be dropped when rendering large ANSI links. (#13869)
 - Fixed an issue preventing users from copying code from the History pane. (#3219)
+- Fixed WSL terminals not starting on RStudio Desktop for Windows. (#13918)
 
 #### Posit Workbench
 - Fixed opening job details in new windows more than once for Workbench jobs on the homepage. (rstudio/rstudio-pro#5179)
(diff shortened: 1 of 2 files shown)


kevinushey:
Do our terminal hooks work with this change? (They're mainly responsible for putting the configured version of Python on the PATH for Terminal sessions.)

I can try to test next week as well.

kevinushey (approved):

gtritchie:
Did terminal hooks ever work with WSL terminals? The python you run from a WSL prompt is the one installed in the Linux subsystem, not the one being used by Windows applications.

gtritchie:
Makes me wonder if WSL terminal support in the IDE is worth keeping. Don't know how widely used it is, and it's quite detached from the rest of the IDE's universe.

Perhaps we might should disable or hide the terminal integration setting in Terminal options when WSL is selected to indicate that this is the case (unless I'm misunderstanding the scenario).

kevinushey (approved):
LGTM!

Confirmed with QA that we're comfortable taking this for Ocean Storm, due to the limited scope of the change.

Issue #13667, opened 2023-09-20
Copilot: Only enable project indexer when Copilot itself is enabled


olivroy:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2023.09.0+401
    OS Version      :  Windows
    R Version       : 4.3.1

### Steps to reproduce the problem

1, Open RStudio project on Network drive (windows)
2. Use setwd() for another place in the network drive. (I use forward slashes to specify the path, and RStudio + `normalizePath()` complain.

### Describe the problem in detail

I get a warning. 

```
Warning message:
File monitoring failed for project at `NW_DRIVE_LOCATION`
Error 2 (Cannot find the specified file) 
Features disabled: R source file indexing, Diagnostics, Copilot indexing
```

This is not bothering me at all. (I am opening the issue more as a heads-up for you).
 So feel free to close if this has been fixed or can't reproduce.

But I never activated Copilot (which is why I think this warning may show up for Windows users that they will not understand once the release is complete.

### Describe the behavior you expected

No warnings. Especially not ones regarding Copilot, when I never used it.

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

If you can't reproduce easily, I will close.

Thank you so much for RStudio.

ronblum:
Thanks @olivroy ! A few things:
1) Are you able to open, edit, and save files after doing the above?
2) Can you include the commands that you use for `setwd()` and `normalizePath()`? Scrubbed is fine, but please include the slashes so that I can try to reproduce exactly.
3) Is this still an issue if you use a mapped drive?
4) What version of Windows are you using?

@kevinushey Note Copilot message even though Copilot wasn't enabled.

olivroy:
> 1. Are you able to open, edit, and save files after doing the above?

Yes, things work. I only have this warning that shows up.

> 2. Can you include the commands that you use for `setwd()` and `normalizePath()`? Scrubbed is fine, but please include the slashes so that I can try to reproduce exactly.

The `setwd()` part was irrelevant, sorry.

It is possibly due to the fact, that since` normalizePath()` thinks that my path does not exist (forward slash vs backslash)
 (since I am using forward slashes or fs paths.

> 3. Is this still an issue if you use a mapped drive?

I have not checked on a mapped drive. But it is for a path that starts with "//"

[...]

PR #13950, opened 2023-11-15
only subscribe to file monitor if indexing enabled

kevinushey:
Addresses https://github.com/rstudio/rstudio/issues/13667.

### DIFF
--- a/src/cpp/session/modules/SessionCopilot.cpp
+++ b/src/cpp/session/modules/SessionCopilot.cpp
@@ -186,6 +186,11 @@ bool s_isSessionShuttingDown = false;
 // Project-specific Copilot options.
 projects::RProjectCopilotOptions s_copilotProjectOptions;
 
+// Forward declaration.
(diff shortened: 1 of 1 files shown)


src/cpp/session/modules/SessionCopilot.cpp
```diff
@@ -985,6 +990,12 @@ void synchronize()
    s_copilotEnabled = isCopilotEnabled();
    s_copilotIndexingEnabled = s_copilotEnabled && isCopilotIndexingEnabled();
    
+   // Subscribe to file monitor if enabled
+   if (s_copilotIndexingEnabled)
+   {
+      static bool once = subscribeToFileMonitor();
```
kevinushey, line 996: This makes use of the fact that, in calls of the form:

```
static T t = f();
```

The function `f()` will only be evaluated only once, on first use.

timtmok (approved):

PR #13956, opened 2023-11-16
Update connection pane deduplication for Posit drivers

bschwedler:
### Intent

The connection pane allows for drivers with a duplicate name to be listed as a duplicate with the suffix of ` with RStudio Drivers`.

Currently duplicates are only allowed if the ODBC installation entries that have `Installer = RStudio`, which is set during the installation of drivers through Desktop Pro.

This change will allow duplicates to be listed with the following entries for `Installer`:
- `RStudio Pro Drivers`
- `Posit Pro Drivers`

### Approach

- Change connection pane suffix to "with Posit Driver"
- Allow duplicate connection pane entries from pro drivers
- Find connection snippet by walking the directory
    The original logic for determining the R snippet path was quite fragile.
    We now walk up the directory tree from driver library until we find the `snippets` directory instead of using a simple regex

- Add `dsn=` to connection snippet from DSNs
    Make the Connection snippet explicit that the second argument is the DSN

- Set path of desktop driver install to match original name
    In cases where the driver name is a duplicate of a previous entry, Desktop Pro would include the `with posit driver` in the directory path
    Installation and uninstallation now only include the original driver name in the path

### Automated Tests

A test was added to [test-connections.R](https://github.com/rstudio/rstudio/commits/main/src/cpp/tests/testthat/test-connections.R) for `.rs.connectionListFilter`

### QA Notes

The easiest way to test this is to add a few placeholder entries in `/etc/odbc.ini` that will then be listed by DSN entries.

```ini
[Athena]
Driver = /my/fake/path

[MySQL]
Driver = /another/fake/path

[PostgreSQL]
Driver = /try/this/path

[Snowflake]
Driver = /yet/another/path
```

#### Test on Workbench

1. Install the Pro Drivers
2. Add the above entries to `/etc/odbc.ini` (on Linux)
3. Start Workbench and open the Connections Pane

Expected result:
* Two entries in the list for every driver that was included in `/etc/odbc.ini`
* The second entry will have ` with Posit Driver` appended to it

#### Test on Desktop Pro

1. Add the above entries to the appropriate `SYSTEM DATA SOURCES` from `odbcinst -j`
2. Start Desktop Pro and open the Connections Pane
3. Install one or more of the drivers that have `with Posit Driver` included in the title

Expected result:
* Two entries in the list for every driver that was included in `/etc/odbc.ini`
* The second entry will have ` with Posit Driver` appended to it
* The installation path of the driver only includes the base name of the driver
* The driver can successfully be installed and uninstalled

### Documentation

No documentation was added because this is an internal behavior change.

### Checklist

### DIFF
--- a/src/cpp/session/modules/SessionConnections.R
+++ b/src/cpp/session/modules/SessionConnections.R
@@ -749,18 +749,31 @@ options(
    })
 })
 
-.rs.addJsonRpcHandler("get_new_connection_context", function() {
-   connectionList <- c(
-      list(),
-      .rs.connectionReadSnippets(),         # add snippets to connections list
-      .rs.connectionReadDSN(),              # add ODBC DSNs to connections list
-      .rs.connectionReadPackages(),         # add packages to connections list
-      .rs.connectionReadOdbc(),             # add ODBC drivers to connections list
-      .rs.connectionReadInstallers(),       # add installers to connections list
-      .rs.connectionReadPackageInstallers() # add package installers to connection list
-   )
+.rs.addFunction("connectionListFilter", function(connectionList) {
+   # Function to filter and deduplicate entries in the connections list
+   # The list may include multiple entries that have the same name,
+   # which may be from the same source or from different sources altogether
+   #
+   # IMPORTANT: We reference and modify the connectionList entries in place,
+   # even though it is less readable so that returned entries are modified
    
    connectionList <- Filter(function(e) !is.null(e), connectionList)
+
+   # Installers that are managed by Pro Drivers or the RStudio IDE
+   pro_driver_installers <- c("RStudio Pro Drivers", "Posit Pro Drivers")
+   posit_installers <- c("RStudio", pro_driver_installers)
+   # Helpers for connection list deduplication
+   is_desktop_install <- function(entry) {
+      identical(as.character(entry$installer), "RStudio") ||
+      (
+         # set by connectionReadInstallers
+         identical(as.character(entry$type), "Install") &&
+         identical(as.character(entry$subtype), "Odbc")
+      )
+   }
+   is_pro_driver <- function(entry) {
+      toString(entry$installer) %in% pro_driver_installers
+   }
 
    # remove duplicate names, in order
    connectionNames <- list()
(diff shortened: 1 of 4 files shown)


src/cpp/session/modules/SessionConnectionsInstaller.R
```diff
@@ -611,5 +612,9 @@
 })
 
 .rs.addFunction("connectionOdbcRStudioDriver", function() {
-   " with RStudio Driver"
+   " with Posit Driver"
+})
+
+.rs.addFunction("connectionStripRStudioDriver", function(connection) {
```
bschwedler, line 618: I'm not a huge fan of this name, but I wanted to keep it similar to `connectionOdbcRStudioDriver`

src/cpp/session/modules/SessionConnections.R
```diff
@@ -552,18 +559,18 @@ options(
...
 
-      hasInstaller <- identical(driverInstaller$value, "RStudio")
+      hasInstaller <- identical(driverInstaller, "RStudio")
       warningMessage <- NULL
 
       if (hasInstaller) {
          installerVersion <- .rs.connectionInstallerInfo(driverId)$version
-
+         currentVersion <- drivers[drivers$attribute == "Version" & drivers$name == driver, ]$value
          currentVersion <- drivers[drivers$attribute == "Version" & drivers$name == driver, ]
```
sharon-wang, line 574: I think only one of these assignments is intentional -- I believe we want the new addition only (line 573)?

src/cpp/session/modules/SessionConnections.R
```diff
@@ -527,20 +527,27 @@ options(
...
+      while (!identical(walkDir, prevDir)) {
+         snippetsDir <- file.path(walkDir, "snippets")
+         if (dir.exists(snippetsDir)) {
+            break
+         }
+         prevDir <- walkDir
+         walkDir <- dirname(walkDir)
+      }
+
+      snippetsFile <- file.path(snippetsDir, paste(tolower(driverId), ".R", sep = ""))
```
GCRev, line 549: Do we want to consider stopping this at the home directory? 

This could also struggle when ACLs are involved since we've had issues before with another piece of (C++) code that walks up the directory structure when ACLs are involved. I don't think it's a huge issue here, though.

bschwedler: Good thinking on the ACLs.

We can't just stop at the user's home directory because it also will read entries from the system ODBC config like
* `/opt/simba/spark/lib/64/libsparkodbc_sb64.so`
* `/opt/rstudio-drivers/postgresql/bin/lib/libpostgresqlodbc_sb64.so`

sharon-wang (approved):
LGTM!

Issue #13969, opened 2023-11-21
Speed up core C++ utility FilePath.isDirectory()


jeffvroom:
The code makes an unnecessary `exists()` call that returns false, when it could just pass in an error code to is_directory() since that returns a "no_such_file_or_directory" error.  The exists() call turns into an unnecessary `stat()` system call.

gtritchie:
Nothing specifically testable here. Unit tests pass, and if this isn't working a lot of things will be noticably broken.

PR #13970, opened 2023-11-21
Remove unnecessary exists() test in FilePath.isDirectory()

jeffvroom:
Not a high priority but I've seen these 2X stat() calls in straces for a while. This seems like a safe change to me that could lower some electricity bills :)

### Intent

Addresses https://github.com/rstudio/rstudio/issues/13969

### Approach

Use error code variant of is_directory to avoid the exception that the exists() tests was likely introduced to avoid.  I left the try/catch there in the unlikely event some exceptions are still thrown.

### Automated Tests

This api is used all over, and there's a basic unit test. 

### QA Notes

No manual testing required

### DIFF
--- a/src/cpp/shared_core/FilePath.cpp
+++ b/src/cpp/shared_core/FilePath.cpp
@@ -1198,16 +1198,12 @@ bool FilePath::isDirectory() const
 {
    try
    {
-      if (!exists())
-         return false;
-      else
-      {
-         return boost::filesystem::is_directory(m_impl->Path)
+      boost::system::error_code ec;
+      return boost::filesystem::is_directory(m_impl->Path, ec);
 #ifdef _WIN32
-            || isJunction()
+         || isJunction()
 #endif
-            ;
-      }
+         ;
    }
    catch(const boost::filesystem::filesystem_error& e)
    {


kevinushey (approved):
LGTM!

src/cpp/shared_core/FilePath.cpp
```diff
@@ -1198,16 +1198,12 @@ bool FilePath::isDirectory() const
 {
    try
    {
-      if (!exists())
-         return false;
-      else
-      {
-         return boost::filesystem::is_directory(m_impl->Path)
+      boost::system::error_code ec;
+      return boost::filesystem::is_directory(m_impl->Path, ec);
```
kevinushey, line 1202: I think the trailing semicolon here will cause an error with the following `|| isJunction()` on Windows. (I wonder if the Boost implementation already supports the detection of junctioned directories.)

kevinushey (changes requested):

src/cpp/shared_core/FilePath.cpp
gtritchie: Boost release notes (1.57, long time ago) say: "The Windows implementation now treats NTFS directory junctions (also known as junctions, also known as mount points) as symlinks. This has the effect of treating directory junctions as directories, and thus supporting all operations suitable for directories."

I'll tweak the PR to ensure it builds on Windows and do a test to double-check that is_directory does indeed treat junctions as directories, and report back.

gtritchie: Confirmed that Boost's `is_directory` does treat junction points as directories.

Issue #13966, opened 2023-11-20
RStudio Desktop doesn't run on Linux if installed from scratch (upgrade works)


ronblum:
### System details

    RStudio Edition : Desktop
    RStudio Version : 2023.12.0-daily+313
    OS Version      : Debian 12.2, Red Hat 9.2, possibly other Linux builds
    R Version       : 4.2.3

### Steps to reproduce the problem

1. Uninstall RStudio Desktop, if already installed.
2. Install RStudio Desktop.
3. `rstudio`

### Describe the problem in detail

RStudio won't start. Instead, this error appears:
```
bash: /usr/bin/rstudio: Permission denied
```

Here's more detail about the file system.

```
[vmware@localhost ~]$ rstudio
bash: rstudio: command not found...

[vmware@localhost ~]$ which rstudio
/usr/bin/which: no rstudio in (/home/vmware/.local/bin:/home/vmware/bin:/usr/local/bin:/usr/local/sbin:/usr/bin:/usr/sbin)

[vmware@localhost ~]$ ls -l /usr/bin/rstudio
lrwxrwxrwx. 1 root root 24 Nov 20 11:26 /usr/bin/rstudio -> /usr/lib/rstudio/rstudio

[vmware@localhost ~]$ ls -l /usr/lib/rstudio/rstudio
ls: cannot access '/usr/lib/rstudio/rstudio': Permission denied

[vmware@localhost ~]$ sudo ls -l /usr/lib/rstudio/rstudio
-rwxr-xr-x. 1 root root 168468592 Nov 19 18:56 /usr/lib/rstudio/rstudio

[vmware@localhost ~]$ sudo ls -ld /usr/lib/rstudio
drwx------. 4 root root 4096 Nov 20 11:26 /usr/lib/rstudio
```

### Describe the behavior you expected

RStudio Desktop should start up.

This was working correctly with version 2023.12.0-daily+212. Let me know if you'd like me to figure out which build it started failing on.

Also note that the first couple of times that I tried, this only happens if RStudio is installed from scratch and upgrading from 2023.09.1 worked. Now that doesn't work, either.

- [ ] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [ ] I have installed the latest version of RStudio, and confirmed that the issue still persists.
[...]

PR #13979, opened 2023-11-22
Fix Linux install folder permissions

timtmok:
### Intent
Address #13966

### Approach
Add directory permissions for the top-level directory. I didn't look too deeply into why the electron-forge/webpack upgrade caused this as a quick check didn't reveal anything obvious.

The group and world have execute set because it's required for listing directory contents. Other programs are installed this way and it's what we used to have before the regression.

### Automated Tests
n/a

### QA Notes
The tarballs were also affected but it wouldn't have been noticed since the permissions were for owner only and the user would be the owner in that case. This has been fixed anyway as the directory permission is set prior to packaging the installer and tarball.

### Documentation
n/a

### Checklist


kevinushey (approved):
LGTM!

kevinushey:
https://github.com/rstudio/rstudio/pull/13979/files#diff-78657f965037cf1abab0e9c739a1079f9239e572d0389e497495a2ece5f643dfR219

Is this still needed?

src/node/desktop/CMakeLists.txt
```diff
@@ -211,6 +211,10 @@ elseif(LINUX)
 
    install(
       DIRECTORY "${ELECTRON_BUILD_DIR}/out/RStudio-linux-${ELECTRON_ARCH}/"
+      DIRECTORY_PERMISSIONS
+      OWNER_READ OWNER_WRITE OWNER_EXECUTE
+      GROUP_READ GROUP_EXECUTE
+      WORLD_READ WORLD_EXECUTE
       DESTINATION "${RSTUDIO_INSTALL_ELECTRON}"
       USE_SOURCE_PERMISSIONS
```
kevinushey, line 219: Is this line still needed? I wonder if the issue reflects an issue with the permissions on the directory created here now that webpack etc. has been updated?

timtmok: I'm not totally familiar with `cmake` so I didn't try without the `USE_SOURCE_PERMISSIONS` option. It seemed that only the top-level directory was affected by the change to be user only accessible.

My thought is that it is needed to preserve the file permissions.

kevinushey: Makes sense! Thanks; let's keep it.

sharon-wang (approved):
LGTM too!

PR #14003, opened 2023-11-28
OS Companion commit for rstudio/rstudio-pro#5490

zachhannum:
Companion commit for rstudio/rstudio-pro#5490


src/cpp/server/include/server/ServerOptions.hpp
```diff
@@ -86,8 +86,6 @@ class Options : public GeneratedOptions,
       return overlayOptions_[name];
    }
 
-   std::string rsessionExecutable() const;
```
zachhannum, line 89: It looks like this was removed from pro at some point? I don't see the function being called anywhere.

skyeturriff (approved):
lgtm!

PR #14030, opened 2023-12-07
allow copilot proxy SSL override; tweak output error display

kevinushey:
- The environment variable might be necessary for RStudio Desktop users.
- Make it clear when an error occurs but no output was available.


kevinushey:
Cherry-picked to main in https://github.com/rstudio/rstudio/commit/55c42e8d9c0df19a6566000f550a0fa6dc519899.

Issue #14029, opened 2023-12-07
Issues showing and updating current branch in external editor windows


gtritchie:
### System details

    RStudio Edition : Desktop or Server
    RStudio Version : RStudio 2024.04.0-daily+74 (or any version, this isn't a new bug)
    OS Version      : any
    R Version       : any

### Note

Noticed while working on #1909, likely end up being fixed as part of that.

### Steps to reproduce the problem

* Open an R project that is using git
* On Desktop (but not server) the current git branch is displayed in the main window title ("project - branch - RStudio")
* Open a source file from that project (e.g. Foo.R), then click "Show in new window" button
* The title of this separate window is "project - branch - RStudio Source Editor" (desktop and server)
* In git pane, change to a different branch
* Examine window titles
* Close and restart RStudio
* Examine window titles

### Describe the problem in detail

* the main window title (on Desktop) updates to show the new branch, but the title on editor window still shows the old branch
* after closing and restarting, the editor window doesn't show the branch at all, just "project - RStudio Source Editor"

### Describe the behavior you expected

* the current branch should be accurate in all windows
* when the app is restarted, the editor window title(s) should show the branch like they did before the restart

- [x] I have read the guide for [submitting good bug reports](https://github.com/rstudio/rstudio/wiki/Writing-Good-Bug-Reports).
- [x] I have installed the latest version of RStudio, and confirmed that the issue still persists.
- [x] If I am reporting an RStudio crash, I have included a [diagnostics report](https://support.posit.co/hc/en-us/articles/200321257-Running-a-Diagnostics-Report).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

Issue #1909, opened 2018-01-04
allow projects to be renamed


kevinushey:
From https://support.rstudio.com/hc/en-us/community/posts/115008315647-Allow-to-rename-projects:

---

Would it be possible to add names for Rstudio projects. Currently the folder name is used. This is for many reasons not always a perfect solution.

Since 2013, many issues have been raised:

https://support.rstudio.com/hc/en-us/community/posts/200663977-Allow-projects-to-be-renamed-independently-of-the-project-folder-in-RStudio

https://support.rstudio.com/hc/en-us/community/posts/200664107-renaming-projects

https://support.rstudio.com/hc/en-us/community/posts/200854386-Naming-Renaming-projects

https://support.rstudio.com/hc/en-us/community/posts/202294727-rename-a-project

Some have been alleviated by adding the full folder name for ambiguous projects. However, I think that many problems can still be solved by just adding a custom name field. This can default to the folder but should be editable.

This would also allow for a quick renaming in Rstudio.

I'm not speaking about packages, since they already use the package name.

philstraforelli:
Just stumbled onto this. This would be very useful for me, as in my work I have to put in my .rproj file within a project's subfolder, that is usually named "Data", and I can't keep track over which "Data" project is which project.

Is it possible to just manually rename the name of the .rproj file?

benscarlson:
I have the same problem, the name in the task bar (OS X) is always "src", because that is where I always place my source code and is where I create the .Rproj file. It seems you can rename the .Rproj file (src.Rproj -> myproj.Rproj), but the name displayed in the task bar is still "src" because the name is taken from the folder. 

Perhaps an option to display the name in the taskbar based on the .Rproj file name instead?

djhurio:
Same here. Thanks!

alperyilmaz:
same here (Linux).. whatever the .Rproj filename is, the folder name is used as project name..

DesiQuintans:
Just ran into this too, it would be nice if the .Rproj's name was inherited, not the parent folder's name.

raubreywhite:
Same here. Project name is always "git" or "src"

Avi-Kenny:
I'll upvote this as well. This would be very useful.

wathnai:
+1, I have same folder structure for all my projects, so all my RStudio projects are "Code"

rmflight:
[...]

PR #14049, opened 2023-12-13
New feature: customize project display name

gtritchie:
### Intent

Addresses #1909 and #14029

### Approach

Users can specify a custom name for an R project in the Project Options dialog. This displays instead of the name of the project folder in the window title for Desktop, the browser tab for Server, and the dock icon on macOS. The custom name is appended (in parentheses) to entries in the File / Recent Projects menu and the Projects dropdown. The project dropdown button itself displays the custom name instead of the folder name (but the tooltip for that button still shows the path including the .Rproj file).

The project name is stored in the `.Rproj` file. If you set a name and then open the project in an older version of RStudio and it rewrites the project file, the name will be lost. This is true for anything we add to this file; a given RStudio version will only write back the fields it knows about and fields added by a newer RStudio will be lost.

The project MRU data is being stored in a new file (`project_name_mru`) where each line starts with the path to the `.Rproj` (exactly the same as in the old `project_mru` file). For each entry with a custom name, a tab character is appended as a separator, followed by the custom name. These files live in `~/.local/share/rstudio/monitored/lists/` for Mac/Linux and `%localappdata%\RStudio\monitored\lists\` on Windows.

If there isn't a `project_name_mru` file at startup it will be seeded with the contents of the old `project_mru` file. Both files will be kept up-to-date as projects are opened and closed (this allows downgrading RStudio without seeing an out-of-date project mru). If the recent projects list is cleared by the user, both files will be emptied.

Here's the options dialog with no custom name entered, followed by an example with a name:

screenshot of project options dialog with no custom name set

screenshot of project options dialog with a custom name

The name updates in all places when the project options dialog is closed (no need to restart). This PR includes minor fixes for satellite editor windows; they weren't updating the title when the git branch changed and didn't show the full title when a session with existing satellite editor windows was started up (#14029).

The title display logic (e.g. for the main window, satellite editor windows, the dock icon, and browser tab for server) is the same as before, except that when a custom name is available for a project, it will be shown where the folder name was previously being shown. Here's an example with Mac Desktop:

mac-title

Here's the project dropdown button/menu showing a custom name for the current project. Note that the differences with more or less path info being shown is existing behavior, I didn't change that. I opted to append the custom name to the MRUs rather than replacing the folder name info (showing both here should provide maximum context for choosing a file, I think).

dropdown

### Automated Tests

None yet; I did create unit tests for the core MRU code (`MruList.cpp`) that I modified before I changed it, then added more tests for the new functionality (which is to support adding extra data after a separator character; the extra data doesn't enter into the uniqueness check for each entry, and can also be updated without moving the entry's position in the list).

### QA Notes

See the spec for more info (search for "IDE Spec: Customize R Project Names" in Confluence).

In general, if a custom name isn't entered, there should be no differences in behaviour (except for the bug fixes related to window titles not always updating when current git branch is changed).

This PR doesn't touch the pro-specific places that show project names: the Workbench Home page and the Project Sharing UI, so those continue showing folder-names for projects; see https://github.com/rstudio/rstudio-pro/issues/5589

### Documentation

TBD

### Checklist


src/gwt/src/org/rstudio/studio/client/projects/StudioClientProjectConstants_fr.properties
```diff
@@ -228,3 +228,5 @@ quartoBlogSubTitle=Créer un nouveau projet blog
 quartoBlogPageCaption=Créer un blog Quarto
 lblOrigin=Origine:
 useCondaenv=Utilisez condaenv avec les packages :
+generalTitle=General
+customProjectNameLabel=Project display name (defaults to folder name):
```
timtmok, line 232: Just a note for others that this string change will show up in the `locdiff` results since it is a new string.

timtmok (approved):

Issue #14044, opened 2023-12-12
Login failed in FreeIPA's password+otp authentication environment


geonmo:
Hello, rstudio community.

I am running an authentication system based on freeipa(LDAP-like). 

In that system, I can select password or password+otp authentication for each user, and some users are logging in with password+otp.

Currently, I have found that users who authenticate with password+otp cannot log in to rstudio.

This is probably because the PAM-related module pam_conversation is not able to handle such logins.

If we connect with password+otp, the connection method will be changed as below.

~]$ pamtester rstudio user authenticate
First Factor: 
Second Factor: 
pamtester: successfully authenticated

Could you please refer to the above and add a feature to allow access from freeipa OTP environment?

gtritchie:
See the associated PR for comments on why we aren't doing this.

PR #14050, opened 2023-12-14
Support sign-in for OTP-enabled accounts in SSSD environments

geonmo:
### Intent

> SSSD supports enforcing OTP for some accounts. Accounts with OTP enabled will not be able to log in with traditional codes, as the password field will look different and require you to enter a password and OTP. 
> Therefore, it was necessary to modify the code.  I have raised an issue in #14044 to request a resolution of this issue.

### Approach

> We had to modify our sign-in webpages and JavaScript code to accommodate OTP input. Since the UI has changed, it's a good idea to use the code as a reference. 
> In this PR, you'll be passing a password and an OTP number that will be separated by a tab character as the password value to pass the OTP information. This way, the information will be passed safely to the PAM processing module and can be separated at the end. 
> In src/cpp/server_core/system/Pam.cpp was incorrectly using the pam_response array, and we fixed it. The existing method causes a segfault error when processing more than one message. 

### Automated Tests

> Since the code is related to login, an SSSD and an account with OTP applied are required for auto-testing. I don't think it's possible to automate it.

### QA Notes

> This PR has fixes for UI things. We raised the PR as a way to deliver the code to the core developers rather than merge it directly.

### Checklist

### DIFF
--- a/src/gwt/www/templates/encrypted-sign-in.htm
+++ b/src/gwt/www/templates/encrypted-sign-in.htm
@@ -87,6 +87,21 @@
                 aria-required="true"
               /><br />
             </p>
+            <p>
+              <label for="otp">OTP:</label><br />
+              <input
+                type="password"
+                name="otp"
+                autocomplete="off"
+                autocorrect="off"
+                autocapitalize="off"
+                spellcheck="false"
+                value=""
+                id="otp"
+                size="45"
+                aria-required="true"
+              /><br />
+            </p>
             <div style="display: #staySignedInDisplay#">
               <input
                 type="checkbox"
--- a/src/cpp/server/pam/PamMain.cpp
+++ b/src/cpp/server/pam/PamMain.cpp
@@ -95,6 +95,14 @@ int main(int argc, char * const argv[])
       // read password (up to 200 chars in length)
       const int MAXPASS = 200;
       std::string password = string_utils::consumeStdin(string_utils::StdinSingleLine);
+      std::string otp;
+      size_t splitTab = password.find('\t');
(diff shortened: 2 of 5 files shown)


gtritchie:
Hi, thank you for your contribution, and sorry we haven't had a chance to evaluate this yet but are hoping to soon.

In the meantime have you filled out and emailed the contributor agreement per the instructions at https://github.com/rstudio/rstudio/blob/main/CONTRIBUTING.md#contributing-code?

Thank you.

gtritchie:
Note to reviewers: this change changes the default sign-in page to look like this, with no way to turn off that extra field:

screenshot of rstudio server signin page with proposed changes

src/gwt/www/templates/encrypted-sign-in.htm
```diff
@@ -87,6 +87,21 @@ <h1 id="caption_header" class="caption">
                 aria-required="true"
               /><br />
             </p>
+            <p>
```
gtritchie, line 90: We can't have this extra UI showing by default. It would need to be controlled via a setting, probably in rserver.conf, and either a have new login page for this mode, or perhaps conditionally show this additional UI.

There is an injection mechanism for this page, see logoHtml for an example (corresponding code is in ServerLoginPages.cpp).

gtritchie (commented):
Unfortunately, after an internal discussion, we are unable to accept this PR. Concerns include the cost of testing (we'd need to setup environments to verify this scenario for each RStudio release), documentation and support, and the impact on derivative products that share this core auth code.

Thank you for your contribution, and sorry we are unable to merge it into the mainline.

PR #14054, opened 2023-12-14
guard against `NA` length()

t-kalinowski:
### Intent

User defined `length()` methods can return things other than a scalar integer. The IDE should guard against that when inspecting the environment with `isSerializable()`.

This was encountered with the IDE inspecting reticulate objects that intentionally raise an exception from `__bool__`(), and thus resetting the value presented by `reticulate::py_last_error()`

### QA Notes

`isTRUE()` was added in R 3.5. If this is an issue, we can avoid using it. 


kevinushey:
Thanks!

PR #14080, opened 2023-12-22
Add snyk code scanning to PR pipeline

GCRev:
See: https://github.com/rstudio/rstudio-pro/issues/5606

### Intent

Implement snyk code scanning in the PR pipeline.

### Approach

Bring over the existing snyk Dockerfile from the pro repo for use in the OS repo, so that we can include it in the PR build pipeline for both OS and Pro. Create a new stage for the snyk code scan.

The theory behind the step is to "whitelist" the changed files so that snyk code does not upload every single file in the codebase for every PR. Snyk doesn't have a way to do this -- it only lets you ignore files or directories. I also wanted to avoid copying these files anywhere inside the existing build repo to avoid the chance that packaging will pull over a directory that we weren't expecting, even if the packaged build output is not uploaded. If we were to copy this stage into another one of our build pipelines, then we could run into that problem.

It doesn't matter where the copied files actually go because snyk code can be told where to scan. As far as I can tell, it will skip over files that are "not code," so documentation and text files are ignored. I'm not totally sure about that, though.

### Automated Tests

N / A

### QA Notes

I've verified that the `sh` steps work, but I haven't tested this in Jenkins directly. At the moment, I don't know enough about the setup of the CI PR process to make changes to it.

### Checklist

### DIFF
--- a/docker/jenkins/Dockerfile.snyk
+++ b/docker/jenkins/Dockerfile.snyk
@@ -1,0 +1,10 @@
+ARG BASE_IMAGE=ubuntu:bionic
+FROM 263245908434.dkr.ecr.us-east-1.amazonaws.com/$BASE_IMAGE
+
+# Install a separate version node and npm because the one used in the base image 
+# is difficult to access due to PATH issues
+RUN curl -sL https://deb.nodesource.com/setup_18.x | bash -
+RUN apt-get install -y nodejs
+
+# install snyk
+RUN npm install -g snyk
--- a/jenkins/Jenkinsfile.pull-request
+++ b/jenkins/Jenkinsfile.pull-request
@@ -12,14 +12,15 @@ pipeline {
   }
 
   environment {
-        GITHUB_LOGIN = credentials('posit-jenkins')
-        OS = 'jammy'
-        ARCH = 'x86_64'
-        FLAVOR = 'Server'
-        TYPE = 'DEB'
-        AWS_ACCOUNT_ID = '749683154838'
-        RSTUDIO_VERSION_FLOWER = ''
-        IS_PRO = false
+    GITHUB_LOGIN = credentials('posit-jenkins')
+    OS = 'jammy'
+    ARCH = 'x86_64'
+    FLAVOR = 'Server'
+    TYPE = 'DEB'
+    AWS_ACCOUNT_ID = '749683154838'
+    RSTUDIO_VERSION_FLOWER = ''
+    IS_PRO = false
+    BASE_IMAGE = ""
   }
 
   stages {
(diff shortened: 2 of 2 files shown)


cm421 (approved):
LGTM!

GCRev:
yolo

GCRev:
I think I'll also wait for @zachhannum to get a look at this before merging.

sharon-wang (approved):
👍

jenkins/Jenkinsfile.pull-request
```diff
@@ -32,10 +33,33 @@ pipeline {
...
+          filename 'docker/jenkins/Dockerfile.snyk'
+          registryCredentialsId 'ecr:us-east-1:aws-build-role'
+          registryUrl 'https://263245908434.dkr.ecr.us-east-1.amazonaws.com'
+          additionalBuildArgs "--build-arg BASE_IMAGE=${BASE_IMAGE}" + ' --build-arg GITHUB_LOGIN=${GITHUB_LOGIN}'
+          reuseNode true
+        }
+      }
+      steps {
+        sh "mkdir --parents ${TMP_SNYK_PATH}"
+        sh "cp --parents -r $(git diff --name-only ${env.CHANGE_TARGET}) ${TMP_SNYK_PATH}"
```
zachhannum, line 57: I'm seeing this error in the PR job:
```
2023-12-22 15:37:57  Obtained jenkins/Jenkinsfile.pull-request from d3bfbced81287930aaf32657ba3137d044e36065
2023-12-22 15:37:57  org.codehaus.groovy.control.MultipleCompilationErrorsException: startup failed:
2023-12-22 15:37:57  WorkflowScript: 57: illegal string body character after dollar sign;
2023-12-22 15:37:57     solution: either escape a literal dollar sign "\$5" or bracket the value expression "${5}" @ line 57, column 9.
2023-12-22 15:37:57             sh "cp --parents -r $(git diff --name-only ${env.CHANGE_TARGET}) ${TMP_SNYK_PATH}"
2023-12-22 15:37:57             ^
2023-12-22 15:37:57  
2023-12-22 15:37:57  1 error
```

I think you just need to escape the first dollar sign since you want Groovy to interpret that as a $ literal instead of string interpolation.

```suggestion
        sh "cp --parents -r \$(git diff --name-only ${env.CHANGE_TARGET}) ${TMP_SNYK_PATH}"
```

zachhannum (changes requested):

jenkins/Jenkinsfile.pull-request
GCRev: Good catch. Added the suggestion in.

docker/jenkins/Dockerfile.snyk
```diff
@@ -0,0 +1,10 @@
+ARG BASE_IMAGE=ubuntu:bionic
+FROM 263245908434.dkr.ecr.us-east-1.amazonaws.com/$BASE_IMAGE
```
zachhannum, line 2: Is there a reason we need to use our actual build image here instead of just a lighter weight base image?

zachhannum (commented):
Looks great, just noticed one more thing!

docker/jenkins/Dockerfile.snyk
GCRev: I don't believe I changed this from the one in the pro repo. We could absolutely change it to something lighter weight, but I haven't looked into that.

zachhannum: Ahh I didn't realize this already existed in the Pro repo.

zachhannum (approved):
🚀

PR #14100, opened 2024-01-08
Feature/release dispatcher

cm421:
## Intent

Addresses https://github.com/rstudio/rstudio-pro/issues/4698 by reorganizing our Jenkins dispatcher jobs. Today our nightly and hourly dispatchers are organized like this with the hourly dispatcher triggering the nightly dispatcher in series:
Screenshot 2024-01-08 at 2 02 29 PM

This PR reorganizes them to be in parallel and trigger one main "Build Dispatcher":
Screenshot 2024-01-10 at 8 45 02 AM

Among other things, this new organization will let us easily fine tune how often `main` builds as opposed to released branches.

### Dispatchers

The Build Dispatcher takes the majority of the Nightly Dispatcher's functionality in that it will build docker images when necessary and triggers all the builds to each of the Linux/Mac/Windows Product build jobs. This dispatcher only runs when triggered by the Hourly/Nightly/Release dispatchers

Hourly and Nightly Dispatchers exclusively build the `main` branch and apart from using the new Build dispatcher to actually trigger product builds will effectively function the same as before: Hourly dispatcher will continue to start a smaller subset of builds on commits to `main` (rate limited to no more than once an hour), and Nightly dispatcher will continue to build a full compliment of builds once a day.

Additionally, since the hourly and nightly dispatchers are so similar, they've been combined into on "Main" dispatcher that will trigger either nightly or hourly builds based on the trigger (or the `DAILY` value if it's started manually)

The Release Dispatcher exclusively builds release branches starting with `rel-*`. It will build a full compliment of builds on commits to these branches (also rate-limited to no more than once an hour), and it will build once a month on the first weekend of the month simply to maintain confidence in our ability to build older releases.

### Avoiding rebuilding an already built binary

In addition to dispatcher reorganizing, the product builds will now check the dailies page to see if a build has already been uploaded, and will skip a build if it already exists. This checks down at the specific OS/ARCH/FLAVOR build level, so if a jammy `x86_64` server build exists, but an `amd64` build does not, the `amd64` build will continue and the `x86_64` build will be skipped. This check can be overridden by the `FORCE_REBUILD` Jenkins build parameter.

#### Rebuilding and the Release Dispatcher

This new dailies check to avoid unnecessary rebuilds will need to be modified for the Release Dispatcher's monthly run, as it will almost certainly be rebuilding an already-built version of builds. The rebuild logic for a monthly rebuild of a release branch will look something like this:
Screenshot 2024-01-10 at 8 24 10 AM

## Automated Tests

N/A build changes

## QA Notes

N/A build changes

## Documentation

N/A build changes

## Checklist


jenkins/Jenkinsfile.linux
```diff
@@ -20,56 +20,6 @@ def package_os =
   'opensuse15': 'OpenSUSE 15'
 ]
 
-// Use a function to keep the pipeline length down.
```
cm421, line 23: These were moved into the utils.groovy script

utils.groovy
```diff
@@ -233,4 +247,97 @@ def getDockerTag() {
...
+  def renamedFile = file.replace('-relwithdebinfo', '')
+  sh "mv ${dir}/${file} ${dir}/${renamedFile}"
+  return renamedFile
+}
+
+/**
+  * Rename the file on the filesystem
+  */
+def renameTarFile(String dir) {
+  if ((FLAVOR == "Desktop") || (FLAVOR == "Electron")) {
```
GCRev, line 264: Putting this `if` here makes me discomfortable. I would recommend putting it around the calling line instead so that nothing unpredictable happens if we use this elsewhere.

GCRev: I checked the code and I see why this if statement made it in here. We can probably leave it, but it still makes me antsy.

utils.groovy
```diff
@@ -233,4 +247,97 @@ def getDockerTag() {
...
+      // build an x86_64 rhel8 WB, x86_64 rhel9 QT RDP, and an arm64 jammy Electron RDP 
+      inHourlySubset = ((env.OS == 'rhel8' && env.ARCH == 'x86_64' && env.FLAVOR == 'Server') ||
+        (env.OS == 'centos7' && env.ARCH == 'x86_64' && env.FLAVOR == 'Desktop') || 
+        (env.OS == 'jammy' && env.ARCH == 'arm64' && env.FLAVOR == 'Electron'))
+    } else {
+      // build an arm64 rhel9 Desktop and an x86_64 focal Server
+      inHourlySubset = ((env.OS == 'rhel9' && env.ARCH == 'arm64' && env.FLAVOR == 'Electron') ||
+        (env.OS == 'focal' && env.ARCH == 'x86_64' && env.FLAVOR == 'Server'))
+    }
+  }
```
GCRev, line 298: I realize I'm nit-picking code we already had, and I hate to be that guy but do you think the legibility of this could benefit from a few early returns?

jenkins/Jenkinsfile.linux
```diff
@@ -141,6 +93,7 @@ pipeline {
                 RSTUDIO_VERSION_SUFFIX) = utils.getVersion(!params.DAILY)
               RSTUDIO_VERSION_FLOWER = utils.getFlower()
               IS_PRO = RSTUDIO_VERSION_SUFFIX.contains('pro')
+              RSTUDIO_VERSION_FILENAME = utils.getVersionFilename(RSTUDIO_VERSION) // Define here for use later in utils.rebuildCheck()
```
GCRev, line 96: 🤔

utils.groovy
```diff
@@ -73,6 +73,13 @@ def getFlower() {
   return readFile(file: 'version/RELEASE').replaceAll(" ", "-").toLowerCase().trim()
 }
 
+def getVersionFilename(String version) {
+  return sh(
+    script: "echo ${version} | sed -e 's/[^a-zA-Z0-9-]/-/g' | sed -e 's/--*/-/g'",
+    returnStdout: true
+  ).trim()
```
GCRev, line 80: Opportunity to use `replaceAll` instead of running this through sh?
```suggestion
  return version.replaceAll('[^a-zA-Z0-9-]', '-').replaceAll('--*', '-').trim()
```

cm421: Much better

utils.groovy
cm421: Tricky since these were originally linux-only. I can add comments explaining when to use them, but anything in particular that would make you less discomfortable?

GCRev: I'm fine with things being specific to Linux, it's more that we have this conditional here that is opaque unless you open this file and know that it's there, rather than being logic performed directly in the pipeline itself. I think you can leave it since we aren't using it anywhere else.

To make it clearer that they are linux-related you could rename the function to `renameTarFileLinux` or something like that, since `Desktop` `Electron` builds on Windows and Mac as well.

jenkins/Jenkinsfile.linux
```diff
@@ -351,9 +307,9 @@ pipeline {
 
                 environment {
                   PACKAGE_DIR = "package/linux/${BUILD_LOCATION}"
-                  PACKAGE_FILE = "${renameFile(PACKAGE_DIR, 'rstudio-*.' + ext_map[env.OS])}"
+                  PACKAGE_FILE = "${utils.renameFile(PACKAGE_DIR, 'rstudio-*.' + ext_map[env.OS])}"
                   TAR_PACKAGE_DIR= "${PACKAGE_DIR}/_CPack_Packages/Linux/${ext_map[env.OS].toUpperCase()}"
-                  TAR_PACKAGE_FILE = "${renameTarFile(TAR_PACKAGE_DIR)}"
+                  TAR_PACKAGE_FILE = "${utils.renameTarFile(TAR_PACKAGE_DIR)}"
```
GCRev, line 312:
```suggestion
                  TAR_PACKAGE_FILE = "${(FLAVOR == "Desktop" || FLAVOR == "Electron") ? utils.renameTarFile(TAR_PACKAGE_DIR) : ''}"
```

Something like this, where we use a ternary here instead of putting it inside the utils function, just to make it clear that we want to perform an action in this case or leave it blank otherwise.

GCRev: I'm _assuming_ that you can do a ternary here. I'm pretty sure you can in ordinary Groovy strings, but I'm not sure about the environment directive.

jenkins/Jenkinsfile.linux
cm421: As per our offline slack discussion, I'm going to ignore this, haha.

The reason this is done is because when `rebuildCheck()` runs in a `when` block, it has limited functionality and can't make these calls. Pre-defining it here allows `rebuildCheck()` to use this value later on.

GCRev (approved):
Overall I think the structure is good.

jenkins/Jenkinsfile.main
```diff
@@ -0,0 +1,89 @@
...
+  // Daily trigger for daily builds
+  triggers {
+    cron 'H H(20-23) * * *'
+  }
+  
+  parameters {
+    booleanParam name: 'DAILY', description: 'Builds daily builds if true; builds hourly builds if false.'
+    booleanParam defaultValue: true, description: 'Publish the build to S3.', name: 'PUBLISH'
+    booleanParam defaultValue: false, description: 'Force build even if there are no changes.', name: 'FORCE_BUILD'
+    booleanParam defaultValue: false, description: 'Rebuild binaries even if they already exist on dailies page' name: 'FORCE_REBUILD'
```
sharon-wang, line 24: I think `FORCE_BUILD_DOCKER` might be missing here!

```suggestion
    booleanParam defaultValue: false, description: 'Force build even if there are no changes.', name: 'FORCE_BUILD'
    booleanParam defaultValue: false, description: 'Force rebuild docker images', name: 'FORCE_BUILD_DOCKER'
    booleanParam defaultValue: false, description: 'Rebuild binaries even if they already exist on dailies page' name: 'FORCE_REBUILD'
```

jenkins/Jenkinsfile.main
```diff
@@ -0,0 +1,89 @@
...
+      steps {
+        build wait: false,
+              job: "IDE/${env.JOB_URL.contains('Pro') ? 'Pro' : 'OS'}-Builds/build-pipeline/${env.BRANCH_NAME.replace('/', '%2F')}",
+              parameters: [
+                string(name: "SLACK_CHANNEL", value: "${SLACK_CHANNEL}"),
+                booleanParam(name: "FORCE_BUILD", value: false),
+                booleanParam(name: "FORCE_BUILD_DOCKER", value: false),
+                booleanParam(name: "DAILY", value: false),
+                booleanParam(name: "PUBLISH", value: params.PUBLISH),
+                booleanParam(name: "FORCE_REBUILD", value: params.FORCE_REBUILD),
```
sharon-wang, line 53: This is a nit suggestion to specify these parameters in the same order as Jenkinsfile.release

```suggestion
                booleanParam(name: "FORCE_BUILD", value: false),
                booleanParam(name: "FORCE_BUILD_DOCKER", value: false),
                booleanParam(name: "FORCE_REBUILD", value: params.FORCE_REBUILD),
                booleanParam(name: "DAILY", value: false),
                booleanParam(name: "PUBLISH", value: params.PUBLISH),

```

jenkins/Jenkinsfile.main
```diff
@@ -0,0 +1,89 @@
...
+      steps {
+        build wait: false,
+              job: "IDE/${env.JOB_URL.contains('Pro') ? 'Pro' : 'OS'}-Builds/build-pipeline/${env.BRANCH_NAME.replace('/', '%2F')}",
+              parameters: [
+                string(name: "SLACK_CHANNEL", value: "${SLACK_CHANNEL}"),
+                booleanParam(name: "FORCE_BUILD", value: params.FORCE_BUILD_BINARIES),
+                booleanParam(name: "FORCE_REBUILD", value: params.FORCE_REBUILD),
+                booleanParam(name: "FORCE_BUILD_DOCKER", value: params.FORCE_BUILD_DOCKER),
+                booleanParam(name: "DAILY", value: true),
+                booleanParam(name: "PUBLISH", value: params.PUBLISH),
```
sharon-wang, line 82: This is a nit suggestion to specify these parameters in the same order as Jenkinsfile.release

```suggestion
                booleanParam(name: "FORCE_BUILD", value: params.FORCE_BUILD_BINARIES),
                booleanParam(name: "FORCE_BUILD_DOCKER", value: params.FORCE_BUILD_DOCKER),
                booleanParam(name: "FORCE_REBUILD", value: params.FORCE_REBUILD),
                booleanParam(name: "DAILY", value: true),
                booleanParam(name: "PUBLISH", value: params.PUBLISH),
```

jenkins/Jenkinsfile.main
```diff
@@ -0,0 +1,89 @@
...
+
+  // Daily trigger for daily builds
+  triggers {
+    cron 'H H(20-23) * * *'
+  }
+  
+  parameters {
+    booleanParam name: 'DAILY', description: 'Builds daily builds if true; builds hourly builds if false.'
+    booleanParam defaultValue: true, description: 'Publish the build to S3.', name: 'PUBLISH'
+    booleanParam defaultValue: false, description: 'Force build even if there are no changes.', name: 'FORCE_BUILD'
```
sharon-wang, line 23: It looks like we expect this parameter to be named `FORCE_BUILD_BINARIES` instead of `FORCE_BUILD` below on line 78

```suggestion
    booleanParam defaultValue: false, description: 'Force build even if there are no changes.', name: 'FORCE_BUILD_BINARIES'
```

jenkins/Jenkinsfile.build
```diff
@@ -241,7 +238,8 @@ pipeline {
                     string(name: "OS_FILTER", value: params.OS_FILTER),
                     string(name: "ARCH_FILTER", value: params.ARCH_FILTER),
                     booleanParam(name: "DAILY", value: params.DAILY),
-                    booleanParam(name: "PUBLISH", value: params.PUBLISH)
+                    booleanParam(name: "PUBLISH", value: params.PUBLISH),
+                    booleanParam(name: "FORCE_REBUILD", value: params.FORCE_REBUILD)
```
sharon-wang, line 242: I suspect you're focusing on Linux at the moment -- just a note to add this `FORCE_REBUILD` parameter to the MacOS and Windows stages as well!

sharon-wang (commented):
Thank you for doing this huge lift!! 🙌 

Overall LGTM, mainly just some nits :)

I wonder if we can reduce possible confusion between the `FORCE_BUILD` `FORCE_REBUILD` and `FORCE_BUILD_BINARIES` parameters via naming/descriptions/docs.
- Could we use one of `FORCE_BUILD` or `FORCE_BUILD_BINARIES` to fold those two into one parameter name?
- Some questions I can foresee myself wondering, in a few months when I've forgotten the details 😛:
    - Which force build parameter(s) should I set?
    - Does one parameter take precedence over another parameter?
    - What's the difference between build and rebuild?
    - Do the parameters interact with the `PUBLISH` parameter differently?

cm421:
@sharon-wang, The `FORCE-*` names are a little messy! I think `FORCE_BUILD`/`REBUILD`/`BUILD_BINARIES` can all be combined so we only have these two:

- `FORCE_BUILD_BINARIES`: Build all binaries even if there are no changes, and even if the binaries have already been built
- `FORCE_BUILD_DOCKER`: Build all docker images, even if there are no changes

sharon-wang:
> @sharon-wang, The `FORCE-*` names are a little messy! I think `FORCE_BUILD`/`REBUILD`/`BUILD_BINARIES` can all be combined so we only have these two:
> 
> * `FORCE_BUILD_BINARIES`: Build all binaries even if there are no changes, and even if the binaries have already been built
> * `FORCE_BUILD_DOCKER`: Build all docker images, even if there are no changes

Nice!! That makes sense to me 👌

PR #14108, opened 2024-01-10
update OS news

GCRev:
Update NEWS to include Databricks integration.

### DIFF
--- a/docs/news/index.qmd
+++ b/docs/news/index.qmd
@@ -34,6 +34,7 @@ This page provides the release notes associated with each release of RStudio and
 - Improved support for development documentation when a package has been loaded via `devtools::load_all()`. (#13526)
 - Better error message when user preferences fail to save due to folder permissions. (#12974)
 - With screen reader support enabled, hitting ESC key allows Tabbing away from editor. [accessibility] (#13593)
+- Added Databricks integration to RStudio Pro Sessions which allows users to seamlessly interact with Databricks clusters and resources (rstudio-pro#4927)
 
 #### Posit Workbench
 - Removed link for opening sessions in RStudio Desktop Pro from Session Info dialog. (rstudio-pro#5263)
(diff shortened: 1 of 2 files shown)


version/news/NEWS-2023.12.0-ocean-storm.md
```diff
@@ -19,6 +19,7 @@
 - Better error message when user preferences fail to save due to folder permissions. (#12974)
 - RStudio now supports pasting of file paths for files copied to the clipboard. (#4572)
 - RStudio now supports duplicate connection names for Posit drivers. (rstudio-pro#5437)
+- Added Databricks integration to RStudio (Pro) which allows users to seamlessly interact with Databricks clusters and resources (rstudio-pro#4927)
```
melissa-barca, line 22:
```suggestion
- Added Databricks integration to RStudio Pro Sessions which allows users to seamlessly interact with Databricks clusters and resources (rstudio-pro#4927)
```
I think "RStudio Pro Sessions" is the correct terminology, and we should move this to the Workbench section since those sessions are only available in Workbench. Unless this is referring to RDP, in which case we should refer to it as such.

melissa-barca (changes requested):
To get these into the official release notes, we also need to update https://github.com/rstudio/rstudio/blob/main/docs/news/index.qmd
Can we do that as part of this PR as well?

docs/news/index.qmd
```diff
@@ -34,6 +34,7 @@ This page provides the release notes associated with each release of RStudio and
 - Improved support for development documentation when a package has been loaded via `devtools::load_all()`. (#13526)
 - Better error message when user preferences fail to save due to folder permissions. (#12974)
 - With screen reader support enabled, hitting ESC key allows Tabbing away from editor. [accessibility] (#13593)
+- Added Databricks integration to RStudio Pro Sessions which allows users to seamlessly interact with Databricks clusters and resources (rstudio-pro#4927)
```
melissa-barca, line 37: I still think we should move this down to the Workbench section. IIUC this section is really just for open source.

GCRev: For sure. Sorry if I didn't notice you mentioning that beforehand.

melissa-barca (approved):
LGTM!

PR #14110, opened 2024-01-10
improve handling of missing / empty workbench lists

kevinushey:
I'm currently getting a "blank screen" on startup due to errors when trying to read workbench lists. This PR makes the IDE startup a bit more permissive, so we can handle the case where a particular workbench list is missing.

I'm not sure if we should be logging or reporting if this happens? I am a bit worried that this might be masking a different sort of error.

Screenshot 2024-01-10 at 2 00 09 PM

### DIFF
--- a/src/gwt/src/org/rstudio/core/client/layout/DualWindowLayoutPanel.java
+++ b/src/gwt/src/org/rstudio/core/client/layout/DualWindowLayoutPanel.java
@@ -14,11 +14,12 @@
  */
 package org.rstudio.core.client.layout;
 
-import com.google.gwt.core.client.JavaScriptObject;
-import com.google.gwt.core.client.Scheduler;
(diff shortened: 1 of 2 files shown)


kevinushey:
I'm going to merge this for now just so I can get RStudio Desktop dev. builds running again; let's revisit if there's a more appropriate fix to be made later.

gtritchie:
I think we should back this change out and work to understand what you were actually encountering and fix that!

I've been using dev builds (dev server and Electron via `npm run`) pretty much daily for past couple months and never hit this.

kevinushey:
I figured out what happened. I have two separate builds of the RStudio session locally. We should normally be picking the more recent build, but apparently that was not functioning as expected -- so in the end, I was mixing "new" UI with "old" session code, hence the failure. Sorry 😞

Issue #14096, opened 2024-01-07
RStudio slow to open R files in directories with many files, especially on network drives


bapike:
I have a network drive containing a single directory with about 2000 files, some of which are small R scripts.  Opening one of these R scripts in RStudio using File->Open File (or the "Open an existing file" button) can take a long time (40 seconds on a LAN, or >20 minutes over a VPN), while opening the same file in Notepad over the same network connection takes 2 seconds.

Note: When a file is opened via File->Recent Files, or when a file opens automatically when RStudio starts (because it was open when RStudio last closed), then the file opens in RStudio within a second or two even over VPN.

After opening a file that demonstrates the problem, using "Help->Diagnostics->Request Log" shows that most of the time is occupied by a `list_files` request for the directory containing the file I'm trying to open.  The response to that request is a list of all the files in that directory, with file metadata.  I'm not sure how this list is useful to RStudio, as it seems unnecessary.  The file seems to be opened only after the `list_files` request completes.

Additionally, this `list_files` request appears to be inefficient.  Running Process Monitor on RStudio while trying to open a problematic file shows a lot of repetition in the requests for information about each file; for example, the attributes for each file seem to be requested about 16 times.  Assuming that the `list_files` request can be run using `.rs.invokeRpc("list_files",mydir,FALSE,FALSE)`, then I've benchmarked `list_files` unfavorably against some similar commands (see below).

I am not using a Project nor using version control from within RStudio.  I turned off as much as possible in my user preferences: R code completion, R diagnostics, checks for missing R packages, spell-checking, version control interface for RStudio projects, and publishing documents/apps/APIs.  I recently updated from RStudio-1.1.442, which may have had a similar problem but perhaps not as severe.

### System details

    RStudio Edition :  RStudio Desktop
    RStudio Version :  RStudio 2023.12.0+369 "Ocean Storm" Release (33206f75bd14d07d84753f965eaa24756eda97b7, 2023-12-17) for windows
    OS Version      :  Windows 10 (22H2)
    R Version       :  R version 4.3.2 (2023-10-31 ucrt)

### Steps to reproduce the problem

1. Create a directory containing lots of files (say, 8000 for a local directory, 1000 for a network drive on a LAN, 30 for a network drive on a VPN, but you may need to adjust).
2. In RStudio, open one of these files using File->Open File.  Time how long it takes to open.
3. In Notepad, open the same file.  Time how long it takes to open.

Here's a script for generating a directory with lots of files:

```
[...]

PR #14111, opened 2024-01-10
avoid listing files in directory when opening a file

kevinushey:
### Intent

Addresses https://github.com/rstudio/rstudio/issues/14096.

### Approach

We have a file system context class, which tries to maintain a working directory + the contents of that directory. I suspect the latter is primarily used for the Files pane; however, because that class is apparently shared in a couple places (e.g. preserving the default directory to be used when opening files) we end up unnecessarily listing files in the directory hosting the file we're opening.

The fix here is fairly narrow; in desktop file dialogs, when we open a document, we define and use a separate helper called `setwd()` rather than `cd()`, and allow `setwd()` to do less work.

An alternate fix would be to defer listing of a directory's contents until `ls()` itself is actually called. That might be preferable, but that also feels like a relatively more dangerous change as it would affect more existing code paths.

### Automated Tests

Not included.

### QA Notes

Test via notes in https://github.com/rstudio/rstudio/issues/14096.

### Documentation

N/A

### Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -20,6 +20,7 @@
 - Fixed GitHub Copilot project preferences not showing correct status message (#14064)
 - Fixed an issue where Quarto chunk option completions were not displayed at the start of a comment (#14074)
 - Fixed an issue where pipes containing a large number of comments were not indented correctly (#12674)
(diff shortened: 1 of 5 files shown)


gtritchie (approved):
LGTM!

